PR #13, opened 2023-02-04
[BUG] Fix tag in FeatureSelection and added tests

aiwalter:
Fixes tag by means of dynamic tag setting. Added some unit tests as well.


lmmentel:
Could you explain what does your change does and why it's needed?

aiwalter:
the `FeatureSelection(method="none")` was not well tested and did therefore not work (anymore?). I added tests for `FeatureSelection` as there were no tests before and also fixed the `method="none"` functionality by dynamically setting the tag.

lmmentel (approved):
:+1:

PR #26, opened 2023-02-06
[ENH] Refactored PCATransformer

aiwalter:
- Removed `numpy` operations in `_fit` and `_transform`. 
- Added unit test
- Added columns naming e.g. `PC_0`, `PC_1` for the first two columns
- Added missing params
- Added attribute `pca_`

I started to refactor bcz I have seen the transformer was not working with my series as intended - however I actually thought due to the datatype module all transformers from `panel` module can also consume `pd.DataFrame`? As this was not the case, I refactored it and moved it to `series` module. In longer perspective we might get rid of the submodules `panel` and `series` anyways and only have `transformations` module

Example before vs. after:
```python
from sktime.utils._testing.panel import _make_nested_from_array
import numpy as np
X = _make_nested_from_array(np.ones(10), n_instances=10, n_columns=4)
pca = PCATransformer(n_components=2).fit(X)
pca.transform(X).head()
```

Output before refactor:
image

Output after refactor:
image

For the series case, its doing the following (which was not possible before):
```pyhton
from sktime.transformations.series.pca import PCATransformer
from sktime.datasets import load_longley
y, X = load_longley()

transformer = PCATransformer(n_components=2)
transformer.fit(X)
Xt = transformer.transform(X)
Xt.head()
```
image

I am not sure if the "before" output was even correct? I am not too familiar with the `panel` module. Maybe @MatthewMiddlehurst you can have a look here?

#### To do:
- [ ] Clean up other references of `PCATransformer`, e.g. in other tests and docs


TonyBagnall:
We dont actually use this transformer, and tbh I dont really understand how he has structured it. The code is simply an scikit wrapper. The structure for datatypes does not make sense to me. It says this

"scitype:transform-input": "Series",
# what is the scitype of X: Series, or Panel
"scitype:transform-output": "Series",
# what scitype is returned: Primitives, Series, Panel

but the transform says 
X : Panel data in 3D np.ndarray format [n_instances, n_variables, n_timepoints]

surely that is a Panel. So the long and short is the whole thing is wrapped up in the complex datatypes conversion thing, and I would not use it myself, I would just use scikit learn

TonyBagnall:
the transformers module is a dogs dinner imo, and there are some significant issues to discuss. We might need to maintain the separation for single instance transformation (series) and multiple instance transformation (so called Panel), because of the different use cases for 2D numpy. If we have n series and t time points, forecasting assumes the data are X[t][n], whereas in classification/regression/clustering we assume it is X[n][t]. I'm not sure about PCATransformer, I'll take a look

aiwalter:
> I would not use it myself, I would just use scikit learn

For forecasting I use this refactored tranformer to reduce exog data size, therefore a transformer is required to put it nicely into the pipeline

> We might need to maintain the separation for single instance transformation (series) and multiple instance transformation (so called Panel), because of the different use cases for 2D numpy. 

I think now we have the `pd.MultiIndex` support for transformers out of the box, so I think there is no difference now anymore between the `panel` module and the `series` module of tranformers? I thought the data type machinery is handling that together with base class stuff?

>  If we have n series and t time points, forecasting assumes the data are X[t][n], whereas in classification/regression/clustering we assume it is X[n][t]. I'm not sure about PCATransformer, I'll take a look

I am a bit confused here, In forecasting this would be a `pd.DataFrame` with `MultiIndex`, I never used `numpy` for forecasting and also I think its BS to even support it but thats a different story.

TonyBagnall:
afaik all machine learning toolkits (tensorflow, pytorch, scikit-learn etc) use numpy almost exclusively, dataframes are much more of an R/stats thing historically (I think scikit support them now as 2D tables). Forecasting and time series machine learning are very different subjects, but thats ok, we can interface. Point is transformations for us should be applied to numpy, because all of the algorithms we implement use numpy and if we use dataframes it becomes very inefficient pulling data in and out, particularly in pipelines. Forecasting and machine learning meet in transformations, but I dont think wrapping everything up in pandas is a good idea at all, we have issues already with the vectorisation of ROCKEt. I think @patrickzib has a better handle on this than me

patrickzib:
Currently, 2d NumPy array support for classification is broken. It is interpreted as a single time series. I did not reach a consensus with FK on how to proceed, as it seems to be useful in Forecasting to have a single time series.

Additionally, the conversion between Pandas and NumPy is slow. I do not know, if it is possible to address this in general. However, the following would work for me: If we input a Pandas DataFrame, we convert the Pandas DataFrames to NumPy Arrays within the first step of the pipeline and return a Pandas DataFrame at the last step of the pipeline, whereas all intermediate steps are on NumPy array.

lmmentel:
I had a quick look and it do not fully understand why this is even needed. It seems that the transformer is a copy of PCA from sklearn with word for word copy of the docs. Surely I'm missing something but given what's in the test, we are testing if sklearn's PCA (with extra steps) works.

Am I wrong assuming that this:

```python
from sktime.transformations.series.pca import PCATransformer
from sktime.datasets import load_longley
_, X = load_longley()
transformer = PCATransformer(n_components=2)
X_hat = transformer.fit_transform(X)
```

would be exactly the same as this

```python
from sklearn.decomposition import PCA
from sktime.datasets import load_longley
_, X = load_longley()
transformer = PCA(n_components=2)
X_hat = transformer.fit_transform(X)
```

If so then why the extra steps?

TonyBagnall:
I'm not sure about your use case, but this does not work in the same way with 3D numpy and dataframes as the existing PCA and this new  so if I run this code

TonyBagnall (approved):
so the use case for this transformer is different to what it was before, but I think this scikit learn wrapper is not really useful for classification and Im happy for it to move to the series side of things.

aiwalter:
>  Surely I'm missing something

The benefit of this PR is more cosmetic, since it basically adds proper column names instead just values of `0`, `1`, `...` as names. What we can already do is this


```python
from sklearn.decomposition import PCA
from sktime.transformations.series.adapt import TabularToSeriesAdaptor
from sktime.datasets import load_longley
_, X = load_longley()
transformer = TabularToSeriesAdaptor(PCA(n_components=2))
X_hat = transformer.fit_transform(X)
X_hat.head()
```
```

	0	1
Period		
1947	-153725.651033	-831.133744
1948	-128579.168413	-294.987174
1949	-129860.699646	-1892.822103
1950	-103301.665927	-1144.347197
1951	-58956.256448	1619.149486
```

E.g.  I use after `PCA` a `WindowSummarizer` where I then need to refer to some column names. Generally I dont mind not to have this PR and completely remove `PCATransformer` then. Not sure whats your opinion @TonyBagnall and @lmmentel  if its worth to have a class just bcz of column names or not?

lmmentel:
Thanks for the extra context @aiwalter. The use case you mention makes perfect sense.

> Not sure whats your opinion @TonyBagnall and @lmmentel if its worth to have a class just bcz of column names or not?

In this specific case I would say that there are 2 separate functionalities: 

- the actual PCA logic
- converting the result and assigning names

Ideally for the case of tabular data (e.g. multivariate time series) we would be able to use sklearn's PCA directly, since that's why we are adopting the api - to be interoperable with sklearn. The type conversion and renaming could be a separate functionality or another step in the pipeline - that way we have less redundant code to maintain and only build things that add value in our specialization domain.

Another case is making PCA work with some of the data structures we support for different tasks, i.e. xarray or dataframes with multiindexes. Here it might make sense to wrap PCA for specific data structure, but I would rather support a few task specific versions of PCA that are designed for a specific data structure that have a hidden do-all converter.

This is probably more of a design question we'll dive into at some point but it's not blocking this PR.

lmmentel (approved):

PR #86, opened 2023-02-20
[ENH] Remove ProximityForest classifier

### LINKED ISSUE #38: [BUG] Deprecate ProximityForest
**Describe the bug**
ProximityForest is broken and at this point it is easiest to deprecate it and maybe look for someone to reimplement it. See here for a discussion

https://github.com/sktime/sktime/issues/3648

**To Reproduce**

```python
from sktime.classification.distance_based import ProximityForest
    from sktime.datasets import load_unit_test
    pf = ProximityForest()
    trainX, trainy = load_unit_test()
    pf.fit(trainX, trainy)```
```
this results in an infinite recursion in _fit leading to stack overflow.
RecursionError: maximum recursion depth exceeded while calling a Python object
this happens on all problems I tried (unit_test, arrow_head, GunPoint, ItalyPowerDemand)

**Expected behaviour**

It should fit and predict with default parameters on data in an allowed format.

**Additional context**
Removal previously blocked. The original author of PF is unavailable,

</details>

TonyBagnall:
ProximityForest is a distance based tree ensemble. It was implemented in 2019, but we could never get accuracy equivalence with the java version. It is also dependent on dataframes internally, and this is hugely inefficient. This implementation is best scrapped, and hopefully someone will take on the challenge of re-implementing it with numpy arrays. 

#### Reference Issues/PRs
Fixes #38


.gitignore
```diff
@@ -161,3 +161,5 @@ envs/
 
 # example mlflow runs
 examples/local/
+sktime/tony_debug/notebook_examples.py
+sktime/tony_debug/__init__.py
```
aiwalter, line 165: just exclude the whole folder `tony_debug`?

TonyBagnall: argg, thought I had sorted that :)

patrickzib (approved):
Looks good to me.

Issue #47, opened 2023-02-08
[ENH] Remove dists_kernels


TonyBagnall:
we currently have a module distances, which contains elastic distance functions, and one called dists_kernels that wraps everything in objects. I would like to remove dists_kernels. It is confusing having both, and I see no benefit from wrapping distances and introducing things like PwTrafoPanelPipeline, BasePairwiseTransformerPanel etc

I think this description would make little sense to most people. 
https://www.sktime.org/en/stable/api_reference/dists_kernels.html

conceptually, insisting that functions be wrapped in objects (functor pattern) is much more Java like than python. Even java recognises functors are a clumsy pattern and introduced lambdas. To go the other way with python just seems perverse. @chrisholder can comment, but I believe all other relevant packages treat distances as functions. 

https://github.com/scikit-time/scikit-time/issues/39
https://github.com/scikit-time/scikit-time/issues/46
https://arxiv.org/abs/2205.15181

PR #90, opened 2023-02-20
[MNT] `convolution_based` rename

MatthewMiddlehurst:
In line with https://github.com/sktime/sktime/pull/3529, but without deprecation.

Moved the ROCKET based classifiers to `convolution_based`, and `RandomIntervalClassifier` to `interval_based`.

The remaining `kernel_based` classifier should be left to #47.


TonyBagnall (changes requested):
fine in principle, but it looks like some of the notebooks need updating

TonyBagnall (approved):

PR #91, opened 2023-02-20
[ENH] _fit_transform method in BaseTransformer 

MatthewMiddlehurst:
Contents of https://github.com/sktime/sktime/pull/3847

Implements a `_fit_transform` method in `BaseTransformer`. `fit_transform` now calls the private method, and contains a set of checks merged from `fit` and `transform`.

This PR also replaces the current `fit_transform` overrides in `SupervisedIntervals` and `TSFreshRelevantFeatureExtractor` with the private method.

Alongside allowing for easier implementation of efficient `fit_transform` methods, the change reduces the repetition of some checks when the method is called.

#### Any other comments?

Happy to make changes or add additional checks if people think they are necessary. As it is, this is essentially a merge of the current `fit` and `transform` methods with some repetition removed.


sktime/transformations/base.py
```diff
@@ -551,9 +534,29 @@ def fit_transform(self, X, y=None):
...
+        # input checks and datatype conversion
+        X_inner, y_inner, metadata = self._fit_checks(X, y, False, True)
+
+        # checks and conversions complete, pass to inner fit_transform
+        ####################################################
+        vectorization_needed = isinstance(X_inner, VectorizedDF)
+        self._is_vectorized = vectorization_needed
+        # we call the ordinary _fit_transform if no looping/vectorization needed
+        if not vectorization_needed:
+            Xt = self._fit_transform(X=X_inner, y=y_inner)
```
patrickzib, line 546: Why does a call to `transform` result in a call to `fit`?

Seems like this was the case already before, but I do not see the reason? Is it done like that in sklearn, too?

Assuming the transformer was fitted, this is kind of unintuitive? Should there be a check, if the transformer was fitted, else apply only transform?

sktime/transformations/base.py
```diff
@@ -1144,6 +1146,17 @@ def _vectorize(self, methodname, **kwargs):
 
             return Xt
 
+        if methodname == "fit_transform":
```
patrickzib, line 1149: Could we discuss `vectorize`?

It states: 
> Uses transformers_ attribute to store one transformer per loop index.

I do not understand the logic behind this. If the transformer is a z-normalization, the mean and std should be computed based on all data. Does vectorize  `fit` multiple instances of the same transform to slices? Or is vectorize only applied to `transform`?

patrickzib (changes requested):

patrickzib:
as a general comment: I do not like all the cloning of transformers to run a simple parallel for-loop (called vectorize). 

The transformer should be fit once on all data. And then it should be stateless and without side-effects, i.e. it should be possible to apply transform to slices without cloning.

patrickzib:
My suggestion: can we get rid of `vectorize`? I do not see any advantage of it. I.e. there is not even any parallelization.

This is the interface of sklearn:
https://github.com/scikit-learn/scikit-learn/blob/8c9c1f27b/sklearn/preprocessing/_function_transformer.py#L20

It is slim and simple.

sktime/transformations/base.py
MatthewMiddlehurst: Not sure if I am understanding. It's a call to `fit_transform`, the expected workflow is `fit` into `transform`?

sktime/transformations/base.py
MatthewMiddlehurst: I don't like it either, but it is what we have currently 🙂.

sktime/transformations/base.py
patrickzib: Sorry. I saw the diff only.

sktime/transformations/base.py
patrickzib: Yes, I am in favor of your chance. 

Would love to see an even bigger change, though. Maybe in a different PR.

patrickzib (approved):
I think this is a good addition. Would love to see an even larger refactor of the BaseTransformer. Maybe in a different PR.

Issue #61, opened 2023-02-14
[ENH] Parallel Backend


patrickzib:
Should we change the parallel backend to threading in the classifiers?

```python
Parallel(n_jobs=n_jobs, backend="threading")(
```

I have noticed issues when calling nested Parallel-constructs.

@MatthewMiddlehurst @TonyBagnall I recall, you noticed similarly?

TonyBagnall:
yes we had issues. Matthew may remember better than me, but I think it tends to time slice and pickle, which can cause permission issues, and that changing the backend to threading ameliorated it. I did not get to the bottom of it, there is a fair amount to do with threading classifiers imo, but we should default to backend threading imo

patrickzib:
Thus, we could change it for all classifiers? Or maybe make it a parameter of each classifier?

TonyBagnall:
I would choose to hard code it for each classifier, I don't think it really needs exposing as a parameter. Having said that, I do not have a full grasp on the ramifications. Testing/benchmarking the parallel is on my (long) list. So perhaps change classifiers where its an issue, run some experiments pre and post with those to make sure its better and correct, and leave the policy change until we can properly evaluate it across the board

MatthewMiddlehurst:
I think we should eventually expose it as a parameter, but just hardcode threading for now. This has solved multithreading issues in the past.

patrickzib:
Has been solved with recent commit

PR #93, opened 2023-02-20
[ENH] Add `prefer="threads"` to classification `Parallel` usage

MatthewMiddlehurst:
Contents of https://github.com/sktime/sktime/pull/3879

Related to #61.


patrickzib (dismissed):
Thank you!

MatthewMiddlehurst:
Just a lot going on. Can only run so many actions at once. Big downside of having 4 hour test runs.

patrickzib:
> B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

I saw this, too, on my code.

It has to be:

```python

warnings.warn(
                ... , stacklevel=1
            )
```

Wonder if this check is new?

patrickzib (dismissed):
Well done!

lmmentel (approved):

patrickzib (approved):

PR #95, opened 2023-02-21
[Bug] ClaSP Bugfixes

patrickzib:
This addresses some bugs in ClaSP:
- Adds a warning if the window size is larger than the time series
- Checks if there are enough points left to do a segmentation

### DIFF
--- a/sktime/transformations/series/clasp.py
+++ b/sktime/transformations/series/clasp.py
@@ -17,9 +17,9 @@ As described in
 __author__ = ["ermshaua", "patrickzib"]
 __all__ = ["ClaSPTransformer"]
 
-import numpy as np
+import warnings
 
-# import numpy.fft as fft
+import numpy as np
 import pandas as pd
 from numba import njit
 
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/clasp.py
```diff
@@ -428,6 +432,11 @@ def _transform(self, X, y=None):
             ClaSP of the single time series as output
             with length as (n-window_length+1)
         """
+        if len(X) - self.window_length < 2 * self.exclusion_radius * len(X):
+            warnings.warn(
+                "Period-Length is larger than size of the time series", stacklevel=1
```
TonyBagnall, line 437: I got warnings before saying stacklevel should be 2 or more, not sure its required though

TonyBagnall (approved):

sktime/transformations/series/clasp.py
patrickzib: Yes, I think, the warning is that stacklevel should be applied as argument:
> B028 No explicit stacklevel keyword argument found. 

I rather not have a stack-trace (stacklevel=2) though, as running clasp on multiple datasets spams the terminal otherwise :)

PR #101, opened 2023-02-21
refactor BaseClassifier attribute _threads_to_use to _n_jobs

TonyBagnall:
refactor the the base classifier attribute _threads_to_use to _n_jobs in order to conform to the general style.

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -53,7 +53,7 @@ class BaseClassifier(BaseEstimator, ABC):
     _X_metadata         : metadata/properties of X seen in fit
     _class_dictionary   : dictionary mapping classes_ onto integers
     0...``n_classes_``-1.
-    _threads_to_use     : number of threads to use in ``fit`` as determined by
+    _n_jobs     : number of threads to use in ``fit`` as determined by
     ``n_jobs``.
     _estimator_type     : string required by sklearn, set to "classifier"
     """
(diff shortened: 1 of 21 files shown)


patrickzib (commented):
I am in favor of this change. However, I do not quite understand, when to use `self._n_jobs` and `self.n_jobs` in classifiers.

As I understand it now:

when we call fit() it will set  

```python
self._threads_to_use = check_n_jobs(self.n_jobs)
```

Thus, from there on, we should always use self._n_jobs instead of self.n_jobs?

(there are a couple of uses of `self.n_jobs` (without underscore) in the classifiers. Thus, maybe these should be changed, too)

patrickzib:
Classes still using Parallel(n_jobs=self.n_jobs) (without underscore):

_tsf.py
_boss.py
_rise.py
_ensemble.py
_muse.py
_weasel.py
clear_sky.py


Maybe I missed some ...

TonyBagnall:
> Classes still using Parallel(n_jobs=self.n_jobs) (without underscore):
> 
> _tsf.py _boss.py _rise.py _ensemble.py _muse.py _weasel.py clear_sky.py
> 
> Maybe I missed some ...

So I guess this is why it hasnt been changed .... I agree that they should all use _n_jobs, but I think I will keep it to classifiers for now, not sure what clear_sky.py is

TonyBagnall:
I'm not touching _tsf as I had to hack it to work, it needs redoing. This refactor is more complex that I imagined. 
image

TonyBagnall:
I'm going to close this for now and just raise the issue https://github.com/scikit-time/scikit-time/issues/102

PR #103, opened 2023-02-22
[MNT] Remove Python 3.7 support

### LINKED ISSUE #84: [ENH] Deprecate Python 3.7 and bump scikit-learn version
Some estimators dont support Python 3.7 anymore, we should think about deprecating it. Also newer `scikit-learn>=1.1.0` has no Python 3.7 anymore.

Probably we can bump then `scikit-learn>=1.1.0` and possibly also other dependencies to require a higher version.

aiwalter:
closes #84

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,7 +115,7 @@ jobs:
     runs-on: windows-2019
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10']
 
     steps:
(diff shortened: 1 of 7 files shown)


lmmentel:
Maybe it's a bit premature to drop support before official EOL. How about prepping a PR and merging closer to actual EOL?

MatthewMiddlehurst:
sklearn dropped 3.7 back in v1.1.0. I think now would be as good a time as any.

lmmentel:
I guess by the time we release we might be quite close to py3.7 EOL so it makes sense.

lmmentel (approved):
:+1:

PR #105, opened 2023-02-23
[ENH] remove TimeSeriesSVC

### LINKED ISSUE #94: [ENH] Remove TimeSeriesSVC

TimeSeriesSVC was introduced on 25th De 2022 without review to sktime as a wrapper for scikit support vector machine. I see no issue associated with this classifier . Its purpose seems to be to find a use case for dists_kerns module, but its problematic and I think it should be removed for the following reasons.

1. It is not properly integrated to distances module, it does everything through AggrDist, which is a BasePairwiseTransformerPanel, which takes a BasePairwiseTransformer. This is a tremendous amount of infrastructure to support creating a distance matrix, and very confusing.
2. Elastic distances do *not* generate semi-positive definite kernels. For people who use SVM would be a deal breaker. The classifier has the comment "typically, SVC literature assumes kernels to be positive semi-definite. However, any pairwise transformer can be passed as kernel, including distances. This will still produce classification results, which may or may not be performant" I dont think we should introduce classifiers on this basis. If there were a paper and related results, then fair enough. However, svm are not used in the TSC literature and svm have historically performed poorly at time series classification.
3. Its just a wrapper. Anyone could just use scikit learn svc if they understood it without having to get their head around the dist_and_kerns module. The core implementation is just this

```python
  def _fit(self, X, y):
        self._X = X
        kernel_mat = self._kernel(X)
        self.svc_estimator_.fit(kernel_mat, y)
        return self

    def _predict(self, X):
        kernel_mat = self._kernel(X, self._X)
        y_pred = self.svc_estimator_.predict(kernel_mat)
        return y_pred
```

Remove this classifier

TonyBagnall:
removes the TimeSeriesSVC, closes #94 

this also removes the kernel_based directory completely, and changes mentions of kernel_based that are incorrect after #90


patrickzib (approved):
Fine with me. I am not aware of a publication for this approach or its practical relevance.

Issue #42, opened 2023-02-08
[DOC] remove references to mtype/machine type scitype/stype


TonyBagnall:
the terms mtype/machine type and scitype/stype were introduced with the datatype module. These are non standard terms that mean nothing to most people and are sure to cause confusion.

For example, the header for the _convert module is this
"""Machine type converters for scitypes."""
and comments like this
"obj : object to convert - any type, should comply with mtype spec for as_scitype

there is no reference here to what these mean nor a link to an explanation. As far as I can tell it is a reinvention of the concept of abstract data types with some weird inclusion of objects. So a scitype would be an abstract data type such as a list, an mtype would be an implementation of a list such as a linked list or an array. I see absolutely no need for this distinction and think all is does is confuse things
 
![image](https://user-images.githubusercontent.com/9594042/217515683-a02241b2-3742-4957-a3c1-a0d29f3e68c0.png)

it basically gives a form out automation to convert, but I find it convoluted and confusing. 

```python
Exports
-------
convert_to(obj, to_type: str, as_scitype: str, store=None)
    converts object "obj" to type "to_type", considerd as "as_scitype"

convert(obj, from_type: str, to_type: str, as_scitype: str, store=None)
    same as convert_to, without automatic identification of "from_type"

mtype(obj, as_scitype: str)
    returns "from_type" of obj, considered as "as_scitype"
---
```

I would like to remove these terms completely, and replace them with, for example
input type/internal type/output type
then have simple converters that convert from one to another and remove the scitype 
just looking at usages of this, to_type is a required parameter, so all this does
```python
def convert_to(
    obj,
    to_type: str,
    as_scitype: str = None,
    store=None,
    store_behaviour: str = None,
):
```
is some convoluted checking, then calls convert! This adds no value I can see. In first instance I would just change the terminology, but really I doubt the below adds any value at all.  
```python
   # input checks on to_type, as_scitype; coerce to_type, as_scitype to lists
    to_type = _check_str_or_list_of_str(to_type, obj_name="to_type")

    # sub-set a preliminary set of as_scitype from to_type, as_scitype
    if as_scitype is not None:
        # if not None, subset to types compatible between to_type and as_scitype
        as_scitype = _check_str_or_list_of_str(as_scitype, obj_name="as_scitype")
        potential_scitypes = mtype_to_scitype(to_type)
        as_scitype = list(set(potential_scitypes).intersection(as_scitype))
    else:
        # if None, infer from to_type
[...]

PR #109, opened 2023-02-25
[ENH] make single problem loaders for equal length problems return numpy arrays

TonyBagnall:
change the return_type from None to numpy3d and remove the very long comments describing possible data structures in unnecessary detail. Part of #42 

*edit*: sorry this touches so many files, I had no idea that the single problem loaders were used so much in testing (other than correctness testing). I think unless testing actual output, tests should use randomly generated data from tests/_testing package. I've had a look through a fair few tests on this epic trawl, and there seems to be a lot of testing for testing sake in obscure code such as benchmarking. There is testing same thing on very similar datasets, using larger datasets unnecessarily etc. I have raised an issue #118


sktime/datasets/_single_problem_loaders.py
```diff
@@ -114,81 +114,36 @@ def load_UCR_UEA_dataset(
     Examples
     --------
     >>> from sktime.datasets import load_UCR_UEA_dataset
-    >>> X, y = load_UCR_UEA_dataset(name="ArrowHead")
+    >>> X, y = load_UCR_UEA_dataset(name="ArrowHead", return_type="numpy3d")
```
patrickzib, line 117: Wouldn't it be better to change the API call instead of the example call? Currently the api call still defaults to None.

```python
def load_UCR_UEA_dataset(
    name, split=None, return_X_y=True, return_type="numpy3d", extract_path=None
):
```

sktime/datasets/_single_problem_loaders.py
```diff
@@ -114,81 +114,36 @@ def load_UCR_UEA_dataset(
...
-    Returns
-    -------
-    X: sktime data container, following mtype specification `return_type`
-        The time series data for the problem, with n instances
-    y: 1D numpy array of length n, only returned if return_X_y if True
-        The class labels for each time series instance in X
-        If return_X_y is False, y is appended to X instead.
+        If True, returns (time series, target) separately as two arrays.
+    return_type: string, optional (default="numpy3d")
+        Data structure to use for time series, should be either "numpy2d" or "numpy3d".
```
patrickzib, line 133: Can we remove `numpy2d`? It will not work with transformers for classification

sktime/datasets/_single_problem_loaders.py
```diff
@@ -114,81 +114,36 @@ def load_UCR_UEA_dataset(
     Examples
     --------
     >>> from sktime.datasets import load_UCR_UEA_dataset
-    >>> X, y = load_UCR_UEA_dataset(name="ArrowHead")
+    >>> X, y = load_UCR_UEA_dataset(name="ArrowHead", return_type="numpy3d")
     """
     return _load_dataset(name, split, return_X_y, return_type, extract_path)
```
patrickzib, line 119: General comment: remove numpy2d for loading datasets?

patrickzib (changes requested):
Two suggestions:
- Change default in load_UCR_UEA_dataset to numpy3d
- Remove support for numpy2d

sktime/datasets/_single_problem_loaders.py
TonyBagnall: need to resolve the unequal length issue before doing that, was going to just get this in first and wait, this is mostly cosmetic

sktime/datasets/_single_problem_loaders.py
TonyBagnall: sure, although it works with all classifiers and is closest to scikit. Maybe I'll just give some examples with reshape

sktime/datasets/_single_problem_loaders.py
TonyBagnall: > 

yes using this PR to find all calls in the notebooks :)

TonyBagnall: but dont want to do too much, next job is to rewrite the classification notebook anyway

sktime/datasets/_single_problem_loaders.py
patrickzib: True. I would love to have numpy2d support, as it it closest so scikit-learn. But then we should got for the transformer instead to support numpy2d input correctly..

TonyBagnall: in time we shall! But this is a very incremental change, just removing the essays on weird supported types and shifting more to numpy

TonyBagnall: actually I think I would prefer to leave the text allowing numpy2d in there, and lets just raise an issue about the transformers, listing all those effected. Is that ok?

TonyBagnall: because if I actually stop the return_type argument and always return 3D numpy, there are a million old legacy calls that will break in places like the "benchmarking" module

TonyBagnall:
> Two suggestions:
> 
> * Change default in load_UCR_UEA_dataset to numpy3d
> * Remove support for numpy2d



> Two suggestions:
> 
> * Change default in load_UCR_UEA_dataset to numpy3d

its in the cards, I think I will change the name to load_tsc_problem, but it needs to also load unequal problems so was going to delay until we remove DataFrames all together

> * Remove support for numpy2d

see my other comments

TonyBagnall:
I had no idea that the single problem loaders were used so much in testing (other than correctness testing), this is surely not a good idea, will raise an issue

TonyBagnall:
well, that tweak to kill time on a saturday morning became a monster!

patrickzib:
I would give you a badge of honour! :)

examples/minirocket.ipynb
```diff
@@ -252,7 +252,7 @@
    },
    "outputs": [],
    "source": [
-    "X_train, y_train = load_basic_motions(split=\"train\", return_X_y=True)"
+    "X_train, y_train = load_basic_motions(split=\"train\")"
```
MatthewMiddlehurst, line 255: Seemed to have missed this and line 330 when replacing the return type, unless the swap to numpy was intentional.

sktime/benchmarking/tests/test_orchestration.py
```diff
@@ -37,55 +37,14 @@ def make_reduction_pipeline(estimator):
     return pipeline
 
 
-# simple test of orchestration and metric evaluation
-@pytest.mark.parametrize("data_loader", [load_gunpoint, load_arrow_head])
-def test_automated_orchestration_vs_manual(data_loader):
-    """Test orchestration."""
-    data = data_loader(return_X_y=False)
```
MatthewMiddlehurst, line 44: Seems like you can add the new return type here instead of removing the test? No idea what the test is doing or if it will stay in the long run.

sktime/datasets/_data_io.py
```diff
@@ -398,22 +398,11 @@ def load_from_tsfile(
...
-        memory data format specification to return X in.
-        str can be any other supported Panel mtype
-            for list of mtypes, see datatypes.SCITYPE_REGISTER
-            for specifications, see examples/AA_datatypes_and_datasets.ipynb
-        commonly used specifications:
-            "nested_univ: nested pd.DataFrame, pd.Series in cells
-            "numpy3D"/"numpy3d"/"np3D": 3D np.ndarray (instance, variable, time index)
-            "numpy2d"/"np2d"/"numpyflat": 2D np.ndarray (instance, time index)
-            "pd-multiindex": pd.DataFrame with 2-level (instance, time) MultiIndex
-        Exception is raised if the data cannot be stored in the requested type.
```
MatthewMiddlehurst, line 410: Think these can be simplified, but not to the current extent. loading into pandas is still available functionality.

examples/minirocket.ipynb
TonyBagnall: want to use numpy where I can, had to switch one above because it used pandas plot, but will tidy up these in a subsequent PR

sktime/benchmarking/tests/test_orchestration.py
TonyBagnall: well I would like shot of it #116 but I guess you are correct, will put it back

sktime/datasets/_data_io.py
TonyBagnall: yes, maybe over zealous there, will rework it

examples/02_classification.ipynb
```diff
@@ -545,7 +545,7 @@
     "`sktime` offers two other ways of building estimators for multivariate time series problems:\n",
     "\n",
     "1. Concatenation of time series columns into a single long time series column via `ColumnConcatenator` and apply a classifier to the concatenated data,\n",
-    "2. Dimension ensembling via `ColumnEnsembleClassifier` in which one classifier is fitted for each time series column/dimension of the time series and their predictions are combined through a voting scheme. \n",
+    "2. Dimension ensembling via `ColumnEnsembleClassifier` in which one classifier is fitted for each time series column/dimension of the time series and their predictions are combined through a voting scheme.\n",
```
patrickzib, line 548: Channel? ;)

examples/AA_datatypes_and_datasets.ipynb
```diff
@@ -262,8 +262,8 @@
     "\n",
     "* structure convention: `obj.index` must be a pair multi-index of type `(Index, t)`, where `t` is one of `Int64Index`, `RangeIndex`, `DatetimeIndex`, `PeriodIndex` and monotonous. `obj.index` must have two levels (can be named or not).\n",
     "* instance index: the first element of pairs in `obj.index` (0-th level value) is interpreted as an instance index, we call it \"instance index\" below.\n",
-    "* instances: rows with the same \"instance index\" index value correspond to the same instance; rows with different \"instance index\" values correspond to different instances. \n",
-    "* time index: the second element of pairs in `obj.index` (1-st level value) is interpreted as a time index, we call it \"time index\" below. \n",
+    "* instances: rows with the same \"instance index\" index value correspond to the same instance; rows with different \"instance index\" values correspond to different instances.\n",
```
patrickzib, line 265: "Index index" ? value

sktime/datasets/_data_io.py
```diff
@@ -154,22 +154,23 @@ def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
...
-            for list of mtypes, see datatypes.MTYPE_REGISTER
-            for specifications, see examples/AA_datatypes_and_datasets.ipynb
-        commonly used specifications:
-            "nested_univ: nested pd.DataFrame, pd.Series in cells
-            "numpy3D"/"numpy3d"/"np3D": 3D np.ndarray (instance, variable, time index)
-            "numpy2d"/"np2d"/"numpyflat": 2D np.ndarray (instance, time index)
-            "pd-multiindex": pd.DataFrame with 2-level (instance, time) MultiIndex
-        Exception is raised if the data cannot be stored in the requested type.
-    extract_path : todo author: please fill in docstring
+    return_data_type : str, optional, default = "nested_univ"
```
patrickzib, line 157: default is wrong.should be `none`

sktime/datasets/_data_io.py
```diff
@@ -154,22 +154,23 @@ def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
     return_X_y: bool, optional (default=True)
```
patrickzib, line 154:
```suggestion
    return_X_y: bool, optional (default=True)
```

There is no default? (it is not true)

sktime/datasets/_data_io.py
```diff
@@ -154,22 +154,23 @@ def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
...
-    extract_path : todo author: please fill in docstring
+    return_data_type : str, optional, default = "nested_univ"
+        "numpy3D"/"numpy3d"/"np3D": recommended for equal length series
+        "numpy2D"/"numpy2d"/"np2d": can be used for univariate equal length series,
+        although we recommend numpy3d, because some transformers do not work with
+        numpy2d.
+        "nested_univ": nested pd.DataFrame, pd.Series in cells, use for unequal
+        length series. There other options, see datatypes.SCITYPE_REGISTER, but these
+        will not be supported longterm.
+    extract_path : path of the location for the data file.
```
patrickzib, line 165: default=None

patrickzib (commented):
Some default types in `_data_io.py` do not seem to match. Might be worth checking? Found some at least.

examples/02_classification.ipynb
TonyBagnall: I'm completely reworking this note book in this https://github.com/scikit-time/scikit-time/pull/123 so I'll leave that until then

examples/AA_datatypes_and_datasets.ipynb
TonyBagnall: dont get me started on this notebook .... job for a different PR

sktime/datasets/_data_io.py
TonyBagnall: good spot, fixed

sktime/datasets/_data_io.py
TonyBagnall: fixed

TonyBagnall:
thanks for the review @patrickzib , think I have addressed all issues now

patrickzib (approved):
well done. thank you.

PR #125, opened 2023-02-28
[ENH] Experimental: TDE using SFA_FAST

patrickzib:
This is PR refactors TDE to use SFA_FAST.

- Added a new class `_tde_fast.py` with `FastTemporalDictionaryEnsemble`
- Removed Multivariate support for now
- Runtime still not optimal


patrickzib:
Histogram Intersection is expensive...

image

TonyBagnall:
this is interesting, but should it be a draft PR?

patrickzib:
Yes, a deaft would be better. but does a draft run ci?

MatthewMiddlehurst:
Ah didnt actually mean to push that, was doing some tidying using this as a base 🙂

TonyBagnall:
conflicted and replaced by #3625 I think closing this is the right thing, can revisit this when we redesign SFA

PR #127, opened 2023-02-28
[ENH] Remove plotting

TonyBagnall:
removes the plotting package dumped in classification from _contrib. These plots can form a part of the new package to benchmarking/comparison, but there is no need for this to be in the classification module atm.


TonyBagnall:
@MatthewMiddlehurst

MatthewMiddlehurst (approved):
This was never meant to go in the package proper. We can reintroduce it later.

PR #135, opened 2023-03-01
[ENH]Update Channel Selection code

### LINKED ISSUE #107: [ENH] Improved version of channel_selection.py transformers
the channel selection algorithm is the only code now dependent on the soon to go (I hope) dists_kernels package. It was contributed by Bhaskar (@haskarb) and they have recently extended the work, published here

https://link.springer.com/article/10.1007/s10618-022-00909-1

Bhaskar has agreed to contribute these improvements, and at the same time remove the dependency on dists_kerns :) Assigned myself to indicate this issue is not up for grabs

haskarb:
#### Reference Issues/PRs

Fixes #107 

Updated channel selection code.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.


sktime/transformations/panel/channel_selection.py
```diff
@@ -45,70 +43,127 @@ def _detect_knee_point(values, indices):
...
 
         return distance_frame
 
 
-class _shrunk_centroid:
+def _clip(x, low_value, high_value):
+    return np.clip(x, low_value, high_value)
+
+
+class Class_Prototype:
```
lmmentel, line 83: I wonder if there would be a more descriptive name for the class? The word `Class_` in the class name is a bit redundant, Could you propose an alternative that would maybe be more meaningful here? Happy to help in case you're stuck but I would need a bit more context about the class.

sktime/transformations/panel/channel_selection.py
```diff
@@ -45,70 +43,127 @@ def _detect_knee_point(values, indices):
...
     """Create centroid."""
 
-    def __init__(self, shrink=0):
-        self.shrink = shrink
+    def __init__(self, cp: str = "mean", mean_centering: bool = False):
```
lmmentel, line 86: Please add a docstring documenting `cp` and `mean_centering` arguments.

sktime/transformations/panel/channel_selection.py
```diff
@@ -45,70 +43,127 @@ def _detect_knee_point(values, indices):
...
+        self.cp = cp
+        self.mean_centering = mean_centering
+
+        assert self.cp in ["mean", "median", "mad"], "Class prototype not supported."
+
+    def _mad_median(self, class_X, median=None):
```
lmmentel, line 92: Please add a docstring and it would be great if you could add type hints for the arguments.

sktime/transformations/panel/channel_selection.py
```diff
@@ -45,70 +43,127 @@ def _detect_knee_point(values, indices):
...
+        low_value = median - _mad * 0.50
+        high_value = median + _mad * 0.50
+        # clip = lambda x: np.clip(x, low_value, high_value)
+        class_X = np.apply_along_axis(
+            _clip, axis=1, arr=class_X, low_value=low_value, high_value=high_value
+        )
+
+        return np.mean(class_X, axis=0), class_X
+
+    def _class_mad(self, X, y):
```
lmmentel, line 104: Missing docstring and type annotations.

sktime/transformations/panel/channel_selection.py
haskarb: Actually it's an implementation from paper.

> The proposed channel selection techniques depend on three components: class prototype, distance matrix and distance aggregation (Fig. [2](https://link.springer.com/article/10.1007/s10618-022-00909-1#Fig2)).

sktime/transformations/panel/channel_selection.py
```diff
@@ -192,27 +248,19 @@ class pair across each channel.
         # can the transformer handle unequal length time series (if passed Panel)?
     }
 
-    def __init__(self, distance=None):
-
+    def __init__(
```
lmmentel, line 251: Any chance you could add type hints to the arguments?

sktime/transformations/panel/channel_selection.py
```diff
@@ -227,29 +275,20 @@ def _fit(self, X, y):
...
-
-        idx = _detect_knee_point(distance, indices)[0]
-
-        self.channels_selected_ = idx
-        self.channels_selected_idx_ = [X.columns[i] for i in idx]
+        centroid_obj = Class_Prototype(
+            cp=self.class_prototype,
+            mean_centering=self.mean_centering,
+        )
+        self.class_prototype_ = centroid_obj.create_class_prototype(X.copy(), y)
```
lmmentel, line 282: It seems there a bit of a name clash here, since you have `Class_Prototype`, `self.class_prototype` and `create_class_prototype`, would it be possible to come up with unique names that reflect what the objects are and what they do?

sktime/transformations/panel/channel_selection.py
```diff
@@ -312,25 +320,45 @@ class ElbowClassPairwise(BaseTransformer):
     Note: Channels, variables, dimensions, features are used interchangeably in
     literature.
 
+    Parameters
+    ----------
+    distance : str
+        Distance metric to use for creating the class prototype.
```
lmmentel, line 326: Could you add what are the allowed values of `distance`?

sktime/transformations/panel/channel_selection.py
```diff
@@ -312,25 +320,45 @@ class ElbowClassPairwise(BaseTransformer):
...
+        Default: 'euclidean'
+    class_prototype : str
```
lmmentel, line 328: Could you add what are the allowed values of `class_prototype`?

sktime/transformations/panel/channel_selection.py
```diff
@@ -267,38 +306,7 @@ def _transform(self, X, y=None):
...
-        params2 = {"distance": DtwDist()}
-
-        # with string shorthand
-        params3 = {"distance": "cosine"}
-
-        return [params1, params2, params3]
+        return X.iloc[:, self.channels_selected_idx]
 
 
 class ElbowClassPairwise(BaseTransformer):
```
lmmentel, line 312: Consider dropping the `Class` from the class name since it's redundant

sktime/transformations/panel/channel_selection.py
```diff
@@ -359,9 +387,27 @@ class ElbowClassPairwise(BaseTransformer):
         # can the transformer handle unequal length time series (if passed Panel)?
     }
 
-    def __init__(self):
+    def __init__(
+        self,
+        distance="euclidean",
```
lmmentel, line 392: Any chance you could add type hints?

sktime/transformations/panel/channel_selection.py
```diff
@@ -359,9 +387,27 @@ class ElbowClassPairwise(BaseTransformer):
...
+        mean_centering=False,
+    ):
+        self.distance = distance
+        self.class_prototype = class_prototype
+        self.mean_centering = mean_centering
+        self._is_fitted = False
+
         super(ElbowClassPairwise, self).__init__()
 
+    def _rank(self):
```
lmmentel, line 403: missing docstring

sktime/transformations/panel/channel_selection.py
```diff
@@ -377,21 +423,32 @@ def _fit(self, X, y):
...
+        )  # Centroid created here
+        obj = _distance_matrix(distance=self.distance)
+        self.distance_frame = obj.create_distance_matrix(
+            self.class_prototype_.copy()
+        )  # Distance matrix created here
+
+        all_chs = np.empty(
+            self.class_prototype_.shape[1] - 1
+        )  # -1 for removing class columsn
+        all_chs.fill(0)
```
lmmentel, line 440: could this be simplified with `np.zeros(self.class_prototype_.shape[1] - 1)` ?

lmmentel (changes requested):

haskarb:
Thanks @lmmentel. I will update the docstring and type hints. However, I'm not sure about changing name of classes, as they refer to class in the dataset, and referenced paper.

TonyBagnall (changes requested):
so this is a great start but will need a bit of tweaking to go in, sorry. I think ultimately we can integrate this with our clustering and distances, but I'm ok for this to go in when ready without that if others are also. As for the name, I agree with @lmmentel that the name would be better changed. The problem is overloading Class == possible values of the target variable and Class as in a template for an object in code. Obviously this is in the algorithm name ECP and I dont think you should change that, since its published name. I'll have a think, but at the minimum, dont mix snake and camel class, 
Class_Prototype -> ClassPrototype
but possibly LabelPrototype
and also can it be a private class? If not, could you add a more descriptive doc string?

TonyBagnall:
can you change this to use numpy3d in shape (n_instances, n_dimensions, series_length)? We are moving to remove nested_univ and pandas from classification related code

sktime/transformations/panel/channel_selection.py
haskarb: Removed this piece of code.

sktime/transformations/panel/channel_selection.py
haskarb: This is fixed.

sktime/transformations/panel/channel_selection.py
haskarb: I'm using sktime's distance function, so if a distance is out of the list, they error should pop from there.

TonyBagnall (dismissed):
great job, thanks

TonyBagnall:
I'm happy with this @lmmentel would you like to review again?

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
+        # return all dimensions if no elbow point is found
+        return indices
+    return best_dims
 
-    return (best_dims,)
 
+def create_distance_matrix(
+    prototype: Union[pd.DataFrame, np.ndarray],
+    class_vals: np.array,
+    distance_: str = "euclidean",
```
lmmentel, line 48: the underscore after `distance` could be dropped, any reason to have it? in scikit-learn it's used to indicate computed attrbutes but it does not seem like the case here

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
+) -> pd.DataFrame:
+    """Create a distance matrix between class_prototypes.
```
lmmentel, line 50:
```suggestion
    """Create a distance matrix between class prototypes.
```

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
 
-class _distance_matrix:
-    """Create distance matrix."""
+    Parameters
+    ----------
+    prototype : pd.DataFrame or np.ndarray
```
lmmentel, line 54: `prototype` and `class` are very generic names that do not say much about what the parameters should be. Could you add an extra sentence to each docstring explaining the meaning or give an example value?

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
-            class_pair = []
-            # calculate the distance of centroid here
-            for _, (q, t) in enumerate(
-                zip(
-                    centroid_frame.drop(["class_vals"], axis=1).iloc[class_[0], :],
-                    centroid_frame.iloc[class_[1], :],
+    """
+    assert prototype.shape[0] == len(
+        class_vals
+    ), "Prototype and class values must be of same length."
```
lmmentel, line 69: Assets are dangerous in production since they can be switched by an interpreter flag, consider changing to a if/except block
```suggestion
    if prototype.shape[0] != len(class_vals):
        raise ValueError(f"Prototype {prototype.shape[0]} and class values {len(class_vals)} must be of same length.")
```

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
-                # print(class_[0])
-
-            distance_frame = pd.concat([distance_frame, pd.DataFrame(dict_)], axis=1)
+            dict_ = {f"Centroid_{idx_class[cls_[0]]}_{idx_class[cls_[1]]}": dis}
+        distance_frame = pd.concat([distance_frame, pd.DataFrame(dict_)], axis=1)
+    return distance_frame
 
-        return distance_frame
 
+def _clip(x: np.array, low_value: float, high_value: float):
```
lmmentel, line 101: this function seems redundant since it call a numpy function directly, why not use `np.clip` instead?

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
+    return np.clip(x, low_value, high_value)
 
-class _shrunk_centroid:
-    """Create centroid."""
 
-    def __init__(self, shrink=0):
-        self.shrink = shrink
+class ClassPrototype:
+    """
+    Class prototype for each class.
```
lmmentel, line 107: this is a bit too self referential, any chance you could some clarification?

sktime/transformations/panel/channel_selection.py
```diff
@@ -43,81 +37,189 @@ def _detect_knee_point(values, indices):
...
+        mean_centering: bool = False,
+    ):
+        self.prototype_type = prototype_type
+        self.mean_centering = mean_centering
+
+        assert self.prototype_type in [
+            "mean",
+            "median",
+            "mad",
+        ], "Class prototype not supported."
```
lmmentel, line 136: consider switching to if/except since assert might be skipped in production

lmmentel (changes requested):
Overall great progress, although it would be great to have some of the new comments addressed before merging

sktime/transformations/panel/channel_selection.py
haskarb: There is a function in sktime which I'm using, and it could conflict with that.

sktime/transformations/panel/channel_selection.py
TonyBagnall: I think this class should be private

sktime/transformations/panel/channel_selection.py
lmmentel: How about changing the import to `from sktime.distances import distance as sktime_distance` that way we can avoid unnecessary underscore.

haskarb: ok. I can do that.

sktime/transformations/panel/channel_selection.py
haskarb: I want to keep this public because I thought it could be used to create just a prototype frame.

sktime/transformations/panel/channel_selection.py
haskarb: Will do this.

sktime/transformations/panel/channel_selection.py
haskarb: Updated the doc string.

sktime/transformations/panel/channel_selection.py
haskarb: Fixed.

lmmentel (approved):
Looks great!

PR #139, opened 2023-03-02
[ENH] Speedup EAgglo by factor 5-10x

patrickzib:
### EAgglo
This PR improves the runtime of EAgglo by a factor of 5-10, by use of:

- NUMBA
- Reimplement Euclidean Distance (instead of scipy cdist)
- Reworked initial distance computations, if `self._member is None` (no groupings + Pandas needed)

### Experiments:

```python
from sktime.datasets import load_gun_point_segmentation
ts, period_size, cps = load_gun_point_segmentation()

# compute a segmentation
model = EAgglo(penalty=lambda x: np.mean(np.diff(np.sort(ts))))
fitted_model = model._fit(pd.DataFrame(ts[:2000]))
```

### After:  12 seconds

### DIFF
--- a/sktime/annotation/tests/test_eagglo.py
+++ b/sktime/annotation/tests/test_eagglo.py
@@ -136,6 +136,6 @@ def test_custom_penalty():
 #     from sktime.datasets import load_gun_point_segmentation
 #     ts, period_size, cps = load_gun_point_segmentation()
 #
-#     # compute a ClaSP segmentation
+#     # compute a segmentation
 #     model = EAgglo(penalty=lambda x: np.mean(np.diff(np.sort(ts))))
 #     fitted_model = model._fit(pd.DataFrame(ts[:2000]))
--- a/sktime/annotation/tests/test_eagglo.py
+++ b/sktime/annotation/tests/test_eagglo.py
@@ -139,7 +139,7 @@ def test_eagglo_large():
 
     # compute a ClaSP segmentation
     model = EAgglo(penalty=lambda x: np.mean(np.diff(np.sort(ts))))
-    fitted_model = model._fit(pd.DataFrame(ts[:500]))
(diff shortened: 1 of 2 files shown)


sktime/annotation/tests/test_eagglo.py
```diff
@@ -126,3 +126,16 @@ def test_custom_penalty():
 
     assert np.allclose(cluster_actual, cluster_expected)
     assert np.allclose(fit_actual, fit_expected)
+
+
+# def test_eagglo_large():
```
lmmentel, line 131: Is the commented test needed?

lmmentel (commented):
Left a small comment but all looks good!

TonyBagnall:
superb, but could this use our numba Euclidean distance in distances rather than reimplement? @chrisholder

sktime/annotation/tests/test_eagglo.py
patrickzib: Thanks. Only for benchmarking. Can remove it, once we finalize the PR :)

patrickzib:
> superb, but could this use our numba Euclidean distance in distances rather than reimplement? @chrisholder

@TonyBagnall tbh: I do not understand how :)

The method I found is not numba`rized
> from sktime.distances._distance import euclidean_distance
> def euclidean_distance(x: np.ndarray, y: np.ndarray, **kwargs: Any) -> float:

Thus, I am not able to use it in a numba'rized function. 
I.e. to use it, I have to remove the decorators from the functions and the runtime increases to 35 seconds (from 15 seconds without).

sktime/annotation/tests/test_eagglo.py
patrickzib: Removed :)

TonyBagnall:
we need a distances notebook @chrisholder :) Will sort it tomorrow

lmmentel (approved):

Issue #141, opened 2023-03-02
[BUG] `NumbaDistance` is potentially slow: Numba recompiles closures upon every call


patrickzib:
I tried to profile Elastic Ensemble to find out why it is so slow in CI.

I suspect the following: There seems to be an issue with numba and nested functions (called closures). Unfortunately, Numba complies the nested function upon every call: 

https://github.com/numba/numba/issues/1810
https://github.com/pola-rs/polars/issues/5872

This causes numba to recompile, i.e. the DTW `numba_dtw_distance()` closure, every time it is invoked. 
**That means for every pair of time series, the closure is compiled first.**

https://github.com/scikit-time/scikit-time/blob/6fd4a729b047ff7974823e1cab6bcd7d3b66b4d2/sktime/distances/_dtw.py#L193
```python
 def _distance_factory(
              [....]
    ) -> DistanceCallable:
         [....]
        @njit(cache=True)
        def numba_dtw_distance(
            _x: np.ndarray,
            _y: np.ndarray,
        ) -> float:
            cost_matrix = _cost_matrix(_x, _y, _bounding_matrix)
            return cost_matrix[-1, -1]

        return numba_dtw_distance
```

I do not have an instant solution for this, but it is a huge bottleneck.

patrickzib:
WHhile profiling, it seems like 90% of the runtime is spent compiling:
<img width="130" alt="image" src="https://user-images.githubusercontent.com/7783034/222537267-30856807-cf1a-435b-ba22-9b87d32ec06c.png">

patrickzib:
Compile is called 4200 times: 
<img width="1102" alt="image" src="https://user-images.githubusercontent.com/7783034/222538967-4352f234-fcce-4e28-9ef9-2962fefe6bdf.png">

TonyBagnall:
I think this is a problem, I thought it was solved previously. One solution we had was to use the distance factory, which builds once I think. @chrisholder can comment much better than me

from sktime.distances import distance_factory

        if isinstance(self.distance, str):
            if self.distance_params is None:
                self.metric_ = distance_factory(X[0], X[0], metric=self.distance)
            else:
                self.metric_ = distance_factory(
                    X[0], X[0], metric=self.distance, **self.distance_params
                )

patrickzib:
[...]

PR #142, opened 2023-03-02
[ENH] [BUG] Improving Runtime of Elastic Ensemble by 20x (by fixing the closure issue)

patrickzib:
Elastic Ensemble was among the slowest methods in CI. 

The root cause are closures in numba, which are recompiled upon every call.

This PR fixes the numba closure issue.

### DIFF
--- a/sktime/distances/_distance.py
+++ b/sktime/distances/_distance.py
@@ -1939,6 +1939,8 @@ def distance_factory(
         If a resolved metric is not no_python compiled.
         If the metric type cannot be determined.
     """
+    global dist_callable
+
     if x is None:
         x = np.zeros((1, 10))
     if y is None:
--- a/sktime/distances/_distance.py
+++ b/sktime/distances/_distance.py
@@ -1948,7 +1948,8 @@ def distance_factory(
 
     callable = _resolve_metric_to_factory(metric, _x, _y, _METRIC_INFOS, **kwargs)
 
-    @njit(cache=True)
+    # TODO Not sure why, but removing this @njit, avoids recompiling the closures
+    # @njit(cache=True)
     def dist_callable(x: np.ndarray, y: np.ndarray):
         _x = _numba_to_timeseries(x)
         _y = _numba_to_timeseries(y)
--- a/sktime/distances/_dtw.py
+++ b/sktime/distances/_dtw.py
@@ -68,7 +68,7 @@ class _DtwDistance(NumbaDistance):
         window: float = None,
         itakura_max_slope: float = None,
         bounding_matrix: np.ndarray = None,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> DistanceAlignmentPathCallable:
         """Create a no_python compiled dtw path distance callable.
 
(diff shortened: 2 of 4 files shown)


patrickzib:
Removing the single njit here, causes a runtime improve of 20-30x for me locally.

https://github.com/scikit-time/scikit-time/blob/c73f14422e38abbdb9f541145e461d272a10c312/sktime/distances/_distance.py#L1951

patrickzib:
> see this #141

Unfortunately, the distance factory is already part of the code in ElacstocEnsemble

patrickzib:
This PR fixes the issue.

Before (Windows 3.9) :
image


After (Windows 3.9):
image


Elastic Ensemble is no longer in the list of slowest 10 methods.

chrisholder:
Ill test EE classifier on the new distances branch and check the performance now though.

patrickzib:
Hi @chrisholder , maybe this is a misconception :) The returned function is still numba-compiled. But the method to return the numba-compiled function is not. This is why it is not breaking anything and we do not have the closure problem any longer

patrickzib:
I.e. `callable(_x, _y)` is still numba-compiled. But the method to return this callable is not any longer.

chrisholder:
Sorry, i dont think I understand your code then.

What is this line doing:

    global dist_callable

chrisholder:
But this

`dist_callable = distance_factory(x, y, metric="dtw")`

This no longer would be a numba distance?

patrickzib:
> But this
> 
> `dist_callable = distance_factory(x, y, metric="dtw")`
> 
> This no longer would be a numba distance?

Were is or called like this?

patrickzib:
> Ill test EE classifier on the new distances branch and check the performance now though.

I am happy to go for your changes, once ready. In the mean time, we could go for this small change though, as it seems to work just fine?

chrisholder:
So lets say I had

```
def example(x, y):
    # With this change this gives a python function
    dist = distance_factory(x, y, metric="dtw")
    
    @njit()
    def some_numba_function(ts1, ts2):
        return dist(ts1, ts2)
```

From my understanding (and sorry if im wrong!) the change would cause this to throw an error because you cant call a non numba function from inside a numba function.

chrisholder:
Sure im happy for this to go in since there are changes coming soon for the distance module and this in the mean time will greatly improve development speed!

chrisholder (approved):
New changes for distance module coming soon but this is a good fix to improve dev speed for the package while waiting for those changes.

patrickzib:
> My opinion on it if you want speed (which I think we do) you just say give it a 2d array or it wont work and then you could delete this additional compilation:
> 
> ```
> @njit()
>   def dist_callable(x: np.ndarray, y: np.ndarray):
>       _x = _numba_to_timeseries(x)
>       _y = _numba_to_timeseries(y)
>       return callable(_x, _y)
> ```
> 
> and just return callable instead

sure! Totally agree. This was not adressed by this PR though, which only solves the numba compilation of the closure on every call as done in EE

patrickzib:
> So lets say I had
> 
> ```
> def example(x, y):
>     # With this change this gives a python function
>     dist = distance_factory(x, y, metric="dtw")
>     
>     @njit()
>     def some_numba_function(ts1, ts2):
>         return dist(ts1, ts2)
> ```


Thank you for approval. Will wait for merging to check this use case. I am interested in the result, too, bit do not have a PC right now :)

Would be great to hear from the results of EE in your branch, too.

PR #143, opened 2023-03-03
Remove the DistanceFeatures transformer

TonyBagnall:
uses Dists_kerns, so remove for now, bring back with numpy later see #144


patrickzib (approved):
Looks good to me. Doesn't even seems to be referenced in any documentation...

Issue #146, opened 2023-03-03
[BUG] scittype:transform tags in BaseTransformer 


TonyBagnall:
**Describe the bug**
Also see here

https://github.com/scikit-time/scikit-time/issues/100

every single transformer has these tags
```python
        "scitype:transform-input": "Series",
        # what is the scitype of X: Series, or Panel
        "scitype:transform-output": "Series",
        # what scitype is returned: Primitives, Series, Panel
```

I have never understood what they meant, since all the panel transformers take Panel input (i.e. a collection of time series). Further digging reveals this

```python
Type and behaviour of transformer is determined by the following tags:
    "scitype:transform-input" tag with values "Primitives" or "Series"
        this determines expected type of input of transform
        if "Primitives", expected inputs X are pd.DataFrame
        if "Series", expected inputs X are Series or Panel
        Note: placeholder tag for upwards compatibility
            currently only "Series" is supported
```

this is next level confusion. Introduce a string variable with a weird "scitype" name, set it to "Series" in base class, which most people wont understand, then set it again to "Series" in every subclass to make sure people are confused, then never use it. I propose just removing this tag all together. 
        "scitype:transform-output": "Series" is listed as having this effect,
 
        type depends on type of X and scitype:transform-output tag:
            |          | `transform`  |                        |
            |   `X`    |  `-output`   |     type of return     |
            |----------|--------------|------------------------|
            | `Series` | `Primitives` | `pd.DataFrame` (1-row) |
            | `Panel`  | `Primitives` | `pd.DataFrame`         |
            | `Series` | `Series`     | `Series`               |
            | `Panel`  | `Series`     | `Panel`                |
            | `Series` | `Panel`      | `Panel`                |

used in 
```python
   def _convert_output(self, X, metadata, inverse=False):
```
 which is called here
```python
        # convert to output mtype
        if not hasattr(self, "_output_convert") or self._output_convert == "auto":
            X_out = self._convert_output(Xt, metadata=metadata)
        else:
            X_out = Xt
```
which is set in base constructor 

```python
    def __init__(self, _output_convert="auto"):
        self._converter_store_X = dict()  # storage dictionary for in/output conversion
        self._output_convert = _output_convert
```
setting a private attribute through constructor seems weird. Is any of this necessary? Can we just remove the output conversion? Or at least specify actual data structures instead of "Series" and "Panel", the above table is meaningless to me. 

I suggest we
1) test setting _output_convert to false by default. If its a parameter, then it should work
[...]

Issue #100, opened 2023-02-21
[ENH] Improve the efficiency of the Transformer base class


TonyBagnall:
**Describe the bug**

I've had a look at the Transformation module, which tbh I have not looked at much since it was completely refactored. tl;dr: its is bloated, hard to understand and very inefficient. I think we should design around it then remove it. So this is my journey through. Lets suppose I build a ShapeletTransformClassifier which extends BaseClassifier. This consists of a pipeline of ShapeletTransform (extends BaseTransform) and RotationForest (a scikit like classifier in our package). 

```python
from sktime.classification.shapelet_based import ShapeletTransformClassifier
from sktime.datasets import load_unit_test

trainx, trainy=load_unit_test(return_type="numpy2d", split="train")
stc = ShapeletTransformClassifier()
stc.fit(trainx,trainy)
```
fit goes to the BaseClassifier fit. This checks the input and finds the meta data, which includes info on data characteristics (number of dimensions, has missing values, unequal length).  

```python
        X, y = self._internal_convert(X, y)
        X_metadata = self._check_classifier_input(X, y)
        missing = X_metadata["has_nans"]
        multivariate = not X_metadata["is_univariate"]
        unequal = not X_metadata["is_equal_length"]
        self._X_metadata = X_metadata
        # Convert data as dictated by the classifier tags
        X = self._convert_X(X)

        # Check this classifier can handle characteristics
        self._check_capabilities(missing, multivariate, unequal)
```
this necessarily requires looking at all data, so if the data has n instances, d dimensions and series length m, this is an O(ndm) operation. Fair enough, a price worth paying I think to check input. Data is also converted here if necessary. Then the private fit is called in ShapeletTransformClassifier. This basically does this (with some other bits), where transformer is ShapeletTransform and estimator is rotation forest. 

```python
        X_t = self._transformer.fit_transform(X, y).to_numpy()
        self._estimator.fit(X_t, y)
```

fit_transform goes to BaseTransformer fit_transform, which calls BaseTransformer fit() then transform(). It gets messy now. fit calls this

 ```python
        # check and convert X/y
        X_inner, y_inner = self._check_X_y(X=X, y=y)
```
now _check_X_y is a 250 line function with this docstring

 ```python
       Returns
        -------
        X_inner : Series, Panel, or Hierarchical object, or VectorizedDF
                compatible with self.get_tag("X_inner_mtype") format
            Case 1: self.get_tag("X_inner_mtype") supports scitype of X, then
[...]

Issue #37, opened 2023-02-07
[ENH] Investigate/improve performance of data checks and pd.MultiIndex operations/iterations


aiwalter:
**Is your feature request related to a problem? Please describe.**
There seems to be quite a problem related to performance of some data checks and `pd.MultiIndex` operations.

**Describe the solution you'd like**
Related: https://github.com/sktime/sktime/issues/4139

TonyBagnall:
does anyone use pd.MultiIndex with forecasting? Its not relevant for classification/regression/clustering

aiwalter:
`MultiIndex` is all over the place for using forecasting/transformations with panel data. And such panel data is very common in industry, we have it also a lot at my day job and there I see severe performance issues with it.

aiwalter:
@ltsaprounis was sharing this link in Slack: https://scikit-learn.org/stable/developers/performance.html#profiling-python-code

it could be a good start point.

aiwalter:
possibly we have to implement some config to store info that some checks have already been done and dont need to be repeated X times again.
We could have a look at newly introduced config from `scikit-learn`: https://scikit-learn.org/dev/modules/generated/sklearn.set_config.html#sklearn.set_config

aiwalter:
todo: additionally improve check exception message. Currently the checks are called in base classes and therefore its not possible to see for user directly in which estimator the check was failing. This could also be improved probably by handing over the parent class name(s) or tracking that automatically with some inspect magic?

PR #147, opened 2023-03-03
[ENH] Include flag to avoid all the conversion calls in BaseTransformer

TonyBagnall:
This is a quick fix to see if we can just circumvent the madness that is the BaseTransformer conversion in check_X_y

See #146 #100 #37 

this is not a long term solution, but I do not have a grasp of the conversion issues for forecasting and this is the easiest way to experiment. For the tsml side of things, all of this is unnecessary and we should move over to use numpy almost exclusively. I will later test what actually happens with our transformers when we use this tag. Not this skips all the so called "vectorization" which is also not needed at all in tsml

skips this 
image
by doing this
image


TonyBagnall:
thinking more, I would probably add conversions based on the "X_inner_type", this at least matches the classification model, where the estimator can specify the internal conversion without reference to concepts like "scitype"

aiwalter:
I think sth like that can be ideally done with the config, either local config or global config, see #148 . The term "passthrough" is used in `sklearn` to not apply any transformation. Als this is what we do in `OptionalPassthough`, so terminology wise skipping a check should probably have another name then.

TonyBagnall:
closing this for now, because its under the old structure and has a conflict, will revisit this idea later once nested_univ sorted

PR #148, opened 2023-03-04
[ENH] Added config adapted from sklearn

aiwalter:
Draft PR for having an sklearn-like config. This was copied 1:1 from new sklearn config.

### DIFF
--- a/sktime/__init__.py
+++ b/sktime/__init__.py
@@ -4,6 +4,7 @@
 
 __version__ = "0.16.0"
 
-__all__ = ["show_versions"]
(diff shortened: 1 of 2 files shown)


lmmentel:
Thanks, any chance you could provide some more info on why this is needed and how it would be used in our case?

aiwalter:
wanted to share this, might be useful for what Tony tried to do in #147 so he can also use this PR for experimenting.

I am still running some experiments to see how we can avoid calling some data check code n times where it would be just sufficient to call it once in the beginning - mostly relevant for pipelines and grid searches

Issue #150, opened 2023-03-04
[ENH] Increment pandas up to 2.0.0


aiwalter:
**Is your feature request related to a problem? Please describe.**
Pandas major release is about to come in few days:
https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html

**Describe the solution you'd like**
Getting ready to support `pandas>=2.0.0`

Seems with the current requirement `pandas<1.6.0` we are already good and no other release will come before going to `2.0.0`.
Some pitfalls might be (soft-)dependencies that still require lower versions of `pandas`.

I started to check with `pandas==2.0.0rc0`

PR #153, opened 2023-03-05
[MNT] Pandas 2.0.0 deprecation actions

aiwalter:
First batch of deprecation changes to support `pandas>=2.0.0`.

related #150


aiwalter:
known issue with `fh`, it seems often `fh.to_absolute()` was used as `index` value for `pd.Series` or `pd.DataFrame`. This is not working anymore and raises an exception. Was even wondering how `pandas` was processing that before? All those cases need to be fixed to `index=fh.to_absolute().to_pandas()`

setup.cfg
```diff
@@ -26,8 +26,7 @@ filterwarnings =
     ignore:numpy.ufunc size changed
 
 [flake8]
-# Default flake8 3.5 ignored flags
-ignore = E121, E123, E126, E226, E24, E704, W503, W504
```
MatthewMiddlehurst, line 30: What are we ignoring here? Do we need to?

sktime/utils/validation/_dependencies.py
```diff
@@ -167,7 +166,8 @@ def _check_soft_dependencies(
                 )
 
             # raise error/warning or return False if version is incompatible
-            if pkg_env_version not in package_version_req:
+            # also accept release candidates as it can be useful for testing
+            if not package_version_req.contains(pkg_env_version, prereleases=True):
                 if severity == "error":
```
MatthewMiddlehurst, line 171: Is this something we want past testing the Pandas prerelease?

MatthewMiddlehurst (commented):
I don't see anything wildly out there, but then again I also don't use most of this code 🙂.

sktime/utils/validation/_dependencies.py
aiwalter: I think we can keep it as we wont release anything with a `rc` dependency. So if people want to test in future other `rc` candidates they wont run into same issue as I did

setup.cfg
aiwalter: > B028: No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.

its a new thing since we bumped `flake8`

aiwalter: default is` stacklevel=1`, if you just add it the flake8 is not complaining anymore

aiwalter: we can add it back again, just today I wont have time to fix it so would appreciate to merge this PR

MatthewMiddlehurst: That works as well I guess, the reasoning they give has a good point though.

PR #156, opened 2023-03-06
[MNT] Rename package to aeon

MatthewMiddlehurst:
Renames the base package to `aeon`. There is no doubt this is missing things, getting every case of sktime/scikit-time may take a bit longer.


TonyBagnall:
772 files changed, lol

Issue #28, opened 2023-02-07
[FORK] Rename the project mentions from `sktime` to new package name


MatthewMiddlehurst:
This includes:
- documentation 
- README #43 
- configuration files: pyproject.toml etc.

MatthewMiddlehurst:
Still deciding on a new project name, see #70

PR #157, opened 2023-03-06
[MNT] Update name to `aeon` in config files

MatthewMiddlehurst:
Renames `scikit-time` to `aeon`. Package is still called `sktime` so that remains the same. Part of #28.


aiwalter:
could we go for lower case name instead `Aeon`?

MatthewMiddlehurst:
should all be lowercase now.

TonyBagnall (approved):

PR #160, opened 2023-03-06
[ENH] Adds WEASEL v2 (with dilation)

patrickzib:
This PR adds WEASEL v2 (with dilation).

This is the currently most accurate dictionary-based classifier.

### DIFF
--- a/sktime/classification/dictionary_based/_weasel_v2.py
+++ b/sktime/classification/dictionary_based/_weasel_v2.py
@@ -1,0 +1,457 @@
+# -*- coding: utf-8 -*-
+"""WEASEL 2.0 classifier.
+
+A Random Dilated Dictionary Transform for Fast, Accurate and Constrained Memory
+Time Series Classification.
+
+"""
+
+__author__ = ["patrickzib"]
+__all__ = ["WEASEL_V2"]
+
+import numpy as np
+from joblib import Parallel, delayed
+from scipy.sparse import hstack
+from sklearn.linear_model import LogisticRegression, RidgeClassifierCV
+from sklearn.utils import check_random_state
+
+from sktime.classification.base import BaseClassifier
+from sktime.transformations.panel.dictionary_based import SFAFast
+
+
+class WEASEL_V2(BaseClassifier):
+    """Word Extraction for Time Series Classification (WEASEL) v2.0.
+
+    Overview: Input 'n' series length 'm'
+    WEASEL is a dictionary classifier that builds a bag-of-patterns using SFA
+    for different window lengths and learns a logistic regression classifier
+    on this bag.
+
+    WEASEL 2.0 has three key parameters that are automcatically set based on the
+    length of the time series:
+    (1) Minimal window length: Typically defaulted to 4
+    (2) Maximal window length: Typically chosen from
+        24, 44 or 84 depending on the time series length.
+    (3) Ensemble size: Typically chosen from 50, 100, 150, to derive
+        a feature vector of roughly 20𝑘 up to 70𝑘 features (distinct words).
+
+    From the other parameters passed, WEASEL chosen random values for each set
+    of configurations. E.g. for each of 150 configurations, a random value is chosen
+    from the below options.
+
+    Parameters
+    ----------
+    min_window : int, default=4,
+        Minimal length of the subsequences to compute words from.
+    norm_options : array of bool, default=[False],
+        If the array contains True, words are computed over mean-normed TS
+        If the array contains False, words are computed over raw TS
+        If both are set, words are computed for both.
+        A value will be randomly chosen for each parameter-configuration.
+    word_lengths : array of int, default=[7, 8],
+        Length of the words to compute. A value will be randomly chosen for each
+        parameter-configuration.
+    use_first_differences: array of bool, default=[True, False],
+        If the array contains True, words are computed over first order differences.
+        If the array contains False, words are computed over the raw time series.
+        If both are set, words are computed for both.
+    feature_selection: {"chi2_top_k", "none", "random"}, default: chi2_top_k
+        Sets the feature selections strategy to be used. Large amounts of memory may be
+        needed depending on the setting of bigrams (true is more) or
+        alpha (larger is more).
+        'chi2_top_k' reduces the number of words to at most 'max_feature_count',
+        dropping values based on p-value.
+        'random' reduces the number to at most 'max_feature_count',
+        by randomly selecting features.
+        'none' does not apply any feature selection and yields large bag of words
+    max_feature_count : int, default=30_000
+       size of the dictionary - number of words to use - if feature_selection set to
+       "chi2" or "random". Else ignored.
+    support_probabilities: bool, default: False
+        If set to False, a RidgeClassifierCV will be trained, which has higher accuracy
+        and is faster, yet does not support predict_proba.
+        If set to True, a LogisticRegression will be trained, which does support
+        predict_proba(), yet is slower and typically less accurate. predict_proba() is
+        needed for example in Early-Classification like TEASER.
+    n_jobs : int, default=4
+        The number of jobs to run in parallel for both `fit` and `predict`.
+        ``-1`` means using all processors.
+    random_state: int or None, default=None
+        Seed for random, integer
+
+    Attributes
+    ----------
+    n_classes_ : int
+        The number of classes.
+    classes_ : list
+        The classes labels.
+
+    See Also
+    --------
+    MUSE
+
+    References
+    ----------
+    .. [1] Patrick Schäfer and Ulf Leser, "WEASEL 2.0 -- A Random Dilated Dictionary
+    Transform for Fast, Accurate and Memory Constrained Time Series Classification",
+    Preprint, https://arxiv.org/abs/2301.10194
+
+    Examples
+    --------
+    >>> from sktime.classification.dictionary_based import WEASEL_V2
+    >>> from sktime.datasets import load_unit_test
+    >>> X_train, y_train = load_unit_test(split="train", return_X_y=True)
+    >>> X_test, y_test = load_unit_test(split="test", return_X_y=True)
+    >>> clf = WEASEL_V2()
+    >>> clf.fit(X_train, y_train)
+    WEASEL_V2(...)
+    >>> y_pred = clf.predict(X_test)
+    """
+
+    _tags = {
+        "capability:multithreading": True,
+        "classifier_type": "dictionary",
+    }
+
+    def __init__(
+        self,
+        min_window=4,
+        norm_options=(False),
+        word_lengths=(7, 8),
+        use_first_differences=(True, False),
+        feature_selection="chi2_top_k",
+        max_feature_count=30_000,
+        random_state=None,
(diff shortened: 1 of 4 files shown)


sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,457 @@
...
+        # avoid overfitting with too many features
+        if self.n_instances < 250:
+            self.max_window = 24
+            self.ensemble_size = 50
+        elif self.series_length < 100:
+            self.max_window = 44
+            self.ensemble_size = 100
+        else:
+            self.max_window = 84
+            self.ensemble_size = 150
```
TonyBagnall, line 191: a lot of hard coded constants here, maybe at least make them class constants?

TonyBagnall (commented):
Might be better to have one configurable WEASEL? I know there are two HIVE-COTE, but I would prefer there to just be one really.

patrickzib:
> Might be better to have one configurable WEASEL? I know there are two HIVE-COTE, but I would prefer there to just be one really.

Yes, agree in general. In this case however, this would be like moving Rocket and MiniRocket to the same classifier.

They are on a conceptional level not very similar:
image

Left (weasel 2.0) vs right (weasel)
image

And the pipeline:
image

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,457 @@
...
+        If set to True, a LogisticRegression will be trained, which does support
+        predict_proba(), yet is slower and typically less accurate. predict_proba() is
+        needed for example in Early-Classification like TEASER.
+    n_jobs : int, default=4
+        The number of jobs to run in parallel for both `fit` and `predict`.
+        ``-1`` means using all processors.
+    random_state: int or None, default=None
+        Seed for random, integer
+
+    Attributes
```
TonyBagnall, line 82: dont need this, these are declared in the base class

sktime/classification/dictionary_based/_weasel_v2.py
TonyBagnall: I meant the thresholds, 250 and 100 etc

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,457 @@
...
+            sfa_words.extend(words)
+
+        # merging arrays from different threads
+        if type(sfa_words[0]) is np.ndarray:
+            all_words = np.concatenate(sfa_words, axis=1)
+        else:
+            all_words = hstack(sfa_words)
+
+        if not self.support_probabilities:
+            self.clf = RidgeClassifierCV(alphas=np.logspace(-1, 5, 10))
```
TonyBagnall, line 246: might be good to give the option for user to pass a classifier? I'd be quite interested to try it with rotation forest, for example, particularly after we adapt it. Alternatively you could encapsulate the transform and use a pipeline?

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,457 @@
...
+        parallel_res = Parallel(n_jobs=self.n_jobs, timeout=99999, backend="threading")(
+            delayed(transformer.transform)(XX) for transformer in self.SFA_transformers
+        )
+
+        all_words = []
+        for words in parallel_res:
+            # words = words.astype(np.float32) / norm
+            all_words.append(words)
+
+        # X_features = self.rocket.transform(X)
```
TonyBagnall, line 314: remove comment

TonyBagnall:
> This PR adds WEASEL v2 (with dilation).
> 
> This is the currently most accurate dictionary-based classifier:
> 
> image
> 
> Tests on the UCR datasets
> 
> [1] Patrick Schäfer and Ulf Leser, "WEASEL 2.0 -- A Random Dilated Dictionary Transform for Fast, Accurate and Memory Constrained Time Series Classification", Preprint, https://arxiv.org/abs/2301.10194

hey, that CD is a spoiler! :)

MatthewMiddlehurst:
Not had a look but just to note, we do actually have ROCKET and MiniROCKET in the same classifier 🙂. I would not be against putting extensions in the same class i.e. HC1+HC2, CIF+DrCIF, STSF+R-STSF

patrickzib:
> Not had a look but just to note, we do actually have ROCKET and MiniROCKET in the same classifier 🙂. I would not be against putting extensions in the same class i.e. HC1+HC2, CIF+DrCIF, STSF+R-STSF

The Transform of Rocket and MiniRocket is not in the same class? The pipeline is...

TonyBagnall:
> > Not had a look but just to note, we do actually have ROCKET and MiniROCKET in the same classifier 🙂. I would not be against putting extensions in the same class i.e. HC1+HC2, CIF+DrCIF, STSF+R-STSF
> 
> The Transform of Rocket and MiniRocket is not in the same class? The pipeline classifier is...

WEASEL is a similar structure to rocket and shapelets, pipeline of transformation+classifier. Perhaps the transform would be better as a transformer? Maybe not now, but I think that should be the goal when we sort out transformations ......

TonyBagnall:
> Not had a look but just to note, we do actually have ROCKET and MiniROCKET in the same classifier 🙂. I would not be against putting extensions in the same class i.e. HC1+HC2, CIF+DrCIF, STSF+R-STSF

I think we should discuss at a classification dev meeting

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: Good idea! 

Yes, I have refactored the class and extracted the transform into an own subclass within the same class.
`WEASEL_V2_TRANSFORM`

This is callable from outside, and I have refactored WEASEL v2 accordingly.

Comments:
It is not so trivial to make it a real `BaseTransformer` though. Issues: 
- `y` is not accepted as numpy array
- return type of WEASEL_V2 is a scipy csr_matrix, which is not supported by current transformers?
- conversion of 2d input data
- fit and transform should not be split. It is fastest to run fit and transform in one go.

Thus, I would like to leave this for the future.

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: Removed `n_jobs`.

I could not find `random_state`  in the base class?

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: Sure! Extracted to constants

patrickzib:
> > > Not had a look but just to note, we do actually have ROCKET and MiniROCKET in the same classifier 🙂. I would not be against putting extensions in the same class i.e. HC1+HC2, CIF+DrCIF, STSF+R-STSF
> > 
> > 
> > The Transform of Rocket and MiniRocket is not in the same class? The pipeline classifier is...
> 
> WEASEL is a similar structure to rocket and shapelets, pipeline of transformation+classifier. Perhaps the transform would be better as a transformer? Maybe not now, but I think that should be the goal when we sort out transformations ......

I agree! Would be great to have a BOSS/WEASEL/v2 transformers! 

There are a couple of issues in the current transformer base class to support it though (see also above in my comment):

- `y` is not accepted as input with numpy array?
- return type of WEASEL_V2 is a scipy csr_matrix, which is not supported by current transformers?
- conversion of 2d input data (ongoing issue)
- fit and transform should not be split for WEASEL. It is fastest to run fit and transform in one go.

Thus, I would like to leave this for the future.

patrickzib:
I did factor out the transform part in WEASEL v2 though. It can be called independently now. It is not BaseTransformer compliant, though.

TonyBagnall:
> I did factor out the transform part in WEASEL v2 though. It can be called independently now. It is not BaseTransformer compliant, though.

thats good, do it like this and then raise an issue tagged with transformers, there is a lot to do in that module ....

TonyBagnall:
it will also make the adaptation to regression much easier

patrickzib:
> thats good, do it like this and then raise an issue tagged with transformers, there is a lot to do in that module ....

Sure: https://github.com/aeon-toolkit/aeon/issues/166

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,592 @@
...
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`.
+        """
+        return {
+            "feature_selection": "none",
+            "support_probabilities": True,
+        }
+
+
+class WEASEL_V2_TRANSFORM:
```
TonyBagnall, line 267: set for putting in a transformer, nice one. Just a quibble on the name, its a bit shouty of being a global constant. 
WEASEL_V2_TRANSFORM -> WEASELTransformerV2 maybe?

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,592 @@
...
+        'none' does not apply any feature selection and yields large bag of words
+    max_feature_count : int, default=30_000
+       size of the dictionary - number of words to use - if feature_selection set to
+       "chi2" or "random". Else ignored.
+    random_state: int or None, default=None
+        Seed for random, integer
+    """
+
+    # _tags = {
+    #     "univariate-only": True,
```
TonyBagnall, line 315: we have not tried this on multivariate, could you roll MUSE functionality into here and make it multivariate capable?

sktime/classification/dictionary_based/_weasel_v2.py
```diff
@@ -0,0 +1,592 @@
...
+        )
+
+        # generate SFA words on sample
+        words = transformer.fit_transform(X, y)
+        all_words.append(words)
+        all_transformers.append(transformer)
+    return all_words, all_transformers
+
+
+def getSFAFast(
```
TonyBagnall, line 554: is this function necessary? Seems to me you may as well call SFAFast on line 527?

TonyBagnall (dismissed):
lgtm, maybe let @MatthewMiddlehurst have a look too, he is more  thorough than me :)

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: Sure, will wait for @MatthewMiddlehurst before I do another push.

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: No, I have not tried. It requires a small redesign. Would rather give it more testing before enabling it :)

sktime/classification/dictionary_based/_weasel_v2.py
patrickzib: Refactored

sktime/classification/dictionary_based/_weasel_v2.py
TonyBagnall: yes thats fine by me I understand

TonyBagnall (approved):
All looks good to me

PR #165, opened 2023-03-07
[ENH] Switch from pandas iteritems() to items() due to warning

TonyBagnall:
switches iteritems() to items() in two places because of Pandas deprecating warning


aiwalter (approved):
looks ogg, I have replaced this also somewhere in my `pandas==2.0.0` PR as in that version it breaks #153

PR #176, opened 2023-03-08
[MNT] Release tool check

MatthewMiddlehurst:
This is more myself trying to understand our current build tools rather than a critique or proposal to overhaul the current workflow, I'm relatively new to this 🙂.

From what I understand, the `make_release.py` file is no longer used, along with the accompanying .sh and Makefile options.

The updated `wheels.yml` is basically what I have been using [here](https://github.com/time-series-machine-learning/tsml-eval/blob/main/.github/workflows/release.yml) (side note I think `release` is a better name for the file, but I have left it for ease of comparison).

Some questions regarding the workflow file:
- I have included the `check_manifest` pre-commit option as a step, any reason not to?
- Is there a reason for using static OS versions rather than latest? I guess the obvious answer is it's more reliable.
- Currently Windows is separated into a separate step using Conda, is this just for testing alternate methods or is it necessary to build the package on Windows?
- For the `pytest` run the Windows step and Makefile creates a testdir, but I don't know if this is actually used anywhere. Would have to change the test .yml to find out, probably.
- Would it be possible to upload the wheels TestPyPi first and verify it can be used to install? Don't know if this is redundant.


MatthewMiddlehurst:
@lmmentel I believe you engineered the current workflow? I more want to understand some of the quirks, the changes are just what I use for other packages I have released based on the current workflow.

lmmentel:
> From what I understand, the `make_release.py` file is no longer used, along with the accompanying .sh and Makefile options.
> 
Yup, I've never used it since I went for automating the releases with CI.

> The updated `wheels.yml` is basically what I have been using [here](https://github.com/time-series-machine-learning/tsml-eval/blob/main/.github/workflows/release.yml) (side note I think `release` is a better name for the file, but I have left it for ease of comparison).

Good idea, go ahead and rename :+1: 

> 
> Some questions regarding the workflow file:
> 
>     * I have included the `check_manifest` pre-commit option as a step, any reason not to?
> 
:+1: 

>     * Is there a reason for using static OS versions rather than latest? I guess the obvious answer is it's more reliable.
> 
The OS change at such a slow pace that we don't need to worry about updating them too often, but if we use `latest` CI will break in some time in the future in a very hard to debug way. I would rather bump manually and validate it still works to be able to rollback.

>     * Currently Windows is separated into a separate step using Conda, is this just for testing alternate methods or is it necessary to build the package on Windows?
> 
There were/are some dependencies that are not supported on windows officially but have conda releases. You could try to try without conda but I would say it's a low priority since it works fine now and switching to an alternative does not benefit us or the users much.

>     * For the `pytest` run the Windows step and Makefile creates a testdir, but I don't know if this is actually used anywhere. Would have to change the test .yml to find out, probably.
This seems like another thing that I would call low priority. Maybe it would be worth going back to this once we look into py3.11 support?
> 
>     * Would it be possible to upload the wheels TestPyPi first and verify it can be used to install? Don't know if this is redundant.
It is certainly possible but I'm not sure how much benefit would that bring, can you see some good reasons to invest time in that?

MatthewMiddlehurst:
made some of the simpler changes in #259 i.e. rename, bump versions, manifest check and remove old tools. 

will revisit the others at some point

MatthewMiddlehurst:
With the changes from #553 and #593, I'm going to try to include this in the latest release in #596.
I agree with your comment that TestPyPi is probably not worth the time or effort, so will just stick to updating the pytest portions to match our current/planned CI.

Feel free to object here or in the release PR, it is not an urgent update really.

Issue #110, opened 2023-02-25
[ENH] removing the use of nested_univ/nested np.DataFrames in time series machine learning


TonyBagnall:
**Is your feature request related to a problem? Please describe.**

Currently, the default is to load tsml problems into nested data frames, where each row is a time series, each column is a dimension/channel and each cell is a pd.Series. I have never liked this usage. It is confusing and inefficient. Furthermore, this use case  is being deprecated in Pandas. 

We could switch to pd.multiindex, but nearly all use cases is tsml are with numpy arrays. scikit, tensorflow, pytorch all use numpy. The only problem is when the problem has unequal length series. I propose switching to numpy whenever possible, removing nested pandas and for the problem case of unequal length using a list of numpy array and providing a good range of transformers to pad/truncate/downsample etc. 

**Describe the solution you'd like**

[ ]. Remove the default behaviour of loading nested DataFramein the univariate single problem loaders #109 
2. Change the transformers that internally use nested DataFrame to use numpy  #85 #98 
3. Check code base for all uses of nested DataFrame and all current internal handling of unequal length (e.g. distances) 
4. Switch unequal length loading to return list of numpy
5. Adapt/provide transformers for list of numpy to numpy (Padding etc)
6. Rewrite notebooks and webpages to reflect the changes
 
I'll add occurrences here as I find them

1. def _fit_predict_boilerplate in BaseClassifier converts everything to nested DataFrame with the following comment
"Convert data to format easily useable for applying cv"

TonyBagnall:
Historically, we have used stored time series in Pandas, where every cell is a pd.Series. This gave maximum flexibility: we could have unequal length and time stamps. However, it adds a level of conversion and checking that is inefficient and does not suit scikit time very well. First stage of removing nested_univ all together is to make sure no classifiers use it. We will restrict to equal length problems until we introduce a nested numpy array format for unequal length series, which will happen once we deprecate explicit Pandas use for the internal fit and predict. Conversion can still happen at the base class level, so user experience will not change.

Classifiers still using nested_univ internal type

1. compose/ColumnEnsembleClassifier
2. compose/ClassifierPipeline
3. compose/ComposableTimeSeriesForestClassifier
4. distance_based/ElasticEnsemble (used internally with transformer DerivativeSlopeTransformer then converted back) #85 
5. distance_based/ShapeDTW (used with several transformers then converted back) #98

TonyBagnall:
this is now complete for classifiers, onwards to transformers

PR #178, opened 2023-03-08
[ENH] Tidy BaseClassifier and classifier template

TonyBagnall:
1. Removes references to nested_univ and scitypes from classification base and template
2. Makes _predict abstract, because having skeleton implementation of both _predict and _predict_proba lead to stack overflow if you dont implement either of them
3. Removes confusing instructions in the template, which is meant to be simple

part of #110 and #42


extension_templates/classification.py
```diff
@@ -3,22 +3,20 @@
 Extension template for time series classifiers.
 
 Purpose of this implementation template:
-    quick implementation of new estimators following the template
-    NOT a concrete class to import! This is NOT a base class or concrete class!
-    This is to be used as a "fill-in" coding template.
+    quick implementation of new estimators following the template. This is NOT a
+    concrete class to import nor is a base class or concrete class. This is to be used
+    as a "fill-in" coding template.
```
MatthewMiddlehurst, line 8: These lines seem a bit off. Mentions concrete class twice and "nor is a base class".

extension_templates/classification.py
```diff
@@ -53,31 +50,42 @@ class MyTimeSeriesClassifier(BaseClassifier):
...
+    param_c : boolean, default= True
+        descriptive explanation of param_c.
+
+    Attributes
+    ----------
+    att_a_ : int
+        attributes are not passed to the constructor. If set in fit use following
+        underscore.
+    att_b_: boolean or None, default=None
+        there may not be any.
```
MatthewMiddlehurst, line 68: Does `sklearn` include optional attributes like this? Not saying its bad just never really encountered or thought about it.

extension_templates/classification.py
```diff
@@ -53,31 +50,42 @@ class MyTimeSeriesClassifier(BaseClassifier):
...
+    --------
+    Optional, related classes
+
+    Notes
+    -----
+    Optional, Anything you want
+
+    References
+    ----------
+    Optional, details of paper where algorithm described
```
MatthewMiddlehurst, line 80: I think we should add an example here. I don't think it will be run in tests (if it is i can add it to the exclude).

extension_templates/classification.py
```diff
@@ -148,15 +131,9 @@ def _fit(self, X, y):
...
-        #
-        # Note: when interfacing a model that has fit, with parameters
-        #   that are not data (X, y) or data-like,
-        #   but model parameters, *don't* add as arguments to fit, but treat as follows:
-        #   1. pass to constructor,  2. write to self in constructor,
-        #   3. read from self in _fit,  4. pass to interfaced_model.fit in _fit
+        # IMPORTANT: avoid side effects to X, y, if changing X and y make local copies.
 
-    # todo: implement this, mandatory
+    # todo: implement this,
```
MatthewMiddlehurst, line 136: Trailing comma, can just use the same comment as _fit probably.

extension_templates/classification.py
```diff
@@ -283,44 +250,3 @@ def get_test_params(cls, parameter_set="default"):
 
         # todo: set the testing parameters for the estimators
         # Testing parameters can be dictionary or list of dictionaries
```
MatthewMiddlehurst, line 252: May be good to have an example dict still

MatthewMiddlehurst (commented):
Good changes

extension_templates/classification.py
TonyBagnall: not sure, probably not. I'm easy either way, although it does stress the underscore suffix pattern

extension_templates/classification.py
TonyBagnall: good idea, done

extension_templates/classification.py
MatthewMiddlehurst: I don't think they do, think we should just keep it as boolean and remove the None and default. If it is conditional, it should just be mentioned in the description.

see `oob_decision_function_ `here https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html

TonyBagnall: removed the None and default versions.

extension_templates/classification.py
```diff
@@ -32,6 +29,8 @@
 
 Testing - implement if aeon classifier (not needed locally):
     get default parameters for test instance(s) - get_test_params()
+
+copyright: aeon developers, BSD-3-Clause License (see LICENSE file)
```
MatthewMiddlehurst, line 33: wouldn't add this. dont really think its needed

MatthewMiddlehurst (approved):

PR #179, opened 2023-03-08
[ENH] Removes the dist_kerns module

### LINKED ISSUE #47: [ENH] Remove dists_kernels

TonyBagnall:
removes the dist_and_kernels module, see #47 for reasons, fixes #47.


TonyBagnall:
begone strange package

PR #180, opened 2023-03-09
[MNT] Fix random test failure in FeatureSelection

### LINKED ISSUE #158: [MNT] Investigate random state failure in test_feature_selection[None-2-random]
test `test_feature_selection[None-2-random]` was failing "randomly", need to investigate it.

pipeline: https://github.com/aeon-toolkit/aeon/actions/runs/4344052231/jobs/7587082413

aiwalter:
closes #158

### DIFF
--- a/sktime/transformations/series/feature_selection.py
+++ b/sktime/transformations/series/feature_selection.py
@@ -67,7 +67,7 @@ class FeatureSelection(BaseTransformer):
     >>> from sktime.datasets import load_longley
     >>> y, X = load_longley()
     >>> transformer = FeatureSelection(method="feature-importances", n_columns=3)
-    >>> Xt = transformer.fit_transform(X, y)
+    >>> X_hat = transformer.fit_transform(X, y)
     """
 
     _tags = {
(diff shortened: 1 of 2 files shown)


lmmentel:
Hey, how about dropping the parametrization over random seeds? Not sure it adds any extra coverage.

TonyBagnall (approved):

TonyBagnall:
> Hey, how about dropping the parametrization over random seeds? Not sure it adds any extra coverage.

I agree, no point testing two random seeds. Fewer tests is a good thing

aiwalter:
> Hey, how about dropping the parametrization over random seeds? Not sure it adds any extra coverage.

The coverage is that this test makes sure that `"random"` is actually doing sth random :) So lets keep it, now it should not fail anymore

MatthewMiddlehurst:
Have to agree with Tony and Lukasz really, it is essentially doubling the time this test takes every PR unnesessarily. If we did this kind of parameterisation in other modules we would be spending 4 hours sitting around. Things like this are prime candidates to cut for reducing the time spent testing.

PR #123, opened 2023-02-27
[ENH] first rework of the classification notebook

TonyBagnall:
I've put this ready to review, but its just a starting point and all notebooks could be improved with more examples. However, I would like to get this in and then improve one at a time as we work on that type of classifier, and obviously let other people get involved. @patrickzib @hadifawaz1999 I've used your images, hope thats ok. 

part of #45

PR #185, opened 2023-03-10
[ENH] Shapelet Transform improvements

TonyBagnall:
the main thing here is that the conversion from numpy to dataframe is removed. All that was happening was the numpy was getting put into a dataframe in the transform, then converted back to numpy in classifier. The base transformer does this conversion automatically, unless you set this argument in the constructor

super(RandomShapeletTransform, self).__init__(_output_convert=False)

other changes:
1. Docstrings tidied up, dimensions changed to channels, all reference to DataFrames removed
2. Helper classes for full ShapeletTransform made private
3. Remove pointless tags

to do in a later PR
Order of classes in file changed to put RandomShapeleTransform first


sktime/transformations/panel/shapelet_transform.py
```diff
@@ -28,1460 +27,1445 @@
...
+        Parameters
+        ----------
+        X : np.ndarray shape (n_time_series, n_channels, series_length)
+            The input data to transform.
 
-                cand_start_pos = candidates_to_visit[candidate_idx][0]
-                cand_len = candidates_to_visit[candidate_idx][1]
+        Returns
+        -------
+        output : pandas DataFrame
```
GuzalBulatova, line 354: Looks like it's a numpy array here?

sktime/transformations/panel/shapelet_transform.py
```diff
@@ -28,1460 +27,1445 @@
...
+        self._random_state = check_random_state(self.random_state)
 
-        # evaluate each split point
-        for split in range(0, len(orderline) - 1):
-            next_class = orderline[split][1]  # +1 if this class, -1 if other
-            if next_class > 0:
-                count_this_class += 1
-            else:
-                count_other_class += 1
+        # Here we establish the order of cases to sample. We need to sample
```
GuzalBulatova, line 712: Might worth putting the description of core logic into the method's docstring, what do you think? *Not a blocker

sktime/transformations/panel/shapelet_transform.py
```diff
@@ -28,1460 +27,1445 @@
...
+                                        )
+                                        * 60
+                                    ),
+                                )
+                            )
+                        time_finished = True
+                        break
+                    else:
+                        if self.verbose > 0:
+                            if candidate_rejected is False:
```
GuzalBulatova, line 1029: Below was probably automatically redacted by a formatter, it was quite hard to read for me. By line 1061 I lost the track of what bracket was closing what and what the operation was referring to. I'd say either override the auto-formatting or break it down. Maybe it's just my perception. What's your view?

sktime/transformations/panel/shapelet_transform.py
```diff
@@ -28,1460 +27,1445 @@
...
-                        break
+        Returns
+        -------
+        zscore : array_like
+            The z-scores, standardized by mean and standard deviation of
+            input array a.
+        """
+        zscored = np.empty(a.shape)
+        for i, j in enumerate(a):
+            # j = np.asanyarray(j)
```
GuzalBulatova, line 1440: Is this line left here on purpose for future work?

GuzalBulatova (changes requested):
I've got a couple of comments and a general suggestion to break down the big PRs into smaller ones where possible. 
Overall this is absolutely fantastic work! Definitely a significant improvement. The code flow is easy to read and I really appreciate the comprehensive documentation.

MatthewMiddlehurst:
I agree that it is extremely difficult to tell what has actually changed without going through the full file (both pre and post change). Maybe moving them around should be a separate thing.

TonyBagnall:
thanks for the feedback @GuzalBulatova, really helpful. I did wonder about the complexity of reviewing when moving classes, just wanted to reprioritise since the ShapeletTransformClassifier is there for legacy only, but it is purely cosmetic, I will shift them back. I have a more systemic issues about the tests for which I will raise an issue

sktime/transformations/panel/shapelet_transform.py
TonyBagnall: fixed, thanks

sktime/transformations/panel/shapelet_transform.py
TonyBagnall: I have no idea! So I have removed it, if it was there for a purpose it should have been commented

TonyBagnall:
> I agree that it is extremely difficult to tell what has actually changed without going through the full file (both pre and post change). Maybe moving them around should be a separate thing.

agreed, I will shift order on its own PR, hopefully more readable now

sktime/transformations/panel/shapelet_transform.py
TonyBagnall: I think that whole comment was left in for debugging, it can go! I have shortened the print, jason put them in before we could assume print(f"ddd") was available I think.

TonyBagnall:
Im working on a notebook change, but I think as I am here with STC I will write a shapelet notebook, which make take a while

sktime/transformations/panel/shapelet_transform.py
TonyBagnall: I'd like to, but I think I will leave it for now, its quite a lot of work for a legacy classifier, and Jason's comments are very good, albeit in a style of inline comments we dont use now

TonyBagnall:
I have added a notebook on using the transform. I want to put more details in, but I need to work out how it interacts with the classification notebooks I'm changing on #123 so would like to leave it at this for now.

MatthewMiddlehurst (approved):
I'm fine with the changes, but the shapelet notebook may conflict with the previous notebook PR. Just not getting picked up because it's in a different dir.

PR #199, opened 2023-03-14
[ENH] switch paa tests to use numpy

TonyBagnall:
another small change, make the tests of PAA use numpy arrays instead of data frames


patrickzib:
Looks good to me. But can you remove the forecasting notebook?

TonyBagnall:
that bloody notebook is following me around! I'll replace it with the one from main

TonyBagnall:
> Looks good to me. But can you remove the forecasting notebook?

@patrickzib so I deleted it, copied it from main and pushed it, but it is still showing changes. No idea

GuzalBulatova:
> > Looks good to me. But can you remove the forecasting notebook?
> 
> @patrickzib so I deleted it, copied it from main and pushed it, but it is still showing changes. No idea
Looks like there are just a bunch of spaces removed, can a pre-hook be a culprit? Although why would it run on a notebook that was reverted, curious.

patrickzib (approved):
👍

PR #200, opened 2023-03-16
[DOC] Refactor documentation out of `source` directory

MatthewMiddlehurst:
There are two common defaults for the documentation location, `docs/` and `docs/source/`. This PR refactors our docs to the first, mainly because it allows for shorter links when referring to the documentation source files. The only two files outside the `source/` directory currently are the `Makefile` and `make.bat`. There is not much organisational benefit to keeping those two separate from the rest of the source files.


TonyBagnall (approved):
makes sense to me, assuming there are no CI type reasons not to

Issue #208, opened 2023-03-17
[ENH] Tidy up ROCKET variants


TonyBagnall:
the convolution_base l Transformers need tidying up and speeding up. Current situation is this

## rocket.py: Rocket
   fit: calls _generate_kernels: randomly generates convolutions and randomly generated from randomly selected subset of channels, stores it all in self.kernels (should be self.kernels_ I think). n_jobs is set, but I cant see where its actually used

  transform: calls _apply_kernels 

## multirocket.py: MultiRocket
  fit: flattens the data, then normalises. This is wrong, should normalise first, although the flattening is pointless as its not multivariate capable. (X.shape[1] == 1), calls _get_parameter

 transform: calls another method called transform, with a long hard coded array 

## multirocket_multivariate.py: MultiRocketMultivariate w

orks directly with 3D numpy, hard coded edge case for short series, seems identical to MultiRocket otherwise. 

## minirocket_multivariate_variable: MiniRocketMultivariateVariable

seems unnecessary.

- [ ] Set up the transformers to only use numpy when equal length
- [ ] Allow for skipping of all conversion and meta data extraction 
- [ ] Merge multivariate and univariate versions
- [ ] Facilitate unequal length series transformers (merge existing multivariate version)
- [ ] Profile and see if can be optimised at all

hadifawaz1999:
@TonyBagnall isn't there a lot of common stuff between all rocket variants ?
why not create a rocket base
for instance applying the convolution kernel can be common with a list of strs as arguments to indicate which poolings to use, which is different from variant to variant

TonyBagnall:
lets make an AEP for convolution based estimators

TonyBagnall:
replaced by #1699

PR #209, opened 2023-03-17
[ENH] Adapt rocket classifiers to return numpy arrays instead of data frames

TonyBagnall:
currently, all rocket transforms are converted into pandas before return in transform, despite all operations being performed on numpy. This PR switches all the rocket variants to return numpy arrays in transform.  The classifiers that use these transforms then just convert it back to numpy. It is much more natural to just not use pandas at all. There are two stages
1. Stop the conversion in transform (remove pd.DataFrame(X_)
2. Stop the automatic conversion in the base class ( super(Rocket, self).__init__(_output_convert=False)) 

 This PR is part of #208. Note I have left the unequal length version for now. For a bonus 10 pts, I will also 

1. speed up the tests prior to merging. they currently take 2 mins locally for 6 rocket tests (even after numba build)  
2. extend the docstring for rocket, which was a bit terse.

I have reverted the MultiRocket because the base class conversions are breaking it with 2D numpy. It breaks them all, but it makes MultiRocket fail tests. We can wait to do this until after we have resolved the conversion problem.


setup.cfg
```diff
@@ -13,13 +13,6 @@ addopts =
     --ignore docs
     --doctest-modules
     --durations 10
-    --timeout 600
-    --cov aeon
-    --cov-report xml
-    --cov-report html
-    --showlocals
-    --matrixdesign True
-    -n auto
```
MatthewMiddlehurst, line 22: These should remain

MatthewMiddlehurst (dismissed):

setup.cfg
TonyBagnall: gets me every time!

MatthewMiddlehurst (approved):

PR #213, opened 2023-03-17
[ENH] remove incorrect iloc calls

TonyBagnall:
this removes a couple of calls to iloc for returned data that is no longer a data frame. It also removes a test that is skipped and listed as being deprecated some time ago.


lmmentel (approved):
:+1:

PR #221, opened 2023-03-22
[MNT] Renamed parameter in mlflow module

aiwalter:
Renamed parameter `aeon_model` formerly `sktime_model` to `forecaster`


aiwalter:
I thought also about it, currently its supporting predict functions related to forecasting. We could add support for `transform()`, will create an issue for it. For the classification I thought it can work with the `scikit-learn` support of `mlflow` but not sure. I can rename the param to `estimator` so then we can also use it in future for transformers

MatthewMiddlehurst (approved):

Issue #128, opened 2023-02-28
[ENH] Remove _DelegatedClassifier


TonyBagnall:
a form of delegation was introduced into the classification module in summer 2022. It is used in RocketClassifier and  Catch22Classifier. It is in my opinion an unnecessary engineering complication that adds no value. I want to remove it and return to the way it was before

TonyBagnall:
so #234 and #235 removed its use, I think this can stay as an unused feature in classification, simply because it is also present in the other packages.

PR #234, opened 2023-03-22
[ENH] Converts RocketClassifier back to being a sklearn pipeline classifier

TonyBagnall:
RocketClassifier had been altered to inherit from _DelegateClassifier in a completely unnecessary complication. This PR reverts it to being an sklearn pipeline classifier, and as a bonus adds the ability to set a different classifier.

Simplified  the docstring, removed the strange option of using a univariate version of MiniRocket on multivariate.

#### Reference Issues/PRs

Part of #128 

 No, already tested in test_all_classifiers, which it passexs

### DIFF
--- a/aeon/classification/convolution_based/_rocket_classifier.py
+++ b/aeon/classification/convolution_based/_rocket_classifier.py
@@ -134,7 +134,7 @@ class RocketClassifier(BaseClassifier):
         self.n_instances_, self.n_dims_, self.series_length_ = X.shape
 
         if self.rocket_transform == "rocket":
-            self.transformer_ = Rocket(num_kernels=self.num_kernels)
+            transformer = Rocket(num_kernels=self.num_kernels)
         elif self.rocket_transform == "minirocket":
             if self.n_dims_ > 1:
                 transformer = MiniRocketMultivariate(
@@ -154,7 +154,7 @@ class RocketClassifier(BaseClassifier):
                     n_features_per_kernel=self.n_features_per_kernel,
                 )
             else:
-                self._transformer = MultiRocket(
+                transformer = MultiRocket(
                     num_kernels=self.num_kernels,
                     max_dilations_per_kernel=self.max_dilations_per_kernel,
                     n_features_per_kernel=self.n_features_per_kernel,
--- a/aeon/classification/convolution_based/_rocket_classifier.py
+++ b/aeon/classification/convolution_based/_rocket_classifier.py
@@ -9,10 +9,9 @@ __all__ = ["RocketClassifier"]
 
 import numpy as np
 from sklearn.linear_model import RidgeClassifierCV
-from sklearn.preprocessing import StandardScaler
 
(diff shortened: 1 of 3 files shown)


aeon/classification/convolution_based/_rocket_classifier.py
```diff
@@ -1,18 +1,19 @@
 # -*- coding: utf-8 -*-
 """RandOm Convolutional KErnel Transform (Rocket).
 
-Pipeline classifier using the ROCKET transformer and RidgeClassifierCV estimator.
+Pipeline classifier using the ROCKET transformer and an sklearn classifier.
 """
 
-__author__ = ["MatthewMiddlehurst", "victordremov", "fkiraly"]
```
MatthewMiddlehurst, line 7: Probably shouldn't remove names from this with the current format.

Though TBH I don't really like the concept, all file specific contributions by users are already recorded in the version control and the general contributors file. It would be more helpful to have a list of current maintainers (`__maintainer__`) instead.

aeon/classification/convolution_based/_rocket_classifier.py
```diff
@@ -92,113 +73,167 @@ class RocketClassifier(_DelegatedClassifier):
...
+        X : 3D np.array of shape = [n_instances, n_dimensions, series_length]
+            The data to make predict probabilities for.
+
+        Returns
+        -------
+        y : array-like, shape = [n_instances, n_classes_]
+            Predicted probabilities using the ordering in classes_.
+        """
+        X_new = self._transformer.transform(X)
+        X_new = self._scaler.transform(X_new)
```
MatthewMiddlehurst, line 227: I think the pipeline can still be used here, would be cleaner that way

aeon/classification/convolution_based/_rocket_classifier.py
```diff
@@ -92,113 +73,167 @@ class RocketClassifier(_DelegatedClassifier):
...
 
+        Notes
+        -----
+        Changes state by creating a fitted model that updates attributes
+        ending in "_" and sets is_fitted flag to True.
+        """
+        self.n_instances_, self.n_dims_, self.series_length_ = X.shape
+
+        if self.rocket_transform == "rocket":
+            self._transformer = Rocket(num_kernels=self.num_kernels)
```
MatthewMiddlehurst, line 152: `random_state` doesn't seem to be set for these? n_jobs for rocket also

aeon/classification/convolution_based/_rocket_classifier.py
TonyBagnall: I am in effect reverting the changes made, so seems reasonable to remove authors of those changes to me

MatthewMiddlehurst: That's fine, won't block. Will probably propose replacing the field for all files at some point.

MatthewMiddlehurst (approved):

PR #242, opened 2023-03-28
[DOC] Remove GSoC and meetups pages, update mentoring with placeholder

### LINKED ISSUE #228: [DOC] Update Mentors
One the webpage there are potential mentors listed via icons, who should not be mentors any longer for this fork.

https://www.aeon-toolkit.org/en/latest/get_involved/mentoring.html

MatthewMiddlehurst:
Resolves #228

We currently do not have a mentoring program, unless someone is willing to update this page and maintain such a system. (previously removed, currently a placeholder. See discussion below.)

The meetups page does not really add anything of value, and the GSoC page is no longe relevant.


MatthewMiddlehurst:
Well we can keep the page, but it should probably be reworked to offer specific programs from specific mentors.

I can replace the current content with a placeholder or just leave it as it is if you want @aiwalter?

Some of the phrasing like "We explicitly encourage students from underrepresented groups or who have experienced discrimination in the past to apply!" makes me feel like im recruiting for a cult 🙂

aiwalter:
yeah that part should go. I can update that page next days. You can make it empty now if you want, e.g. write "tbd"

TonyBagnall (approved):

Issue #197, opened 2023-03-14
[ENH] Transformers that need conversion from nested pd.DataFrame to numpy arrays


TonyBagnall:
nested_univ is no longer going to be supported in aeon, switching fully to numpy. Some transformers still store things internally in this format. Nearly all of them just iterate and convert to numpy, so can just be converted to "X_inner_mtype": "numpy3D", and then adjusted when we facilitate unequal length. I'll list here as I find them (would search tags but doesnt work, see #182 )

- [x] DerivativeSlopeTransformer _extract.py  #392 
- [x] PlateauFinder in _extract.py #392 
- [x] RandomIntervalFeatureExtractor in compose.py #391
- [x] RandomIntervalSegmenter  in segment.py #391 
- [x] IntervalSegmenter  in segment.py #391
- [x] TSInterpolator (why is it named this? We do not use TS suffix elsewhere?) interpolate.py #388 
- [x] PaddingTransformer (again, may refactor the name) in padder.py  #364
- [x] TruncationTransformer in truncation.py #364 
transformations/Panel/reduce.py
- [x] SFA
- [x] SlidingWindowTransformer
- [x] DerivativeSlopeTransformer #261  
- [x] Catch22
- [x] Catch22Wrapper
- [x] DWTTransformer
- [x] HOG1DTransformer #261 
- [x] SlopeTransformer
- [x] TruncationTransformer
- [x] _TSFreshFeatureExtractor, TSFreshFeatureExtractor, TSFreshRelevantFeatureExtractor
- [x] PAA #269 
- [x] All the rocket transformers #209 
- [x] ShapeTranformer #250 

Leaving these
Tabularizer reduce.py
its embedded in the reduction stuff in forecasting, so I dont want to mess with it
ColumnTransformer
can sort it out when we address #171

basically need to switch from 
        "X_inner_mtype": "nested_univ",  # which mtypes do _fit/_predict support for X?
to this
        "X_inner_mtype": "numpy3D",  
and obviously make it all work and update any tests. Also a good time to add examples. Also need to do this in the constructor to stop nonsense
        super(PAA, self).__init__(_output_convert=False)

then treat X as numpy in fit and transform. Some much easier than others, any that work with unequal length (e.g. truncate and pad) leave for now because need to work with lists of numpy

aiwalter:
Actually it seems nested pandas is still working with pandas 2.0.0

TonyBagnall:
still want to get rid of them :)

[...]

Issue #98, opened 2023-02-21
[ENH] Slow transformers that may not be needed


TonyBagnall:
there are a range of transformers that are a) very simple algorithms, b) not used much/at all and c) involve conversion to dataframe. I think they could be converted to simple numpy/numba functions (whichever is faster). I will list as I find them, I dont mind in principle having functionality that is not explicitly used, but not when it is so inefficient. A lot of this code has been sat here for years and the doc strings are not well configured. Many are in there own file, causing code bloat. All involve a lot of tests (which are quick, but nevertheless ...). For me, anything not used and X_inner_mtype = nested_univ is a candidate for removal. 

as far as I can tell, used only in ShapeDTW:
1. sktime.transformations.panel.SlopeTransformer:
2.  sktime.transformations.panel.DWTTransformer:
3.  sktime.transformations.panel.HOG1DTransformer
4.  sktime.transformations.panel.SlidingWindowSegmenter
used in EE (see #85 ) and ShapeDTW:
5.  sktime.transformations.panel.DerivativeSlopeTransformer
others to investigate that seem to be only used in testing
PlateauFinder:
RandomIntervalSegmenter: 
DistanceFeatures: 
TSInterpolator:
IntervalSegmenter
RandomIntervalFeatureExtractor

TonyBagnall:
this has now been completed

Issue #85, opened 2023-02-20
[ENH] Write a function to perform the logic of DerivativeSlopeTransformer using numpy arrays instead of pd.DataFrame


TonyBagnall:
quick review reveals a few possible improvements to EE, which was one of the earliest classifiers we introduced

1When calculating derivatives, EE uses a DerivativeSlopeTransformer which has internal type nested_univ. This means this calculation is shoving a numpy back into a panda, doing inefficient calculations, then pulling ti back into numpy.

```python
           # X is a 3D numpy that then gets internally converted to pandas
            der_X = DerivativeSlopeTransformer().fit_transform(X)
            # it convert back to numpy: remove this and just create differences
            if isinstance(der_X, pd.DataFrame):
                der_X = from_nested_to_3d_numpy(der_X)
```

Rather than changing DerivativeSlopeTransformer, which is embedded in a morass of code, I think this really simply transform should just be done in a static method in ElasticEnsemble using numpy only. Basically convert this 

```python
            def get_der(x):
            der = []
            for i in range(1, len(x) - 1):
                der.append(((x[i] - x[i - 1]) + ((x[i + 1] - x[i - 1]) / 2)) / 2)
            return pd.Series([der[0]] + der + [der[-1]])

        return [get_der(x) for x in X]
```
to use numpy. There are other possible optimisations in EE, there is no testing for correctness for EE and it does not include all of the original

PR #246, opened 2023-03-29
[ENH] init file for Panel transformers

TonyBagnall:
addpopulates the init file for the directory transformations/Panel to match the standard patterns of import via init. I will take the files private and refactor out of the Panel directory individually as each is refactored to not use nested_univ 

#### Reference Issues/PRs

related to #98 #197 #85


TonyBagnall:
mixed this with other changes

Issue #249, opened 2023-03-29
[ENH] Improve the efficiency of SlopeTransformer


TonyBagnall:
**Is your feature request related to a problem? Please describe.**
the SlopeTransformer in transformations/panel/slope.py has been converted to use numpy arrays rather than data frames internally in PR #250 . However, it still relies on a lot of looping and lists, which I'm sure is inefficient. It could be done much faster with numpy operations.

**Describe the solution you'd like**
I would like a timing experiment with current code, reworking using numpy ops, then a timing test after the refactor to show speed up. 

**Describe alternatives you've considered**
Leave it as is. It is used in ShapeDTW, but nowhere else. This is a good first issue for someone wanting to get involved in algorithm development, but it is hardly pressing.

PR #250, opened 2023-03-29
[ENH] Convert ShapeTransformer to use numpy internally

TonyBagnall:
ShapeTransformer transforms by finding slope over intervals. This PR converts it to use numpy3D internally as part of ongoing refactoring. ShapeDTW uses it, and still passes all the tests after the changes. ShapeDTW should be refactored to use numpy3D after all transformers I think.

#### Reference Issues/PRs
Part of #197 See #249 for future possible improvements to this transformer.


patrickzib:
Something seems wrong. Tests take more than 5 hours now. Not sure, if it is related to this PR or due to an error on main-branch:
image

MatthewMiddlehurst (approved):

PR #252, opened 2023-04-01
[ENH] add numba function to find slope derivative

### LINKED ISSUE #85: [ENH] Write a function to perform the logic of DerivativeSlopeTransformer using numpy arrays instead of pd.DataFrame

TonyBagnall:
currently, ElasticEnsemble finds the slope derivative for ddtw using DerivativeSlopeTransformer. This takes a 3d numpy, internally converts everything into nested pandas, then loops through converting. Once returned, the derivative dataframe is converted back to nump.

functions: series_slope_derivative(x) takes 3D time series, loops over all and calls private function _der(x). 

This replaces the use of DerivativeSlopeTransformer in ElasticEnsemble with a numba compiled  function defined in the ElasticEnsemble function. This is done with loops and numba. Its very possible it might be a bit faster with numpy vectorised operations, but this suggests the difference will be minimal, and this is easier to read (for me at least). 
https://numba.pydata.org/numba-doc/latest/user/performance-tips.html

If this is acceptable, I will do the same with ShapeDTW and decide if we want to keep DerivativeSlopeTransformer

x#### Reference Issues/PRs

Fixes #85 
Part of #197

whether it would be better to vectorise one or both of the functions 

I tried, but dont think pytest works with numba functions?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,14 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
+    --timeout 600
+    --cov aeon
+    --cov-report xml
+    --cov-report html
+    --showlocals
+    --matrixdesign True
+    -n auto
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
-    --timeout 600
-    --cov aeon
-    --cov-report xml
-    --cov-report html
-    --showlocals
-    --matrixdesign True
-    -n auto
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
--- a/aeon/classification/distance_based/_elastic_ensemble.py
+++ b/aeon/classification/distance_based/_elastic_ensemble.py
@@ -11,7 +11,7 @@ import time
 from itertools import product
 
 import numpy as np
-import pandas as pd
+from numba import njit
 from sklearn.metrics import accuracy_score
 from sklearn.model_selection import (
     GridSearchCV,
@@ -27,14 +27,19 @@ from aeon.classification.distance_based._time_series_neighbors import (
 )
 
 
+@njit(fastmath=True, cache=True)
 def _der(x: np.ndarray):
-    """Perform derivative slope transform on single series."""
-    d = ((x[1:-1] - x[:-2]) + ((x[2:] - x[:-2]) / 2.0)) / 2.0
-    d = np.concatenate(([d[0]], d, [d[-1]]))
-    return d
+    """Loop based Derivative Slope transform."""
+    m = len(x)
+    der = np.zeros(m)
+    for i in range(1, m - 1):
+        der[i] = ((x[i] - x[i - 1]) + ((x[i + 1] - x[i - 1]) / 2.0)) / 2.0
+    der[0] = der[1]
+    der[m - 1] = der[m - 2]
+    return der
 
 
-def series_slope_derivative(X: np.ndarray):
+def series_slope_derivative(X: np.ndarray) -> np.ndarray:
     """Find the slope derivative of collection of time series.
 
     Parameters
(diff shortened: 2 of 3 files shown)


TonyBagnall:
so chatGPT said I should do it like this. Just running some timing tests, the chaptGPT version seems a little bit slower than the numba one. but its probably preferable, because I can write a test method for it, although I should check its correct too

```python
def _der2(x: np.ndarray):
    """Vectorized Derivative Slope transform."""
    der = ((x[1:-1] - x[:-2]) + ((x[2:] - x[:-2]) / 2.0)) / 2.0
    der = np.concatenate(([der[0]], der, [der[-1]]))
    return der

def series_slope_derivative2(X: np.ndarray):
    """Vectorized slope derivative of 3D time series.

    Parameters
    ----------
    X: np.ndarray shape (n_time_series, n_channels, series_length)

    Returns
    -------
    np.ndarray shape (n_time_series, n_channels, series_length)
    """
    return np.apply_along_axis(_der, axis=-1, arr=X)

**Comment by TonyBagnall** [comment]
actually, in my very ad hoc timing experiment, vectorised is looking a little faster, I'll set up a proper test and include the pandas version too

**Comment by TonyBagnall** [comment]
so first experiment, thought I would put in current implementation. Columns are 
image

**Comment by patrickzib** [comment]
Nice result! Just wondering: you could also decorate the vectorized function _der2(x: np.ndarray): using numba?

**Comment by TonyBagnall** [comment]
well that was a bad idea! njit and vectorisation dont work, this actually crashes
def _der3(x: np.ndarray):
    """Vectorized Derivative Slope transform."""
    der = ((x[1:-1] - x[:-2]) + ((x[2:] - x[:-2]) / 2.0)) / 2.0
    der = np.concatenate(([der[0]], der, [der[-1]]))
    return der

seems like a straight choice between vectorised or numba. I'll run a longer timing comparison. I think its probably better to vectorise if the differences are small, because it can be tested. Does anyone have an opinion about this?

**Comment by patrickzib** [comment]
I see. I will look into the code later on today or tomorrow, too. I feel like the np.concantenate is not necessary. Will try to come up with a solution, too.

**Comment by patrickzib** [comment]
Final note: 

```python

def _der3(x: np.ndarray):
    m = len(x)
    der = np.zeros(m)
    der[1:-1] = ((x[1:-1] - x[:-2]) + ((x[2:] - x[:-2]) / 2.0)) / 2.0
    der[0] = der[1]
    der[-1] = der[-2]
    return der

```

This vectorized code would be faster than the ChatGPT-version but is slightly slower than the for-loop. So the for-loop is perfectly fine to go in. Well done!

patrickzib:
CI runtimes are way too high with > 5 hours. Not sure, what is wrong.

See also:
https://github.com/aeon-toolkit/aeon/pull/250

setup.cfg
```diff
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
```
patrickzib, line 15: Please undo these changes :)

patrickzib (approved):
Looks good to me. Impressive speedup

PR #253, opened 2023-04-04
[ENH] New data loaders

TonyBagnall:
the data loaders in datasetys/_data_io.py are a mess, with several changes back and forward, and lots of old code written to comply with python versions we dont have to support any more. 

The standard way is to load everything into nested datatframes, then allow the classifiers to sort it out. In the move away from nested data frames, I have started a rewrite. The basic approach I want is to load into a list of numpy, that can then be converted to 3D numpy if equal length, otherwise left as list of numpy. 

I still need to tidy this up a bit. Ive put in a separate file for now. But its clearly better. For example, when loading FaceDetection
http://timeseriesclassification.com/description.php?Dataset=FaceDetection
```python
from aeon.datasets._data_loader import _load_data, _load_header_info, load_from_tsfile
from aeon.datasets._data_io import load_from_tsfile_to_dataframe
file ="C:\\Data\\FaceDetection\\FaceDetection_TRAIN.ts"
import time
from aeon.datasets._data_io import load_from_tsfile_to_dataframe
start=time.time()
data, y = load_from_tsfile_to_dataframe(file)
print(" Dataframe load = ",time.time()-start)

start=time.time()
data, y, meta_data = load_from_tsfile(file)
print(" Numpy load = ",time.time()-start)
```
loading into data frames gives warnings 

image

and is much slower

image


TonyBagnall:
its about three times faster on the univariate data, but the most important thing is to move away from nested_univ

TonyBagnall:
also first loads direct to numpy and passes that to the classifier, second loads to DF and passes that to the classifier, which will then convert it to numpy. 
 Numpy rocket build =  213.1187698841095
 Dataframe rocket load =  723.0161702632904
thats three times slower, just because of base class nonsense! 

```python
    problem = "FaceDetection"
    start = time.time()
    data, y, meta_data = load_from_tsfile(f"{file}{problem}/{problem}_TRAIN.ts")
    print(" Numpy load = ", time.time() - start)
    print(" Shape of data = ", data.shape)
    print(" Building rocket .....")
    r = RocketClassifier()
    start = time.time()
    r.fit(data, y)
    print(" Numpy rocket build = ", time.time() - start)
    start = time.time()
    data, y = load_from_tsfile_to_dataframe(f"{file}{problem}/{problem}_TRAIN.ts")
    print(" Dataframe load = ", time.time() - start)
    print(" Building rocket .....")
    r = RocketClassifier()
    start = time.time()
    r.fit(data, y)
    print(" Dataframe rocket build= ", time.time() - start)

**Comment by patrickzib** [comment]
Nice. How would this affect classifiers that cannot handle lists of numpy arrays?

> The basic approach I want is to load into a list of numpy, that can then be converted to 3D numpy if equal length, otherwise left as list of numpy.

Not sure, if I understand correctly: does it automatically converted to 3d numpy array in case of equal length?

File: aeon/datasets/_data_loader.py [hunk 1]
```diff
@@ -0,0 +1,168 @@
...
+from aeon.datasets._data_io import load_from_tsfile_to_dataframe
+
+VALID_RETURN_TYPES = ["numpy3d", "numpy2d", "np3d", "np2d", "numpy_list", "nested_univ"]
+
+
+def _load_header_info(file):
+    """Load the meta data from a .ts file and advance file to the data.
+
+    Parameters
+    ----------
```
chrisholder, line 13: Comment style needing types for parameters and return

aeon/datasets/_data_loader.py
```diff
@@ -0,0 +1,168 @@
...
+    return meta_data
+
+
+def _load_data(file, meta_data):
+    """Load data from a file  with no header.
+
+    this assumes each time series has the same number of channels, but allows unequal
+    length series between cases.
+
+    Parameters
```
chrisholder, line 73: same change here comment style

aeon/datasets/_data_loader.py
```diff
@@ -0,0 +1,168 @@
...
+    return data, np.asarray(y_values), meta_data
+
+
+def load_from_tsfile(
+    full_file_path_and_name,
+    replace_missing_vals_with="NaN",
+    return_meta_data=True,
+):
+    """Load time series .ts file into X and (optionally) y.
+
```
chrisholder, line 146: Parameters + return

chrisholder (changes requested):

aeon/datasets/_data_loader.py
TonyBagnall: done

aeon/datasets/_data_loader.py
TonyBagnall: done

chrisholder (approved):

PR #261, opened 2023-04-05
[ENH] Convert HOG1DTransformer and DerivativeSlopeTransformer to use numpy arrays internally

TonyBagnall:
Switches HOG1DTransformer to use numpy internally. All it does at the moment is pull everything out pandas and uses numpy transforms. No doubt this could all be done more efficiently, but its a minor thing only used in one classifier

Switches DerivativeSlopeTransformer to use numpy by simply calling the function introduced in #252 Need to keep the wrapper for now because of shape dtw

#### Reference Issues/PRs
Part of #197


aeon/tests/_config.py
```diff
@@ -31,7 +31,6 @@
 
 
 EXCLUDED_TESTS = {
-    # issue when predicting residuals, see #3479
```
MatthewMiddlehurst, line 34: This would be useful to keep

aeon/transformations/panel/hog1d.py
```diff
@@ -4,111 +4,77 @@
...
+
         """
         # Get information about the dataframe
-        num_insts = X.shape[0]
-        col_names = X.columns
-        num_atts = len(X.iloc[0, 0])
-
+        n_cases, n_channels, series_length = X.shape
+        if n_channels > 1:
+            raise ValueError("HOG1D does not support multivariate time series.")
```
MatthewMiddlehurst, line 66: Any reason for this? Would it result in different length dimensions possibly?

aeon/transformations/panel/segment.py
```diff
@@ -393,6 +393,14 @@ class SlidingWindowSegmenter(BaseTransformer):
...
 
     Proposed in the ShapeDTW algorithm.
+
+    Examples
+    --------
+    >>> from aeon.datasets import load_unit_test
+    >>> from aeon.transformations.panel.hog1d import HOG1DTransformer
+    >>> data = np.array([[[1, 2, 3, 4, 5, 6, 7, 8]], [[5, 5, 5, 5, 5, 5, 5, 5]]])
+    >>> seggy = HOG1DTransformer(n_intervals=4)
+    >>> data2 = seggy.fit_transform(data)
```
MatthewMiddlehurst, line 403: I think this is the wrong class, the example is currently under SlidingWindowSegmenter

TonyBagnall: thanks, good spot

aeon/transformations/panel/hog1d.py
TonyBagnall: so its a 1D transform by name that produces a multivariate data set. It is not exactly clear how it would work for multivariate

MatthewMiddlehurst (approved):

PR #262, opened 2023-04-05
[ENH] Convert SlidingWindowTransformer to use numpy internally

TonyBagnall:
simple change, since the first line in fit was to convert to numpy!

#### Reference Issues/PRs
#197


aeon/classification/distance_based/_shape_dtw.py
```diff
@@ -253,9 +252,10 @@ def _preprocess(self, X):
         # the test/training data. It extracts the subsequences
         # and then performs the shape descriptor function on
         # each subsequence.
-        X = convert(X, from_type="numpy3D", to_type="nested_univ")
+        # X = convert(X, from_type="numpy3D", to_type="nested_univ")
```
GuzalBulatova, line 255: Is this line commented out on purpose or can be deleted?

setup.cfg
```diff
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
```
GuzalBulatova, line 15: Just curious, what is the reason for deleting this pytest options?

GuzalBulatova (approved):
Looks good to go! Only have a couple of questions, not blockers.

setup.cfg
TonyBagnall: there is no reason, its a mistake I keep forgetting to correct, thanks!

aeon/classification/distance_based/_shape_dtw.py
TonyBagnall: it can go, shapeDTW needs to be done after all these, thank you.

chrisholder (approved):

PR #269, opened 2023-04-06
[ENH] Convert PAA to use numpy internally

TonyBagnall:
switches PAA to numpy, simple again, since it internally converts everything to numpy. @patrickzib any thoughts on numba here? I've left the core logic alone for now, can njit it after

#### Reference Issues/PRs
#197

### DIFF
--- a/aeon/transformations/panel/dictionary_based/_paa.py
+++ b/aeon/transformations/panel/dictionary_based/_paa.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
 """Piecewise Aggregate Approximation Transformer (PAA)."""
-import pandas as pd
+import numpy as np
 
-from aeon.datatypes._panel._convert import from_nested_to_2d_array
 from aeon.transformations.base import BaseTransformer
 
 __author__ = "MatthewMiddlehurst"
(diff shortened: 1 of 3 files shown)


patrickzib:
It should be fine even without numba. :)

aeon/transformations/panel/dictionary_based/_paa.py
```diff
@@ -11,93 +10,89 @@
...
-            dataFrames.append(self._perform_paa_along_dim(pd.DataFrame(X[x])))
-
-        # Combine the dimensions together
-        result = pd.concat(dataFrames, axis=1, sort=False)
-        result.columns = col_names
+        channels = []
+        for i in range(n_channels):
+            channels.append(self._perform_paa_along_dim(_X[i]))
+        result = np.array(channels)
+        result = np.swapaxes(result, 0, 1)
```
chrisholder, line 78: This is kind of strange to do right after you create the array. Why does perform paa return something with the axis wrong? Shouldn't that be fixed so you don't need to do this. I don't really understand the internals of this algorithm so maybe this is unavoidable

chrisholder (approved):

PR #270, opened 2023-04-06
[ENH] convert SlopeTransformer to use numpy internally

TonyBagnall:
there is a theme here :)

This is the last of the ShapeDTW transformers that need conversion. As with others, there is a lot of scope for speeding up with vectorisation or numba, but we can leave that for another day, too many of these to do first! It pleases me each time I delete import pandas as pd though :)

Note I have also decreased the test problem size as this PR was failing because of it, see #272 

#### Reference Issues/PRs
#197 #98


aeon/transformations/panel/slope.py
```diff
@@ -51,54 +52,28 @@ def _transform(self, X, y=None):
...
+        collection of time series to transform
         y : ignored argument for interface compatibility
-            Additional data, e.g., labels for transformation
 
         Returns
         -------
-        Xt : nested pandas DataFrame of shape [n_instances, n_features]
-            each cell of Xt contains pandas.Series
-            transformed version of X
+        3D np.ndarray of shape = [n_instances, n_channels, series_length] collection
```
chrisholder, line 61: Maybe a weird question but why are we naming n_dims as n_channels? This isnt wrong but i know is classification and clustering we use n_dims so whats with the inconsistency

chrisholder (dismissed):

aeon/transformations/panel/slope.py
chrisholder: Ok upon inspecting some of the updated code turns out we're now using n_channels - good to know ill make sure to keep consistent with this

PR #273, opened 2023-04-07
[ENH] remove test_fit_transform_output

TonyBagnall:
this PR simply removes the Transformer test function 

test_fit_transform_output

this function, added in April 2022, imposes the twisted input/output matching of the Panel/Series/Table/Primitive restriction on what transformers can return based on their input. It is not needed imo, but it is also deeply flawed. For example if I want to transform series to series, if I input 3D numpy, it will not let me return 3D numpy, whether I set the tags to "Series" or "Panel" . Digging down, it is, as usual, riddled with exceptions and special cases. For example, at the heart is this

```python
def check_is_scitype(
    obj,
    scitype: Union[str, List[str]],
    return_metadata=False,
    var_name="obj",
    exclude_mtypes=AMBIGUOUS_MTYPES,
):
```
which is deeply hidden in registry 
```python
# mtypes to exclude in checking since they are ambiguous and rare
AMBIGUOUS_MTYPES = ["numpyflat", "alignment_loc"]
```
Ive never seen alignment_loc, and numpyflat is just a 2D numpy array. This kind of strict type testing with hidden exceptions is just bad code imo. On a wider note, transformers will be design return different things, depending onto the definition. If we assume series are input 3D numpy, dictionary transformers transform to a dictionary, series-to-series transform to time series (3D numpy) and series-to-vector go 3D -> 2D. They just need to be used correctly and exceptions thrown if used incorrectly. 

#### Reference Issues/PRs
all part of #197

### DIFF
--- a/aeon/transformations/tests/test_all_transformers.py
+++ b/aeon/transformations/tests/test_all_transformers.py
@@ -55,30 +55,6 @@ class TestAllTransformers(TransformerFixtureGenerator, QuickTester):
         if fit_empty_tag and remember_data_tag:
             raise AssertionError(msg)
 
-    def _expected_trafo_output_scitype(self, X_scitype, trafo_input, trafo_output):
-        """Return expected output scitype, given X scitype and input/output.
-
-        Paramaters
-        ----------
-        X_scitype : str, scitype of the input to transform
-        trafo_input : str, scitype of "instance"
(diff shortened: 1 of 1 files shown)


GuzalBulatova (approved):
Looks good to me, thank you Tony! Agree that the functionality is twisted and isn't worth the amount of complications this was adding.

PR #283, opened 2023-04-10
[BUG] `attrs` quick fix

### LINKED ISSUE #282: [BUG] `attrs` dependency discussion
The `attrs` dependency is no longer being installed on CI, we should add it as an actual dependency or remove its usage.
It is currently only used in 2 annotation estimators (GGS and IGTS) which I believe were developed by @lmmentel.

----

Issue for discussion of `attrs` as a dependency.

MatthewMiddlehurst:
Not sure if this will work, but crudely removes any imports for the `attrs` module.
Related #282.

### DIFF
--- a/aeon/annotation/ggs.py
+++ b/aeon/annotation/ggs.py
@@ -36,11 +36,10 @@ References
 
 import logging
 import math
-from typing import Dict, List, Tuple
+from typing import List, Tuple
 
 import numpy as np
 import numpy.typing as npt
-from attrs import asdict, define, field
 from sklearn.utils.validation import check_random_state
 
 from aeon.base import BaseEstimator
@@ -48,7 +47,6 @@ from aeon.base import BaseEstimator
 logger = logging.getLogger(__name__)
 
 
-@define
 class GGS:
     """
     Greedy Gaussian Segmentation.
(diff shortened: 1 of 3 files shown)


MatthewMiddlehurst:
There is probably a better way to handle this, I am not familiar with the removed package or its usage in these classes. Probably best for @lmmentel to suggest the best course.

MatthewMiddlehurst:
The codecov checks are back... yay, I guess?

lmmentel:
I would be for taking attrs as a dependency instead of removing the use.

MatthewMiddlehurst:
Ok, I have just added it as a soft dependency for now. Think it is worth discussion in the issue though.

MatthewMiddlehurst:
never mind, it was failing the soft-dep test anyway! would have to be an actual dependency.

TonyBagnall (approved):
so I guess we have this as a hard dep for now, but I cant really see the benefit it gives myself.

lmmentel (approved):

lmmentel:
Thanks for putting it in @MatthewMiddlehurst! 

@TonyBagnall this was an experiment in a bit more modern definition of estimator classes. [attrs](https://www.attrs.org/en/stable/index.html) has a pretty good docs where they explain what it is good for if you're interested.

I'm all for having a discussion whether it fits the overall design we're going for. I'm happy to make an argument for it but in the context of multiple inheritance madness we have now it shouldn't be that hard of a sell.

PR #292, opened 2023-04-13
[MNT] Added new logo

aiwalter:


MatthewMiddlehurst:
removing the current images would break stuff as it is and some of the colours are weird on current versions. working on it in another branch

MatthewMiddlehurst:
Versions with more consistent font colours and replacement for current usage were merged in the PR above, if there are issues with those it can be altered in a separate PR.

PR #294, opened 2023-04-13
[ENH] Convert Catch22 and Catch22Wrapper to use numpy internally

TonyBagnall:
very simple one this 

#197


aeon/transformations/panel/catch22.py
```diff
@@ -78,13 +76,20 @@ class Catch22(BaseTransformer):
...
     time-series analysis: the empirical structure of time series and their methods.
     Journal of the Royal Society Interface, 10(83), 20130048.
+
+    Examples
+    --------
+    >>> from aeon.transformations.panel.catch22 import Catch22
+    >>> import numpy as np
+    >>> data = np.array([[[1,2,3,4,5,6,7,8,9,10]],[[5,5,5,5,5,5,5,5,5,5]]])
+    >>> c22 = Catch22(replace_nans=True)
+    >>> data2 = c22.fit_transform(data)
```
chrisholder, line 86: Maybe better to just have:
```
    Examples
    --------
    >>> from aeon.transformations.panel.catch22 import Catch22
    >>> import numpy as np
    >>> data = np.array([[[1,2,3,4,5,6,7,8,9,10]],[[5,5,5,5,5,5,5,5,5,5]]])
    >>> c22 = Catch22(replace_nans=True)
    >>> c22.fit_transform(data)
    #Result output here
```
And add in the result output so people know what that looks like

aeon/transformations/panel/catch22.py
```diff
@@ -144,13 +149,13 @@ def _transform(self, X, y=None):
 
         Parameters
         ----------
-        X : 3D numpy array of shape [n_instances, n_dimensions, n_features],
-            input time series panel.
+        X : 3D numpy array of shape [n_instances, n_channels, n_features],
```
chrisholder, line 152: Up to you but I personally think having shape with (n_instances, n_channels, n_features) is better than using square brackets because square brackets imply a list.

Id also personally use the type X: np.ndarray of shape (n_instances, n_channels, n_features)  when talking about parameter types. This applies to all of the other comments as well.

chrisholder (approved):

chrisholder (approved):

PR #296, opened 2023-04-13
[ENH] Introduce list of numpy arrays data type for classification/regression/clustering

TonyBagnall:
as part of #110 we are going to stop supporting "nested_univ". The only real use case atm for "nested_univ" is for unequal length series. To achieve this functionality, we will use lists of numpy or either 1D (univariate, e.g. PLAID) or 2D (multivariate, e.g. JapaneseVowels). If multivariate, we assume the length is the same for any one instance.

This PR introduces the type in the datatypes module and provides a check function. The meta data calculation could probably be improved. This is first stage though. Next is to load into this format, then we provide support for it in classifiers and clusterers.


chrisholder (dismissed):

MatthewMiddlehurst:
Is allowing a list of 1d arrays a good idea? You would have to check in every estimator which uses the shape whether it is 1d or 2d.

TonyBagnall:
> Is allowing a list of 1d arrays a good idea? You would have to check in every estimator which uses the shape whether it is 1d or 2d.

you are right, at least until we sort out 2D with transformers, we could do what we do in the classifier base class and just reshape, but less efficient with them in a list

aeon/datasets/tests/test_data_io.py
```diff
@@ -31,7 +31,14 @@
...
 )
-from aeon.datatypes import check_is_mtype
+from aeon.datatypes import check_is_mtype, scitype_to_mtype
+
+# using this and not a direct import
+# in order to avoid mtypes that require soft dependencies
+MTYPE_LIST_PANEL = scitype_to_mtype("Panel")
+
+# Disabling test for these mtypes since they don't support certain functionality yet
+_TO_DISABLE = ["pd-long", "pd-wide", "numpyflat", "df-list", "pd-multiindex"]
```
MatthewMiddlehurst, line 41: Are these used for anything?

aeon/datatypes/_panel/_check.py
```diff
@@ -126,6 +126,51 @@ def check_dflist_panel(obj, return_metadata=False, var_name="obj"):
...
+
+    if len(bad_inds) > 0:
+        msg = f"{var_name}[i] must np.ndarray, but found other types at i={bad_inds}"
+        return _ret(False, msg, None, return_metadata)
+
+    bad_inds = [
+        i
+        for i in range(n)
+        if not isinstance(np_list[i], np.ndarray) and np_list[i].ndim < 3
+    ]
```
MatthewMiddlehurst, line 154: If we are sticking with 2d for now, may want to change this

aeon/datasets/tests/test_data_io.py
TonyBagnall: not to my knowledge, this was just moved from somewhere else

MatthewMiddlehurst: Should probably remove it if its not being used in anything. Cant see it being removed anywhere in the PR.

chrisholder (approved):

PR #300, opened 2023-04-14
[ENH] combine test_returns_self with test_fit_updates_state

TonyBagnall:
test_all_estimators calls fit a very large number of times. Whislt I appreciate having clear tests that do things independently, the state of testing is so slow, I think its worth combining some of them to reduce the calls to fit. 

test_returns_self takes about a minute and calls fit on everything, testing if self is returned. I have rolled this into test_fit_updates_state, using the call to fit there to test self and to print the same error message. This takes it locally down from 13mins 6 sec to 11m 43s (not including loading). Every little helps!

there are other similar compromises we could make here to reduce the number of fit calls. Real pig is forecasting though, need to address that at some point

### DIFF
--- a/aeon/tests/_config.py
+++ b/aeon/tests/_config.py
@@ -142,7 +142,6 @@ EXCLUDED_TESTS = {
         "test_inheritance",
         "test_create_test_instance",
     ],
-    "SAX": "test_fit_transform_output",  # SAX returns strange output format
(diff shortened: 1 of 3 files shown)


chrisholder (approved):
I think this is a great idea - checking if it returns self can be inferred by other tests passing

PR #302, opened 2023-04-15
[ENH] distances module update

chrisholder:
The previous distances module had problems I had been aware of for some time but didnt want to change as I knew the pr would be very large and hard to get in the old code base with all the constraints. This pr attempts to address some of the following problems:
1) The numba distances were being compiled every time they were called (as they were creating numba closures every time they were called so that polymorphism could be used). This led to large bottlenecks in performance.
2) The fastmath flag in numba was unable to be used because the bounding matrices used infinite and finite values which caused bounding windows to be ignored when using the fastmath flag
3) Twe had a bug in its implementation.
4) msm and twe were seg faulting (for unknown reasons) in very large datasets
5) The API using distance_factory was not very intuitive
6) Optimisations were sacrificed for the sake of using an oop based design.

This large pr completely overhauls the entire module and moves away from classes and instead uses functions. This means numba can be used end to end without the need to recompile every time the function is called as they now can be cached. 

Additionally, large optimisations have been made within the code to achieve the best performance.

Furthermore, the api has been greatly simplified (while expanded) to be easier for users to use.

Finally, all the distances now support univariate, multivariate and unequal length properly (the previous version only some distances properly supported unequal length).

There is a lot but Ill attempt to summaries it in a couple points:

1) Reimplemented all the distances using functions so they can be cached.
2) Implemented pairwise, single to multiple ts distances, and multiple ts to multiple ts distances utility function for each specific distance. This has allowed for optimisations to be made per specific distance for large computations.
3) Cleaned up the api with these utility functions and additionally added a lot more functions that just say what they do rather than hiding their logic inside one function that takes a string parameter to tell it what to do.
4) Fixed minor bugs in TWE. I outline the exact bug further down.
5) Added multivariate support for all distances and ensured unequal length series works for all distances. This includes adding a new algorithm for msm (dependent msm) that was proposed to handle multivariate (while still keep indepdent msm).
6) Fixed the bug that was causing seg faults for very large datasets (I hope it near impossible to debug because I think it's in the c code generated by numba so swapped some stuff around and try to clean up the areas it could seg fault).

No

I added all the tests and I checked for equality between the new and the old distances which were achieved (barring the distances that had minor bugs that have now been fixed). 

Additionally, I updated the tests for elastic ensemble as it wasn't passing anymore due to the bugs that have now been fixed in the distances.

#### Any other comments?

The implementations for Wddtw, erp and edr do not give the same result as the old version. The reason for this is the previous version used euclidean distance whereas the new versions use the squared euclidean distance. This doesn't impact the actual quality of the distance result (I do not have the paper on me right now happy to dig it up if someone wants to see it that did a study on the impact of using L1 L2 L3 L4 etc distance with elastic distance measure and found L1 and L2 are equivalent). This simply saves the computation of having to do a square root every iteration.

Bugs fixes explained:

### TWE match bug

Twe had a bug in the 'match' stage where it wasn't considering the previous values. This was the old code:

```
                match = (
                    cost_matrix[i - 1, j - 1]
                    + match_same_euclid_dist
                    + match_previous_euclid_dist
                    + (nu * (2 * abs(i - j)))
                )
```
Note the final line (nu * (2* abs(i-j))). This was an error when implementing as match should be:

```
                match = (
                    cost_matrix[i - 1, j - 1]
                    + match_same_squared_d
                    + match_prev_squared_d
                    + nu * (abs(i - j) + abs((i - 1) - (j - 1)))
                )
```
Where the previous value abs((i-1) - (j - 1))) should be considered. This has now been fixed.

##### For all contributions


chrisholder:
Just some numbers in terms of speed up and performances the new distances gain:

The top table shows the raw seconds gained in execution. The bottom table shows the % increase in speed.

For univariate series compared to the old distances:
image

For multivariate series compared to the old distances:

image

PR #305, opened 2023-04-18
[ENH] Bounding matrix for distances module

chrisholder:

This is the first of several pull requests that update the distance module. In this particular PR, the bounding matrix code has been updated to use boolean values instead of finite and infinite values. There are two reasons for this update:

1) Checking if a value is finite or infinite is slower than simply checking if a boolean is true or false.
2) The use of infinite values prevents us from using the fastmath flag in numba, as it is unable to check if a value is infinite or finite when fastmath is active.

Additionally, the option to pass a custom bounding matrix and an Itakura parallelogram bounding matrix parameter have been temporarily removed. These will be added back in later, but it would require a bigger refactor that I will do when introducing the updated distances. I plan to add the option to pass a custom bounding matrix back, but the Itakura parallelogram bounding matrix parameter will remain permanently removed as it adds unnecessary complexity for something that very few people will use. If users want to use it, they can pass in a custom bounding matrix.

As an example, the old bounding matrix produced something that looked like this:
```
[
[ 0. 0. inf inf inf],
[ 0. 0. 0. inf inf],
[inf 0. 0. 0. inf],
[inf inf 0. 0. 0.],
[inf inf inf 0. 0.]
]
```

The new bounding matrix now produces:

```
[
[ True  True False False False], 
[ True  True  True False False], 
[False  True  True  True False], 
[False False  True  True  True], 
[False False False  True  True]
]
```

Many of the changes to the files are simply refactoring comments/parameters to use the new bounding most of the files outside of _bounding_matrix.py and tests/test_bounding.py can be ignored.

Yes tests/test_bounding.py

##### For all contributions


TonyBagnall:
lgtm

PR #308, opened 2023-04-19
[ENH]  Update euclidean and squared distance

chrisholder:
#### Reference Issues/PRs
This pr depends on pr #305

This pr begins the refactor of the distances starting with euclidean distance and squared distance.

The changes improve the efficiency for 1d, 2d and 3d. Furthermore, it adds the new api for the distances from_single_to_multiple_distance (from a single time series to a dataset of time series), and from_multiple_to_multiple_distance (from a dataset fo time series to a dataset of time seres).

It also improves the test coverage by adding new tests for 3d datasets and unequal length tests.

Yes the existing tests have been updated and further testing has been added to ensure coverage.

##### For all contributions


aeon/distances/_distance.py
```diff
@@ -732,109 +737,6 @@ def twe_distance(
...
-        If the value of x or y has more than 2 dimensions.
-        If a metric string provided, and is not a defined valid string.
-        If a metric object (instance of class) is provided and doesn't inherit from
-        NumbaDistance.
-        If a resolved metric is not no_python compiled.
-        If the metric type cannot be determined.
-
-    Examples
-    --------
-    >>> import numpy as np
```
TonyBagnall, line 824: could you explicitly import the distance in the example so it can be cut and paste into external code?

aeon/distances/_euclidean.py
```diff
@@ -1,87 +1,240 @@
...
+    distances = np.zeros(n_instances)
+
+    for i in range(n_instances):
+        distances[i] = euclidean_distance(x, y[i])
+
+    return distances
+
+
+@njit(cache=True, fastmath=True)
+def euclidean_from_multiple_to_multiple_distance(
```
TonyBagnall, line 179: isnt this the same as pairwise?

aeon/distances/_euclidean.py
```diff
@@ -1,87 +1,240 @@
...
-            distance += difference * difference
-    return np.sqrt(distance)
+    if y.ndim == 3 and x.ndim == 3:
+        return _euclidean_from_multiple_to_multiple_distance(x, y)
+    if y.ndim == 2 and x.ndim == 2:
+        _x = x.reshape((x.shape[0], 1, x.shape[1]))
+        _y = y.reshape((y.shape[0], 1, y.shape[1]))
+        return _euclidean_from_multiple_to_multiple_distance(_x, _y)
+    if y.ndim == 1 and x.ndim == 1:
+        _x = x.reshape((x.shape[0], 1, 1))
```
TonyBagnall, line 223: this looks like its reshaping a series so that n_instances == series_length, i.e. creating n_instances series of length 1? I see in the comment its not, I guess there is a good reason to transform to (series)length,1,1)?

aeon/distances/_euclidean.py
```diff
@@ -1,87 +1,240 @@
...
+    return distances
+
+
+@njit(cache=True, fastmath=True)
+def euclidean_from_multiple_to_multiple_distance(
+    x: np.ndarray, y: np.ndarray
+) -> np.ndarray:
+    """Compute the euclidean distance between two sets of time series.
+
+    If x and y are the same then you should use euclidean_pairwise_distance.
```
TonyBagnall, line 184: ah I see the difference now :)

TonyBagnall (dismissed):

TonyBagnall:
so to clarify, pairwise is for when but sets of series contain the same number of instances, whereas from_multiple_to_multiple_distance is for when the two collections have different number of instances? Why do we need separate methods for these operations?

aeon/distances/_distance.py
chrisholder: This is removed code but in all of the new code I have ensured to import it properly so it can be copied and pasted.

aeon/distances/_euclidean.py
chrisholder: This is a very good spot. I just checked and this was in the old code as well. I think it was just a typo when I added univariate support to pairwise since it was a very last minute addition because originally if you gave a 1d array to pairwise it would throw an error.

This should be 
`_x = x.reshape((1, 1, x.shape[0]))`

chrisholder:
So the only difference in terms of code between from_multiple_to_multiple and pairwise is in the pairwise because you 
know it's to itself you can make this optimisation:
```
def _euclidean_pairwise_distance(X: np.ndarray) -> np.ndarray:
    n_instances = X.shape[0]
    distances = np.zeros((n_instances, n_instances))

    for i in range(n_instances):
        for j in range(i + 1, n_instances):
            distances[i, j] = euclidean_distance(X[i], X[j])
            distances[j, i] = distances[i, j]

    return distances
```

 "distances[j, i] = distances[i, j]" because you know the matrix is reflected you dont have to compute the values again (therefore you do half the computation).

The main reason for the split is because I didnt want pairwise to take a performance hit and greatly increase the complexity just for the sake of utility. If we were to add multiple_to_multiple in then arguably we should also add single_to_multiple because it's just another case of computing pairs. 

This would greatly bloat the pairwise_distance function with lots of conditional checks and furthermore bloat the documentation for that one function greatly.

Im happy to move it in so you only have pairwise_distance and distance (and single_to_multiple and multiple_to_multiple exist under pairwise_distance) but I think it's overcomplicating thing when you can just have 3 separate distinct functions that all just do one thing.

TonyBagnall:
> 

ah ok yes that makes sense, thanks

TonyBagnall (approved):

PR #310, opened 2023-04-20
[ENH] RDST transformer

baraline:

Implementation of the RDST transformer from convst package, with additional optimization and unit tests.

No additional dependency

Conformity with the standard of Aeon for code documentation and tests, (I followed guidelines, used pre commit etc....)

All functions should be unit tested and passing, at least locally.

#### Any other comments?

- How do we specify in the tags that y not mandatory but accepted ? `"requires_y": False` ?
- Should the added utils in `utils/numba/rdst_utils.py` be dispatched to other files ?
- I have forced 64bit dtypes in the numba code to avoid mixing 32 and 64 bit dtype which can cause error with numba. Is there a way to set this with more flexibility depending on the system ?

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc).
- [X] For new estimators, I've added the estimator to the online documentation.


MatthewMiddlehurst:
Thanks for creating the PR, it looks like a very nice contribution! Had a quick look through and will give it a more thorough look later.

To answer your questions:
    

> How do we specify in the tags that y not mandatory but accepted ? "requires_y": False ?

`"requires_y": False` is the tag yes, the methods should accept a `y` argument even if it is not used for pipelines etc.
    
> Should the added utils in utils/numba/rdst_utils.py be dispatched to other files ?

It depends on whether you think they are generally applicable, I guess. There are current a stats and general file where they can go, or if you don't think it would fit in either of those it can go in a new file. 

If they are only applicable to RDST, I would just have them in the transformer file rather than a rdst utils file.

>I have forced 64bit dtypes in the numba code to avoid mixing 32 and 64 bit dtype which can cause error with numba. Is there a way to set this with more flexibility depending on the system ?

Not currently I think. Where is this happening? The only error I can really remember regarding dtypes for stuff I have implemented is when you have different returns and you have to ensure they are of the same type.

MatthewMiddlehurst:
It seems like you are getting different results on MacOS for some reason based on the test error.

baraline:
Hi Matthew, 

> It depends on whether you think they are generally applicable, I guess. There are current a `stats` and `general` file where they can go, or if you don't think it would fit in either of those it can go in a new file.
> If they are only applicable to RDST, I would just have them in the transformer file as private functions rather than a rdst utils file.

In that case, I will dispatch them into the according files in a future commit.

> Not currently I think. Where is this happening? The only error I can really remember regarding dtypes for stuff I have implemented is when you have different returns and you have to ensure they are of the same type.

Indeed, after testing other cases (arithmetic operation between arrays etc...) do not raise any error. I guess it is then only a matter of desired precision and lower/upper bounds for the values. I'll try to change it to resort to 32 bit where 64bit is not necessary.

> It seems like you are getting different results on MacOS for some reason based on the test error.

One test fails due to the shapelet length array being empty :
`FAILED aeon/tests/test_all_estimators.py::TestAllEstimators::test_non_state_changing_method_contract[RandomDilatedShapeletTransform-TransformerFitTransformPanelUnivariateNumpyWithClassYOnlyFit-transform] - ValueError: Shapelet lengths array is empty, did you give shapelets lengths superior to the size of the series ?`

It must be because the input data is either smaller than the default size (11), I'll check which dataset are used in these tests.

The MacOS one is failing because the output results are different for basic motion and unit test datastet. Could the numpy random generator (with seed) be different in MacOS compared to Unix and Windows ? I need to look into it, but that seems strange.

MatthewMiddlehurst:
Would it be possible to lower the length used if the series are too short? Super short series are an edge case and rather unrealistic but its good to account for it.

baraline:
Time series length is now supported a length of 2. The default is set to `min(series_length,11)` if shapelets_length param is left to None.

Issues with float32 tests seems to be resolved after converting dtypes to 32bit (which, for integer, is equivalent to a C int for 32bit rather than a long for 64bit cf https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.intc )

Removed the test on basic motion that was failing on Mac, the other tests checking the transformation and feature extraction are OK. Should we raise an issue to try to re-introduce them in a future PR ?

aeon/transformations/panel/dilated_shapelet_transform.py
```diff
@@ -0,0 +1,895 @@
...
+        "y_inner_mtype": "numpy1D",
+        "requires_y": False,
+        "capability:inverse_transform": False,
+        "handles-missing-data": False,
+    }
+
+    def __init__(
+        self,
+        max_shapelets=10_000,
+        shapelet_lengths=None,
```
MatthewMiddlehurst, line 145: Think the docs above need to be updated to reflect the change of default to None.

aeon/transformations/panel/dilated_shapelet_transform.py
```diff
@@ -0,0 +1,895 @@
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        if parameter_set == "default":
+            params = {"max_shapelets": 10}
```
MatthewMiddlehurst, line 313: Not a change request, but an alternative to using the minimum series length as the shapelet length if its too small is to just set a very small shapelet length here for testing.

aeon/transformations/panel/dilated_shapelet_transform.py
```diff
@@ -0,0 +1,895 @@
...
+
+        self._check_input_params()
+
+        self._n_jobs = check_n_jobs(self.n_jobs)
+        set_num_threads(self._n_jobs)
+
+        if y is None:
+            y = np.zeros(self.n_instances)
+        else:
+            y = LabelEncoder().fit_transform(y)
```
MatthewMiddlehurst, line 195: Not a change request, just interested on the actual impact of excluding labels. May be useful to include in the docstring if there is one.

Is it just a semi-guided random search vs completely random?

MatthewMiddlehurst (changes requested):
I think it would be good to create an issue about re-adding the test, it will take a bit of exploration to figure out what is happening with MacOS but good to have a note of it.

`min(max(series_length//2,2),11)` looks better to me, don't really want full length shapelets.

aeon/transformations/panel/dilated_shapelet_transform.py
MatthewMiddlehurst: This is only if you want to throw an error for invalid length inputs rather than adjust to smaller series.

MatthewMiddlehurst:
Looking good! Only really a few small comments from my side, 

I am just going to trust that all the inner working i.e. shapelet extraction, distances are working fine, but if you want me to take a look at any particular function let me know!

aeon/transformations/panel/dilated_shapelet_transform.py
baraline: Close to it, the only impact of class during shapelet extraction is to guide the choice of the threshold for the shapelet occurrence feature:

- With class, another sample of the same class as the one used for shapelet value extraction is used (or the same sample if there is only one sample per class)
- Without class, the choice of sample is random.

This information is given in the docstring of the transformer as :
`
Threshold is randomly chosen between two percentiles of the distribution of the distance vector between the shapelet and another time series, of the same class if they are given during fit, or to another random sample.
`

Is it clear enought ?

baraline: Changed for clarity to:
`
- Threshold is randomly chosen between two percentiles of the distribution
        of the distance vector between the shapelet and another time series. This time
        serie is drawn from the same class if classes are given during fit. Otherwise,
        a random sample will be used. If there is only one sample per class, the same
        sample will be used.
`

baraline:
Requested changes are integrated in the last commit. I'll create the issue for the macOS test and try to have a look at it later. Next PR on this subject will be about implementing the classifiers.

Concerning the inner workings, i'll try to run a validation on the UCR archive to see if there is any deviation from original results, but the unit tests on the extraction and generation of shapelets seems to be OK for now. 

It would also like to make a version that support variable length inputs as in the original package. Any specific requirements for variable length transformers ? (e.g. make them into a new file with a prefix/suffix etc...) ?

MatthewMiddlehurst (approved):
Thanks for the contribution.

I think the best way to deal with unequal length is to include it in the same class really, unless it substantially changes the algorithm.

TonyBagnall:
this is great, thanks. I'm working on properly enabling unequal class, we can add this later once it is stable.

PR #329, opened 2023-04-24
[ENH] Removes nested dataframes from shape dtw

TonyBagnall:
This removes the nested_univ from ShapeDTW (also fixes a couple of bugs I found there with "compound" option) and also removes doc string mentions to using nested Pandas dataframes. There is no usage of the nested_univ input type in classification, regression or clustering. There are still some transformations to do, and some testing to revise

#### Reference Issues/PRs
Part of #110

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,14 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
+    --timeout 600
+    --cov aeon
+    --cov-report xml
+    --cov-report html
+    --showlocals
+    --matrixdesign True
+    -n auto
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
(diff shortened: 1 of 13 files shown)


MatthewMiddlehurst:
Test changes in setup.cfg got pushed 🙂 

Are there results from the previous version to compare against for ShapeDTW in the expected results file?

chrisholder (approved):

setup.cfg
```diff
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
-    --timeout 600
-    --cov aeon
```
chrisholder, line 17: you need to put these lines back in

TonyBagnall: how many times have I done that! We should try fix it :)

TonyBagnall:
> Test changes in setup.cfg got pushed 🙂
> 
> Are there results from the previous version to compare against for ShapeDTW in the expected results file?

dont think so, couldnt find any

MatthewMiddlehurst:
looks like there is. results are in tests/_expected_outputs.py

image

MatthewMiddlehurst (approved):

Issue #111, opened 2023-02-25
[ENH] what to do with the param_est module?


TonyBagnall:
**Is your feature request related to a problem? Please describe.**
there is a module in the toolkit called param_est. It was introduced in August 2022. It seems to be a wrapper for statsmodels.tsa.stattools import acf

      acf_series, confint = acf(
            x=X,
            adjusted=adjusted,
            nlags=nlags,
            fft=fft,
            missing=missing,
            alpha=p_threshold,
        )
  
 with an OO infrastructure built to support it,  

SeasonalityACF->BaseParamFitter->BaseEstimator

it may be people like this kind of wrapper, but personally I don't understand why you wouldn't just call statsmodels directly. Its not used anywhere else in the toolkit that I can find and there is no associated notebook. I think it should either move to the forecasting module or it should just go, I dont mind which personally (although I would vote removal).

ltsaprounis:
Sorry for the late comment, just saw the PR. 

I don't see why it needs its own base class etc. For example we have this: 
https://github.com/aeon-toolkit/aeon/blob/5683099a306118b87462f09c7ae72fa9bed07d4c/aeon/utils/seasonality.py#L17 
since forever and it's just a function, as it should be imo. I'd be in favour of removing the entire module as well.

GuzalBulatova:
Maybe it had something to do with datatypes? Don't see how exactly though... I agree that this wrapper should go @TonyBagnall , I don't understand the motivation behind it.

TonyBagnall:
ok, I moved it but now I will delete it. Its OO gone mad, and I'm saying this as a Java programmer!

PR #330, opened 2023-04-24
[ENH] refactor param_est to live in forecasting module

TonyBagnall:
#### Reference Issues/PRs
 see #111 

param_est is designed entirely for forecasting: the base class doc string says         
        Parameters
        ----------
        X : time series in aeon compatible data container format
                Time series to which to fit the forecaster in the update.

as I say in #111 I dont really get it, but IMO if it stays it should live in forecasting.


GuzalBulatova (approved):
Looks good to me, I'm fine with deleting it altogether too.

PR #339, opened 2023-04-26
[MNT] Update file change action and cancel workflow 

MatthewMiddlehurst:
Replaces the current `cancel.yml` file with a step in `tests.yml` (see https://github.com/sktime/sktime/issues/3999). Hopefully this will cancel previous runs prior to running the new workflow. This could be further guaranteed by adding a `needs:` to the next few steps.

Replaces the action used to find changed files for pre-commit. https://github.com/trilom/file-changes-action does not seem to be actively maintained and is raising deprecation warnings.  https://github.com/tj-actions/changed-files works just as well for our purpose (hopefully) and is still being developed.


lmmentel:
Looks good overall. A few small things you might want to consider before merging:

1. there's a warning from code-quality action, that might be worth addressing 
```
code-quality
The `python-version` input is not set. The version of Python currently in `PATH` will be used.
```

2. an alternative to third party GHA for canceling jobs it might be worth considering [concurrency option](https://docs.github.com/en/actions/using-jobs/using-concurrency) that's native to GHA. Here's how sklearn uses it [wheels,.yml](https://github.com/scikit-learn/scikit-learn/blob/144324b328bf1767d027d4f7f19572b701928b55/.github/workflows/wheels.yml#L20)

MatthewMiddlehurst:
Good ideas. Added a Python version to the code quality test and swapped to the GHA concurrency.

Difficult to figure out if these are working as intended, as the old workflow is still active. Can just revert if it doesn't I guess.

lmmentel (approved):
Looks good :+1: Let's monitor the jobs being cancelled to verify concurrency config works.

PR #340, opened 2023-04-26
[MNT] Consolidate config files and add new pre-commit functionality

MatthewMiddlehurst:
Adds some more checks to `pre-commit-config.yaml` and moves configuration currently in `setup.cfg` to `pyproject.toml` and `pre-commit-config.yaml`. Some options are no longer present, but from what I have checked these are either default values or deprecated for our method of building. 

Added checks (description from pre-commit page):
check-ast - simply checks whether the files parse as valid python. 
check-docstring-first - checks a common error of defining a docstring after code. 
check-toml - checks toml files for parseable syntax. 
fix-byte-order-marker - removes utf-8 byte order marker. 
fix-encoding-pragma - removes # -*- coding: utf-8 -*- from the top of python files. 
name-tests-test - verifies that test files are named correctly. 

Adds a hook for [mypy](https://github.com/python/mypy), which is a type checker which will check that variables and functions are being used correctly based on type hints. As we are adding more type hints to the repo, this will help with quality control.


TonyBagnall:
I'm in principle fine with the new tests, but don't really understand the other set up changes. @lmmentel any thoughts?

TonyBagnall (approved):

TonyBagnall (approved):

TonyBagnall (approved):

Issue #45, opened 2023-02-08
[DOC] Notebooks structure and content in examples directory


TonyBagnall:
I would like to take the opportunity of the reboot to restructure and partially rewrite the notebooks in the directory examples. There is an issue of whether we need separate notebooks and webpages, lets discuss, I dont feel strongly, but whichever, the structure needs looking at  (scikit learn has an examples and docs dir)

Currently they look like this, just a dumping ground with inconsistent naming. such as AA_datatypes_and_datasetds
![image](https://user-images.githubusercontent.com/9594042/217547519-9f01c196-f159-4767-ad1c-9d993f3bdf5c.png)

I think at the root dir we should have some basic info and then one notebook per module, e.g. 

getting_started.ipynb
data_input.ipynb
annotation.ipynb
classification.ipynb
clustering.ipynb
distances.ipynb
forecasting.ipynb
networks.ipynb
regression.ipynb
transformations.ipynb

then subfolder more detailed ones by module. 

with links to subdirectories with more detail.

PR #346, opened 2023-04-27
& TonyBagnall [DOC] Remake `get_started` page

MatthewMiddlehurst:
Completely remakes the `get_started` page. Converts the `glossary` page to .md and adds some placeholder terms.


TonyBagnall:
I have a few changes to make actually, I will do so on this PR

docs/getting_started.md
```diff
@@ -0,0 +1,236 @@
...
+Additionally, it provides numerous algorithms for {term}`time series transformation`,
+altering time series into different representations and domains or processing
+time series data into tabular data.
+
+The following provides introductory examples for each of these modules. In the examples
+used the datatypes most commonly used for the task, but a variety of input types for
+data are available. See [here](/examples/datasets/AA_datatypes_and_datasets.ipynb) for
+more information on input datatypes. For more information on the variety of estimators
+available for each task, see the [API](api_reference) and [examples](examples) pages.
+
```
hadifawaz1999, line 35: Should we probably add a section on loading data with maybe a visual example ? Maybe its better to do that before the user goes into Forecasting, classification etc. ?

docs/glossary.md
```diff
@@ -0,0 +1,141 @@
...
+
+Time series collection
+    todo
+
+Univariate time series
+    A single time series. While univariate analysis often only uses information
+    contained in the series itself, univariate time series regression and forecasting
+    can also include {term}`exogenous` data.
+
+Multivariate time series
```
hadifawaz1999, line 62: Given the sorting attribute, the Multivariate TS definition comes way before the Time series definition, maybe linking the term 'time series' to its definition will make the reading easier to follow

TonyBagnall:
See #45

docs/getting_started.md
TonyBagnall: I agree, but it will be easier to do this once this is in. The whole data loading examples revamp will happen once #355 is in. Currently, there is a lot of baggage here. The notebook AA-datatypes_and_datasets is a huge, confusing mess and is going in the doc revamp. There is also a division in loading between forecasting and machine learning/dataframes vs arrays

docs/glossary.md
TonyBagnall: yes agreed, again it will be easier to do once this page is in the package.

docs/getting_started.md
MatthewMiddlehurst: I think this would better fit in the user guide page (i.e. longer, in-depth notebook). Currently, this page is just an introduction to each module, and I think it would be better that way.

There definitely should be a guide for datatypes and dataloading, however. As Tony says, there is a lot of baggage there.

docs/glossary.md
MatthewMiddlehurst: Yeah, these all need to be reworked really. But i think that is better for another PR, I just want the terms in there so I can create links in the getting started page.

I can create an issue so it is not forgotten.

docs/getting_started.md
MatthewMiddlehurst: We give a small intro to an accepted data type for each module (1d pd.Series and 2d/3d numpy)

hadifawaz1999: Yeah you're right, we can re-visit this after the page is in

docs/glossary.md
hadifawaz1999: Its done @MatthewMiddlehurst #361

hadifawaz1999 (approved):

Issue #187, opened 2023-03-10
[BUG] Deep learning classifier and regression test exclusions


MatthewMiddlehurst:
Some issues still remain with the classification deep learners after #134. A number of tests are currently skipped, and the `callbacks` parameter is present but its functionality is commented out.

Skipped tests in `aeon/tests/_config.py`:
>  "test_fit_idempotent",
>   "test_persistence_via_pickle"
>   "test_save_estimators_to_file"
>   "test_fit_does_not_overwrite_hyper_params"
>   "test_methods_have_no_side_effects"

TonyBagnall:
all part of test issues, about hald the classifiers are excluded from all tests! I think we will need to get these working asap

hadifawaz1999:
should we close this @TonyBagnall @MatthewMiddlehurst ?

MatthewMiddlehurst:
These still remain for various reasons. Most are done, but we should tackle each of these eventually.

```
EXCLUDE_ESTIMATORS = [   
    "TapNetRegressor",
    "TapNetClassifier",
    "ResNetClassifier",  # known ResNetClassifier sporafic failures, see #3954
]

EXCLUDED_TESTS = {
    # InceptionTimeClassifier contains deep learners, it isnt one itself, so still
    # exclude
    "InceptionTimeClassifier": [
        "test_fit_deterministic",
        "test_persistence_via_pickle",
        "test_save_estimators_to_file",
    ],
    "InceptionTimeRegressor": [
        "test_fit_deterministic",
        "test_persistence_via_pickle",
        "test_save_estimators_to_file",
    ],
   "CNNNetwork": "test_inheritance",  # not a registered base class, WiP, see #3028
}
```

TonyBagnall:
ResNetClassifier is still failing a lot of tests, just seeing if we can include the others now

TonyBagnall:
so InceptionTime fails because it inherits from BaseClassifier rather than BaseDeepClassifier, I'd forgotten that. I dont think this is a problem. ResNet and TapNet could possibly be made to work properly. I think given the changes, its probably best to close this issue and https://github.com/aeon-toolkit/aeon/issues/173 and open a new, updated one

PR #348, opened 2023-04-27
[MNT] Remove test_fit_does_not_overwrite_hyper_params and test_methods_have_no_side_effects from config for most deep learners CNN MLP Encoder and FCN classifiers

hadifawaz1999:
This is a followup to #338 and #315 as mentioned in #187


TonyBagnall (approved):
looks good to me. these ones are not likely to have intermittent fails I think

PR #350, opened 2023-04-28
[MNT] Fix workflow concurrency

MatthewMiddlehurst:
#339 did not cancel the previous run on main when a new PR was merged. This attempts to fix that.


MatthewMiddlehurst:
This does work for PRs at least, I read up a little bit more and I imagine  ``${{ github.workflow }}-${{ github.run_id }}`` is always going to be unique. ``${{ github.workflow }}-${{ github.ref }}`` should cancel all runs for this workflow on the same branch, which is what we want I'm pretty sure if ``${{ github.workflow }}-${{ github.head_ref }}`` fails.

Issue #118, opened 2023-02-26
[ENH] test functions using baked in datasets


TonyBagnall:
During what I thought would be a minor tweak refactoring the loading #109 I realised that a huge number of test functions load one of the baked in classification data sets, assume it is a dataframe and use it to test output structure. Obviously baked in problems are good for testing correctness, but I dont think it is a good idea to use them just to check, for example formatting of input/output. There are functions to generate random datasets, and it would be better to use them if the tests are not at all dependent on the values in the datset. I think basic_motion is loaded several hundred times during testing. This will carry a (probably small) overhead, and I think generally file loading/writing should be minimised in CI. I may be wrong and its no problem. If its not worth the effort I will just leave it as is. Appreciate thoughts on the issue.

Here is an incomplete list of test functions that use one of the TSC single problem loaders.

![image](https://user-images.githubusercontent.com/9594042/221425401-fac08c45-7e3b-4797-ac57-4d5cb71ca533.png)
![image](https://user-images.githubusercontent.com/9594042/221425624-2c2710f3-65d3-4247-9848-e2f80635926f.png)

TonyBagnall:
I dont think this is a problem any more, since we have purged nested_univ, seems very limited benefit to gain from large changes

Issue #116, opened 2023-02-26
[ENH] Benchmarking module


TonyBagnall:
**Is your feature request related to a problem? Please describe.**
The majority of the benchmarking module has sat there for two or three years with no one maintaining or developing it. The majority of the code is an object model of Task, Strategy, Orchestrator, Evaluator, Datasets, Results. I think we should just remove it all, and adopt a much simpler model for benchmarking that focussing on running experiments, storing results and comparing results to other benchmarks provided. In support

1. The code is not well documented, the doc strings are all one liners
2. The code is disconnected from the rest of the tool kit. Not a bad thing per se, but it indicates no one has used it much
3. There is a lot of testing that involves writing and reading, and these tests sometimes take a while

**Describe the solution you'd like**
I would like to strip out the whole Strategy model and replace it with simple functions, with code to reproduce forecasting, classification, regression and clustering, along the lines started here

https://github.com/sktime/sktime/issues/1123

one of the reasons Ive not got round to this is the weight of all this legacy code. 

**Describe alternatives you've considered**
If someone wants to adopt this module as it is, I'd be happy

patrickzib:
I have not been using it.

IMO: It might be worth to consider removing it and moving the new module to an own repository? This way the documentation could also be improved.

TonyBagnall:
I think I would like some benchmarking code here, not least because it will be a USP, we are quite good at benchmarking after all :) Ultimately, I would like something equivalent to our java MultipleEstimatorEvaluator module, but in a much more functional, interactive way. So given your own set of results, you can pull results of other classifiers off tsc.com, draw CDs, plot results, find different metrics etc. I would also like single problem evaluators to draw AUROC etc.  I think it could make good intern  projects. But I do not want to do it in this design

DBCerigo:
Hi, we (myself and pals) implemented a simple replacement benchmarking module in sktime in late 2022/early 2023. I think benchmarking "results collator" (as per https://github.com/sktime/sktime/issues/1123) would be an apt description of what we were attempting. You can see an example usage/tutorial usage notebook for it [here](https://github.com/datavaluepeople/sktime/blob/kotsu-benchmarking-v2/examples/04_benchmarking_v2.ipynb), and the code for it [here](https://github.com/sktime/sktime/pull/2977/files).

[...]

PR #355, opened 2023-04-29
[ENH] Change data loaders and writers to minimize use of "nested_univ" input type

TonyBagnall:
#### Reference Issues/PRs
#110 
Note this has introduced a bug in TSFreshExtractor described in #381. It will be easier to fix this once the rest is in.
 
Sorry its so big. 

This PR does quite a lot to change the way collections are loaded and written to file. It is a little over complex to maintain backward compatibility. I am a little nervous about this, because it turns out the problem loaders are called many hundreds of times (see #118). Some of this may have been done in a previous PR, but this characterises all of the changes to the datasets package

1) meta data is loaded from the header file, and there is an option to return it
2) data is always loaded into a list of 2D numpy, one for each instance. If data is equal length, this is converted to 3D numpy. Need to do list -> numpy in all cases because we do not store number of instances in header, but its not too inefficient
3) retains the ability to return nested_univ (DataFrame with each cell a series) for backward compatibility, but otherwise removes all general usage of nested_univ. 
4) Refactors data_io into data_loaders and data_writers (it was too big)
5) Removes a load_TSC_problems function, can put it back if anyone wants it (raise an issue, see #382 )
6) Removes a test in TSFresh that seemed pointless. Testing may need further tidying/improvement
7) Single problem loaders for PLAID and japanese vowels now return list of numpy
8) The option for return_X_y == False still returns nested_univ. This need fixing (see #383), but the benchmarking module makes a lot of use of this, so will wait to fix this until benchmarking is sorted out (see #116)
9) Added tests for the arff and tsv format (see #363)
10) Simplified the writers

there may be some other changes, and I'm sure there is much that can be improved, but this is a start. Note we are not supporting time stamped data with this (we never did, but there is a tag for timestamps). This would be a good thing to implement, but it is really a separate issue.


aeon/datasets/_data_generators.py
```diff
@@ -0,0 +1,137 @@
...
+    n_cases: int
+        Number of cases.
+    series_length: int
+        Length of the series.
+    n_channels: int
+        Number of dimensions.
+
+    Returns
+    -------
+    DataFrame
```
chrisholder, line 22: Maybe have a description on the DataFrame

aeon/datasets/_data_loaders.py
```diff
@@ -1,69 +1,308 @@
...
+def load_from_tsfile(
+    full_file_path_and_name,
+    replace_missing_vals_with="NaN",
+    return_meta_data=False,
+):
+    """Load time series .ts file into X and (optionally) y.
+
+    Parameters
+    ----------
+    full_file_path_and_name : string
```
chrisholder, line 173: Descriptions would be nice

aeon/datasets/_data_loaders.py
```diff
@@ -1,69 +1,308 @@
...
+        although we recommend numpy3d, because some transformers do not work with
+        numpy2d. If None will load 3D numpy or list of numpy
+        There other options, see datatypes.SCITYPE_REGISTER, but these
+        will not necessarily be supported longterm.
+    local_module: default = os.path.dirname(__file__),
+    local_dirname: default = "data"
+
+    Raises
+    ------
+    Raise ValueException if the requested return type is not supported
```
chrisholder, line 244: Raises
-------
ValueError
    if the requested return type is not supported

aeon/datasets/_data_writers.py
```diff
@@ -0,0 +1,609 @@
...
+    series_length: int, default=-1
+        Indicates the series length if they are of equal length.
+    missing_values: str, default="NaN"
+        Representation for missing values.
+    comment: str or None, default=None
+        Comment text to be inserted before the header in a block.
+    fold: str or None, default=None
+        Addon at the end of the filename, i.e. _TRAIN or _TEST.
+
+    Returns
```
chrisholder, line 166: Probably dont need this

aeon/datasets/_data_writers.py
```diff
@@ -0,0 +1,609 @@
...
+def _write_data_to_tsfile(
+    X,
+    path,
+    problem_name,
+    y=None,
+    missing_values="NaN",
+    comment=None,
+    suffix="",
+):
+    """
```
chrisholder, line 277: Should be on first line

aeon/datasets/_data_writers.py
```diff
@@ -0,0 +1,609 @@
...
+    y: list, ndarray or None, default=None
+        The class values for each case, optional.
+    missing_values: str, default="NaN"
+        Representation for missing values.
+    comment: str or None, default=None
+        Comment text to be inserted before the header in a block.
+    suffix: str or None, default=None
+        Addon at the end of the filename, i.e. _TRAIN or _TEST.
+
+    Returns
```
chrisholder, line 300: Again probably dont need this

aeon/datasets/_data_writers.py
```diff
@@ -0,0 +1,609 @@
...
+        The class values for each case, optional.
+    comment: str or None, default=None
+        Comment text to be inserted before the header in a block.
+    fold: str or None, default=None
+        Addon at the end of the filename, i.e. _TRAIN or _TEST.
+    fit_transform: bool, default=True
+        Whether to fit the transformer prior to calling transform.
+
+    Returns
+    -------
```
chrisholder, line 513: Again

chrisholder (dismissed):

setup.cfg
```diff
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
```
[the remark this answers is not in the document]
TonyBagnall: sigh, always!

aeon/datasets/_data_writers.py
TonyBagnall: I'll leave them I think

chrisholder (approved):

Issue #321, opened 2023-04-22
[ENH] add the model checkpoint for deep learning classifiers


hadifawaz1999:
This was removed when getting in inceptionTime at first because of fails in the tests. The idea behind this fail i think was because of writing to file (saving the model in hdf5 format). To see if this can be go back in because it is essential to deep learners.

hadifawaz1999:
- [ ] InceptionTime
- [ ] CNN
- [ ] Encoder
- [ ] FCN
- [ ] ResNet
- [ ] MLP

PR #362, opened 2023-05-01
[ENH] add model checkpoint to inceptionTime deep classifier

hadifawaz1999:
Part of #321 
to see if it works fine, if yes than I should add the removal of the file saved.
Even if the tests get checked, this should get in before the addition of the removal of the saved model using `os.remove`


hadifawaz1999:
This should be ready now to get in.

TonyBagnall (dismissed):

TonyBagnall:
Have you removed it from excluded list?

hadifawaz1999:
@TonyBagnall which test do we exclude, the `test_save_estimators_to_file` only or the `test_persistence_via_pickle` as well ?

MatthewMiddlehurst (approved):
We should probably `import time` at the top of the file but I'm not gonna make this rerun CI for that.

hadifawaz1999:
Yeah @MatthewMiddlehurst i can keep it in mind for the next PR that includes deep learning models

Issue #360, opened 2023-04-29
[ENH] Switch transformers for unequal length to use list of numpy


TonyBagnall:
### Describe the feature or idea you want to propose

the old nested_univ format is being deprecated (see #110). The classifiers, regressors and clusterers are now isolated from nested_univ. Most  the panel transformers are also free from the format. In transformers, the last remaining job is to redesign the transformers meant for unequal length (listed below) to work with the new format

### Describe your proposed solution

Each of these needs change unequal to equal and need converting

- [ ] TSInterpolator (why is it named this? We do not use TS suffix elsewhere?) interpolate.py
- [ ] PaddingTransformer (again, may refactor the name) in padder.py  #364
- [ ] TruncationTransformer in truncation.py #364 

others in the panel directory are
- [ ] ColumnTransformer in compose.py
- [ ] IntervalSegmenter in segment.py
- [ ] _TSFreshFeatureExtractor (tsfresh.py) and two subclasses

nested_univ is not used anywhere else in transformations I can find (bootstrap, hierarchical, series)

### Describe alternatives you've considered, if relevant

np-list has been introduced as an alternative to pd.multiindex which is being used in the other place. No one I have spoken to over here likes it.

TonyBagnall:
this is a duplicate of #197

PR #364, opened 2023-05-02
[ENH] TruncationTransformer, PaddingTransformer and TSFresh internal type to np-list

TonyBagnall:
#### Reference Issues/PRs
relies on #355 
part of #360 

Converts the TruncationTransformer and PaddingTransformer to work with np-list (list of 2D numpy) and 3D numpy input types. Its so much simpler than dataframes. 

I have also removed the lower and upper bounds in TruncationTransformer. I found it confusing: if just lower was set it truncated to 0 ... lower-1, but if lower and upper were set it extracted lower, lower+1, ...., upper-1. These are not the same behaviour to me.  I dont see much use for wrapping like this, if you want to slice you are more likely to do it directly. This function is primarily to convert list of 2D numpy arrays to a 3D numpy array. 

I have added extra tests and removed a test based on multiindex which seemed to be testing conversion rather than this classes functionality


hadifawaz1999 (approved):
It all looks good for these changes.
Should we maybe in the future add a feature where the user can choose a method for the `fill value` ? For instance padd with the average of the series, min, max etc. Or maybe add a feature where the `fill value` can be an array ? where the user can pad different `fill value` to each series differently ?

MatthewMiddlehurst:
Would be nice to add, not high priority, though. Would create a feature request issue if you want it implemented.

PR #365, opened 2023-05-02
[MNT] Unstable extras

### LINKED ISSUE #289: [MNT] Include optional dependency group for unstable dependencies.
We have some dependencies not included in `all_extras` for whatever reason. I.e. `pycatch22` which fails to install on some setups.

We should at least track these, my current idea is an `unstable_extras` dependency group.

MatthewMiddlehurst:
Includes a dependencies groups for imports which exist in the package, but cause issues when included in `all_extras` currently.

Resolves #289

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,6 @@ all_extras = [
     "pyod>=0.8.0",
     "scikit_posthocs>=0.6.5",
     "seaborn>=0.11.0",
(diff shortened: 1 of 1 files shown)


TonyBagnall:
I like the title of this PR :) thats all I have ....

chrisholder (approved):

PR #377, opened 2023-05-03
[ENH] refactor _threads_to_use to _n_jobs

### LINKED ISSUE #102: [ENH] refactor BaseClassifier attribute _threads_to_use to _n_jobs

way back when, the internal copy of number of jobs in the BaseClassifier was set to be _threads_to_use (I think this was my bad). It should be _n_jobs to be consistent with other notation. Refactoring touches a lot of files, but also highlights that many classifiers are using their child class n_jobs  variable rather than the parent class _n_jobs. Thus it becomes more disruptive and dangerous to refactor,. See here
https://github.com/scikit-time/scikit-time/pull/101
I have closed the above PR while other large changes go on and just raised the issue here to be solved once the code base settles down.

TonyBagnall:
#### Reference Issues/PRs

fixes #102 

Simple refactor of _n_threads_to_use to _n_jobs

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,14 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
+    --timeout 600
+    --cov aeon
+    --cov-report xml
+    --cov-report html
+    --showlocals
+    --matrixdesign True
+    -n auto
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,14 +12,6 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
-    --durations 10
-    --timeout 600
-    --cov aeon
-    --cov-report xml
-    --cov-report html
-    --showlocals
-    --matrixdesign True
-    -n auto
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
--- a/aeon/classification/base.py
+++ b/aeon/classification/base.py
@@ -53,7 +53,7 @@ class BaseClassifier(BaseEstimator, ABC):
     _X_metadata         : metadata/properties of X seen in fit
     _class_dictionary   : dictionary mapping classes_ onto integers
         0...``n_classes_``-1.
-    _threads_to_use     : number of threads to use in ``fit`` as determined by
+    _n_jobs     : number of threads to use in ``fit`` as determined by
         ``n_jobs``.
     _estimator_type     : string required by sklearn, set to "classifier"
     """
@@ -76,7 +76,7 @@ class BaseClassifier(BaseEstimator, ABC):
         self.fit_time_ = 0  # time elapsed in last fit call
         self._X_metadata = []  # metadata/properties of X seen in fit
         self._class_dictionary = {}
-        self._threads_to_use = 1
+        self._n_jobs = 1
 
         # required for compatibility with some sklearn interfaces e.g.       #
         # CalibratedClassifierCV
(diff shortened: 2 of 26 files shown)


setup.cfg
```diff
@@ -12,14 +12,6 @@ addopts =
...
     --ignore docs
     --doctest-modules
-    --durations 10
-    --timeout 600
-    --cov aeon
-    --cov-report xml
-    --cov-report html
-    --showlocals
-    --matrixdesign True
-    -n auto
```
MatthewMiddlehurst, line 22: The classic

MatthewMiddlehurst (changes requested):

patrickzib:
Hey @TonyBagnall , thanks a lot for touching these again! It is a monster.

I am afraid we are not there yet. There are quite a few classes using `self.n_jobs`. 

For the base classes, I do not even understand where the variable `self.n_jobs` was set (possibly being a bug?):

- Clustering: https://github.com/aeon-toolkit/aeon/blob/e7e4592f6135f4fe3c5cb220a60541002719ec5f/aeon/clustering/base.py#L71

- Classification:  https://github.com/aeon-toolkit/aeon/blob/e7e4592f6135f4fe3c5cb220a60541002719ec5f/aeon/regression/base.py#L153

- Regression: https://github.com/aeon-toolkit/aeon/blob/e7e4592f6135f4fe3c5cb220a60541002719ec5f/aeon/regression/base.py#L153


IMO: it would be best to refactor all occurrences of `self.n_jobs` to `self._n_jobs`, too, and there are > 100 occurrences of self.n_jobs :/

patrickzib:
Sorry, if I am missing something. But how is `self.n_jobs` initialized in these base classes?

setup.cfg
TonyBagnall: every. bloody. time. :)

patrickzib:
If we agree on something that is meaningful, I can pick it up and do the changes to the estimators.

Could you explain the logic behind self.n_jobs for me? It seems like a FK twist.

Right now, it seems like every estimator has to set `self.n_jobs` from the same parameter passed to the class - next, each base-class checks  `self.n_jobs` and sets it to `self._n_jobs`. Yet, most classifiers still use `self.n_jobs`, as they are not aware, that the base class set `self._n_jobs`.

MatthewMiddlehurst:
This was actually me. I put it in the base class to avoid coping the check method in every estimator. I feel having a method will just run into the same issue of people using the estimator `self.n_jobs` over it.

Can either make people process `self.n_jobs` in their own estimator, or just document better/mention it in reviews.

TonyBagnall:
> 

I dont feel too strongly about it, but I think it should probably all be dealt with in the base class and also be a base class parameter. That or take it out of base class, its neither one nor the other at the moment.

TonyBagnall:
> 

if we defaulted it to None in the base class it would then fail tests if people reset it in their subclasses?

TonyBagnall:
what do you think @patrickzib ? Do this PR now and sort out later or close this PR and fix all together? I'm fine either way, I just dont like PRs having around :)

MatthewMiddlehurst:
I think we should merge this and create a separate issue to convert classifiers/regressors which use `self.n_jobs`. I think the new name is better.

It's difficult to enforce the usage of `check_n_jobs`/`self._n_jobs` other than by review and documentation IMO. I think the current base class way works fine in reducing the amount of copied boilerplate, it just has a bad name.

MatthewMiddlehurst:
Can go through each estimator and change it in this PR also, I don't mind. Not sure how much work/extra file changes that would be though.

MatthewMiddlehurst (approved):

PR #379, opened 2023-05-04
[ENH] Implement simple(r) benchmarking

DBCerigo:
#### Reference Issues/PRs
https://github.com/aeon-toolkit/aeon/issues/116 for general context. 
https://github.com/aeon-toolkit/aeon/issues/116#issuecomment-1509958967 for context specific to this PR and implementation.  

- a core benchmarking module
- forecasting benchmarking module (using the core module)
- unit tests
- example notebook showing usage

Optional, to `all_extras`: `kotsu>=0.3.1`

Reviewing the example notebook first to get a sense of usage of the API and if it fits with how you envisaged it.
Then the implementation.  

Considerations for further work: 
- Deleting/archiving all the previous benchmarking code?
- Implementing benchmarking modules for classification, clustering, etc.

Yes.

##### For all contributions

### DIFF
--- a/aeon/benchmarking/benchmarks.py
+++ b/aeon/benchmarking/benchmarks.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+"""Benchmarking interface for use with aeon objects.
+
+Wraps kotsu benchmarking package.
+"""
+from typing import Callable, Optional, Union
+
+import pandas as pd
+
+from aeon.base import BaseEstimator
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
+class BaseBenchmark:
+    """Base class for benchmarks.
+
+    A benchmark consists of a set of tasks and a set of estimators.
+    """
+
+    def __init__(self):
+        _check_soft_dependencies("kotsu")
+        import kotsu
+
+        self.estimators = kotsu.registration.ModelRegistry()
+        self.validations = kotsu.registration.ValidationRegistry()
+        self.kotsu_run = kotsu.run.run
+
+    def add_estimator(
+        self,
+        estimator: BaseEstimator,
+        estimator_id: Optional[str] = None,
+    ):
+        """Register an estimator to the benchmark.
+
+        Parameters
+        ----------
+        estimator : BaseEstimator object
+            Estimator to add to the benchmark.
+        estimator_id : str, optional (default=None)
+            Identifier for estimator. If none given then uses estimator's class name.
+
+        """
+        estimator_id = estimator_id or f"{estimator.__class__.__name__}-v1"
+        estimator = estimator.clone()  # extra cautious
+        self.estimators.register(id=estimator_id, entry_point=estimator.clone)
+
+    def _add_task(
+        self,
+        task_entrypoint: Union[Callable, str],
+        task_kwargs: Optional[dict] = None,
+        task_id: Optional[str] = None,
+    ):
+        """Register a task to the benchmark."""
+        task_id = task_id or (
+            f"{task_entrypoint}-v1"
+            if isinstance(task_entrypoint, str)
+            else f"{task_entrypoint.__name__}-v1"
+        )
+        self.validations.register(
+            id=task_id, entry_point=task_entrypoint, kwargs=task_kwargs
+        )
+
+    def run(self, output_file: str) -> pd.DataFrame:
+        """Run the benchmark.
+
+        Parameters
+        ----------
+        output_file : str
+            Path to write results output file to.
+
+        Returns
+        -------
+        pandas DataFrame of results, with columns; "validation_id", "model_id",
+            "runtime_secs", and columns for each of the metrics returned by the
+            benchmark task.
+        """
+        results_df = self.kotsu_run(self.estimators, self.validations, output_file)
+        return results_df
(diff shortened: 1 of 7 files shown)


aeon/benchmarking/benchmarks.py
```diff
@@ -0,0 +1,76 @@
...
+        """Run the benchmark.
+
+        Parameters
+        ----------
+        output_file : str
+            Path to write results output file to.
+
+        Returns
+        -------
+        pandas DataFrame of results
```
hadifawaz1999, line 73: I think it would be better to show the output format of the Data Frame here.

hadifawaz1999: For instance the column names, rows etc. one line sentence

TonyBagnall:
one thing is that you need a callable dataset_loader. For machine learning we have a single function (load_ts_file) we pass the location of the file to load as a string. This implies you need a different function for each datasets, not sure how that would work?

ltsaprounis:
> one thing is that you need a callable dataset_loader. For machine learning we have a single function (load_ts_file) we pass a string of the location of the file to load. This set up implies you need a different function for each datasets, not sure how that would work?

Could always do something like this to make it a callable.

```python
# Option 1
from functools import partial

load_data = partial(load_ts_file, path="path")

# Option 2
def load_data():
    return load_ts_file("path")
```

But I guess it's more of a headache...

@DBCerigo what's the idea behind having the loader as a callable and not just passing the data?

aeon/benchmarking/benchmarks.py
DBCerigo: Done [e3f2bde](https://github.com/aeon-toolkit/aeon/pull/379/commits/e3f2bde24ecede2cf081c8641fd09283063c2063)

DBCerigo:
> one thing is that you need a callable dataset_loader. For machine learning we have a single function (load_ts_file) we pass a string of the location of the file to load. This set up implies you need a different function for each datasets, not sure how that would work?

@TonyBagnall @ltsaprounis 

As shown in the example notebook https://github.com/datavaluepeople/aeon/blob/benchmarking/v2/examples/04_benchmarking_v2.ipynb the intention with using `dataset_loader` is to match the module `aeon.datasets` (for example `from aeon.datasets import load_airline` is used in the notebook) in keeping with interfacing directly with `aeon` objects and classes. The notebook also shows a method for adding your own datasets - the same as Option 2 in the above. 

If `aeon` is moving away from using the `aeon.datasets` module, then I suggest we alter the implementation to not be so focussed on interfacing well with that module. If instead we wanted to only support usage via `load_ts_file` we could make `add_task` take a `dataset_path: str` (instead of `dataset_loader: Callable`) and alter this line
 https://github.com/datavaluepeople/aeon/blob/e3f2bde24ecede2cf081c8641fd09283063c2063/aeon/benchmarking/forecasting.py#L37
 to use `load_ts_file`. 

Let me know yours thought/preferences.

hadifawaz1999:
How about taking the dataset as parameter ? Because not all users work with ts files.

ltsaprounis:
I don't mind it being callable. It allows for some flexibility down the line e.g. if `load_ts_file` returns metadata too etc. For forecasting benchmarking I think this is good to go, there's going to be some more work to be done to allow for hierarchical data benchmarking etc. but for now I think this gives as a good starting point to start working on this.

@DBCerigo - that being said, I suggest we add a warning the public methods / classes, saying that this interface is experimental and we might change the API in the future.
```python
warn("<name of class/function> is experimental, and interfaces may change.")
```

DBCerigo:
> How about taking the dataset as parameter ? Because not all users work with ts files.

I don't think we would want to have the datasets in memory at the time we are add benchmarking tasks, given we might be adding tasks for many multiple different datasets to compare models over etc. 

@ltsaprounis all sounds good to me. I'll wait for a day or two to see if anyone else pitches in further, and if not then I'll add the warnings and leave it to be merged if so.

DBCerigo:
@ltsaprounis now done. (And fixed missing [ENH] tag on previous commit message.)

TonyBagnall:
I'm happy for this to go in, but will clarify our use case. The datasets module is staying, but it only contains toy examples. Most of the problems we use are on timeseriesclassification.com, and we load them from disk with

def load_from_tsfile(
    full_file_path_and_name,
    replace_missing_vals_with="NaN",
    return_meta_data=False,
)
I'm sure we can work something out for classification, if this is ok for forecasting then go for it.

DBCerigo:
> I'm happy for this to go in, but will clarify our use case. The datasets module is staying, but it only contains toy examples. Most of the problems we use are on timeseriesclassification.com, and we load them from disk with

I'm confident we could (easily) extend the interface in the future to be able to take just the path to a file on disk and load it (with `load_from_tsfile`) at validation run time (while still keeping the funcs which return data option as well), if we find that that functionality is much desired. 

> I'm sure we can work something out for classification, if this is ok for forecasting then go for it.

The intention with the implementation and this PR, is that `aeon/benchmarking/forecasting.py` serves as a template for implementing benchmarking for classification (and clustering, etc.). It won't be quite as simple as replacing `forecasting` with `classification` in the file, but it shouldn't be too too far off :)

ltsaprounis (approved):
good stuff 🚀

PR #380, opened 2023-05-04
[MNT] Remove any reference of pykalman

### LINKED ISSUE #376: [BUG] 'pykalman' optional dependency unmaintained and unsupported python 3.11
### Describe the bug

`pykalman` related tests are currently not passing on python 3.11 due to an error coming from https://github.com/pykalman/pykalman/blob/8d3f8e498b64d902016a0216bf2bcc8b262d917b/pykalman/utils.py#L111

`pykalman` repo seems abandoned and [latest release](https://pypi.org/project/pykalman/#history) was almost 10 years ago (Jul 7 2013). Long term we need a solution that's compatible with py3.11 onwards.

_No response_

### Expected results

Tests passing 

### Actual results

```python-traceback
obj = <pykalman.standard.KalmanFilter object at 0x7fbc3d0b6dd0>

    def get_params(obj):
        '''Get names and values of all parameters in `obj`'s __init__'''
        try:
            # get names of every variable in the argument
            args = inspect.getargspec(obj.__init__)[0]
            args.pop(0)   # remove "self"
    
            # get values for each of the above in the object
            argdict = dict([(arg, obj.__getattribute__(arg)) for arg in args])
            return argdict
        except:
>           raise ValueError("object has no __init__ method")
E           ValueError: object has no __init__ method

obj        = <pykalman.standard.KalmanFilter object at 0x7fbc3d0b6dd0>
```

### Versions

_No response_

hadifawaz1999:
Fix #376 
Basically purging any reference to `pykalman` imports, functions, classes, tests, examples, anything

No but it removes one, `pykalman`

To check if i missed anything, i used `grep -rnw ./ -e 'pykalman' ` to check the files that i should modify

No, removed anything related


lmmentel:
Nice! I think you could also remove `pykalman` from the list of dependencies in `pyproject.toml`. We should also cleanup the coda recipe, but that's a separate issue.

hadifawaz1999:
> Nice! I think you could also remove `pykalman` from the list of dependencies in `pyproject.toml`. We should also cleanup the coda recipe, but that's a separate issue.

Just pushed the removal of pykalman from the `project.toml` file, this should be ready after the CI check .

aeon/transformations/series/kalman_filter.py
```diff
@@ -3,15 +3,14 @@
 """Kalman Filter Transformers.
 
 Series based transformers, based on Kalman Filter algorithm. Contains Base class
-and two transformers which are each Adapters for external packages
-pykalman and FilterPy.
+and a transformer which is an Adapters for the external package
```
lmmentel, line 6:
```suggestion
and a transformer which is an adapter for the external package
```

lmmentel (approved):
LGTM!

hadifawaz1999:
nice catch @lmmentel of the `Adapters`  :) an extra eye is always helpful, thanks !

PR #195, opened 2023-03-13
[GOV] New `aeon` governance document.

MatthewMiddlehurst:
This PR contains changes to the governance document in the style on `Scikit-learn`, based off my opinions from #2. All of this is flexible, but I think we should cut down the size of the original document whatever we go with.

This is a full rewrite of the governance document. Diff can be seen here but its not very helpful https://gist.github.com/MatthewMiddlehurst/b51db154402ecd88ca7587133b65e28a/revisions

Key changes:
- Removes algorithm maintainer role
- Removes CC observer role
- Removes some CC power i.e. veto right

current gov page: https://www.aeon-toolkit.org/en/latest/get_involved/governance.html
new gov page: https://aeon-toolkit--195.org.readthedocs.build/en/195/get_involved/governance.html

PR #389, opened 2023-05-06
[GOV] Governance workgroups

MatthewMiddlehurst:
An alternative to the current governance roles in #195 inspired by Pandas workgroups. This would replace the currently inherited Community Council (CC) role.

The exact workgroups and details surrounding them are all up for debate.

### DIFF
--- a/docs/get_involved/governance.md
+++ b/docs/get_involved/governance.md
@@ -28,7 +28,7 @@ process.
 Core developers are community members that have made significant contributions and are
 trusted to continue the development of the project and engage with its community.
 Core developers are granted write access to the `aeon` repository and voting rights
-on most project decisions. They are expected to review code contributions
+on all project decisions. They are expected to review code contributions
 and engage with topics or code they are knowledgeable about.
 
 New core developers are nominated by existing core developers and are subject to a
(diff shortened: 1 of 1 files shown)


docs/get_involved/governance.md
```diff
@@ -39,32 +39,65 @@ their role after a long period of inactivity and lack of engagement with fellow
...
 
-### Community Council
+### Workgroups
 
-The Community Council (CC) consists of core developers who have additional
-responsibilities to ensure the smooth running of the project. This includes strategic
-planning, project infrastructure management, and release management. In the event core
-developers fail to reach consensus, the CC is responsible for resolving the issue.
+Extra responsibilities are delegated to workgroups, which are groups of contributors
+with a core developer lead. Workgroups are responsible for specific areas of the project
```
TonyBagnall, line 45: I wqould just say a group of contributors including at least one core developer

TonyBagnall (approved):
I think its fine. So its a choice between a CC and working groups. I think I prefer this, but do not feel strongly.

MatthewMiddlehurst:
Just to confirm, this would just alter the roles portion of https://github.com/aeon-toolkit/aeon/pull/195 if merged. But i agree, we should give everyone a chance to discuss. If this does go through, i imagine there will not be much remaining on #195 to block it.

docs/get_involved/governance.md
MatthewMiddlehurst: Is it the group leadership role itself or a core dev needing to fill the role the issue? 

When it comes to account access some sort of designated group member is required IMO

PR #417, opened 2023-05-11
[ENH] Refactor sax

### LINKED ISSUE #257: [ENH] Breakpoints calculation in SAX
In [this line](https://github.com/aeon-toolkit/aeon/blob/548dd73153a6277d456b0ae1bd8fb545d74385de/aeon/transformations/panel/dictionary_based/_sax.py#L175)
it seems that the breakpoints for the SAX algorithm are hardcoded, someone correct me if am wrong. If this is the case, why dont we do it using the scipy.norm.ppf ? as it does generate the breakpoints we need for the algorithm. As follows:

`scipy.norm.ppf(np.arange(1, number_of_segments) / number_of_segments)`

hadifawaz1999:
#### Reference Issues/PRs
Part of #197  and fix #257 

Refactoring SAX to
1. Stop using pandas but uses numpy instead
2. Use scipy to generate the breakpoints
3. Make the distribution used to generate breakpoints controlled by the user
4. Make transformation faster using `np.digitize`
5. Add inverting the SAX to approximate the time series following the chosen distribution (not technical inverse transform)
6. Add test for sax
7. Add example for sax in root

no

The optimization of the code

yes `test_sax.py` in dictionary based panel (collection) transformations

##### For all contributions


aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +18,132 @@ class SAX(BaseTransformer):
...
-            Nested dataframe with univariate time-series in cells.
+        paa = PAA(n_intervals=self.n_segments)
+        X_paa = paa.fit_transform(X=X)
 
-        Returns
-        -------
-        dims: Pandas data frame with first dimension in column zero
-        """
-        X = X.squeeze(1)
+        self.breakpoints, self.breakpoints_mid = self._generate_breakpoints(
```
patrickzib, line 92: Why are breakpoints generated in transform? Shouldn't this be done once in the constructor and then the same breakpoints are used in transform?

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +18,132 @@ class SAX(BaseTransformer):
...
+            alphabet_size=self.alphabet_size,
+            distribution=self.distribution,
+            distribution_params=self.distribution_params,
+        )
+
+        return self._inverse_sax_symbols(
+            self._get_sax_symbols(X_paa=X_paa), series_length=series_length
+        )
+
+    def _get_sax_symbols(self, X_paa):
```
patrickzib, line 102: add @njit to it?

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +18,132 @@ class SAX(BaseTransformer):
...
+        # the -1 is because breakpoints have self.alphabet_size - 1 elements
 
-        bags = pd.DataFrame()
-        dim = []
+        return sax_symbols
 
-        for i in range(n_instances):
-            bag = {}
-            lastWord = -1
+    def _inverse_sax_symbols(self, sax_symbols, series_length):
```
patrickzib, line 114: Could you add documentation to the method?

aeon/transformations/panel/dictionary_based/_sax.py
patrickzib, line 102: Could you add documentation to the method?

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +18,132 @@ class SAX(BaseTransformer):
...
-        dims: Pandas data frame with first dimension in column zero
-        """
-        X = X.squeeze(1)
+        self.breakpoints, self.breakpoints_mid = self._generate_breakpoints(
+            alphabet_size=self.alphabet_size,
+            distribution=self.distribution,
+            distribution_params=self.distribution_params,
+        )
+
+        return self._inverse_sax_symbols(
```
patrickzib, line 98: Could you explain what this line does? :)

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +18,132 @@ class SAX(BaseTransformer):
...
+        )
+
+    def _get_sax_symbols(self, X_paa):
+        sax_symbols = np.zeros(X_paa.shape, dtype=int) - 1
 
-        if self.alphabet_size < 2 or self.alphabet_size > 4:
-            raise RuntimeError("Alphabet size must be an integer between 2 and 4")
-        if self.word_length < 1 or self.word_length > 16:
-            raise RuntimeError("Word length must be an integer between 1 and 16")
+        for i_bp, bp in enumerate(self.breakpoints):
```
patrickzib, line 105: Is this the same as: `np.digitize`?
https://numpy.org/doc/stable/reference/generated/numpy.digitize.html

patrickzib (changes requested):
Nice. Thank you. WOuld be even better if we used @njit and/or numpy methods like `np.digitize`

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: Yes you are right did not think of that, thanks for noting that out

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: Actually there is a version with numba usage njit, its [this commit](https://github.com/aeon-toolkit/aeon/pull/417/commits/eb18f88bb6f00b56fc1a3595cdf334892672ed7f), i did a time analysis on this commit and without using numba, it didnt change the time or make it any faster. The only thing is the code becomes with a lot of details not needed so i decided to drop njit, what do you think ?

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: will do

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: basically, if SAX transforms a time series of length 10 to a series of length 5 (n_segments=5), you may need the original length back.

For example if you TS in sax form gave (a, b,a,c,c) and your original length is 10, then the inverse transform will give you (a,a,b,b,a,a,c,c,c,c) this is the idea used in tslearn, the only difference is that i use them both togethere (check return of transform function). Let me know if you prefer to keep them separated such as in tslearn implementation.

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: I never heard of this function my bad :) will look into it and see if i can add it here

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: will do

aeon/transformations/panel/dictionary_based/_sax.py
patrickzib: I see. Never thought of that before. Could you give an example where this might be beneficial?

my personal preference would be to add a parameter that can turn this on and off, with the default being to reduce the dimensionality.

aeon/transformations/panel/dictionary_based/_sax.py
patrickzib: Good point. There is so much framework going on there with the call to PAA. Would be interesting to profile your refactored class to see where most time is spent.

hadifawaz1999: my assumption is that PAA is taking too much time, my goal is to refactor it as well like i did for SAX because I noticed that the output of PAA in aeon is different than tslearn.

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: Well actually when applying sax transform it generates a symbol which in this case will just be an index pointing to the breakpoint value.

For example if we have only 1 breakpoint which  is 0, if the time series value is positive it will be replaces by 1 and if it is negative it will be replaces by 0.
This is the way SAX will make the time series in a more 'digital' space instead of a continuous one.

However, you may don't want to represent each segment by an index that points to the breakpoint, maybe you want a real value to represent each segment, in this case you need to define the middle points between the breakpoints interval. For instance if your alphabet size is 2, your breakpoint will be 0 defining 2 intervals, `]-inf,0]` and `]0,+inf[`. For each segment value given by PAA we will compare the value `x` to 0, if its positive we want it to be replaced this time by a real value that represents this interval `]0,+inf[`. The formula of the calculation is taken from the `tslearn` implementation actually  and this was my understanding of the inverse function. This can be helpful where you want to represent the values of the time segment by a single real value, this value is chosen by creating the breakpoints thresholds. For example in the figure below you can see this being applied on a UCR example, its a way of going back to a continuous form but also still in an approximation form using Gaussian values. So its not completely the inverse because it is not able to give you the orignal time series of course.
image

Basically its like defining that each 10 consecutive time stamps (segment length) can be represented by one real value, kind of like a sampling theorem. This is the the way PAA does it, but going also through SAX will have the Gaussian distribution representation instead of only averaging the segment values.

Am not sure if it makes sense what I just explained, I completely understand if it doesn't because it took me some time to understand the concept of this implementation before i decided to port it in. I even have an application for it that I wanted to try, where I use a VAE to go from this discritized space to the original one, modeling it as a generative model from discrete to continuous space using SAX symbols.

But I agree it should be a parameter as you said

patrickzib: Sure, I agree, it makes sense to have the inverse for visualization purposes. :) 

But if SAX is applied as a transformer, I would expect it to reduce dimensionality via PAA (i.e. without padding the length back to the original length). padding to the original length increases memory and does not add much additional information for the downstream task.

Maybe even providing two methods transform and inverse_transform might be best? Not sure how other transformers do this.

patrickzib: IMO transform should return indices.

hadifawaz1999: I agree, it would be better to divide them because usually users will want the original transform for their work.

hadifawaz1999: And actually your proposal of dividing them is what is done in tslearn as well.

hadifawaz1999: will take care of that and will add documentation providing detailed explanation of the purpose of each function

patrickzib: Thank you :)

hadifawaz1999: Should we add the ability for the user to choose the type of the symbols ? What if the user wants the SAX to give back a series of strings for instance. That could be handled easily from the user side but what if its better to make it internal, what do you think @patrickzib ?

patrickzib: True, there are many possible return types:

- Array of Indices (Ints)
- Array of characters (String)
- Concatenated into one large binary number: One Long

In SFA, we store an array of indices internally, when a class parameter is set to TRUE, and always return the long representation of the SFA word. But it does not mean that this has to be the case always.

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: it is resolved now

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: all documentation is added to all functions

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: its taken care of, let me know what you think

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: It works perfectly in one line :) Thanks for the advise of using `np.digitize` !

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,194 @@ class SAX(BaseTransformer):
...
+            self.distribution_params_ = (
+                dict(scale=1.0)
+                if self.distribution_params is None
+                else self.distribution_params
+            )
+
+        else:
+            raise NotImplementedError("still not added")
+
+        self.breakpoints, self.breakpoints_mid = self._generate_breakpoints(
```
patrickzib, line 89: One last thought: SAX with Gaussian-breakpoints assumes that the data is z-normalized prior to transform. 

Is that taken care of? Sorry, if I miss it or it is hidden in the PAA-transform. Might be more safe to  apply zscore to the data first?

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: The issue comes from the fact that the number of segments will not always divide the time series length. Given this issue, we can parametrize PAA in order to choose whether or not to discard or pad the transformation till the last time stamps ignored in the segments repartitions. This can and i think should be done in the paa refactor, and then included in the sax refactor.
Thanks for pointing this out !

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: It should be good now, i emphasized about the returned symbols and what they represent, let me know if its okay

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: Thanks for reminding me that, i added a parameter to choose whether or not to z-normalize the input data.

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
+    distribution : str, default = "Gaussian",
+        the distribution function to use when generating the
+        alphabet
+    distribution_params : dict, default = None,
+        the parameters of the used distribution, if the used
+        distribution is "Gaussian" and this parameter is None
+        then the default setup is {"scale" : 1.0}
+    znormalized : bool, default = True,
+        this parameter is set to True when the input time series
+        are supposed to be z-nromalized, i.e. the mean of each
```
patrickzib, line 39: assumed to be z-normalized
(wording and typo)

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
+        time series should b 0 and the standard deviation should be
```
patrickzib, line 40: be instead of b (typo)

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
+        self.breakpoints, self.breakpoints_mid = self._generate_breakpoints(
+            alphabet_size=self.alphabet_size,
+            distribution=self.distribution,
+            distribution_params=self.distribution_params_,
+        )
+
+        super(SAX, self).__init__()
+
+    def _get_paa(self, X):
+        """Transform the input time seires to PAA segments.
```
patrickzib, line 106: series (typo)

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
-
-    Attributes
+    n_segments : int, default = 8,
+        number of segments for the PAA, each segement is represented
+        by a symbol
+    alphabet_size : int, default = 4,
+        size of the alphabet to be used to create the bag of words
+    distribution : str, default = "Gaussian",
+        the distribution function to use when generating the
+        alphabet
```
patrickzib, line 32: what other values are supported?

patrickzib: I.e. currently only Gaussian is supported.

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
+
+        if self.distribution == "Gaussian":
+            self.distribution_params_ = (
+                dict(scale=1.0)
+                if self.distribution_params is None
+                else self.distribution_params
+            )
+
+        else:
+            raise NotImplementedError("still not added")
```
patrickzib, line 95: raise NotImplementedError(self.distribution, "is still not added")

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -26,176 +19,210 @@ class SAX(BaseTransformer):
...
+        X : np.ndarray of shape = (n_instances, n_channels, series_length)
+            The input time series
+
+        Returns
+        -------
+        X_paa : np.ndarray of shape = (n_instances, n_channels, n_segments)
+            The output of the PAA transformation
+        """
+        if not self.znormalized:
+            X = scipy.stats.zscore(X, axis=2)
```
patrickzib, line 119: Sorry for asking. Isn't is axis=1? I.e. z-norm is applied for each series.

patrickzib: (or is the input 3d?)

aeon/transformations/panel/dictionary_based/_sax.py
```diff
@@ -216,6 +243,43 @@ def get_test_params(cls, parameter_set="default"):
...
         """
-        # small word length, window size for testing
-        params = {"word_length": 2, "window_size": 4}
+        params = {"n_segments": 10, "alphabet_size": 8}
         return params
+
+
+@njit(parallel=True, fastmath=True)
+def _invert_sax_symbols(sax_symbols, series_length, breakpoints_mid):
+    """Produce the original time series using a Gaussian estimation.
```
patrickzib, line 252: Reconstruct (?) the ...

patrickzib (commented):
Found some minor typos. And zscore is not totally clear to me. If the data is a 2d numpy array, it should be zscore(X, axis=1), I guess? I.e. row-wise. Or is the data assumed to be a 3d numpy array?

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: I have been meaning to test with other distributions, such as uniform for instance or beta. I can add at least 2 more now, been planning to make them step by step to avoid including a distribution that doesn't make sense to use.

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: I assumed that aeon does an internal re-shaping, i remember that was the case in the baseClassifier, will check if its the same in transformer

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: Yes makes sense :p thanks

aeon/transformations/panel/dictionary_based/_sax.py
hadifawaz1999: thanks ! will take care of that

aeon/transformations/panel/dictionary_based/_sax.py
patrickzib: Sure, fine. Just meant to add a docstring saying which ones are supported right now. No need to implement others. :)

hadifawaz1999: Oh okay, will do that then

aeon/transformations/panel/dictionary_based/_sax.py
patrickzib: @TonyBagnall I think, we are planning to remove the reshaping for transformers and accept numpy 2d input? Or what is the current state here?

hadifawaz1999: yeah makes sense to put -1, this way in the future we do not edit it

patrickzib (approved):

Issue #347, opened 2023-04-27
[MNT] Estimators excluded from tests


hadifawaz1999:
### Describe the issue

In the `_config.py` file, a lot of esimtators are excluded including ResNet (see [this old issue](https://github.com/sktime/sktime/issues/3954) but actually "locally" ResNet passed the tests 2 times.

TonyBagnall:
the list of fully excluded estimators is this. I'm not sure why HIVECOTE is excluded, something about macs. https://github.com/sktime/sktime/issues/3103  
we should look into that, although slightly loath to introduce *more* testing ....
EXCLUDE_ESTIMATORS = [
    # SFA is non-compliant with any transformer interfaces, #2064
    "SFA",
    # Interface is outdated, needs a rework.
    "ColumnTransformer",
    # PlateauFinder seems to be broken, see #2259
    "PlateauFinder",
    # below are removed due to mac failures we don't fully understand, see #3103
    "HIVECOTEV1",
    "HIVECOTEV2",
    "RandomIntervalSpectralEnsemble",
    "RandomInvervals",
    "RandomIntervalSegmenter",
    "RandomIntervalFeatureExtractor",
    "RandomIntervalClassifier",
    "MiniRocket",
    "MatrixProfileTransformer",
    # tapnet based estimators fail stochastically for unknown reasons, see #3525
    "TapNetRegressor",
    "TapNetClassifier",
    "ResNetClassifier",  # known ResNetClassifier sporafic failures, see #3954
]

TonyBagnall:
note that this is just the interface testing, the localised testing still runs I think

hadifawaz1999:
Maybe we can leave this issue till after the testing in general is re-visited and wrapped up ?

hadifawaz1999:
@TonyBagnall maybe we can add a comment here to summarize what was un-excluded and what is still excluded ? to keep track

TonyBagnall:
good idea: currently excluded

```python
EXCLUDE_ESTIMATORS = [
    # SFA is non-compliant with any transformer interfaces, #2064
    "SFA",
    # Interface is outdated, needs a rework.
    "ColumnTransformer",
    # PlateauFinder seems to be broken, see #2259
    "PlateauFinder",
    # below are removed due to mac failures we don't fully understand, see #3103
    "HIVECOTEV1",
    "HIVECOTEV2",
    "RandomIntervalSpectralEnsemble",
    "RandomInvervals",
[...]

Issue #271, opened 2023-04-07
[ENH] Refactor all tests


TonyBagnall:
**Is your feature request related to a problem? Please describe.**

Once we have sorted out transformers, I would like to start thinking about testing. Two issues are the very slow CI and the huge over complexity of the code. I favour refactoring the whole testing mechanism. Its a huge job which will need to be done incrementally, but I don't think anyone here likes the current object oriented scenario model. It makes it very hard to find out why a test is failing and is fundamentally linked to the whole mtype/stype paradigm, which personally I don't support.

Just looking at scikit learn. They have a much simpler and more pythonesque model: a test directory in the source root testing general functionality using a collection of test functions (and some other stuff has crept in it seems) then a test package in every sub package testing specific functionality. 

The current system is in my opinion grossly over engineered and unmaintainable by anyone but the person who wrote it. It is in fact not working currently.  There is a huge list of excluded tests, which to me suggests the generalisation has failed. Just like with all the attempts at monolithic centralised control (e.g. datatypes), the result is a horrible huge set of exceptions. Its just bad design and an imposition of one world view. Note scikit learn has no exceptions I can see. For example, I have had to exclude SlidingWindowTransformer (that takes a univariate time series collection and converts it to a multivariate time series collection where each channel is a window) because it tells me "Scitype Series input must output Scitype Panel" or if I change tags   "Scitype Panelinput must output Scitype Hierarchical". I refuse to buy in to all of that and figure out how to make it work.  

I think we aim to strip it all back to function based testing. It wont be quick, but it will be worth it for reduced CI and easier maintenance of testing. But if someone thinks we should go another way, lets discuss. Forecasting is obviously going to be the biggest problem given it takes 70% of CI time, and it is not my area, so input from those on that side would be very welcome.

TonyBagnall:
generic issue that is not that relevant any more, closing

PR #419, opened 2023-05-12
[EN] Hard code deep learning exclusions from tests

TonyBagnall:
see 
https://github.com/aeon-toolkit/aeon/issues/271
https://github.com/aeon-toolkit/aeon/issues/347

The deep learning algorithms are fundamentally not deterministic. They also cannot be pickled. There are three relevant tests in test_all_estimators

"test_fit_idempotent",
"test_persistence_via_pickle" 
"test_save_estimators_to_file",

firstly, I have refactored the poorly named test_fit_idempotent to test_fit_deterministic. I have then escaped tests if the estimator is a DeepClassifier or DeepRegressor. Turns out this is done all over the testing. Currently, InceptionTime is still excluded because technically it is an ensemble of deep learners, not a deep learner itself.

### DIFF
--- a/aeon/tests/_config.py
+++ b/aeon/tests/_config.py
@@ -34,11 +34,13 @@ EXCLUDED_TESTS = {
     # exclude
     "InceptionTimeClassifier": [
         "test_fit_deterministic",
-        "test_persistence_via_pickle" "test_save_estimators_to_file",
+        "test_persistence_via_pickle",
+        "test_save_estimators_to_file",
     ],
     "InceptionTimeRegressor": [
         "test_fit_deterministic",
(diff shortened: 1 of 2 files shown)


hadifawaz1999 (approved):
LGTM! We should keep this in mind to add in the future the `DeepClustering` base class when they get in.

MatthewMiddlehurst:
It would be good to give people a bit more time to view these.

1. Does the new test name fit? They will only be deterministic if you set `random_state`
2. Instead of skipping individual base classes and having to maintain it, We could have a `BaseDeepEstimator` at some point?
3. The skip on `test_save_estimators_to_file` seems to be above the comment instead of grouped with other skips

PR #430, opened 2023-05-13
[ENH] Parametrize model saving for deep classifiers and regressors

### LINKED ISSUE #407: [ENH] Add parameter to choose to save best and last models in base deep classifiers and regressors

If the user wants to save the best model than it shouldnt be deleted as done by default. Saving the last model should also be an option.

This will remove the dependency on the pickling save estimator to file test, this way we can skip it.

_No response_

### Additional context

_No response_

hadifawaz1999:
#### Reference Issues/PRs
Fix #407 

1. Remove unnecessary pickle stuff from base deep classifier
2. same as 1. but for base deep regressor
3. Add parameters to save best and last models
4. If the best model is to be saved then the user chooses a name for the file (by default now its deleted, this will happen if the parameter is set to False)
5. Use base class method to save last model with a chosen file name
6. Added some missing tags for deep classifiers

no

if they agree with code architecture of model saving and removal of pickle stuff

no need

##### For all contributions
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

### DIFF
--- a/aeon/classification/deep_learning/base.py
+++ b/aeon/classification/deep_learning/base.py
@@ -28,6 +28,14 @@ class BaseDeepClassifier(BaseClassifier, ABC):
     ----------
     batch_size : int, default = 40
         training batch size for the model
+    save_last_model     : bool, default = False
+        Whether or not to save the last model, last
+        epoch trained, using the base class method
+        save_last_model_to_file
+    last_file_name      : str, default = "last_model"
+        The name of the file of the last model, if
+        save_last_model is set to False, this parameter
+        is discarded
 
     Arguments
     ---------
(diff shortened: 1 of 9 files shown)


TonyBagnall:
all looks good, does this mean it will now pass test_save_estimators_to_file? Or is that dependent on pickle?

hadifawaz1999:
> all looks good, does this mean it will now pass test_save_estimators_to_file? Or is that dependent on pickle?

I think this test is for pickle only, @MatthewMiddlehurst can you confirm ?

MatthewMiddlehurst:
Not had a look through this properly yet. I think the test would currently require pickling, but the methods used here and the ones in the test don't currently match up anyway. We should eventually enable testing for saving deep learners however we decide to do it though.

hadifawaz1999:
I can take care of the test once we get this in.

TonyBagnall (approved):
LGTM

Issue #154, opened 2023-03-05
[MNT] Reduce the forecasting test suite execution time


lmmentel:
Currently testing the forecasting subpackage takes up  ~60% of the total testing time, which varies between 1-3 h :scream_cat:  and therefore need to be cut down to a reasonable number to improve dev experience and productivity.

This issue aim to collect ideas and suggestions  on where could we look for opportunities to cut down the testing time, so please share your thoughts.

Related:
- #151

patrickzib:
I want to look into the issue, too. 

For the long running tests, I have checked, 50% of the time is spent in validation of inputs rather than the downstream task. The same checks are repeated first for the full input X and y, then for conversion of X and y, then for each fold of X and y, then for the conversion of each fold, …

lmmentel:
Test named `test_predict_time_index` is run 7236 times, which is 134 estimators x 9 types x 6 forecasting horizons.

https://github.com/scikit-time/scikit-time/blob/5ede43c1f28c6ec7a7af3ba4d8350dd7cdbdba5b/sktime/forecasting/tests/test_all_forecasters.py#L202-L206


 There is however a separate test for forecasting horizons `test_fh` that is run 54 time for the same 9 types x 6 forecasting horizons.

https://github.com/scikit-time/scikit-time/blob/5ede43c1f28c6ec7a7af3ba4d8350dd7cdbdba5b/sktime/forecasting/base/tests/test_fh.py#L54-L58

It feels like those two are testing the same thing since `ForecastngHorizon` is a part of the `BaseForecaster` class

https://github.com/scikit-time/scikit-time/blob/5ede43c1f28c6ec7a7af3ba4d8350dd7cdbdba5b/sktime/utils/validation/forecasting.py#L265

and arguments passed as `fh` get coerced to `ForecastingHorizon` for all estimators using `BaseForecaster` by calling `check_fh`. 

It seems the original `test_predict_time_index` only tests forecasters inheriting from `BaseForecaster` and essentially testing the base class instead of actual implementation. Therefore I would drop the test entirely, unless I'm missing something.

MatthewMiddlehurst:
I think we have accomplished this somewhat with the `PR_TESTING` setup? I'm sure the individual tests could be improved a lot though.

TonyBagnall:
closing this, testing now in hand

Issue #151, opened 2023-03-04
[MNT] Cut down number of parameter combinations for testing


lmmentel:
We are currently running same tests with different parameter values that are not bringing much value in terms of coverage. 

Consider this test from forecasting

```python
@pytest.mark.parametrize("y", TEST_YS)
@pytest.mark.parametrize("fh", [*TEST_FHS, *TEST_FHS_TIMEDELTA])
@pytest.mark.parametrize("window_length", TEST_WINDOW_LENGTHS)
@pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
@pytest.mark.parametrize("initial_window", TEST_INITIAL_WINDOW)
def test_sliding_window_splitter_with_initial_window(
    y, fh, window_length, step_length, initial_window
):
```

and the corresponding parameter sets:

```python
TEST_OOS_FHS = [1, np.array([2, 5], dtype="int64")]  # out-of-sample
TEST_INS_FHS = [
    -3,  # single in-sample
    np.array([-2, -5], dtype="int64"),  # multiple in-sample
    0,  # last training point
    np.array([-3, 2], dtype="int64"),  # mixed in-sample and out-of-sample
]
TEST_FHS = [*TEST_OOS_FHS, *TEST_INS_FHS]
TEST_OOS_FHS_TIMEDELTA = [
    [pd.Timedelta(1, unit="D")],
    [pd.Timedelta(2, unit="D"), pd.Timedelta(5, unit="D")],
]  # out-of-sample
TEST_INS_FHS_TIMEDELTA = [
    pd.Timedelta(-3, unit="D"),  # single in-sample
    [pd.Timedelta(-2, unit="D"), pd.Timedelta(-5, unit="D")],  # multiple in-sample
    pd.Timedelta(0, unit="D"),  # last training point
    [
        pd.Timedelta(-3, unit="D"),
        pd.Timedelta(2, unit="D"),
    ],  # mixed in-sample and out-of-sample
]
TEST_FHS_TIMEDELTA = [*TEST_OOS_FHS_TIMEDELTA, *TEST_INS_FHS_TIMEDELTA]
TEST_WINDOW_LENGTHS_INT = [1, 5]
TEST_WINDOW_LENGTHS_TIMEDELTA = [pd.Timedelta(1, unit="D"), pd.Timedelta(5, unit="D")]
[...]

Issue #59, opened 2023-02-14
[ENH] CI is slow


patrickzib:
CI is slow on some platforms with up to 100 minutes. 

Among the slowest tests is `sktime/forecasting/tests/test_all_forecasters.py::TestAllForecasters`
<img width="1085" alt="image" src="https://user-images.githubusercontent.com/7783034/218746195-ce992e40-ba56-4754-959b-71d72cb116a1.png">

Screenshot taken from: `[test-unix (3.9, macOS-11)]`

To start with: can we do something about these forecasting tests? They take >1000 seconds.

patrickzib:
Another example: 
<img width="542" alt="image" src="https://user-images.githubusercontent.com/7783034/218746980-21e6766b-2fe2-4962-9145-6390f169f4cf.png">

<img width="1073" alt="image" src="https://user-images.githubusercontent.com/7783034/218747089-8122723a-46b0-43f6-bb72-621d17191d65.png">

lmmentel:
Yup this is way too long to be practical. I would say 10 min is closer to a comfortable upper limit but we are far away from that. 

I suspect that the way tests are run is partly responsible to the horrendous execution time since the test config has 1400+ LOC and it imports **everything** and dispatches estimator to test suites (or something like that). One performance penalty here is that **all** numba functions are compiled which in turn make the test collection step very long. It is also very hard to modify and maintain the testing monstrosity since it's partly centralized and party distributed into individual modules.

I don't know I am able to offer a strategy here to deal with that but I would probably distribute the test to subpackages or individual modules and move away from the centralized runner. The aim would be to enable running different granularity of tests e.g. test only the `classification` subpackage or a single module file. I don't think this is easy to do now.

Let me know if there are better alternatives.

TonyBagnall:
the testing is a monster, and personally I would like to scrap it all and go back to something closer to what it was like, which was based on scikit-learn. There are also a lot of possibly unnecessary tests, and a confusion between interface testing and use case testing. It is not feasible to test edge cases in general CI imo, but having different testing parameters seems to be trying to half heartedly do just that. I would like levels of testing

patrickzib:
Linux is worst now with **3 hours**?
<img width="716" alt="image" src="https://user-images.githubusercontent.com/7783034/220332407-354e8012-e7d8-4abe-8cf5-92428162f007.png">

MatthewMiddlehurst:
[...]

PR #437, opened 2023-05-15
[MNT] Switch test configs on MATRIXDESIGN

TonyBagnall:
Reopened this due to overnight testing. The idea is to switch on MATRIXDESIGN. If True, we try minimal configurations aimed at coverage rather than completeness. If False, it does the full tests. We then run the full tests with configs on the overnight runs, and a much faster daily test. 

test_all_forecasters:
uses the forecasting/tests/_config.py for the the matrix tests, forecasting/tests/_config_full.py for the overnight tests. 
 
#59 #151 #154 #271

to come: running off main test times 
test_all_forecasters:
test_all_estimators: 
test_all_classifiers:

running with MATRIXDESIGN:
test_all_forecasters:
test_all_estimators: 
test_all_classifiers:


ltsaprounis (commented):
Left some suggestions mainly for adding some things back in the config or replacing something with something else. This is because so of the removed test_configs are things that can fail if someone adds a new forecaster etc.

TonyBagnall:
> 

great, thanks

aeon/forecasting/tests/_config.py
```diff
@@ -67,43 +67,35 @@
...
     # index type, fh type, is_relative
     ("int", "int", True),
     ("int", "int", False),
-    ("range", "int", True),
-    ("range", "int", False),
-    ("period", "int", True),
-    ("period", "period", False),
-    ("datetime", "int", True),
-    ("datetime", "datetime", False),
-    ("datetime", "timedelta", True),
```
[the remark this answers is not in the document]
TonyBagnall: I guessed as much, why I commented it rather than delete :)

TonyBagnall:
@ltsaprounis I think I reverted those, could you check when you get a chance?

TonyBagnall:
these reverts mean we are back to 17k tests and taking an hour. It has at least identified the prime suspect, testing nine variants of VALID_INDEX_FH_COMBINATIONS creates 10k extra tests and takes it back to an hour. If I run with just these

VALID_INDEX_FH_COMBINATIONS = [
    # index type, fh type, is_relative
    ("int", "int", True),
    ("int", "int", False),
 #   ("range", "int", True),
 #   ("range", "int", False),
 #   ("period", "int", True),
 #   ("period", "period", False),
 #   ("datetime", "int", True),
 #   ("datetime", "datetime", False),
 #   ("datetime", "timedelta", True),
]

it takes 15 mins. I will close this PR for now, I feel there is no enthusiasm for the _confg and_config_full solution, it is a hack really, and the important BATS param changes are now in https://github.com/aeon-toolkit/aeon/pull/568, which does not change coverage. I think we should use the above settings but introduce a one off test that just checks 
VALID_INDEX_FH_COMBINATIONS
without all combinations.

aeon/forecasting/tests/_config.py
```diff
@@ -87,29 +85,19 @@
...
     ("int", "int", True),
-    ("int", "int", False),
-    ("range", "int", True),
-    ("range", "int", False),
-    ("period", "int", True),
-    ("period", "period", False),
-    ("datetime", "int", True),
     ("datetime", "datetime", False),
-    ("datetime", "timedelta", True),
 ]
```
MatthewMiddlehurst, line 98: based on the above comments the main issue seems to be this, may be worth adding back some of the important parameters based on @ltsaprounis comments above?

MatthewMiddlehurst (commented):
I think this is good in principle, but am unsure about the actual values for each config set.

Ideally, we want the most common/easy/important failures in the PR config

aeon/forecasting/tests/_config.py
TonyBagnall: I have no feel for it tbh, happy for anyone to advise, but remember this is meant to be minimal combinations and the expectation is people run test_all_forecasters before a forecasting PR

aeon/forecasting/tests/_config_full.py
```diff
@@ -0,0 +1,115 @@
...
+    "TEST_INS_FHS",
+    "TEST_OOS_FHS",
+    "TEST_WINDOW_LENGTHS_INT",
+    "TEST_WINDOW_LENGTHS",
+    "TEST_INITIAL_WINDOW_INT",
+    "TEST_INITIAL_WINDOW",
+    "VALID_INDEX_FH_COMBINATIONS",
+    "INDEX_TYPE_LOOKUP",
+    "TEST_N_ITERS",
+]
```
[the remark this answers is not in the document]
TonyBagnall: good spot, was there from ages ago I suspect

aeon/forecasting/tests/_config.py
```diff
@@ -2,52 +2,35 @@
...
-    pd.to_datetime(["2000-01-04", "2000-01-08", "2000-01-11"]),
+    pd.to_datetime(["2000-01-23"]),
+    pd.to_datetime(["2000-01-04", "2000-01-08"]),
 ]
 TEST_CUTOFFS = [*TEST_CUTOFFS_INT_LIST, *TEST_CUTOFFS_INT_ARR, *TEST_CUTOFFS_TIMESTAMP]
 
-TEST_WINDOW_LENGTHS_INT = [1, 5]
-TEST_WINDOW_LENGTHS_TIMEDELTA = [pd.Timedelta(1, unit="D"), pd.Timedelta(5, unit="D")]
-TEST_WINDOW_LENGTHS_DATEOFFSET = [pd.offsets.Day(1), pd.offsets.Day(5)]
+TEST_WINDOW_LENGTHS_INT = [1]
```
MatthewMiddlehurst, line 22: Maybe this should still be 5 to test a different length?

MatthewMiddlehurst (commented):
Think this is fine for now, one comment.

MatthewMiddlehurst (approved):

aeon/forecasting/tests/_config.py
TonyBagnall: changed back to [1 5[

Issue #438, opened 2023-05-16
[ENH] Classifiers overriding base class fit


TonyBagnall:
### Describe the issue

I would like to impose that fit/predict/predict_proba are final in the base class. This will reduce the amount of interfacing testing we have to do: If a classifier cant override fit, it cannot replace base class functionality. However, there are some classifiers that override fit. This is related to the strange duplication of time series forest! They overrid fit because of the multiple inheritance problems

Ones Ive found so far:
class ComposableTimeSeriesForestClassifier(BaseTimeSeriesForest, BaseClassifier):
class TimeSeriesForestClassifier(
    BaseTimeSeriesForest, ForestClassifier, BaseClassifier
):

### Suggest a potential alternative/fix

I think we should either fix these classifiers so they inherit fit and implement abstract _fit, or remove them.

TonyBagnall:
fixed by #575

Issue #435, opened 2023-05-15
[ENH] Tabularizer transformer still uses nested_univ data type


TonyBagnall:
### Describe the feature or idea you want to propose

Tabularizer still has inner data type
        "X_inner_mtype": ["nested_univ", "numpy3D"],

this transformer is used all over t he place, so we need to be careful changing it. It has 19 uses in benchmarking, forecasting and transformations

### Describe your proposed solution

it needs to be converted to 
        "X_inner_mtype": ["np-list", "numpy3D"],

TonyBagnall:
Just clarifying where and why this transformer is used. 

in file forecasting/compose/reduce
it is used as a local import in get_test_params

Issue #383, opened 2023-05-04
[ENH] Loading data when return_X_y is false


TonyBagnall:
### Describe the feature or idea you want to propose

post #355 if return_X_y is true (default) all collections loaders return either 3D numpy for equal length X or list of 2D numpy for unequal length X, and y as a numpy array. However,  if return_X_y is False, it still loads it into a nested_univ (pd.DataFrame), This needs to be done better. The problem is that Y can contain strings or floats(e.g. BasicMotions has string class labels).

I want to do this post #355. The only place return_X_y is used is in the moribund part of the benchmarking module. 

### Describe your proposed solution

I'm assuming you cannot mix floats and strings in the same numpy array (although could it an object array? Not sure how python deals with primitives as objects, weirdly I would imagine). So then we could

1. No longer allow the combined return. Butting labels and time series 
2. Encode the strings to map them onto integers. This is what happens internally with the label encoder. However, this may have side effects. We could return the string labels as meta data
3. Leave it as nested univariate until it breaks for someone

chrisholder:
So you could use numpy for this something like this:

`res = np.array(("TEXT", 1, 1), dtype='|S4, i4, i4')
`
See: https://stackoverflow.com/questions/24832715/numpy-array-matrix-of-mixed-types for more examples

I don't know how much numba would like this? I don't know if there really is an instance where you'd be calling a numba function with the label in the same array as the time series (most of the numba stuff in aeon I see assumes the data is without the label or has it split already) so it might not be an issue.

Other suggestions could be having a tuple returns like:

(ts_collection, labels) 

where ts_collection is 3D np.ndarray (or List[np.ndarray]) of floats that is the data and labels is 1D np.ndarray of str that .

TonyBagnall:
so we now just return it as a tuple if return_X_y is true, and thats ok for me, alternative would be just remove the option, but no need for that

PR #439, opened 2023-05-16
[ENH] remove unsupported legacy benchmarking code

### LINKED ISSUE #116: [ENH] Benchmarking module

TonyBagnall:
#### Reference Issues/PRs

Fixes #116 
Related to #383 #435 
Follows #379 

I think now is the time to remove the legacy benchmarking code no one is supporting. Why now?
1. We have the kotsu framework to work with
2. this is about the only place nested_univ is still required
3. it has to assumes nested_univ because it stores X and y together in a dataframe. This is hard to reconcile with numpy storage (#383)
4. it uses ComposableTimeSeriesForest a lot, which breaks the classifier contract (#438)
5. I dont like the structure, its just over engineered imo. Orchestration, strategy, task. Just run some experiments and display results!  (#116)
I'll leave this up for a bit, but I really dont think anyone wants to champion this code


ltsaprounis (approved):
LGTM - I have never used the benchmarking module.

MatthewMiddlehurst (commented):
Quite a large removal, so should probably be left up for a while to allow for comment, I agree.

I have not used any of this also, and I'm fine with removing it. I think some of the individual functions may be salvageable, though? i.e. the metrics, significance testing and plotting (`plot_critical_difference_diagram`). Though I have not had a thorough look, so maybe this is not the case.

PR #450, opened 2023-05-19
[ENH] refactored predict in `BaseDeepClassifier`

### LINKED ISSUE #448: [ENH] in BaseDeepClassifier the `_predict` functions needs refactoring

The current `_predict` function in base deep learning models.
Currently it checks for random state for some reason and it has a for loop ...

It should simply return `np.argmax(probs, axis=1)`

_No response_

### Additional context

_No response_

hadifawaz1999:
#### Reference Issues/PRs
Fix #448 

The implementation of `_predict`  method in `BaseDeepClassifier` is not optimal, it uses a for loop for some reason and checks the random state.. It simply should use `np.argmax`.

### DIFF
--- a/aeon/classification/deep_learning/base.py
+++ b/aeon/classification/deep_learning/base.py
@@ -19,7 +19,6 @@ from abc import ABC, abstractmethod
 
 import numpy as np
 from sklearn.preprocessing import LabelEncoder, OneHotEncoder
-from sklearn.utils import check_random_state
 
 from aeon.classification.base import BaseClassifier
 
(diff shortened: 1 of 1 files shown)


TonyBagnall:
ah I think I remember why it does it like this: to choose ties randomly. I think np.argmax does not do this, returns the first if equal?

MatthewMiddlehurst:
These should match `BaseClassifier` as much as possible.

hadifawaz1999:
Closing this. For now it looks like there is no other way to implement this more efficiently.

PR #463, opened 2023-05-30
[ENH] data loaders and writers

### LINKED ISSUE #382: [ENH] Loading direct from website

After #355 the function that loaded direct from timeseriesclassification.com needs reworking. I want to tidy that up and include the regression problems, but want to do it after #355 which is too big already. They can already be loaded locally. 

properly load all classification and regression problems either from a local path or from the website, with a notebook explaining data format etc
 

_No response_

### Additional context

_No response_

### LINKED ISSUE #440: [MNT] `load_TSC_dataset` is not imported in `__init__.py`
### Describe the issue

After refactoring the data loading module #355 , the function `load_TSC_dataset` that replaced `load_UCR_UEA_dataset` is missing, should get back in.

_No response_

### Additional context

_No response_

### LINKED ISSUE #449: [MNT] `test_write_to_tsfile_unequal_length` occasionally raises a `FileNotFoundError`
### Describe the issue

This happens sporadically, see:

https://github.com/aeon-toolkit/aeon/actions/runs/5013217453/jobs/8986467056
https://github.com/aeon-toolkit/aeon/actions/runs/4990819828/jobs/8995302918?pr=318

_No response_

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs
Fixes #382 
Fixes #449 
Fixes #440 

Tidies up the dataset loading, allow setting return type and bring back loading directly from tsc.com. Also tidies up data writing, including ability to write regression problems, and improves testing for writing. I'm still not sure what to do about #383, I think we should possibly remove the ability to download X and y as a single data structure. 

This PR contains some rough notebooks. I started them here but I want to finish these in a separate PR to avoid scope creep

Became a bit larger than I thought it would be. In datasets we know have

**loading**
_data_loaders with public functions that load into numpy or list of numpy
load_from_tsfile (I would refactor to load_from_ts_file but its used in too many places!)
load_from_arff_file
load_from_tsv_file
load_from_tsc (rename?)
load_from_tsf_file (this simplifies load_from_tsf_fto_dataframe)

I have not converted load_from_longtable_to_dataframe. It had no way of determining the class variable so everything is in a dataframe. I think we assume that if someone has data in this format they can work out how to format it into numpy. All of the load_X_to_dataframe are still there, just moved to _dataframe_loaders

test_data_loaders

I have removed an insane test that was 1000 lines of code and testing dataframe compatibility for the removed function and greatly simplified the other tests. Adapted the tsf test to also test the new tsf loader.

**writing**
_data_writers has two public methods
write_to_tsfile
write_results_to_uea_format
the latter can move to benchmarking with the redesign of this. The former now supports regression problems. 

I have also deleted the setup.py file in datasets which was made redundant a long time ago. On slack, @lmmentel says "This is an artifact from the previous, scipy inspired, config where each subpackage had it's own setup.py If there isn't any other module within the subpackage using it, it should be safe to delete."
 
there is a file data_generators.py which can probably also go to benchmarking, or the new simulatation module. There are generators in testing that could also be co located with these at a later date.


aeon/datasets/_data_loaders.py
```diff
@@ -202,14 +217,74 @@ def load_from_tsfile(
...
+    -------
+    X: np.ndarray or list
+    y: numpy array
+        The class labels for each case in X, returned separately if return_X_y is
+        True, or appended to X if False
+
+    Examples
+    --------
+    >>> from aeon.datasets import load_from_tsc
+    >>> X, y = load_from_tsc(name="ArrowHead")
```
MatthewMiddlehurst, line 283: We don't want this to run in tests

aeon/datasets/tests/test_data_loaders.py
```diff
@@ -15,28 +14,24 @@
...
-    _load_provided_dataset,
-    load_from_tsfile,
-)
+from aeon.datasets._data_loaders import DIRNAME, MODULE, _load_provided_dataset
 from aeon.datatypes import check_is_mtype
 
 
 @pytest.mark.parametrize("return_X_y", [True, False])
-@pytest.mark.parametrize("return_type", ["nested_univ", "numpy3D"])
+@pytest.mark.parametrize("return_type", ["nested_univ", "numpy3D", "numpy2D"])
```
MatthewMiddlehurst, line 34: I think this should be "numpyflat"

aeon/datasets/_data_loaders.py
TonyBagnall: I guess I can doctest skip this

aeon/datasets/tests/test_data_loaders.py
TonyBagnall: # Return appropriate return_type in case an alias was used
def _alias_datatype_check(return_type):
    if return_type in ["numpy2d", "numpy2D", "np2d", "np2D"]:
        return_type = "numpyflat"
    if return_type in ["numpy3d", "np3d", "np3D"]:
        return_type = "numpy3D"
    return return_type

MatthewMiddlehurst (approved):

PR #465, opened 2023-05-31
[ENH] `numba` utility functions

MatthewMiddlehurst:
Adds some new numba utility functions, and adds examples to most existing ones.

### DIFF
--- a/aeon/utils/numba/general.py
+++ b/aeon/utils/numba/general.py
@@ -1,6 +1,26 @@
 # -*- coding: utf-8 -*-
 """General numba utilities."""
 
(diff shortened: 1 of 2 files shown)


patrickzib:
LGTM. 

Just a small comment: first order differences can also be computed with np.diff along any axis. And z-normalization with zscore from scipy. 

Would be nice to benchmark the differences between numba‘rized implementations and these default implementations.

MatthewMiddlehurst:
I think I did benchmark all these functions a while back, and they either had a miniscule difference or were much faster.

Main benefit is having them available for use in other numba functions really.

PR #331, opened 2023-04-24
[ENH] purge mentions of Panel in classification

TonyBagnall:
#### Reference Issues/PRs

Part of #42 

#### What does this implement/fix? Explain your changes.

Purely cosmetic PR removes mentions of the term Panel from classification, clustering and regression which to machine learning people  has no meaning at all. mtype generally changed to input type. Left the pipelines alone, they are of marginal use. Not changed any code, conversion and checking changes comes later.

PR #466, opened 2023-06-01
[MNT] Rename `panel` transformers folder to `collection`

MatthewMiddlehurst:
See #263 and #331.

The current `panel` transformers are essentially only used in the machine learning modules (there is only 1 transformer used in the forecasting package in 2 files). With the new collection base class changing the functionality to differentiate these from regular transformers, this PR also renames the `panel` folder to match the new terminology used. 

I have nothing against keeping the `panel` folder for certain transformers if people want, and these transformers would then be excluded from the collection base class changes.


MatthewMiddlehurst:
Didn't mean to do that, oops!

TonyBagnall:
its a monster!

TonyBagnall (approved):

Issue #472, opened 2023-06-05
[BUG] ResNetClassifier fails tests


TonyBagnall:
### Describe the bug

ResNetClassifier fails all the tests due to an issue cloning the classifier

E RuntimeError: Cannot clone object ResNetClassifier(batch_size=4, n_epochs=10, optimizer=(None,)), as the constructor either does not set or modifies parameter optimizer

### Steps/Code to reproduce the bug

_config, comment out resnet, then run test_all_estimators

### Expected results

pass the tests

### Actual results

estimator = ResNetClassifier(batch_size=4, n_epochs=10, optimizer=(None,))

    def clone(estimator, *, safe=True):
        """Construct a new unfitted estimator with the same parameters.
    
        Clone does a deep copy of the model in an estimator
        without actually copying attached data. It returns a new estimator
        with the same parameters that has not been fitted on any data.
    
        Parameters
        ----------
        estimator : {list, tuple, set} of estimator instance or a single \
                estimator instance
            The estimator or group of estimators to be cloned.
        safe : bool, default=True
            If safe is False, clone will fall back to a deep copy on objects
            that are not estimators.
    
        Returns
        -------
        estimator : object
            The deep copy of the input, an estimator if input is an estimator.
    
        Notes
        -----
        If the estimator's `random_state` parameter is an integer (or if the
        estimator doesn't have a `random_state` parameter), an *exact clone* is
        returned: the clone and the original estimator will give the exact same
        results. Otherwise, *statistical clone* is returned: the clone might
        return different results from the original estimator. More details can be
        found in :ref:`randomness`.
        """
        estimator_type = type(estimator)
        # XXX: not handling dictionaries
        if estimator_type in (list, tuple, set, frozenset):
            return estimator_type([clone(e, safe=safe) for e in estimator])
        elif not hasattr(estimator, "get_params") or isinstance(estimator, type):
            if not safe:
                return copy.deepcopy(estimator)
            else:
                if isinstance(estimator, type):
                    raise TypeError(
                        "Cannot clone object. "
                        + "You should provide an instance of "
                        + "scikit-learn estimator instead of a class."
                    )
                else:
                    raise TypeError(
                        "Cannot clone object '%s' (type %s): "
                        "it does not seem to be a scikit-learn "
                        "estimator as it does not implement a "
                        "'get_params' method." % (repr(estimator), type(estimator))
                    )
    
        klass = estimator.__class__
        new_object_params = estimator.get_params(deep=False)
[...]

PR #473, opened 2023-06-05
[ENH] Tidy up test exclusions

TonyBagnall:
This PR tidies up some of the exclusions in testing/_config.py and updates some of the issues

1. Removes exclusions I dont think we need any more. Many referenced issues in the bad place which have since been closed. If they cause intermittent failure, I would like to see the fails, raise an issue and re-exclude again.
 2. Removes a redundant function from RISE
 3. Changes PlateauFinder to not store the intervals it uses
 

#### Reference Issues/PRs
#472 #347

### DIFF
--- a/aeon/classification/deep_learning/tapnet.py
+++ b/aeon/classification/deep_learning/tapnet.py
@@ -270,11 +270,4 @@ class TapNetClassifier(BaseDeepClassifier):
             "dilation": 2,
             "layers": (32, 16),
         }
-        # param2 = {
-        #     "n_epochs": 20,
-        #     "use_cnn": False,
-        #     "layers": (25, 25),
-        # }
-        test_params = [param1]
-
-        return test_params
+        return [param1]
--- a/aeon/classification/interval_based/_rise.py
+++ b/aeon/classification/interval_based/_rise.py
@@ -204,14 +204,6 @@ class RandomIntervalSpectralEnsemble(BaseClassifier):
 
         super(RandomIntervalSpectralEnsemble, self).__init__()
 
-    @property
-    def feature_importances_(self):
-        """Feature importance not supported for the RISE classifier."""
-        raise NotImplementedError(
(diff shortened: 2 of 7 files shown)


TonyBagnall:
@hadifawaz1999 I think this is just a random DL fail, I can see no reason for it failing, but could you take a look and see if there is anything 
https://github.com/aeon-toolkit/aeon/actions/runs/5186055397/jobs/9346945374?pr=473

image

hadifawaz1999:
> @hadifawaz1999 I think this is just a random DL fail, I can see no reason for it failing, but could you take a look and see if there is anything 
> https://github.com/aeon-toolkit/aeon/actions/runs/5186055397/jobs/9346945374?pr=473
> 
> image
> 
> 

Hmm thats super weird it shouldn't do that, are you running it locally or is this on CI ? will try the test multiple times locally today, will let you know what happens.

TonyBagnall:
This is from the CI, I think its a form of interference from saving on CI, but I can honestly not see how the network would be uninitialised

TonyBagnall:
more limited un-exclusion:
1. passes
2.

hadifawaz1999:
> This is from the CI, I think its a form of interference from saving on CI, but I can honestly not see how the network would be uninitialised

The CI saving model is now fixed in #485

MatthewMiddlehurst:
I don't think the latest failure is on our end, so i would disregard it. Seems like its not only us https://github.com/codecov/codecov-action/issues/1003.

The PR doesnt actually change the removed estimators list, so not sure why failure #7 happened! I have a hunch some of the DL ones are memory hungry and could be reduced, but thats just speculation

hadifawaz1999:
Do we have some clue about the other ones still excluded such as HC1 ? @TonyBagnall

TonyBagnall:
@hadifawaz1999 I'm waiting for @MatthewMiddlehurst 's big revamp of the tree ensembles before going there

Issue #451, opened 2023-05-19
[ENH]  Benchmarking tools for download, collation and display of bake off results


TonyBagnall:
### Describe the feature or idea you want to propose

I would like to be able to pull reference results from timeseriesclassification.com, compare against my own results, perform tests of difference, plot results and generally auto-generate a report on differences. 

### Describe your proposed solution

suite of tools in the revamped benchmarking module

TonyBagnall:
so my wish list of things I would like to be able to draw from data drawn from tsc.com

1. box/violin plots (@patrickzib maybe?)
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/bea09214-3212-423c-967f-579f5f70be29)

2. pairwise scatter plots (but with the w/d/l)
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/6107ccf0-9294-47fd-8f8a-b57ea1e43bf7)

3. Funky heat maps (@hadifawaz1999)
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/9bc295b4-83f0-4b46-b978-a1b6edfdfae3)

4. Time accuracy graphs (clearly problematic with hardware!)
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/da4095de-cf97-499b-9e12-4e4421c59a22)

TonyBagnall:
I was wondering whether to actually put the results in aeon in the benchmarking module. They are small, one tiny csv per classifier, and it would remove the need to go via timeseriesclassification.com

patrickzib:
> I was wondering whether to actually put the results in aeon in the benchmarking module. They are small, one tiny csv per classifier, and it would remove the need to go via timeseriesclassification.com

Just a small comment: if you load data from the tsc.com website, via:

```python
    URL = "https://..."
    data = np.loadtxt(URL, ...)
```

It will get cached locally by numpy. Thus, all subsequent requests by the same script are can be answered via the local cache.

hadifawaz1999:
![image](https://github.com/aeon-toolkit/aeon/assets/54309336/b74146e5-ab14-47b2-a8bb-9ffda4024a46)

TonyBagnall:
so we now can load results and have the visualisation module, so I think we can close this and anything specific to those modules can be a separate issue

PR #481, opened 2023-06-09
[ENH] Introduce alternative clique finding methods

TonyBagnall:
#### Reference Issues/PRs
#451 
dependent on #452 

This will allow CD cliques to be formed using pairwise test results, either with Bonferroni adjustment or Holm correction. I am also tidying up the CD drawer.

One thing with the post hoc nemenyi test: the qalpha values are hard coded for three alpha levels and set by n_estimators only (not number of problems/rows). I assume these are for very large n, but there seem to be methods that set it by both n_estimators (columns) and n_problems (rows). I will investigate further


hadifawaz1999:
Small question, its better to use the Friedman test from scipy isn't it ? As done in [here](https://github.com/hfawaz/cd-diagram/blob/1b48b4b2ff262de502757db357d1247d56fdc068/main.py#L21) for instance. Not sure if you agree, but it would reduce the amount of code. Correct me if this is a version that is different than the one you are using in the cdd code.

aeon/benchmarking/_critical_difference.py
```diff
@@ -109,7 +109,7 @@ def pairwise_cliques(
         if correction == "holmes":
             corr = n_estimators - i
         elif correction == "bonferroni":
-            corr = n_estimators * (n_estimators) - 1
+            corr = (n_estimators * (n_estimators) - 1) / 2.0
```
hadifawaz1999, line 112: Unless am wrong, but the `corr` value should be the number of null hypothesis to test, and this number should be equal to the number of pairs of estimators, basically the combination of 2s from `N` estimators. So the best thing is to do `math.comb(n_estimators, 2)` because I think the formula used now in the code produces something else. Please correct me if am saying nonsense :)

TonyBagnall: @hadifawaz1999 thats a bonferroni adjustment, and its conservative. You do not actually need to perform all tests to find a single clique. The Holme correction takes a different approach. For each clique, you take the highest ranked as a control, and compare in order to the remaining estimators. Thus the correction for first clique is n_estimators-1. Same for second, correction n_estimators-2.

TonyBagnall: there is no right answer to this, but I do think Bonferroni is overly conservative, and I think most of the literature backs me up

TonyBagnall: but on the above, the number of pairwise tests is (n_estimators * (n_estimators) - 1) , but we divide by 2 because we dont do tests both ways (which I had forgotten to do abovew, hence the correction :)

hadifawaz1999: Oh okay I see,  I always thought this correction just divides by `m` for some reason, thanks for the info !

TonyBagnall:
> 

yes definitely, this may not be the version going in, frankly doesnt look very good. We should also use the proper qstat values for the posthoc test, rather these hard coded values which I think are for very large n only and a weird legacy of the original matlab code :)

TonyBagnall:
made redundant by #582

Issue #474, opened 2023-06-06
[BUG] RDST transformer crashes without error message


MatthewMiddlehurst:
### Describe the bug

When trying to run `RandomDilatedShapeletTransform` on ArrowHead, the process crashes on two different Windows machines and Python versions for me.

### Steps/Code to reproduce the bug

```python
from aeon.datasets import load_arrow_head
from aeon.transformations.panel import RandomDilatedShapeletTransform

if __name__ == "__main__":
    X_train, y_train = load_arrow_head(split="train", return_X_y=True)

    r = RandomDilatedShapeletTransform()

    print(r.fit_transform(X_train, y_train))
```

### Expected results

The transformer does not crash or outputs a suitable error message.

### Actual results

```python-traceback
Process finished with exit code -1073740940 (0xC0000374)
```

### Versions

<details>
System:
    python: 3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)]

executable: D:\CMP Machine Learning\Repositories\aeon\venv\Scripts\python.exe
   machine: Windows-10-10.0.19045-SP0

Python dependencies:
          pip: 22.3.1
   setuptools: 65.5.1
      sklearn: None
         aeon: 0.0.1
  statsmodels: None
        numpy: 1.23.5
        scipy: 1.10.1
       pandas: 1.5.3
   matplotlib: 3.7.1
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: None
      tsfresh: None
<\details>

MatthewMiddlehurst:
FYI @baraline

TonyBagnall:
I can reperoduce this, doesnt happen with UnitTest though!

TonyBagnall:
```python
from aeon.transformations.collection import RandomDilatedShapeletTransform
from aeon.datasets import load_unit_test, load_arrow_head

X, y = load_unit_test(split="train")
rdst=RandomDilatedShapeletTransform()
print(" X type = ",type(X), "Shape X = ",X.shape, "y type = ",type(y)," Y shape = ",
      y.shape)
X2= rdst.fit_transform(X,y)
X, y = load_arrow_head(split="train")
print(" X type = ",type(X), "Shape X = ",X.shape, "y type = ",type(y)," Y shape = ",
      y.shape)
X2= rdst.fit_transform(X,y)
print(X2)
```

TonyBagnall:
[...]

PR #483, opened 2023-06-11
[BUG] RDST bugfixes

baraline:
This PR introduce bug fixes for RDST transformer to make its performance equal to the performance of the original implementation in the convst package.

- Additional Fixes for #474 
- Distance function as argument, with Manhattan distance as default, as in convst.
- Added Manhattan distance to aeon distances
- Fixed the initialization of the lambda threshold, it was only considering the lower bound and not both the lower and upper bounds.
- Refactored distance vector and shapelet feature extraction
- Removed the tentative optimization of using MASS algorithm to compute normalized distance profiles, this will be introduced in a future PR using the distance function argument to specify if it shall be used or not.

No.

- @MatthewMiddlehurst noted some huge accuracy difference between the convst implementation and this one, I validated locally that the differences are gone, but a check from someone else would be nice.

Unit test were added for the new functions, and I updated the one for #474

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc).
- [X] For new estimators, I've added the estimator to the online documentation.

### DIFF
--- a/aeon/transformations/collection/dilated_shapelet_transform.py
+++ b/aeon/transformations/collection/dilated_shapelet_transform.py
@@ -14,6 +14,7 @@ import numpy as np
 from numba import njit, prange, set_num_threads
 from sklearn.preprocessing import LabelEncoder
 
+from aeon.distances import manhattan_distance
 from aeon.transformations.base import BaseTransformer
 from aeon.utils.numba.general import (
     choice_log,
@@ -21,7 +22,6 @@ from aeon.utils.numba.general import (
     get_subsequence,
     get_subsequence_with_mean_std,
     set_numba_random_seed,
-    sliding_dot_product,
     sliding_mean_std_one_series,
 )
 from aeon.utils.numba.stats import prime_up_to
@@ -67,6 +67,10 @@ class RandomDilatedShapeletTransform(BaseTransformer):
         initialized such as it will use a z-normalized distance, inducing either scale
         sensitivity or invariance. A value of 1 would mean that all shapelets will use
         a z-normalized distance.
+    distance_function: function
+        A distance function defined as a numba function with signature as
+        (x: np.ndarray, y: np.ndarray) -> float. The default distance function is the
+        manhattan distance.
     threshold_percentiles : array, default=None
         The two percentiles used to select the threshold used to compute the Shapelet
         Occurrence feature. If None, the 5th and the 10th percentiles (i.e. [5,10])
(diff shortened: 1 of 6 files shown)


MatthewMiddlehurst (dismissed):
Looks good to me, will give it a run through. Don't have to wait for that to merge this though,

@chrisholder @TonyBagnall may be better suited to reviewing the new distance, so will wait for them.

baraline:
Tests seem to fail on test_constructor for all estimators, is a numba function (the newly introduced distance parameter) as an argument not accepted ? It don't get why It worked on some version and failed on others, am I missing something ?

MatthewMiddlehurst:
The tests are spread to only be on certain OS/Python combos. I would make the default None or str.

MatthewMiddlehurst:
Tried to run it on our cluster and we have arrived back at https://github.com/baraline/convst/issues/24 🙂 

I think it doesnt like caching+function input. Perhaps it breaks one of the requirements here. https://numba.pydata.org/numba-doc/latest/developer/caching.html

aeon/transformations/collection/dilated_shapelet_transform.py
```diff
@@ -645,37 +686,106 @@ def dilated_shapelet_transform(X, shapelets):
...
+                X_subs[i_sub, i_channel, i_length] = X[
+                    i_channel, i_sub + (i_length * dilation)
+                ]
+    return X_subs
+
+
+@njit(fastmath=True, cache=True)
+def compute_shapelet_features(
+    X_subs, values, length, dilation, threshold, distance_function
+):
```
MatthewMiddlehurst, line 788: I don't think the dilation input is required for this anymore, same for the functions below.

MatthewMiddlehurst:
If i remove the distance parameter and just use the Manhattan function it seems fine. Perhaps instead of inputting a function it could be string only and select based on that inside the function based on that?

I have never run into this numba quirk before, rather annoying.

baraline:
It indeed seems to be the reason for the issue we were experimenting ... Indeed, a bit annoying but at least it's nice to have part of the answer. I'll try to come up with something using a str argument,
It would be annoying to have to define individual transforms functions for each distance.

MatthewMiddlehurst:
After running on some UCR datasets the fix appears to have worked and it performs as expected accuracy wise, however.

baraline:
All my attempts in including the choice of distance function in the Numba functions caused problems with the caching, so I made the manhattan distance by default for all functions for consistency with convst.

> After running on some UCR datasets the fix appears to have worked and it performs as expected accuracy wise, however.

However ?

MatthewMiddlehurst:
ah just meant that despite the new issue the accuracy looks good. Think after this the transformer should be fine

MatthewMiddlehurst:
@TonyBagnall @chrisholder the transform side of things looks good to me. Any comments on the new distance?

TonyBagnall (approved):
LGTM

PR #486, opened 2023-06-13
[MNT] tidy up resnet and fix failures

### LINKED ISSUE #472: [BUG] ResNetClassifier fails tests

hadifawaz1999:
This should fix #472


aeon/classification/deep_learning/resnet.py
```diff
@@ -221,11 +219,11 @@ def build_model(self, input_shape, n_classes, **kwargs):
 
         input_layer, output_layer = self._network.build_network(input_shape, **kwargs)
 
-        output_layer = keras.layers.Dense(
-            units=n_classes, activation=self.activation, use_bias=self.use_bias
+        output_layer = tf.keras.layers.Dense(
+            units=n_classes, activation="softmax", use_bias=self.use_bias
```
MatthewMiddlehurst, line 223: Is the activation parameter used anywhere with this change? it also seems to be set twice in the init

hadifawaz1999: Hey @MatthewMiddlehurst , thanks for noticing the double setting ! will fix this now, this might be a reason as well for failing. And yes the activation parameter is used for the ResNetNetwork class object (the same setup for all deep learners). I just had forgotten to fix this issue when doing the refactor, should always check everything :)

TonyBagnall (approved):
great job

Issue #434, opened 2023-05-15
[ENH] Refactoring Piecewise Transformation aka PAA and SAX


hadifawaz1999:
### Describe the feature or idea you want to propose

PAA should:

- [x] Use numpy internally #269 
- [x] Be refactored to use numba -- No need its already vectorized see #489 
- [x] add an inverse function to approximate TS #489
- [ ] Should be put in collection, not collection/dictionary_based
- [x] Refactor tests to use numpy #199
- [x] Refactor tests to new implementation usage #489

SAX should:

- [x] Use numpy internally #417 
- [x] Be refactored to use numba #417 
- [x] Use scipy to calculate breakpoints automatically #417 
- [x] add an inverse function to approximate  #417 
- [x] Refactor tests to use numpy #417 
- [x]  Refactor tests to new implementation usage #417 

- Add everything to codeowners after collections/panel thing is sorted out

### Describe your proposed solution

Doing it myself, once SAX is done, PAA should be addressed. A time comparison should be done after all of this is done.

MatthewMiddlehurst:
Is this finished or are there other things planned?

hadifawaz1999:
yup its finished @MatthewMiddlehurst thanks for the reminder

PR #489, opened 2023-06-14
[ENH] Refactoring PAA

hadifawaz1999:
#### Reference Issues/PRs
Part of #434 

Refactors PAA to become faster

The only input parameter is the number of segments, or in other words, the target dimension.
Basically for the transformation can be done in:
- One line, parallelized over all samples and dimensions if and only if the length of the TS is divisible by the number of segments
- A for loop on the segments split indices, still parallelized, only in the case of non divisibility

The optimality and correctness of the PAA refactor

Refactored the PAA test

##### For all contributions

### DIFF
--- a/aeon/transformations/collection/dictionary_based/_paa.py
+++ b/aeon/transformations/collection/dictionary_based/_paa.py
@@ -4,19 +4,20 @@ import numpy as np
 
 from aeon.transformations.base import BaseTransformer
 
-__author__ = "MatthewMiddlehurst"
+__author__ = ["MatthewMiddlehurst", "hadifawaz1999"]
 
 
 class PAA(BaseTransformer):
     """Piecewise Aggregate Approximation Transformer (PAA).
 
     (PAA) Piecewise Aggregate Approximation Transformer, as described in [1]. For
-    each series reduce the dimensionality to num_intervals, where each value is the
+    each series reduce the dimensionality to n_segments, where each value is the
     mean of values in the interval.
 
     Parameters
     ----------
-    n_intervals   : int, dimension of the transformed data (default 8)
+    n_segments   : int, default = 8
+        Dimension of the transformed data
 
     Notes
     -----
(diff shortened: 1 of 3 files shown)


hadifawaz1999:
The implementation includes as well an `inverse_paa` function to produce the following, this is mostly for visual aid:

image

patrickzib:
It is a great improvement over the previous version. Well done. 

One issue in combination with SAX: where do we make sure that the input data is z-normalized when transformed with SAX? As of now, PAA does not apply z-normalization to the time series. Same for SAX, right?

patrickzib:
> @patrickzib I had the same question in my head :p is where to include it now, the simplest solution is to have `znormalized` parameter into both, this way if the user is using SAX or PAA, the data will always be znormalized. But wanted to see what you think before doing that.

Yes, maybe add a parameter to SAX: z-normalize, defaulted to True. When set to true, it applies z-norm to the input time series. This way, you do not have to modify PAA.

patrickzib:
The more sophisticated option would be to set the SAX breakpoints from the actual data. I.e. compute the mean and std from the time series and set it for the Gaussian distribution as input prior to computing the breakpoints.

hadifawaz1999:
> > @patrickzib I had the same question in my head :p is where to include it now, the simplest solution is to have `znormalized` parameter into both, this way if the user is using SAX or PAA, the data will always be znormalized. But wanted to see what you think before doing that.
> 
> Yes, maybe add a parameter to SAX: z-normalize, defaulted to True. When set to true, it applies z-norm to the input time series. This way, you do not have to modify PAA.

But if someone uses PAA only (no SAX involved), shouldn't the input time series also be znormalized ? I should check the original paper because not sure if its an obligation such as in SAX

hadifawaz1999:
> 

Thats actually a good idea, will make it an argument maybe ? to either manually choose them or automatically generate them from the data

patrickzib:
> But if someone uses PAA only (no SAX involved), shouldn't the input time series also be znormalized ? I should check the original paper because not sure if its an obligation such as in SAX

PAA should not depend on the normalization of the data. It was a trick in SAX to be able to have constant breakpoints.

hadifawaz1999:
> > But if someone uses PAA only (no SAX involved), shouldn't the input time series also be znormalized ? I should check the original paper because not sure if its an obligation such as in SAX
> 
> PAA should not depend on the normalization of the data. It was a trick in SAX to be able to have constant breakpoints.

Oh okay, thanks for the info, so yeah will make a SAX obligation only then.

hadifawaz1999:
I remember now, we did add it to SAX, here is the [line](https://github.com/aeon-toolkit/aeon/blob/254758ed57f05066af8df8027ff8050ae937fa91/aeon/transformations/collection/dictionary_based/_sax.py#L119)

patrickzib (approved):
LGTM

PR #497, opened 2023-06-16
[REF] change single class use case

TonyBagnall:

We have an internal mechanism in base classifier for the edge case when all the y values seen in fit are the same. The predict and predict_proba then just return that value or prob = 1 of that value.

this was done by a call to check_is_scitype, which involves multiple scans of the data. This simplifies it by using len, except for the edge case of using multi index DataFrame. 

Also removed a pointless duplication in tests and moved the overloaded operator to the bottom of the file and changed series_length to n_timepoints.

### DIFF
--- a/aeon/classification/base.py
+++ b/aeon/classification/base.py
@@ -91,9 +91,13 @@ class BaseClassifier(BaseEstimator, ABC):
         Parameters
         ----------
         X : 3D np.array (any number of channels, equal length series)
-                of shape [n_instances, n_channels, n_timepoints]
+                of shape (n_instances, n_channels, n_timepoints)
             or 2D np.array (univariate, equal length series)
-                of shape [n_instances, n_timepoints]
+                of shape (n_instances, n_timepoints)
+            or list of numpy arrays (any number of channels, unequal length series)
+                of shape [n_instances], 2D np.array n_channels, n_timepoints_i), where
+                n_timepoints_i is length of series i
+            other types are allowed and converted into one of the above.
         y : 1D np.array, of shape [n_instances] - class labels for fitting
             indices correspond to instance indices in X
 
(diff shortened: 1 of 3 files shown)


TonyBagnall:
> Changes to `setup.cfg` are included in the PR 🙂

*sigh*, fixed :)

aeon/classification/base.py
```diff
@@ -85,54 +85,20 @@ def __init__(self):
         super(BaseClassifier, self).__init__()
         _check_estimator_deps(self)
 
-    def __rmul__(self, other):
-        """Magic * method, return concatenated ClassifierPipeline, transformers on left.
-
```
MatthewMiddlehurst, line 90: I think it is neater to keep the `__X__` type methods at the top, even if we are not the biggest fans of the contents 🙂

aeon/classification/base.py
```diff
@@ -85,54 +85,20 @@ def __init__(self):
...
         X : 3D np.array (any number of channels, equal length series)
-                of shape [n_instances, n_channels, series_length]
+                of shape (n_instances, n_channels, n_timepoints)
             or 2D np.array (univariate, equal length series)
-                of shape [n_instances, series_length]
-        y : 1D np.array of int, of shape [n_instances] - class labels for fitting
+                of shape (n_instances, n_timepoints)
+            or list of numpy arrays (any number of channels, unequal length series)
+                of shape [n_instances], 2D np.array n_channels, n_timepoints_i), where
+                n_timepoints_i is length of series i
```
MatthewMiddlehurst, line 99: This looks like it's missing a ( in the middle

Same for other cases where this comment appears

extension_templates/classification.py
```diff
@@ -97,10 +97,13 @@ class MyTimeSeriesClassifier(BaseClassifier):
         "python_version": None,  # PEP 440 python version specifier to limit versions
     }
 
-    # todo: add any hyper-parameters and components to constructor
+    # todo: add any parameters to constructor
     def __init__(self, param_a, param_b="default", param_c=None):
         # estimators should precede parameters
         #  if estimators have default values, set None and initalize below
+        # Note that parameters passed and set in constructor should not be changed
+        # in order to comply withe scikit learn strucutre. Intead, copy into local
```
MatthewMiddlehurst, line 105: Typo

aeon/classification/base.py
TonyBagnall: I looked, it's not a convention that I could see, I prefer it out the way

TonyBagnall: I mean its not a python convention afaik, aeon conventions are inherited from the bad place. But if its a blocker, I'll move it back

aeon/classification/base.py
TonyBagnall: good spot

extension_templates/classification.py
MatthewMiddlehurst: "in order to comply withe scikit learn"

MatthewMiddlehurst (approved):

PR #371, opened 2023-05-03
[MNT] Add `non-deterministic` and `cant-pickle` tag for estimators

hadifawaz1999:
#### Reference Issues/PRs

Resolves #317 

#### What does this implement/fix? Explain your changes.

Adds `non-deterministic` and `cant-pickle` tag for estimators. For instance deep learners are non-deterministic, different runs produce different outputs. This is used to skip the `test_fit_idempotent`.

PR #505, opened 2023-06-22
[MNT] Replace `handles-missing-data` tag with `capability:missing_values`

MatthewMiddlehurst:
#### Reference Issues/PRs

Contains changes from #371, which should be merged first.

Removes the `handles-missing-data` tag in favour of `capability:missing_values`. Both of these currently duplicates of each other. Replaces all usage of the `handles-missing-data` with `capability:missing_values`.

Updates the estimator type section for some tags.

The removal of the `handles-missing-data` tag as a general idea and the changes in the `registry/_tags.py` file.

#### Any other comments?

The majority of usage for the `handles-missing-data` tag is in the forecasting module. I selected `capability:missing_values` as the tag to keep as most modules have "capability" tags and because of a comment in the `BaseTransformer` class selecting it as the preferred tag.


TonyBagnall (approved):
Its a monster! Agree its good to do though

PR #508, opened 2023-06-23
[ENH] Add some simple collection transformers

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #510 

Part of #499.

Adds the following collection transformers. 
- ACF
- Autoregression coefficients
- Periodogram
- Seven number summary stats


aeon/transformations/collection/ar_coefficient.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if order > n_timepoints - 1:
+            raise ValueError(
+                f"order ({order}) must be smaller than n_timepoints - 1 "
+                f"({n_timepoints - 1})."
+            )
+
+        Xt = np.zeros((n_instances, n_channels, order))
+        for i in range(n_instances):
+            for n in range(n_channels):
```
TonyBagnall, line 82: can we vectorise or njit these loops?

aeon/transformations/collection/periodogram.py
```diff
@@ -0,0 +1,112 @@
...
+                        0,
+                        int(
+                            math.pow(2, math.ceil(math.log(X.shape[2], 2))) - X.shape[2]
+                        ),
+                    ),
+                ),
+                **kwargs,
+            )
+
+        if self.use_pyfftw:
```
TonyBagnall, line 98: is there some benefit from pyfftw as opposed to numpy function?

aeon/transformations/collection/summary.py
```diff
@@ -0,0 +1,115 @@
...
+    Transforms a time series into seven basic summary statistics.
+
+    Parameters
+    ----------
+    summary_stats : ["default", "percentiles", "bowley", "tukey"], default="default"
+        The summary statistics to compute.
+        The options are as follows, with float denoting the percentile value extracted
+        from the series:
+            - "default": mean, std, min, max, 0.25, 0.5, 0.75
+            - "percentiles": 0.215, 0.887, 0.25, 0.5, 0.75, 0.9113, 0.9785
```
TonyBagnall, line 32: any reason for these values?

aeon/transformations/collection/summary.py
```diff
@@ -0,0 +1,115 @@
...
+            functions = [
+                row_mean,
+                row_std,
+                row_numba_min,
+                row_numba_max,
+                0.25,
+                0.5,
+                0.75,
+            ]
+        elif self.summary_stats == "percentiles":
```
TonyBagnall, line 69: maybe put this switch in a function? its a bit ugly here

TonyBagnall (approved):
some minor non blocking comments

aeon/transformations/collection/summary.py
MatthewMiddlehurst: default is our current default for the summary classifier (basic stats and quartiles)

others are from the most trusted of sources, [wikipedia](https://en.wikipedia.org/wiki/Seven-number_summary#cite_ref-1)

aeon/transformations/collection/periodogram.py
MatthewMiddlehurst: It's a wrapper for https://www.fftw.org/ which is all C, so probably faster? I have not benchmarked it though. 

It was used in the RSTSF code, so I included it here as well as an option.

aeon/transformations/collection/ar_coefficient.py
MatthewMiddlehurst: Not sure. We can't njit `burg`, so it would only be the looping really. Would have to find a way to write the first part of the function output efficiently while discarding the second part

aeon/transformations/collection/summary.py
MatthewMiddlehurst: Done

aeon/transformations/collection/periodogram.py
TonyBagnall: worth figuring out which is best, I'll raise an issue

TonyBagnall (approved):

PR #509, opened 2023-06-23
[ENH] Updates to `Catch22` and fixes to collection converters

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #510 

Part of #499.

Updates the Catch22 related classes, these are the major changes:
- Updates the parameters and documentation for `Catch22Classifier` (renames the file name so it's shorter)
- Implements a `Catch22Regessor` class using a `RandomForestRegressor` as a default estimator
- Removes the `Catch22Wrapper` class
- Implements the `pycatch22` package as a parameter to replace the above class. Updates the `_transform` method to accommodate this. Adds new parallel backend parameter to match the classification package.
- Adds `np-list` and unequal length support for all the above

Moves the 'expected results' generators for transformers currently in the classification package into a new file

During my unequal length testing I found that the `np-list` to `df-list` converter did not work correctly. `df-list` has cases as columns and timepoints as rows. This has been fixes and tests have been added.

While fixing the above I noticed an automatic converter from 3D numpy to 2D numpy which literally just concatenated all of the channels into a single series. This has been purged with fire.


TonyBagnall (approved):
lgtm

Issue #168, opened 2023-03-07
[DOC] Website redesign


MatthewMiddlehurst:
The website still currently uses the `sktime` documentation. Rather than just rename to `aeon`. IMO the current documentation is incredibly bloated, with a lot of information most users will not care about at a front.

As well as overhauling the look to differentiate ourselves from `sktime`, we should improve and reduce the content where necessary.

A first attempt on the front/index page is put together by myself in #122.

MatthewMiddlehurst:
Here is a checklist of pages which have been updated so far and ones that should be updated or removed.

- [x] Index
- [ ] Installation
- [x] Get Started
- [ ] User Guide
- [ ] API
- [ ] Examples
- [ ] Get Involved
- [ ] Development Guide
- [x] Contributors
- [x] About Us
- [ ] Glossary
- [x] Changelog
- [ ] Estimator Overview
- [ ] Related Software

PR #527, opened 2023-06-30
[DOC] Remove mentoring and estimator overview webpages

MatthewMiddlehurst:
Part of #168.

Removes two pages from the website. More of an argument to keep these than #525 and #526 though, IMO.

`get_involved/mentoring.md` currently a TBD page, we decided to keep it in #242, but I don't think we have any plans to accept mentees after 2.5 months. I think we can remove it for now and reintroduce when/if we have a bigger user base and/or plans to introduce projects for applicants. @aiwalter was interested, so pinging for comment if they want to keep. (https://www.aeon-toolkit.org/en/latest/get_involved/mentoring.html)

`estimator_overview.md` is a table of estimators which links to the API. I think it is rather redundant with the API, and automatically picks up items which are not estimators (i.e. performance metrics for forecasting). I don't know why someone would search this list for what they want rather than just go to the relevant API page i.e. https://www.aeon-toolkit.org/en/latest/api_reference/classification.html. (https://www.aeon-toolkit.org/en/latest/estimator_overview.html)


TonyBagnall (approved):
lots of redundant info to go

PR #534, opened 2023-07-04
[MNT] Set upper bound for `tslearn` to fix CI

MatthewMiddlehurst:
New version has a bug importing `torch` which breaks CI.


TonyBagnall (approved):

TonyBagnall:
we should let romain know

MatthewMiddlehurst:
https://github.com/tslearn-team/tslearn/issues/462

PR #535, opened 2023-07-04
[DOC] fixed path of image

hadifawaz1999:
Fixed an image path for the deep learning notebook.


TonyBagnall:
for the other folders, I have created an img directory in the sub dir and moved pictures there, so that img/H-Inception.png would work. Might be better to duplicate this structure for networks

hadifawaz1999:
> for the other folders, I have created an img directory in the sub dir and moved pictures there, so that img/H-Inception.png would work. Might be better to duplicate this structure for networks

Oh okay will do that then thanks ! do you use the H-Inception image in some notebooks in root ? if not i can move it instead of duplicate the image.

TonyBagnall:
> > for the other folders, I have created an img directory in the sub dir and moved pictures there, so that img/H-Inception.png would work. Might be better to duplicate this structure for networks
> 
> Oh okay will do that then thanks ! do you use the H-Inception image in some notebooks in root ? if not i can move it instead of duplicate the image.

move it rather than copy, dont think its used elsewhere, will check after

hadifawaz1999:
> > > for the other folders, I have created an img directory in the sub dir and moved pictures there, so that img/H-Inception.png would work. Might be better to duplicate this structure for networks
> > 
> > 
> > Oh okay will do that then thanks ! do you use the H-Inception image in some notebooks in root ? if not i can move it instead of duplicate the image.
> 
> move it rather than copy, dont think its used elsewhere, will check after

Should be good now, let me know. I checked and its not used in other notebooks.

hadifawaz1999:
Should i wait till we fix the kalman stuff and other packages that were updated @TonyBagnall @MatthewMiddlehurst ? because the fails are because of that.

TonyBagnall (approved):

Issue #507, opened 2023-06-23
[DOC] Examples notebooks


TonyBagnall:
### Describe the issue linked to the documentation

our docs

https://www.aeon-toolkit.org/en/latest/examples.html

link into the example notebooks. Restructure is to create a sub directory of notebooks for each module, then link them from the examples. This is happening in #506 

To do:
- [ ] Restructure forecasting
- [x] Remove redundant notebooks
- [ ] Make sure examples.md is correctly and completely linked in

Missing notebooks
- [x] Regression

Missing images
- [x] Early classification

I'll just update this list as I find missing things

PR #536, opened 2023-07-04
[DOC] Move forecasting notebooks

TonyBagnall:
#### Reference Issues/PRs
part of https://github.com/aeon-toolkit/aeon/issues/507

moves forecasting notebooks into their own dir, renames them to remove bizarre prefixes and adds an image for proabilistic. No content changed at all


MatthewMiddlehurst:
May be other concerns by @ltsaprounis and @aiwalter, but i think the only potential issue with this move is that I cant see any change to the usage of the `glocal_fc` folder and some of the images. Either they are unused or the path to use them has not been updated.

MatthewMiddlehurst:
How did Markus get pinged as a code owner 🤔

TonyBagnall:
> How did Markus get pinged as a code owner 🤔

codeowners I guess, will give it a tidy up soon

MatthewMiddlehurst (approved):
The only issue I can see is that the change broke the links in the user guide page (the datasets one seems broken also). Will approve but would be good to fix those even if the page is still to be updated.

TonyBagnall:
I'd not looked at the user guide at all, will update it now

PR #542, opened 2023-07-05
[MNT] Setting more upper bounds to fix CI

MatthewMiddlehurst:
An update to tensorflow has broken some of the deep learning tests, and numpy 1.24 has broken the kalman filter transformer.


TonyBagnall (approved):

TonyBagnall:
weird! Maybe is not quite formatted correctly, I'll take a look

PR #511, opened 2023-06-23
[ENH] Update interval extraction transformers and pipelines

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #509 and #508 

Part of #499.

#### What does this implement/fix? Explain your changes.

Updates the `RandomIntervals` and `SupervisedIntervals` transformers to add more configuration, documentaton and `_fit_transform`.

With this, updates the `RandomIntervalClassifier` class to use these new options, and adds `RandomIntervalRegressor` and `SupervisedIntervalClassifier` estimators.

PR #545, opened 2023-07-06
[ENH] Implement configurable interval base class with classifier and regressor

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #511

Part of #499.

Implements a configurable interval forest base class, with a classifier and regressor which use it. This is set up in a way that you could configure any of our current interval based approaches using the base class.

There is a lot going on here, if any additional explanation is required I can provide.


aeon/base/estimator/interval_based/base_interval_forest.py
```diff
@@ -0,0 +1,1030 @@
...
+        Number of estimators to build for the ensemble.
+    interval_selection_method : "random", "supervised" or "random-supervised",
+            default="random"
+        The interval selection transformer to use.
+            - "random" uses a RandomIntervalTransformer.
+            - "supervised" uses a SupervisedIntervalTransformer.
+            - "random-supervised" uses a SupervisedIntervalTransformer with
+                randomised elements.
+
+        Supervised methods can only be used for classification tasks, and require
```
TonyBagnall, line 49: not for regression?

aeon/base/estimator/interval_based/base_interval_forest.py
```diff
@@ -0,0 +1,1030 @@
...
+        Valid options are "loky", "multiprocessing", "threading" or a custom backend.
+        See the joblib Parallel documentation for more details.
+
+    Attributes
+    ----------
+    n_instances_ : int
+        The number of train cases.
+    n_channels_ : int
+        The number of channels per case.
+    series_length_ : int
```
TonyBagnall, line 151: just generally, trying to convert series_length to n_timepoints

TonyBagnall (approved):
this is a monster. I'm not sure about this design tbh, with the weird sidewise _fit overriding, but its better than before. It would be good to add a notebook examples/classification/interval_based on using it to configure a new classifier.

MatthewMiddlehurst:
In the classifier/regressor the base class is the leftmost when inheriting, so it should have priority when calling methods. https://www.programiz.com/python-programming/multiple-inheritance

Many of the interval classifiers follow the approach `transform series -> select intervals -> extract features -> build tree` with various forest-like qualities. It made sense just to have a single class that follows that formula, IMO.

aeon/base/estimator/interval_based/base_interval_forest.py
MatthewMiddlehurst: No not currently, would need to implement a new scoring method. Updated the docs on the actual classifier/regressor to better reflect the available methods.

TonyBagnall:
> In the classifier/regressor the base class is the leftmost when inheriting, so it should have priority when calling methods. https://www.programiz.com/python-programming/multiple-inheritance
> 
> Many of the interval classifiers follow the approach `transform series -> select intervals -> extract features -> build tree` with various forest-like qualities. It made sense just to have a single class that follows that formula, IMO.

TonyBagnall:
sorry, accidentally closed it!

PR #551, opened 2023-07-07
[ENH] Revisions to classifier base class

### LINKED ISSUE #51: [ENH] base classifier input checks

There are two checks I think should happen in the classifier base class
1. We have a clear distinction between classifiers and regressors. I think we should check that the input is not continuous in fit (and predict), since everything will behave strangely if a classifier is built on a regression problem. However, I'm not sure how best to do this, since input can be a Series or a numpy array of floats. How many unique values make it a regression problem? 
2. We should check in predict that the input X is the same dimensions as the X passed in fit. This should only be allowed with early classification

TonyBagnall:
Fixes #51 part of #42 

this can be advanced now no classifiers now override fit

1. Make fit/predict/predict_proba final
2. Add a test to check subclasses dont overrride fit/predict/predict_proba (since, python)
3. Remove the tests test_multivariate_input_exception and test_handles_single_class
since these are handled in the base class and tested in base class tests
4. Fixes https://github.com/aeon-toolkit/aeon/issues/51 by checking y input type
5. Adds a test to test_all_classifiers to make sure tags capability:unequal_length and x_internal_type are consistent
6. Restructure BaseClassifier fit and predict to this type of thing (see code for exact), using the consolidated conversion in aeon.utils.validation.collections.

```python
fit(X, y):
  metadata = checkX(X)  # checks X a valid input type and against the capabilities against the data
  X = convertX(X)   # converts X to the tag x_inner_type
  y = check_y(y)  # checks y
#Other stuff
 self._fit(x,y)  
```
note that early classification needs looking at, because currently it makes a static call to BaseClassifier object methods such as fit, which causes errors if object attributes are altered in object methods. It does not extend BaseClassifier for valid reasons, so I have had to clone the checking methods which use data stored in the object. I think at some point we should redesign base early classifier

this is largely cosmetic tbh, but I think its much clearer and will make it easier to insert a collection class between base estimator and the collection base classes


hadifawaz1999:
Do deep learners follow these new adaptations ?

TonyBagnall:
> Do deep learners follow these new adaptations ?

deep learning classifiers will simply inherit all this, so no need to change them at all. Similar for regressors and clusterers, for the next stage

aeon/classification/base.py
```diff
@@ -389,168 +379,156 @@ def _predict_proba(self, X) -> np.ndarray:
 
         return dists
 
-    def _check_convert_X_for_predict(self, X):
-        """Input checks, capability checks, repeated in all predict/score methods.
+    def checkX(self, X):
```
MatthewMiddlehurst, line 382: I don't think this rename matches the naming conventions elsewhere. `_check_X` or `check_X` seems fine?

Making this method public means it will appear in the documentation for every classifier. If we want it to be public but not do that, maybe it can be a function outside the class. Not pushing for either public or private, just a warning.

aeon/classification/base.py
```diff
@@ -389,168 +379,156 @@ def _predict_proba(self, X) -> np.ndarray:
...
-        -------
-        metadata : dict with metadata for X
-
-        Raises
-        ------
-        ValueError
-            If y or X is invalid input data type, or there is not enough data
-        """
-        # Check X is valid input type and recover the data characteristics
-        X_valid, _, X_metadata = check_is_scitype(
```
MatthewMiddlehurst, line 507: I can understand the desire to remove these bits, but I can't see where we actually check whether the input datatype is correct (i.e. 3d or 2d) without them. It does not seem to be in checkX at least.

aeon/classification/base.py
```diff
@@ -389,168 +379,156 @@ def _predict_proba(self, X) -> np.ndarray:
...
+                f"but {type(self).__name__} cannot handle {problems_or}. "
             )
+            raise ValueError(msg)
+        return metadata
 
-            if self.is_composite():
-                warn(msg)
-            else:
-                raise ValueError(msg)
+    def convertX(self, X):
```
MatthewMiddlehurst, line 439: Same as checkX comment

aeon/classification/base.py
MatthewMiddlehurst: I may be wrong, let me know if I am

aeon/classification/tests/test_all_classifiers.py
```diff
@@ -141,6 +96,7 @@ def test_classifier_on_unit_test_data(self, estimator_class):
         # train classifier and predict probas
         estimator_instance.fit(X_train, y_train)
         y_proba = estimator_instance.predict_proba(X_test[indices])
+        #
```
MatthewMiddlehurst, line 99: stray comment

aeon/classification/tests/test_all_classifiers.py
```diff
@@ -271,3 +227,26 @@ def test_classifier_train_estimate(self, estimator_class):
         else:
             # skip test if it can't produce an estimate
             return None
+
+    def test_classifier_tags_consistent(self, estimator_class):
+        """Test that the tag x_inner_type is consistent with capability:multivariate."""
```
MatthewMiddlehurst, line 232: Wrong capability name here i think

aeon/utils/_testing/scenarios.py
```diff
@@ -269,7 +269,6 @@ def _check_dict_of_dict(obj, name="obj"):
         and np.all([isinstance(x, str) for x in obj.keys()])
     ):
         raise TypeError(
-            f"Argument {name} with must be a dict of dict, with str keys, "
-            f"found {type(obj)}"
+            f"{obj} must be a dict of dict, with str keys, " f"found {type(obj)}"
```
MatthewMiddlehurst, line 272: Think this is leftover from another PR?

MatthewMiddlehurst (commented):
Mostly small things. Would it be possible to summarise the changes to test_base.py? The GitHub diff does not help very much.

aeon/classification/base.py
TonyBagnall: if a completely invalid type (e.g. int) is passed as X, it is picked up in checkX

``python
import numpy as np


X = np.random.random(size=(10,1,200))
X2 = np.random.random(size=(10,1,200))
X3 = 55

knn = KNeighborsTimeSeriesClassifier()
print(knn.checkX(X))
print(knn.checkX(X2))
print(knn.checkX(X3))
```
If a valid type is passed (e.g. numpy2D, or 3D), if is dealt with in convertX

```python
print(knn.convertX(X).shape)
print(knn.convertX(X2).shape)
```

MatthewMiddlehurst: My concern is more with wonky dataframes, lists of lists and 1d arrays etc.

This may be tested in the new tests, i have not checked.

TonyBagnall:
@MatthewMiddlehurst I think I have dealt with the comments and changes. Re the docstrings, they need some work in formatting but will wait until after #650 is in. Re test changes
test_all_classifiers:
test_handles_single_class removed: this is now tested in the base class only, not for all classifier
test_classifier_output: no longer runs the whole check is sytype, which was used purely to get number of cases
test_classifier_tags_consistent: added to make sure data tags do not contradict eachother. So if capability:unequal length is true, then the inner_type must contain a data structure able to handle unequal length

test_base
switched from DummyClassifier to _TestClassifier to avoid confusion with concrete DummyClassifier
removed capability testing, replaced with testing for checkX and convertX (these are moving in #650 so will revise this then)
added testing for previously uncovered or new methods

aeon/classification/base.py
TonyBagnall: Guess it can be _check_X and _convert_X, but I generally dont like mixing camel and snake

MatthewMiddlehurst: It's currently camel and we usually use snake, no?

aeon/classification/tests/test_base.py
```diff
@@ -53,183 +46,236 @@ def _fit(self, X, y):
...
+    y = np.array([[0, 0], [1, 1], [1, 1]])
+    with pytest.raises(TypeError, match=m5):
+        dummy.fit(X, y)
+    # Continuous y
+    y = np.random.random(5)
+    with pytest.raises(ValueError, match=m6):
+        dummy.fit(X, y)
+
+
+def test_checkX():
```
MatthewMiddlehurst, line 125: My remaining concern is that this test/method does not check a broad set of invalid inputs. Ideally it should check input types valid elsewhere in the package to reject them (1d arrays etc.), incorrect input types (i.e. dimensionality > 3) and maybe some common types we do not support (i.e. 3 tier nested list, though probably easier to just reject anything that is not accepted rather than go for specifics.).

I think we already have most of this already, but it may be lost in the transition. This can be part of another PR if this is getting a bit long.

Also the test names were missed in the rename 🙂

MatthewMiddlehurst:
I ran through some of the edge cases locally on the new base class, and I'm happy that they are being covered sufficiently and tested elsewhere. Fine to approve when tests pass (seems like EC stuff missed the rename).

MatthewMiddlehurst (approved):

PR #553, opened 2023-07-07
[MNT] Add periodic test workflow and remove Makefile

MatthewMiddlehurst:
#### Reference Issues/PRs

Part of my suggestion in #59

Implements a set of tests which run every day at 01:30 UTC. These do not use the "matrixdesign" option and include an extra test without numba compilation for code coverage.

Updates the code coverage files, and prevents other tests which use numba from writing coverage files.

To prevent the new tests from using matrix design and coverage, renovates the `setup.cfg` options nad moves the relevant matrix and coverage calls to the test yml.

Removes Makefile, which currently contains options which each run pytest in different ways. The original "test" option creates a test directory and copies files, but at some point changing directory to it was removed. I'm pretty sure these work just as well in the pytest yml files.

The new periodic tests yml file.
The test configuration change.
The removal of Makefile.

@lmmentel it would be good if you could comment if you get some time. I can probably break this up a bit if there too much going on, but some of the config changes are required to stop the new tests skipping stuff like the PR CI does currently.


TonyBagnall:
Any idea how longthe full tests will take?

lmmentel:
I looked though the change but haven't had enough time for a deep dive. I think the direction is good, but would probably try to reduce complexity in the nightly tests config even further. We are probably no there yet but maybe we could articulate a few design principles for the different types of testing regimes. For example:

## PR test
- ensure minimal compatibility with the code base while prioritizing speed of execution.
- Run essential tests with gates ensuring CI fails as soon as possible. 

## Nightly test
- ensure complete coverage taking the time to run though all edge cases.
- Run full suite of tests against all supported platforms with a `fail-fast: false` flag set up to make sure all tests are executed and we have a complete log on what's currently failing.

A few more practical ideas:
- From what I saw the nighly CI could be simplified a bit by dropping `nosoftdeps` tests/gate since we either way want to run everything we have.
- In PR test we could probably pull all platforms under `matrix` strategy since only invoking the makefile was stopping the same invocation (as far as I remember).

If this is not time sensitive I'm happy to have a deeper look next weekend. Hope this is somewhat helpful.

MatthewMiddlehurst:
> Any idea how longthe full tests will take?

No, we would have to see them run. Longer than the current PR ones for sure as there is no splitting based on Python version/OS.

There is a concern this would hit the max time limit, but even if it does there is no damage being done to the current CI workflow other than potentially code coverage reporting.

MatthewMiddlehurst:
@lmmentel Yes that is what I am going for really. Thanks for taking a look, no rush on this so it can wait. 

I think when we have the periodic test up and people are happy with it, it will be a lot easier to start considering ways to cut back on the PR side of testing. 

In the issue I mentioned `pytest-testmon` as probably the easiest to implement, but that has some drawbacks and is likely to miss stuff.

> From what I saw the nighly CI could be simplified a bit by dropping nosoftdeps tests/gate since we either way want to run everything we have.

I would have to delve into what exactly is being run in these tests a bit more really, but for now I think its fine to keep the testing with minimal dependencies.

The gate preventing other tests from being run until it completes can go for sure though, good catch.

> In PR test we could probably pull all platforms under matrix strategy since only invoking the makefile was stopping the same invocation (as far as I remember).

I think this would be better as another PR after we let the new workflow run a couple of times. I want to make this change and I don't think it would fail (it hasn't in my other tests), but better safe than sorry.

TonyBagnall:
I'm fine with this as an incremental improvement for more testing, but does not remove (or increases) the need to cut down testing permutations, esp in forecasting, see #151 #437

lmmentel:
@MatthewMiddlehurst after reading your comments I started thinking that probably the safest approach would be to do make a first pass simply adding the nightly tests without modifying the existing CI. Let it run for a few day for us to build the confidence it does what's intended and then refactor the PR test removing the Makefile. That being said I see the modified pipeline run OK so maybe I'm being too conservative. Happy to approve as is if you want to take it for a spin over the next few days and then I can have look during the weekend if we didn't miss anything.

lmmentel (approved):

MatthewMiddlehurst:
There should now be minimal changes to the original tests, with the exception of explicitly calling --matrixdesign and removing coverage collection.

The first needs to be removed from the general config to stop the nightly tests from usingit, and the coverage does not really work as intended with the current set up either because of numba.

lmmentel (approved):

Issue #537, opened 2023-07-05
[ENH] numpy array vs pd.DataFrame storage for a single series with `n_channels` and `n_timepoints`


TonyBagnall:
### Describe the feature or idea you want to propose

there is a fundamental difference in how a series is represented in a 2D numpy and a pd.DataFrame. For 2D numpy, we store it in shape `(n_channels,  n_timepoints)`, and in pd.DataFrame we store it in shape `( n_timepoints, n_channels)`.

For example
```python
import numpy as np
from aeon.datatypes import convert
# 10 multivariate time series with 3 channels of length 100 in "numpy3D" format
multi = np.random.random(size=(10,3,100))
np_list = convert(multi, from_type="numpy3D", to_type="np-list")
print(f" Type = {type(np_list)}, type first {type(np_list[0])} shape first "
      f"{np_list[0].shape}")
df_list = convert(multi, from_type="numpy3D", to_type="df-list")
print(f" Type = {type(df_list)}, type first {type(df_list[0])} shape first "
      f"{df_list[0].shape}")
```

the reason for this difference is historic and fundamental. It would be weird to store data in `(n_channels,  n_timepoints)` in a data frame, and all our classification algorithms are designed to work efficiently with `(n_channels,  n_timepoints)`.

### Describe your proposed solution

I think we just keep this distinction but we need to make it crystal clear and do checks. The reason this arises for machine learning side is when dealing with time stamped data. We would naturally store that in a DataFrame, but then need converters to numpy.

### Describe alternatives you've considered, if relevant

I think the only viable alternative is to convert the whole machine learning side to `( n_timepoints, n_channels)`. This is in fact the format used in keras and tslearn. However, it would require a lot of internal reshaping, since many algorithms iterate over channels acting on each series. We could do this with an axis tag, i.e. specify for an estimator how to view the data, and reshaping numpy is very efficient. It just seems like too much change for two little reward to me though. Changing to a front facing  `( n_cases, n_timepoints, n_channels)` is a seismic change.

TonyBagnall:
so I will close this general issue, since we now have BaseSeriesEstimator with an axis to specify the shape. Specific axis issues raise different issues

PR #555, opened 2023-07-08
[MNT] Add test for transformer numpy inpuy

MatthewMiddlehurst:
#### Reference Issues/PRs

Related to #537 (2d numpy is also m,d rather than d,m)

Implements a test to ensure that the `BaseTransformer` broadcasting for numpy array input is the same for 1d, 2d (transposed) and 3d inputs.

There is a potential issue when broadcasting 3d numpy arrays, as they are stored using (n,d,m) as a format while 2d numpy arrays use (m,d). Broadcasting the 3d array would build a transformer on individual series in some cases, which introduces the possibility to introduce problems with this difference. While it does not appear to be an issue currently, I would like that to remain the case.

The implemented test, and whether there are any other similar scenarios that could also use a test.

### DIFF
--- a/aeon/transformations/tests/test_base.py
+++ b/aeon/transformations/tests/test_base.py
@@ -17,10 +17,12 @@ from inspect import isclass
 
 import numpy as np
 import pandas as pd
+from numpy.testing import assert_array_equal
 
 from aeon.datatypes import check_is_scitype, get_examples, mtype_to_scitype
 from aeon.transformations.base import BaseTransformer
 from aeon.transformations.compose import FitInTransform
+from aeon.transformations.series.boxcox import BoxCoxTransformer
 from aeon.utils._testing.scenarios_transformers import (
     TransformerFitTransformHierarchicalMultivariate,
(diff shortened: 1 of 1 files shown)


aeon/transformations/tests/test_base.py
```diff
@@ -593,3 +595,25 @@ def test_vectorize_reconstruct_unique_columns():
     t = Detrender.create_test_instance()
     Xt = t.fit_transform(X)
     assert set(Xt.columns) == set([0, 1])
+
+
+def test_numpy_format_outputs():
+    """Test that all numpy formats return the same output when converted."""
+    X = np.random.random(size=(2, 1, 8))
+    bc = BoxCoxTransformer()
```
TonyBagnall, line 603: can we generalise this rather than just test box cox? Should it not be in the full transformer test suite?

MatthewMiddlehurst: this is all base class conversion, since we are comparing output using a dummy probably is not helpful

TonyBagnall (approved):

PR #561, opened 2023-07-11
[DOC] Utils notebook directory

TonyBagnall:
setting up a directory for utils, with the mlflow example. Be good to get more examples in here, there are a fair few files in utils such as plotting, profiling, sampling, seasonality, slope_and_trend and stats. Also datetime, sklearn, multiindex and stats files, which could probably be rationalised, but not in this PR.


TonyBagnall:
> The notebook viewer does not like this probably because of the move, has anything in it changed?

simply refactored, they look the same to me?

MatthewMiddlehurst (approved):
There is quite a large difference in lines removed and added, but if you made no changes may just be output?

TonyBagnall:
> There is quite a large difference in lines removed and added, but if you made no changes may just be output?

pycharm does weird things to notebooks. I corrected a couple of typos and think it changed formatting.

PR #562, opened 2023-07-11
[MNT] Exclude results loading notebook

MatthewMiddlehurst:
The notebook loads data from an external source, so we should not run it in our CI. It is currently failing, but this is for multiple PRs so it may require looking at if it's not an external issue @TonyBagnall.


TonyBagnall:
resolved now, but agree it should be excluded

PR #569, opened 2023-07-14
[MNT] params forecasting

TonyBagnall:
changes test parameters for two forecasters


TonyBagnall:
will come back to the slow squared residuals tests another time

Issue #570, opened 2023-07-14
[ENH] Load bake off results from tsc.com


TonyBagnall:
### Describe the feature or idea you want to propose

as part of the benchmarking revamp, we want to be able to load results directly from tsc.com for recently published bake offs. 

[ ] Univariate classification for 112 standard with 32 classifiers and 30 new data with eight classifiers, results in 
https://arxiv.org/abs/2304.13029
[ ] Univariate clustering for kmeans and kmedians with elastic distances on 112 standard, results presented in
https://arxiv.org/abs/2205.15181
[ ] Regression results for maybe 12 regressors on 61 existing and new TSER datasets presented in 
https://arxiv.org/abs/2305.01429
(all three papers are under review)
also need to sort out
[ ] Multivariate classification on 26+ 30 odd new ones, recreation of 
https://link.springer.com/article/10.1007/s10618-020-00727-3
plus a bunch unpublished as yet, need to revisit MTSC

### Describe your proposed solution

need a full list of the algorithms available, aliased names and offline testing.

TonyBagnall:
Done this :)

PR #571, opened 2023-07-14
[ENH] Full classifier results

TonyBagnall:
#### Reference Issues/PRs
part of #570 

adds a full list of 32 classifiers evaluated in bakeoff redux. Some were not generated with aeon (used java based tsml on some) and we do not guarantee the results are 100% deterministic with current code base, not quite got to that level of organisation, but certainly not significantly different.

No

Not sure whether the results loader should offer the option of returning results for all resamples rather than average.

Added a test which is skipped, since we do not want it running in CI.


MatthewMiddlehurst:
looks good, but edits test config

TonyBagnall:
> looks good, but edits test config

sigh. fixed

MatthewMiddlehurst (approved):

Issue #40, opened 2023-02-08
[ENH] Redesign of time series forest classifier/regressor


TonyBagnall:
We currently have duplicate versions of the same algorithm TSF, and for some reason I have never understood it plays a very prominent role in examples etc. I think we should simplify to a single version and not use it so much, it is a very obscure algorithm not used at all in practice.

PR #575, opened 2023-07-16
[ENH] Replace TSF estimators with base class version and remove `series_as_features` package

MatthewMiddlehurst:
Replace the 2 TSF classifier algorithms and 1 TSF regressor algorithms with the new interval base class versions.

This involves deleting the `series_as_features` package, which currently only contains old TSF base classes and the `_split.py` file (this could be salvaged, but I have never used it or seen it be used).

Includes docstring improvements for other interval classes.


TonyBagnall (approved):
Closes #40

PR #580, opened 2023-07-20
[ENH] Removes the delegate structure from regression

### LINKED ISSUE #578: [ENH] Remove usages of the delegate model from regression

Remove the use of delegates in regression. An (IMO) over complex delegate design was introduced in the other place before the split. We have removed it in Classification, reverting to a simpler (and as efficient) solution with fewer lines of code. It is still used in Regression but only used in the RocketRegressor

revert RocketRegressor to just be a pipeline

Could leave it in if anyone wants to support this design choice

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs
Fixes #578 

Removes the delegate regressor structure and reverts is sole use with RocketRegressor to match the structure used by RocketClassifier

Existing tests should pass, logically identical to delegate version,


MatthewMiddlehurst:
I think we should just align with classification for now, so change ROCKET and leave the base class for if anyone wants to use the structure.

TonyBagnall:
> I think we should just align with classification for now, so change ROCKET and leave the base class for if anyone wants to use the structure.

ok, although no one is going to use it

MatthewMiddlehurst (approved):

PR #582, opened 2023-07-20
[ENH] Adds CD diagrams, violin plots, boxplots and some other graphs. 

dguijo:
This PR adds new diagrams. Issues #451 (open) and #116 (closed). PR #452 (merged).

- CDDs with Wilcoxon + Holm correction. Several features such as highlighting in colours or adding aliases are included. 
- Violin plots.
- Scatter plots.
- Different types of boxplots.


TonyBagnall:
@dguijo looking great, could you post some screenshot examples here?

dguijo:
> @dguijo looking great, could you post some screenshot examples here?

Thank you! Sure, here are some examples of the different plots that can be made:


**aeon.benchmarking.results_plotting.plot_scatter**

The first one is a scatter plot comparing the results achieved by two different classifiers. In this plot, it can be observed that the HC2 is compared to the InceptionT approach. The first one wins in 63 datasets, losses in 32 and ties in 16. The opposite is obtained for the second one (32 wins, 16 ties and 63 losses)
Wins obtained by the method shown in the y-axis, HC2 (or losses of the x-axis method, InceptionT) are in blue color, whereas green is used for losses (or wins of the x-axis method) and orange for ties.

image


**aeon.benchmarking.results_plotting.plot_scatter_predictions**

The second one is also a scatter plot. This plot will generally be used for TSER approaches as the output is continuous. In the case of TSC approaches, the output would be a confusion matrix. This time, the scatter shows the predictions made by a specific regressor over a dataset. In the example shown below, the method is the DrCIF and the dataset is the HouseholdPowerConsumption1. Predicted values are shown in the y-axis, whereas actual values are shown in the x-axis. 

image


**aeon.benchmarking.results_plotting.plot_boxplot_median**

The third one plots the distribution of the deviations from the median. Even though the goal is the same, different types of boxplots could be generated (violin, boxplots or swarm). Below is the example for the violin using the last results of the bakeoff redux for the main TSC methods.

image

If the swarm type is chosen, the plot would be the following:

image


**aeon.benchmarking.results_plotting.plot_critical_difference**

The last type of plots are the Critical Differences diagrams. This type of plots were already in aeon, though I've reworked it to include the Wilcoxon with Holm posthoc test, which is the one used in the bakeoffs of TSC and TSER. Here is one example using the last results of the bakeoff redux for the main TSC methods. As can be seen, aliases and colours can be given to the different approaches:

image

Cliques can also be formed using nemenyi instead of Wilcoxon+Holm, in case of any interest. 


Finally, I included a placeholder for the Multi-comparison Matrix (**aeon.benchmarking.results_plotting.plot_multi_comparison_matrix**) developed by @hadifawaz1999. He is going to put the code in this package when this PR is in.

TonyBagnall (approved):
LGTM

PR #585, opened 2023-07-20
[ENH] Updates the STSF classifier to use new interval base class

MatthewMiddlehurst:
Replace the STSF classifier with the new interval base class version. There is no regression version of the supervised extraction process currently, so no regressor.


aeon/classification/interval_based/_tsf.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# copyright: aeon developers, BSD-3-Clause License (see LICENSE file)
```
TonyBagnall, line 2: I thought we were not putting this line in any more?

TonyBagnall (approved):

aeon/classification/interval_based/_tsf.py
MatthewMiddlehurst: Are we not? I don't really remember if we walked about it.

PR #593, opened 2023-07-23
[MNT] DL OOM test failure fix and remove Makefile

MatthewMiddlehurst:
There are currently testing failures in the periodic tests with various crashes and memory related errors.

This PR adds `gc.collect()` to the end of fit for all deep learning. Running this on the full suite of tests with no matrix testing (same as the periodic tests) results in no crashes after 3 runs.

As part of the testing I had to redo some of the initial changes from #553 to ensure the tests workflow process was similar, I propose keeping these changes and removing the Makefile. While there have been failures in the periodic tests, after this PR those are legitimate and unrelated to the workflow itself.


TonyBagnall (approved):
after this is in, we should at some point see if we could move the garbage collection call to the base class, even if this involves extending fit. But this should go in I think, to progress the testing changes

hadifawaz1999 (approved):
LGTM! Thanks for fixing this, I agree with Tony, will put up an issue on making it into the base class at some point.

PR #576, opened 2023-07-17
[ENH] centralise and tidy up the converters

TonyBagnall:
This PR has been split into two and is part of a sequence of four. 

1. https://github.com/aeon-toolkit/aeon/pull/606 implements utility functions to test data characteristics and perform conversions, all still built on datatypes. Basically duplicates some functionality currently done by multiple calls to checkisscitype
2. This PR pulls all the conversions for collections into _convert_collections, and it also provides the missing conversions (listed in NOT_IMPLEMENTED_CONVERSIONS)

Note  these are not yet called, it is a set up before revising classifier base then adding a new base class. this PR pulls the converters used by the collections estimators into one place, simplifies things in relation to datatypes and provides comprehensive conversion. Note that conversions of some of the more unusual types are not done as efficiently as they could be, since I have just combined two other conversions. Given we allow seven different data structures and some of them are never used, it seemed reasonable to me, I can create an issue to reimplement after this is in.

Next steps are:

3. Tidying up classification base class to use these 
https://github.com/aeon-toolkit/aeon/pull/551

4. generalise to a common base class inbetween BaseEstimator and BaseClassifier/Regressor/Clusterer/CollectionsTransformer to share the same conversion normalisation. 

#### Reference Issues/PRs

part of #42 see #557

## Did you add any tests for the change?

yes I think I covered it all with tests

#### Any other comments?

Wasnt sure where to put these

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.
- [ ] For new estimators, I've added the estimator to the online documentation.

PR #594, opened 2023-07-24
[BUG] Scenarios string fix

TonyBagnall:
from the other place, there is an unnecessary conversion of complex dictionary to string which slows down test set up by >60 secs (2 minutes on my desktop)

https://github.com/sktime/sktime/issues/4900


TonyBagnall:
> I prefer the solution here https://github.com/aeon-toolkit/aeon/pull/576/files#diff-308afa8181ee011a686a7b578ef3f76aa941d794f768d8d599cac8382b147f08
> 
> Keeps the informative message without loading it every time. I think you tested and it works as well?

I dont think it matters,  but change it if you want, they both have the same effect.

TonyBagnall:
> I prefer the solution here https://github.com/aeon-toolkit/aeon/pull/576/files#diff-308afa8181ee011a686a7b578ef3f76aa941d794f768d8d599cac8382b147f08
> 
> Keeps the informative message without loading it every time. I think you tested and it works as well?

this solution still prints the whole data structure to a string unnecessarily. Hang on, will alter it now

MatthewMiddlehurst (approved):

PR #606, opened 2023-07-27
[DOCS] pre-processing notebook and collections validations

TonyBagnall:
1. correct some things on classification notebook
2. Adds a first go at a preprocessing notebook to show how to deal with missing values and unequal length. More to come of this at a later date
3. Adds a util/validation file called collection.py to contain simple ways of checking for unequal and missing (related to #576  with associated tests
4. Fixes minor bug on distances notebook 
5. Removes big picture from front of docs, changes first sentence (from slack)

step (3) was necesssary to write the notebook, but morphed into something a bit more, sorry. The file collection.py and the full test suite associated it is a precursor to #551 and #576 

after these two PRs, I'll introduce a new base class between BaseEstimator and BaseClassifier so these changes can be easily extended to BaseRegressor, BaseClusterer and BaseCollectionTransformer.


aeon/utils/validation/collection.py
```diff
@@ -0,0 +1,359 @@
...
+            return "nested_univ"
+        if isinstance(X.index, pd.MultiIndex):
+            return "pd-multiindex"
+        elif _is_pd_wide(X):
+            return "pd-wide"
+        raise ValueError(
+            "ERROR unknown pd.DataFrame, contains non float values, "
+            "not hierarchical nor is it nested pd.Series"
+        )
+    #    if isinstance(X, dask.dataframe.core.DataFrame):
```
chrisholder, line 185: there is a dask dataframe?

chrisholder (approved):

aeon/utils/validation/collection.py
TonyBagnall: there is, but its not supported at all, so will only add back if there is some demand once the rest is fixed

PR #608, opened 2023-07-28
[MNT] Explicitly include and exclude items from the build MANIFEST.in

MatthewMiddlehurst:
Tidies up `MANIFEST.in` to list whether items are included or excluded from the build rather than being inferred in some cases. Also removes the `include` for items which no long exist.

The current ignore settings for `check-manifest` don't actually change whether file are included or not, it just stops them from appearing in the pre-commit check. Rather than just ignoring them in the check, they should be explicitly included/excluded.

The difference in the actual contents of the project build after this are as follows:
- The benchmarking examples .csv files are included in the build (currently causing the release to fail)
- The `build_tools` folder is no longer included
- Some of the top level markdown files i.e. GOVERNANCE.md and CONTRIBUTORS.md are no longer included in the build


TonyBagnall (approved):

MatthewMiddlehurst:
Only impacts release, no testing required.

PR #619, opened 2023-07-31
[MNT] Add runtime list for notebook test

MatthewMiddlehurst:
Records the runtime of notebooks in the notebook test workflow and prints it out at the end. i.e. 
`205	examples/networks/deep_learning.ipynb` 
which shows runtime (seconds) and then the notebook which was run.

Should help with #59, as people should be able to tell how long each notebook is taking now.


MatthewMiddlehurst:
this is what the output looks like (not the full list)

image

TonyBagnall (approved):
really useful!

TonyBagnall:
dont think you wanted the print?

image

MatthewMiddlehurst:
it outputs the actual command being run as well for other stuff, not sure that can be prevented

PR #636, opened 2023-08-03
[MNT] Skip tests for PRs which do not touch any important or `aeon/` files

MatthewMiddlehurst:
This PR adds new `paths:` configuration to the PR testing workflow. This means that PRs which do not change any files listed within `paths:` will not be run.

I have split the current workflows into three different files, the test steps themselves are the same.

The pytest tests will only run if one of the following is changed:

- `aeon/**` (any package files)
- `.github/workflows/**` (any test workflows)
- `pyproject.toml` (important config and dependencies)

The notebook examples test will be run with the above, but also

- `examples/**` (any examples file)

The pre-commit checks (code-quality) will always run still.

The periodic/nightly tests remain the same and will always run everything.


MatthewMiddlehurst:
i.e. for a PR not touching any of the listed files:
image

TonyBagnall (approved):

PR #643, opened 2023-08-04
[MNT] Collection transformation `summarize` refactor and removal

MatthewMiddlehurst:
Either removes or moves the contents of the collections transformer `summarize` directory. 

- `DerivativeSlopeTransformer` is a simple function wrapper, moves the function to the numba utils file and replaces the transformer with a `FunctionTransformer`
- `RandomIntervalFeatureExtractor` is a duplicate transformer for extracting random interval summary stats. It was used in the old version of TSF, but this PR just removes it as it is no longer used in the codebase.
- `FittedParamExtractor` and `PlateauFinder` are not used in any ML package classifiers currently, and I don't think they have to use the new base class, so they have just been moved.


TonyBagnall:
@MatthewMiddlehurst would it be possible to fix #644 in a separate PR? There is a lot of other stuff going on here and dont have a huge amount of time atm, would prefer #644 fixed quickly

TonyBagnall (approved):
lgtm

PR #652, opened 2023-08-10
[ENH] Test what happens when we include excluded estimator: 1. MiniRocket

TonyBagnall:
Many of these exclusions go back years. This PR is to remove exclusions and raise new issues if they still fail tests, starting with the six completely excluded estimators

*Minirocket*
1. passes
2. passes
3. passes: good enough for me!
4. passes


TonyBagnall:
@hadifawaz1999 if you get a chance to sort out tapnet, that would be brilliant

TonyBagnall:
passed 5 or 6 times, see #661

PR #597, opened 2023-07-24
[ENH] Testing revision version 1

TonyBagnall:
#### Reference Issues/PRs

Dependent on: #595 and #437 

See also:
https://github.com/aeon-toolkit/aeon/issues/418
https://github.com/aeon-toolkit/aeon/issues/59
https://github.com/aeon-toolkit/aeon/issues/151
https://github.com/aeon-toolkit/aeon/issues/154

#### What does this implement/fix? Explain your changes.

This is the version of the new testing model I am proposing. Basically everything switches on MATRIXDESIGN. By day, this boolean is set to True, dramatically reducing CI time. By night, it becomes a monster, and runs full tests on everything, everywhere, all the time.

PR #654, opened 2023-08-11
[ENH] Test what happens when we include excluded estimator: 2. MatrixProfileTransformer

TonyBagnall:
see #347 

MatrixProfileTransformer may well have problems, its  old and unloved. This is to diagnose them. Note I have included the changes in #597 for my sanity.

1. Passes
2. Passes
3. Passes
4. Passes


TonyBagnall:
five passes enough for me, see #661

PR #655, opened 2023-08-11
[ENH] Test what happens when we include excluded estimator: 3. ColumnTransformer

TonyBagnall:

ColumnTransformer has the comment     # Interface is outdated, needs a rework.

docs state "Simply takes the column transformer from sklearn  and adds capability to handle pandas dataframe."
it uses         "X_inner_mtype": "nested_univ",  

Its not used anywhere. Maybe we can remove this and rework it as a ChannelTransformer that independenty transforms channels?

Anyway, lets see what it fails first


TonyBagnall:
well that was quick

FAILED aeon/tests/test_softdeps.py::test_est_fit_without_modulenotfound[ColumnTransformer] - ValueError: X must be a pd.DataFrame or a np.array, but found: <class 'pandas.core.series.Series'>

its very outdated, I think it best to remove this transformer.

PR #656, opened 2023-08-11
[ENH] Test what happens when we include excluded estimator: 4. TapNet

TonyBagnall:

TapNet needs a reworking anyway, but I thought I would run it through testing to see if it fails. 

1. Passes
2. Passes
3. Passes
4.


TonyBagnall:
Tests pass repeatedly

PR #657, opened 2023-08-11
[ENH] remove ColumnTransformer

### LINKED ISSUE #171: [ENH] Update `ColumnTransformer`
The current `ColumnTransformer` is outdated and inherits from `sklearn`. We should find out 1. why we cant just use the `sklearn` transformer, and if we cant 2. update our version to fit the current interface (most likely as a wrapper).

TonyBagnall:
fixes #171 
the ColumnTransformer was a very early introduction to aeon. I think we should just remove it, and this is what this PR does. It is not used anywhere else in the toolkit.

Docstring reads    "Applies transformations to columns of an array or pandas DataFrame. Simply takes the column transformer from sklearn and adds capability to handle pandas dataframe."

reasons for removal
1. Built to work with nested_univ, which is not used anywhere else any more
2. Fails all the tests and has been excluded for over a year
3. Its just an sktime wrapper and I think we should move away from that
4. We now call multivariate time series variables channels not columns, which always seemed weird to me.

If someone wants to keep it, I would appreciate it if they at least updated it to pass the tests.


hadifawaz1999:
Seems that also `ColumnConcatenator` also considers channels as columns, naming it differently is better i think. But I agree with the removal of this one.

hadifawaz1999 (approved):
LGTM!

hadifawaz1999 (approved):
LGTM!

TonyBagnall:
> setup.cfg

sigh, and two other ones. I really should make the effort to sort that out

PR #661, opened 2023-08-13
[ENH] remove exclusions from testing in tests/_config.py

TonyBagnall:
see #347. This PR removes exclusions for

    "MiniRocket", see #652 
    "MatrixProfileTransformer", see #654 
    "TapNetRegressor", see #656 
    "TapNetClassifier", see #656 

since they have all passed multiple times. If they fail on overnight, we can exclude them again. I would like direct information from aeon, not some obscure issues from the bad place, since so much has changed. Only remaining exclusion is ColumnTransformer, which is removed in #657 

1. passes
2. passes
3.


hadifawaz1999 (approved):
LGTM!

hadifawaz1999 (approved):

Issue #653, opened 2023-08-10
[ENH] Speed up notebooks


TonyBagnall:
### Describe the feature or idea you want to propose

Notebook testing now takes around 40-50 mins, which is excessive imo. Matt has added a breakdown to the time each takes. It would be good to try speed up the worst offenders.

### Describe your proposed solution

speed up some of these
<img width="319" alt="image" src="https://github.com/aeon-toolkit/aeon/assets/9594042/ab9545b1-ca3e-49db-ad66-89edc7c139a5">
Start with those over 100 secs:

- [ ] Clustering (585 secs): see #659 
- [ ] forecasting_proba (442)
- [ ] forecasting (353)
- [ ] deep_learning (319)
- [ ] dictionary_based (141)
- [x] signature_method (116) see #662
- [ ] sklearn_distances (111)

TonyBagnall:
can close this, things seem fine atm

PR #662, opened 2023-08-13
[ENH] speed up signatures notebook

TonyBagnall:
see #653. This PR reduces the search space used for signatures for the example tuning, reducing notebook CI time.

There is an image missing for this notebook I cannot find anywhere. I have asked @jambo6 if he has a copy, otherwise will just remove the image and related text


hadifawaz1999 (approved):
LGTM!

PR #678, opened 2023-08-21
[DOC] Add @dguijo to core developer table and CODEOWNERS

MatthewMiddlehurst:
Adds @dguijo to the About webpage as a core developer and adds them to the CODEOWNERS for regression after the recent core developer vote.

If there are any other changes required/requested just post them here.


dguijo (approved):
Everything is correct. You can also add me to the ordinal classification module. 

Thank you, @MatthewMiddlehurst!

Issue #676, opened 2023-08-20
[BUG] extract_path in load_classification() requires a forward slash


BenGravell:
### Describe the bug

Datasets do not load properly when using the `extract_path` kwarg in `aeon.datasets.load_classification()` when the `extract_path` does not end with a forward slash `/` to indicate that it is a directory. 

This is a sharp edge for users, as the output of `os.path.join()` does not produce a forward slash. A basic OS-independent example of the `extract_path` is given in the [attached gist](https://gist.github.com/BenGravell/9a0a8257d4fed5f543a721a2ad6d0ee0). This would be a more useful example for the docs at https://www.aeon-toolkit.org/en/latest/api_reference/auto_generated/aeon.datasets.load_classification.html#aeon.datasets.load_classification instead of the ad-hoc `C:/Temp/` which is Windows-specific and is not even a default created directory for new installs of Windows (at least on my machine).

### Steps/Code to reproduce the bug

https://gist.github.com/BenGravell/9a0a8257d4fed5f543a721a2ad6d0ee0

### Expected results

No error thrown & dataset loaded properly as
```
X, y, meta_data = load_classification(dataset_name, extract_path=extract_path)
```
when `extract_path` does not end in a forward slash, e.g. with `extract_path=os.path.join(tempfile.gettempdir(), "aeon", "datasets")`.

### Actual results

```python-traceback
PS C:\Users\bjgra\GitHub\vibes> python .\extract_test.py
Traceback (most recent call last):
  File "C:\Users\bjgra\GitHub\vibes\extract_test.py", line 23, in <module>
    X, y, meta_data = load_classification(dataset_name, extract_path=extract_path)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\bjgra\venv\vibes\Lib\site-packages\aeon\datasets\_data_loaders.py", line 1279, in load_classification
    return _load_tsc_dataset(
           ^^^^^^^^^^^^^^^^^^
  File "C:\Users\bjgra\venv\vibes\Lib\site-packages\aeon\datasets\_data_loaders.py", line 466, in _load_tsc_dataset
    return _load_saved_dataset(
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\bjgra\venv\vibes\Lib\site-packages\aeon\datasets\_data_loaders.py", line 301, in _load_saved_dataset
[...]

PR #665, opened 2023-08-16
[ENH] Improve coverage for datasets module

TonyBagnall:
#### Reference Issues/PRs

Part of #664 
Fixes #668 

#### What does this implement/fix? Explain your changes.

This adds some tests for uncovered data loaders. It is tricky to get high coverage because much of the untested code relies on external websites, but this should catch 100+ lines with new tests.

PR #679, opened 2023-08-21
[BUG] remove unnecessary call to os.path.dirname

TonyBagnall:
#### Reference Issues/PRs

fixes (hopefully) #676 

load_classification/regression/forecasting had a redundant call to os.path.dirname for the extract path, perhaps a legacy of previously passing the file name with the path. This meant this fails

https://gist.github.com/BenGravell/9a0a8257d4fed5f543a721a2ad6d0ee0

removing the call makes no difference I can see, apart from removing the need for the suffix slash

### DIFF
--- a/aeon/datasets/_data_loaders.py
+++ b/aeon/datasets/_data_loaders.py
@@ -219,7 +219,7 @@ def load_from_tsfile(
     """
     # Check file ends in .ts, if not, insert
     if not full_file_path_and_name.endswith(".ts"):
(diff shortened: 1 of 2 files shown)


MatthewMiddlehurst:
Don't see anything wrong with this if it resolves the issue. Is there a test we can add for different extract paths?

TonyBagnall:
> Don't see anything wrong with this if it resolves the issue. Is there a test we can add for different extract paths?

I can look at that in #665

MatthewMiddlehurst (approved):
ok, sounds fair.

PR #714, opened 2023-09-01
[ENH] skip tests in tests_lookup for PR_TESTING

TonyBagnall:
CI run testnosoftdeps runs before everything else and the tests of lookup function take around 5 mins.
usually 
test_all_estimators_by_scitype
or
test_all_estimators_return_tags

all_estimators is a confusing function (to me at least) and I suspect somewhere it is doing much more work than necessary (repeated scans maybe), but simplest solution for me is to just run them in the overnight tests by switching on PR_TESTING


TonyBagnall:
doesnt skip the tests, killing this

PR #717, opened 2023-09-02
[ENH] remove TagAliasMixin

TonyBagnall:
the base class has a thing called TagAliaserMixin, that is meant to be used to deprecate tags. This PR removes it for the following reasons

1. It is not at all clear how to use it, and the documentation has no examples. The developer guide says this

### Deprecating tags
To deprecate tags, it needs to be ensured that warnings are raised when the tag is used. There are two common scenarios: removing a tag, or renaming a tag.

For either scenario, the helper class TagAliaserMixin (in aeon.base) can be used.

To deprecate tags, add the TagAliaserMixin to BaseEstimator, or another BaseObject descendant. It is advised to select the youngest descendant that fully covers use of the deprecated tag. TagAliaserMixin overrides the tag family of methods, and should hence be the first class to inherit from (or in case of multiple mixins, earlier than BaseObject).

alias_dict in TagAliaserMixin contains a dictionary of deprecated tags: For removal, add an entry "old_tag_name": "". For renaming, add an entry "old_tag_name": "new_tag_name" deprecate_dict contains the version number of renaming or removal, and should have the same keys as alias_dict.

The TagAliaserMixin class will ensure that new tags alias old tags and vice versa, during the deprecation period. Informative warnings will be raised whenever the deprecated tags are being accessed.

When removing/renaming tags after the deprecation period, ensure to remove the removed tags from the dictionaries in TagAliaserMixin class. If no tags are deprecated anymore (e.g., all deprecated tags are removed/renamed), ensure to remove this class as a parent of BaseObject or BaseEstimator.

I dont really understand why you would change BaseEstimator to remove a tag. There is no alias_dict declared in TagAliaserMixin, so I assume you have to create it? All in all, the above is not sufficient for me to use this class. Maybe thats just me, but there are other reasons.

2. It is not declared in all at the top of the file, nor is it in init. This makes it harder to use

3. It is not tested at all. I came to this class from the coverage stats, and it has no coverage at all

I dont mind keeping it if someone likes it, but if someone does want this, could they please

a) give an example of how to use it
b) write some tests for the functionality


MatthewMiddlehurst:
I think we should keep this for now, but maybe move it to its own file or utils.

TonyBagnall:
> I think we should keep this for now, but maybe move it to its own file or utils.

ok, but do you understand how to use it? Seems to require

TonyBagnall:
@MatthewMiddlehurst could you take a look at this if you get a chance and you want to keep it

PR #721, opened 2023-09-04
[ENH] Convert remaining collection transformers

MatthewMiddlehurst:
Converts the remaining transformers in the `collection` sub package to use the new base class.

Returns the `PlateauFinder` class to collections for now.


TonyBagnall:
could you add the capability:multivariate tag to any you are converting that miss it at the moment?

MatthewMiddlehurst:
I think that would be better in its own PR. We still have the duplicate tag which I want to deal with, will create an issue.

TonyBagnall:
I think this has been subsumed by #825 and we can close this PR @MatthewMiddlehurst ?

MatthewMiddlehurst:
I will take a look, I think there is a change to PlateuFinder I would like to revert.

Issue #711, opened 2023-08-31
[BUG] Intermittent test fail from shapelet transform


TonyBagnall:
### Describe the bug

the test, 
aeon/transformations/collection/tests/test_shapelet_transform.py::test_st_on_basic_motions

fails occasionally with results different from expected. Possible obscure lack of seeding, perhaps when shapelets are equal? Both of below are on ubuntu 22.04, but different python versions. If we see more, post link here

### Steps/Code to reproduce the bug

see
https://github.com/aeon-toolkit/aeon/actions/runs/6040287686/job/16391060741?pr=673
and
https://github.com/aeon-toolkit/aeon/actions/runs/6019651707/job/16329767004

### Expected results

not fail tests. 

### Actual results

fails test

TonyBagnall:
https://github.com/aeon-toolkit/aeon/actions/runs/6058143042/job/16440031225?pr=650

TonyBagnall:
https://github.com/aeon-toolkit/aeon/actions/runs/6119665293/job/16610201937

MatthewMiddlehurst:
https://github.com/aeon-toolkit/aeon/actions/runs/6121348243/job/16615093102

MatthewMiddlehurst:
https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py#L29

The test itself. Not gone through the logs, but wonder how small the difference in results is

PR #734, opened 2023-09-08
set random_state in rotation forest

TonyBagnall:
#### Reference Issues/PRs
see #711 

attempts to fix the sporadic fail by setting the random_state in RotationForestClassifier

### DIFF
--- a/aeon/classification/shapelet_based/_stc.py
+++ b/aeon/classification/shapelet_based/_stc.py
@@ -219,7 +219,9 @@ class ShapeletTransformClassifier(BaseClassifier):
         )
 
         self._estimator = _clone_estimator(
-            RotationForestClassifier() if self.estimator is None else self.estimator,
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst:
`_clone_estimator` should already be setting the random state, if it was not, I think we would be seeing issues with other estimators.

Have you managed to reproduce the failure locally? Just wondering if it is OS specific really.

MatthewMiddlehurst:
Ah never mind, I see you have linked in the issue. Appears everywhere.

TonyBagnall:
just had a quick scan and took a punt on this, its appearing frequently, but not so frequently to be an obvious bug. Cant reproduce it, although a lot of the fails are windows. Ah this fails for something else!

MatthewMiddlehurst:
I think the test is just the transform portion, not the classifier

TonyBagnall:
just ran it 20 odd times locally, no fails

PR #738, opened 2023-09-09
[BUG] Fix EE test edge case

TonyBagnall:
fixed a test for error raised by sklearn so it always raises same error, even if all labels the same


MatthewMiddlehurst:
I'm not quite sure what this is fixing, is there a failing workflow?

TonyBagnall:
under the old generator, two different errors could be raised using 10% parameter search at different points in scikit. If I fix the class values like this, it doesnt,

MatthewMiddlehurst (approved):
Do you remember where the errors are or will it happen if I run the old version enough?

I assume its due to the random number of values for each class generated (guaranteed at least 1 for each, but can be 1). I can approve if you just want the test to cover 50/50 class values but the underlying subsampling error would still exist.

TonyBagnall:
> Do you remember where the errors are or will it happen if I run the old version enough?
> 
> I assume its due to the random number of values for each class generated (guaranteed at least 1 for each, but can be 1). I can approve if you just want the test to cover 50/50 class values but the underlying subsampling error would still exist.

this is a new test, so its not like a new fail, just a scenario not covered. Basically if only one of one class  it fails in the place Im not trying to test, which is in the cross validation, if you take 10% of 10 cases, it should fail there whatever the class distribution

PR #739, opened 2023-09-10
[DOC] Update documentation for new devs  and documentation build fixes

hedeershowk:

Some notes on the PR:

- added `docs/symlink_examples.sh` from sktime but this may not be correct?
- `docs/Makefile`:
    * Switched SOURCE in the Makefile to '.'
    * Added `mkdir -p source/examples` (because symlink_examples.sh seems to want it there)
- generated `docs/developer_guide/dev_installation.rst` from `docs/developer_guide/dev_installation.md` and removed the latter
- `Git and GitHub Workflow` looks like a fork and PR model (what I expected) but the developer installation guide looks like just pulling the main repo and PR against that.  Which method should new contributors use?
- switched Argument(s) -> Parameters in docstrings everywhere so long as there was no ambiguity at all and the parameters matched the function signature.  Otherwise, I left it as it was.  There are other doc build cleanup tasks like this remaining it seems.

I had to manually install pandoc as well on the system in order for the documentation build to work.  I'm thinking I've maybe done something wrong in my initial setup (works after I installed).

##### For all contributions


MatthewMiddlehurst:
Ah someone else pointed out the docs Makefile recently actually, were you able to build the documentation locally without issue with the edited file?

When editing docs I mostly use the [readthedocs](https://readthedocs.org/projects/aeon-toolkit/builds/) builds, so I would miss these issues!

docs/Makefile
```diff
@@ -22,6 +22,7 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	mkdir -p source/examples
 	./symlink_examples.sh
```
MatthewMiddlehurst, line 26: So I believe what this script did was create an examples folder in the documentation, then create an individual symlink for each notebook in examples. This would need to be updated constantly as the notebooks were changed.

This shouldn't be required anymore, as we just replaced it with a symlink to the examples directory (https://github.com/aeon-toolkit/aeon/blob/main/docs/examples).

hedeershowk: Oh great okay then I'll just remove it completely!

hedeershowk:
> Ah someone else pointed out the docs Makefile recently actually, were you able to build the documentation locally without issue with the edited file?

Ya with the edis the html build worked although there were lots of warnings from Sphinx.  I should run the rtd build too and see what that looks like.

hedeershowk:
Matthew, I got rid of symblinks and made one more minor update (link to readthedocs in `Documentation Build` wasn't correct).

Not sure how the readthedocs build is different from the normal `make html` build though and if there's a separate path here?  I can check it out if so and maybe add some notes to the `Documentation Build` page too

MatthewMiddlehurst:
This is the link for the latest build: https://readthedocs.org/projects/aeon-toolkit/builds/21876413/

I don't think it is that much different, just does not use the Makefile

MatthewMiddlehurst:
Thanks for the fixes. Is the new [docs/source/index.md](https://github.com/aeon-toolkit/aeon/pull/739/files#diff-fe2f3439fff0648f2185e11c2a7fc6c0be3109e06acdcf9a5ed16bf73bdc2198) necessary?

I went through the local docs build (I'm on windows so its make.bat though), and it seemed to work fine except for the example notebooks and images. 

Im unsure if the extras on the current Makefile are actually required. it may just be better to use somthing closer to the `sphinx-quickstart` default i.e.
```
# Minimal makefile for Sphinx documentation
#

# You can set these variables from the command line, and also
# from the environment for the first two.
SPHINXOPTS    ?=
SPHINXBUILD   ?= sphinx-build
SOURCEDIR     = .
BUILDDIR      = _build

# Put it first so that "make" without argument is like "make help".
help:
	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

.PHONY: help Makefile

# Catch-all target: route all unknown targets to Sphinx using the new
# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
%: Makefile
	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
 
No need to worry about that here though I think.

hedeershowk:
> Thanks for the fixes. Is the new [docs/source/index.md](https://github.com/aeon-toolkit/aeon/pull/739/files#diff-fe2f3439fff0648f2185e11c2a7fc6c0be3109e06acdcf9a5ed16bf73bdc2198) necessary?

Ya it looks like sphinx wants it to be in there to properly build.  I can check one more time to make sure.


> Im unsure if the extras on the current Makefile are actually required. it may just be better to use somthing closer to the `sphinx-quickstart` default i.e.

I'm happy to open a new branch and test this out if you'd like?  I agree and wanted to make minimal changes in this PR just to get it working with the existing config.

MatthewMiddlehurst:
If you want to test it out that would be great, does not have to be in this PR though. The simple version `make.bat html` builds without error but fails to link the images and examples in the local `_build` html for some reason.

I think having a duplicate index file is not a good idea, hopefully it is not needed, but if it is currently maybe there is a setting pointing to `docs/source/index.md` rather than just `docs/index.md`?

MatthewMiddlehurst:
Ah figured out the example/image thing. Its Windows (not uncommon). Fixing it did give the pandox warning for the annotation notebook, will look into that.

hedeershowk:
> I think having a duplicate index file is not a good idea, hopefully it is not needed, but if it is currently maybe there is a setting pointing to `docs/source/index.md` rather than just `docs/index.md`?

Pushed with the source/index.md file removed

MatthewMiddlehurst (approved):
Great, thanks for the changes. Started the tests, don't think anything should fail though.

MatthewMiddlehurst:
@all-contributors add @hedeershowk for bug fix and documentation

Issue #732, opened 2023-09-08
[BUG] _write_dataframe_to_tsfile does not work correctly with multivariate series


TonyBagnall:
### Describe the bug

the function _write_dataframe_to_tsfile  writes the wrong header, it should work out the characteristics of the data before calling _write_header, but it does not. 

### Steps/Code to reproduce the bug

```python

import os
from aeon.datasets import load_from_tsfile
from aeon.datasets._data_writers import _write_dataframe_to_tsfile
from aeon.utils._testing.collection import make_nested_dataframe_data

X, y =make_nested_dataframe_data(n_cases=10, n_channels=3, n_timepoints=10)
_write_dataframe_to_tsfile(X,"./Temp",problem_name="testy.ts", y=y)
assert os.path.exists("./Temp/testy.ts")
X, y = load_from_tsfile("./Temp/testy.ts")
```

### Expected results

load should work

### Actual results

```python-traceback
 File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 228, in load_from_tsfile
    data, y, meta_data = _load_data(file, meta_data)
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 154, in _load_data
    raise IOError(
OSError: Seen 3 in case 2.Expecting univariate from meta data
```

Lymah123:
Can I work on the issue?

MatthewMiddlehurst:
Apologies, I was not tracking this issue and missed the comment. Someone else on the Slack picked this up and merged it. Thanks for offering though, if you want other issue suggestions feel free to ask!

PR #740, opened 2023-09-11
[BUG] fixes multivariate dataframe tsfile writes (#732)

hedeershowk:
#### Reference Issues/PRs
https://github.com/aeon-toolkit/aeon/issues/732

Added appropriate parameters for the dataframe ts file writer to output the proper header.  Also one minor fix to `write_to_tsfile()` (see comment below).  

Yes in `aeon/datasets/tests/test_data_writers.py` I added `test_write_dataframe_to_ts_and_load`.  However, this test is currently failing, see comment below.

#### Any other comments?

* In `write_to_tsfile()` the initial check was just calling `isinstance(pd.Dataframe)` which I switched to `isinstance(X, pd.DataFrame)`.  It should have raised at some point though but maybe that function has never been called before without X being a `list` or `numpy.ndarray`? 
* I don't expect my current test to pass because the behavior of `load_from_tsfile()` hasn't been changed in this PR.  I wasn't sure if the idea was to keep `load_from_tsfile()` and `load_from_tsfile_to_dataframe()` as explicitly separate operations or if a user writing a dataframe to a a ts file should expect `load_from_tsfile()` to also return them a dataframe (right now it's np.ndarray).  I would personally expect a dataframe if I've written a dataframe out but I'm not sure whether that's the intention and if so how to store that information within the header (or elsewhere?) so the loader can be made aware.  Maybe a `@datatype` in the header?   

##### For all contributions


TonyBagnall:
hi, thanks for this, its very welcome. The dataframe loaders/writers are legacy really, back to the time when we stored collections in nested dataframes when each cell was a pd.Series. Good spot on the isinstance check, also found it here when looking to improve coverage. https://github.com/aeon-toolkit/aeon/pull/733. I am happy to leave 733 and update from yours once we get it in. I'll take a look today.

aeon/datasets/tests/test_data_writers.py
```diff
@@ -98,3 +98,27 @@ def test_write_dataframe_to_ts():
...
+    Tests whether a dataframe written to a .ts file can be read with
+    load_from_tsfile
+    """
+    # load an example dataset
+    problem_name = "Testy.ts"
+    X, y = make_nested_dataframe_data()
+    # output the dataframe in a ts file
+    _write_dataframe_to_tsfile(
+        X=X,
+        path="./Temp/",
```
TonyBagnall, line 114: using "./Temp/" occasionally fails on CI, in #733 I switched to

    path = os.path.dirname(__file__) + "/Temp/"

which seems more reliable

hedeershowk: Okay I'll update.  How do you feel about os.path.join instead of +?

TonyBagnall:
>I wasn't sure if the idea was to keep load_from_tsfile() and load_from_tsfile_to_dataframe()

hi @hedeershowk, just to clarify whats going on here, because its not necessarily intuitive and is a bit of a legacy to the origins of aeon  ....

load_from_tsfile() just loads data into either numpy (if series are equal length) or list of numpy (if series are unequal). These are the standard formats for the machine learning side (i.e. not forecasting): for equal length series stored as numpy is shape (n_cases, n_channels, n_timepoints). This structure is not amenable to standard pd.DataFrames, which in the simplest case for equal length with n_channels == 1 would store in shape (n_cases, n_timepoints) or (n_timepoints, n_cases), and if n_channels > 1 it gets complicated. You can use hierarchical/long formats,  but its messy.

Originally, some contributors really wanted to use data frames, so overcame these problems by storing data in a pd.DataFrame shape (n_cases, n_channels), with each entry a time series of length n_timepoints. This non standard usage tended to confuse everyone, so we have more recently attempted to implicitly deprecate it by removing usage. The x_to_dataframe functions are a legacy feature to load and read in this format. It means in recent refactoring I tried to ignore them, but you have found me out :)

So back to your PR. ITs useful, thanks, and good spot with the bug. I'd very much like to encourage your participation, we try to be as welcoming as possible. I will try take a look at how it fits in tomorrow, but I have quite a lot on, sorry if it takes me a little longer. 

If you want to  discuss further, please join our slack, as in #general or #contributors, or direct message me.

aeon/datasets/tests/test_data_writers.py
TonyBagnall: I think its good :)

hedeershowk:
>  >I wasn't sure if the idea was to keep load_from_tsfile() and load_from_tsfile_to_dataframe()
> 
> hi @hedeershowk, just to clarify whats going on here, because its not necessarily intuitive and is a bit of a legacy to the origins of aeon  ....
> 
> load_from_tsfile() just loads data into either numpy (if series are equal length) or list of numpy (if series are unequal). These are the standard formats for the machine learning side (i.e. not forecasting): for equal length series stored as numpy is shape (n_cases, n_channels, n_timepoints). This structure is not amenable to standard pd.DataFrames, which in the simplest case for equal length with n_channels == 1 would store in shape (n_cases, n_timepoints) or (n_timepoints, n_cases), and if n_channels > 1 it gets complicated. You can use hierarchical/long formats,  but its messy.
> 
> Originally, some contributors really wanted to use data frames, so overcame these problems by storing data in a pd.DataFrame shape (n_cases, n_channels), with each entry a time series of length n_timepoints. This non standard usage tended to confuse everyone, so we have more recently attempted to implicitly deprecate it by removing usage. The x_to_dataframe functions are a legacy feature to load and read in this format. It means in recent refactoring I tried to ignore them, but you have found me out :)
> 
> So back to your PR. ITs useful, thanks, and good spot with the bug. I'd very much like to encourage your participation, we try to be as welcoming as possible. I will try take a look at how it fits in tomorrow, but I have quite a lot on, sorry if it takes me a little longer. 
> 
> If you want to  discuss further, please join our slack, as in #general or #contributors, or direct message me.    
> 
> 
>   

Ah okay thanks for the background information.  Makes sense and I can appreciate the desire to get things into data frames (in some cases) and the complexity of getting that done properly with the shape. 

I could update the test to more explicitly test the header construction the way you did in the bug report or just use the dataframe loader.  No rush here, just let me know what you'd prefer to get the PR how it can be most useful to you.

I've also joined the Slack (thanks) and look forward to participating and contributing more.

hedeershowk:
cleaned up that test and I switched to [tempfile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory).  I was actually running into a bunch of file not found errors when running tests in this directory (even in the main branch).  Not sure why just running `pytest` from the module root didn't fail but when I explicitly ran pytest on `test_data_writers.py` I had a bunch of fails.  This PR fixes that.

hedeershowk:
hm haven't dug into it but looks like maybe the test fail in `test_proportion_train_in_param_finding()` is related to the shortcut in base when dealing with a single class?  I'll try to reproduce and resolve it locally.

hedeershowk:
Not clear how my changes could have an impact on the current failing test in CI?  I'll run locally to confirm and see if I can figure out the issue as well.

TonyBagnall (commented):
it looks good but I'm not sure about the changes to the Makefile @MatthewMiddlehurst can you take a look please?

TonyBagnall:
@hedeershowk could you update this with main (and resolve contributors conflict)? I think a lot of these changes were also in your other PR. Thanks, Tony

TonyBagnall (approved):

Issue #726, opened 2023-09-05
[BUG] CBOSS crashes with small train set siz


TonyBagnall:
### Describe the bug

found when writing tests for HC1, CBOSS crashes with train always with trains set 4, often set size of 10, not with train size 15

### Steps/Code to reproduce the bug

```python
    from aeon.classification.dictionary_based import ContractableBOSS
    from aeon.utils._testing.collection import make_2d_test_data
    X, y = make_2d_test_data(n_cases=4, n_timepoints=100, n_labels=2)
    cboss = ContractableBOSS()
    cboss.fit(X, y)
```

### Expected results

either informative exception about train set size or should work.

 

### Actual results

```python-traceback
Traceback (most recent call last):
  File "C:\Code\aeon\venv\lib\site-packages\IPython\core\interactiveshell.py", line 3460, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-c35d1f6140d9>", line 1, in <module>
    runfile('C:\\Code\\aeon\\aeon\\local\\debug.py', wdir='C:\\Code\\aeon\\aeon\\local')
  File "C:\Program Files\JetBrains\PyCharm 2020.3\plugins\python\helpers\pydev\_pydev_bundle\pydev_umd.py", line 198, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "C:\Program Files\JetBrains\PyCharm 2020.3\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:\Code\aeon\aeon\local\debug.py", line 6, in <module>
    cboss.fit(X, y)
  File "C:\Code\aeon\aeon\classification\base.py", line 186, in fit
    self._fit(X, y)
  File "C:\Code\aeon\aeon\classification\dictionary_based\_cboss.py", line 266, in _fit
    boss._clean()
  File "C:\Code\aeon\aeon\classification\dictionary_based\_boss.py", line 691, in _clean
    self._transformer.words = None
AttributeError: 'NoneType' object has no attribute 'words'
```

hedeershowk:
[...]

PR #741, opened 2023-09-12
[BUG] remove base classifier shortcut and new test

hedeershowk:
#### Reference Issues/PRs

#726 

1.  Adds a guard to `_clean()` in cBOSS in case the transformer is not yet initialized.
2.  Removes the shortcut for single class cases in BaseClassifier.  After discussing with Matthew it seems like this may introduce the need for further updates on many classifiers. 

Whether or not to make the change in BaseClassifier or instead fix up cBOSS and IndividualBOSS (see comment below)

Yes I added a new test `test_boss_empty_fit()` in `test_boss.py`. edit: I can make this test more specific to the issue by forcing a single class in the data.

#### Any other comments?

Guarding against an empty transformer only partially resolves the bug reported in #726 (and reproduced in a test case I added).  After that, there's a new error when the ContractableBOSS code tries to calculate accuracy of the IndividualBOSS instances.  This is due to the same root cause though:

The root cause is that because IndividualBOSS subsamples, there are situations where it tries to fit data with a single class.  This causes the BaseClassifier to not call `_fit()` on the IndividualBOSS instance.  The `_transformer` and `_transformer_data` properties of IndividualBOSS are initialized to `None` and `[]` respectively in the `__init__()` and then updated in `_fit()`.  Because `_fit()` is never called though, the IndividualBOSS instance doesn't update those two variables and this causes breaks further down.

In this PR I'm proposing to changing BaseClassifier mainly to initiate a discussion on the topic but I think an equally viable path is to go into IndividualBOSS and make sure that `_transformer_data` and `_transformer` are initialized in the `__init__` in such a way that the rest of the code works.  I've already started looking into that and don't think it's a huge lift (probably a lot simpler than the refactor implied by updated BaseClassifier tbh).  Happy to update this PR with that instead of this change to `BaseClassifier`.

Happy to discuss on slack as well

##### For all contributions
- [X ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.


aeon/classification/base.py
```diff
@@ -150,10 +150,10 @@ def fit(self, X, y):
...
         # escape early and do not fit if only one class label has been seen
         #   in this case, we later predict the single class label seen
-        if len(self.classes_) == 1:
-            self.fit_time_ = int(round(time.time() * 1000)) - start
-            self._is_fitted = True
-            return self
+        # if len(self.classes_) == 1:
+        #     self.fit_time_ = int(round(time.time() * 1000)) - start
+        #     self._is_fitted = True
+        #     return self
```
MatthewMiddlehurst, line 156: I think we should keep this, having the same functionality and output for single class label inputs for all classifiers is a good trait I think we should aim to keep. 

However, as you have discovered there is a hole in our testing currently where we do not test small input sizes where this can appear for cases like subsampling.

Im open to discussion, but my current suggestion would be to generalise your test and move it to [here](https://github.com/aeon-toolkit/aeon/blob/main/aeon/classification/tests/test_all_classifiers.py#L42) so it is run on every classifier. This still puts the onus on the individual classifier to deal with the single label case like your change to `_clean(self)`, but will actively enforce that it is being done rather than the current situation.

I can help with this if you need.

aeon/classification/dictionary_based/tests/test_boss.py
```diff
@@ -23,3 +24,9 @@ def test_boss_train_estimate():
     assert train_probas.shape == (20, 2)
     train_preds = boss.classes_[np.argmax(train_probas, axis=1)]
     assert accuracy_score(y_train, train_preds) >= 0.6
+
+
+def test_boss_empty_fit():
+    X, y = make_2d_test_data(n_cases=4, n_timepoints=100, n_labels=2)
+    cboss = ContractableBOSS()
+    cboss.fit(X, y)
```
MatthewMiddlehurst, line 32: I think we can reduce the data size for what we want to test, i.e. cases=3 and timepoints=20.

It would also be good to test `predict_proba` as well to make sure that runs as intended.

MatthewMiddlehurst (commented):
Hello, thanks for the PR and sorry for the late reply, I have been away!

MatthewMiddlehurst:
If anyone has any comments on the current method for dealing with single class label input i.e. @TonyBagnall or @patrickzib feel welcome to post here.

TonyBagnall:
> If anyone has any comments on the current method for dealing with single class label input i.e. @TonyBagnall or @patrickzib feel welcome to post here.

I think we should keep it for now, since it deals with things that can break in fairly obscure places.

hedeershowk:
Sorry for the late response here.  So how about I revert the base changes and as Matthew suggested move the test and keep the specific fix for the bug?

MatthewMiddlehurst:
Yes I think keep the base as is for now.

TonyBagnall:
> Sorry for the late response here. So how about I revert the base changes and as Matthew suggested move the test and keep the specific fix for the bug?

sounds good to me, thanks.

TonyBagnall:
> Sorry for the late response here. So how about I revert the base changes and as Matthew suggested move the test and keep the specific fix for the bug?

hi @hedeershowk are you still working on this? Can we help in any way?

hedeershowk:
> > Sorry for the late response here. So how about I revert the base changes and as Matthew suggested move the test and keep the specific fix for the bug?
> 
> hi @hedeershowk are you still working on this? Can we help in any way?

Sorry I haven't followed up.  I can clean it up and PR again in the next few days but if it's holding anything back please obviously go ahead.  Apologies for such a long delay here.

MatthewMiddlehurst:
Thanks for the reply, no rush really. Just checking in to see if you still want to work on it 🙂

MatthewMiddlehurst:
I have resolved the conversations from before, I think we want to keep the base class functionality for now. This can cause some issues, but with proper testing they should be caught.

MatthewMiddlehurst:
@all-contributors please add @hedeershowk for code.

MatthewMiddlehurst (approved):

PR #754, opened 2023-09-20
[ENH] ShapeDTW as a distance function !

hadifawaz1999:

Implements ShapeDTW in numba inside the distances module

I did have to ignore the test_cost_mat because it will never work with ShapeDTW, the last element in the ShapeDTW cost mat, is different than the actual shapeDTW distance given that the algorithm transfers the alignment path back to the original series.

Made a small change in the test_transformation of sklearn compatibility, using assert_almost_equal now because for some reason the assertion came to a shapeDTW distance 0.6 != 0.60000000000000000000001 (and other examples like that) so wasn't sure the reason for that , probably some floating point stuff.

#### Any other comments?
Let me know if more information should be added to explain the implementation.
PS: this is not the implementation form the ShapeDBA paper, because ours only work in the case where the descriptor is identity, leaving that as future work and add it as a condition to use the more efficient code. But i made sure my implementation gives the same results as ours in the paper :)


aeon/distances/_shapedtw.py
```diff
@@ -0,0 +1,423 @@
...
+    # loop through each data point
+    for i in range(n_timepoints):
+        # loop through each dimension of the MTS
+        for j in range(n_channels):
+            val = descriptor_function(x[j, i : i + sliding_window])
+            out_mts[j * dim_desc : (j + 1) * dim_desc, i] = val
+    return out_mts
+
+
+def shapedtw_distance(
```
TonyBagnall, line 76: maybe shape_dtw_distance?

hadifawaz1999:
Fixed the caching issue @chrisholder , was using a dictionary for the choice of descriptor function which is considered a global large array apparently and can't use it in numba if i want to cache, so replaced it with if statements.

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+    y : np.ndarray
+        Second time series, either univariate, shape ``(n_timepoints,)``, or
+        multivariate, shape ``(n_channels, n_timepoints)``.
+    window : float or None, default=None
+        The window to use for the bounding matrix. If None, no bounding matrix
+        is used. window is a percentage deviation, so if ``window = 0.1`` then
+        10% of the series length is the max warping allowed.
+        is used.
+    descriptor : str, default=None (if None then identity is used).
+        This defines which transformation is applied on the sub-sequences.
```
chrisholder, line 136: What are the options include a list of valid options?

I also wouldn't describe my parameter with 'this'. Just say 

```
Defines which transformation is applied on the sub-sequeces. Valid descriptors are: ['identity', ...]. 

Identity is ....<insert description of identity>
other is... <insert description of what the others do>

If not specified then identity is used.
```

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+    reach : int, default=30.
+        This is the length of the sub-sequences.
```
chrisholder, line 138: Same here just say 

`The length of the sub-sequences to consider`

chrisholder: Change what is mentioned here in all places

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+
+    Returns
+    -------
+    float
+        DTW distance between x and y, minimum value 0.
```
chrisholder, line 143: Shape-DTW

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+    """
+    if x.ndim == 1 and y.ndim == 1:
+        _x = x.reshape((1, x.shape[0]))
+        _y = y.reshape((1, y.shape[0]))
+
+        x_pad = _pad_ts_edges(x=_x, reach=reach)
+        y_pad = _pad_ts_edges(x=_y, reach=reach)
+
+        bounding_matrix = create_bounding_matrix(
+            _x.shape[1], _y.shape[1], window, itakura_max_slope
```
chrisholder, line 168: Assuming this is intentional that you don't use the padded arrays for the bounding matrix?

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+
+    Raises
+    ------
+    ValueError
+        If x and y are not 1D or 2D arrays.
+    """
+    new_x = _transform_subsequences(x=x, descriptor=descriptor, reach=reach)
+    new_y = _transform_subsequences(x=y, descriptor=descriptor, reach=reach)
+
+    shapedtw_cost_mat = _dtw_cost_matrix(
```
chrisholder, line 238: I don't exactly know what the transform does but it is important whatever it spits out is the same shape as your bounding matrix provided (some maybe you want to make it here if this does something weird with x and y)

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+        x=new_x, y=new_y, bounding_matrix=bounding_matrix
+    )
+
+    i = shapedtw_cost_mat.shape[0] - 1
+    j = shapedtw_cost_mat.shape[1] - 1
+
+    shapedtw_dist = 0
+
+    while i >= 0 and j >= 0:
+        shapedtw_dist += np.square(np.linalg.norm(x[:, reach + i] - y[:, reach + j]))
```
chrisholder, line 248: Just for consistency across the distance module can you use:

`from aeon.distances._squared import _univariate_squared_distance
`

I guess here then you'd do 

` shapedtw_dist += _univariate_squared_distance(x[:, reach + i], y[:, reach + j])`

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+
+        a = shapedtw_cost_mat[i - 1, j - 1]
+        b = shapedtw_cost_mat[i, j - 1]
+        c = shapedtw_cost_mat[i - 1, j]
+        if a < b:
```
chrisholder, line 253: This can be simplified to:

```
if a < b and a < c:
    i -= 1
    j -= 1
elif b < c:
    j -= 1
else:
    i -= 1
```

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+    window : float or None, default=None
+        The window to use for the bounding matrix. If None, no bounding matrix
+        is used. window is a percentage deviation, so if ``window = 0.1`` then
+        10% of the series length is the max warping allowed.
+        is used.
+    descriptor : str, default=None (if None then identity is used).
+        This defines which transformation is applied on the sub-sequences.
+    reach : int, default=30.
+        This is the length of the sub-sequences.
+
```
chrisholder, line 139: Missing docstring for itakura para:

    itakura_max_slope : float, default=None
        Maximum slope as a proportion of the number of time points used to create
        Itakura parallelogram on the bounding matrix. Must be between 0. and 1.
        
 Add this to all the functions that it is specified in below (think you accidently deleted it in all the functions during refactor)

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+@njit(cache=True, fastmath=True)
+def _shape_dtw_cost_matrix(
+    x: np.ndarray,
+    y: np.ndarray,
+    bounding_matrix: np.ndarray,
+    descriptor: str = "identity",
+    reach: int = 30,
+) -> float:
+    """Compute the ShapeDTW cost matrix between two series x and y.
+
```
chrisholder, line 357: Just remove this docstring for private method more docs to maintain that no one will see on the API

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+def _shape_dtw_distance(
+    x: np.ndarray,
+    y: np.ndarray,
+    bounding_matrix: np.ndarray,
+    descriptor: str = "identity",
+    reach: int = 30,
+) -> float:
+    """Compute the ShapeDTW distance function between two series x and y.
+
+    Parameters
```
chrisholder, line 207: I would just remove this docstring on the private methods no one will see them on the API

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+    cost_matrix = shape_dtw_cost_matrix(
+        x=x,
+        y=y,
+        window=window,
+        descriptor=descriptor,
+        reach=reach,
+        itakura_max_slope=itakura_max_slope,
+    )
+
+    shapedtw_dist = shape_dtw_distance(
```
chrisholder, line 446: Here effectivly your recomputing the cost matrix again to just get the distance. Is it possible to split up the _shape_dtw_distance function so that the part that gets the distance from the cost matrix is a function that we can call here instead of recomputing the whole cost matrix?

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,594 @@
...
+@njit(cache=True, fastmath=True)
+def _shape_dtw_pairwise_distance(
+    X: np.ndarray,
+    y: np.ndarray = None,
+    window: float = None,
+    descriptor: str = "identity",
+    reach: int = 30,
+    itakura_max_slope: float = None,
+) -> np.ndarray:
+    """Compute the ShapeDTW pairwise distance among a set of series.
```
chrisholder, line 544: Remove the docstring for a private method

chrisholder (changes requested):

aeon/distances/_shape_dtw.py
hadifawaz1999: yup because the bounding matrix should have the same height/width as the cost matrix which has a height/width set to the length of time series on both, so i use the length of the original series instead of the padded one.

aeon/distances/_shape_dtw.py
hadifawaz1999: it wouldn't make a difference, the bounding matrix takes as input the height/width

the transformation function takes a series of length L + 2 * reach (where L is the length of the original series) and outputs a series of length L again.

aeon/distances/_shape_dtw.py
hadifawaz1999: but this a multivariate distance not univariate.

`x[:, reach + i]` is a vector in the general case

aeon/distances/_shape_dtw.py
hadifawaz1999: oh yeah good spotting, i forgot to copy it.

aeon/distances/_shape_dtw.py
hadifawaz1999: makes sense

aeon/distances/_shape_dtw.py
hadifawaz1999: for now i didn't add new options, but i added through out all the docstrings what you mentioned

aeon/distances/_shape_dtw.py
hadifawaz1999: done

aeon/distances/_shape_dtw.py
hadifawaz1999: done

aeon/distances/_shape_dtw.py
hadifawaz1999: done

aeon/distances/_shape_dtw.py
hadifawaz1999: makes sense, i made a function for that now let me know if thats better

aeon/distances/_shape_dtw.py
hadifawaz1999: done

aeon/distances/_shape_dtw.py
hadifawaz1999: makes sense :) thanks

aeon/distances/_shape_dtw.py
chrisholder: Is it?

Isn't x here if it's a multivariate:

x = [[1, 2, 3, 4], [5, 6, 7, 8]]

And lets say rech = 1 and i = 0

x[:, 1+0] = x[:, 1]

This would return 

x[:, 1] = [2, 6]

aeon/distances/_shape_dtw.py
```diff
@@ -0,0 +1,551 @@
...
+    Parameters
+    ----------
+    x : np.ndarray
+        First time series, either univariate, shape ``(n_timepoints,)``, or
+        multivariate, shape ``(n_channels, n_timepoints)``.
+    descriptor : str, default=None (if None then identity is used).
+        Defines which transformation is applied on the sub-sequences.
+        Valid descriptors are: ['identity']
+
+        Identity is ....<insert description of identity>
```
chrisholder, line 75: When I said i mean <insert description of identity> I meant for you to fill in the description :joy:

aeon/distances/_shape_dtw.py
chrisholder: Fill in the description of identity

aeon/distances/_shape_dtw.py
hadifawaz1999: oh never mind, i misunderstood :) i meant to say that x is multivariate but once you index the time axis its like having a univariate now even it its a vector now

aeon/distances/_shape_dtw.py
hadifawaz1999: i actually wanted to add it but i guess i forgot after copying it 😂

chrisholder (approved):
Great addition to the distance module!

PR #724, opened 2023-09-05
[ENH] Experimental new similarity search

TonyBagnall:
this is just a working area to experiment with a new similarity search module, want a PR to run it through the CI.

PR #756, opened 2023-09-21
Similarity search

baraline:
Overview of the content:

- Changed base shape expected for X to (n_samples, n_channels, n_timestamps), and for Q to (n_channels, q_length) to handle the more general case. 
- The choice of how a candidate of X is matched with Q is left to the child classes (e.g.  TopKSimilaritySearch class will ask for a k argument and return top-k matches). A possibility would be to make a base class for distance profiles to force them having the same API and have attributes indicating if they are normalized or not.
- Normalization happens in the predict method every time it is called. It stores the mean and std of each channels of Q and of all possible candidates of X as attributes (e.g _X_means). They are stored as argument to be called in the child classes of the base class when the distance profile is normalized. It may be possible to avoid recomputing it by checking if the length of Q and the stored input _X have not changed.
- Added naive Euclidean and normalized naive Euclidean distance profiles with basic tests
- Added top-k similarity search class with basic tests. The return of predict is an array of tuple as (with k=1) [(id_sample, id_timestamp)] 

TODOs left : 
- Make documentation for module and estimator
- Decide how to handle/check when there is not input or query length change to avoid computing normalization again


MatthewMiddlehurst:
I think this branch may be more up to date with main than the experimental one? @TonyBagnall may have to update the original one to make sure they have the same commits if you want a clearer "files changed" view

baraline:
> I think this branch may be more up to date with main than the experimental one? @TonyBagnall may have to update the original one to make sure they have the same commits if you want a clearer "files changed" view

In order to make a PR, I add to make a fork from the main branch (couldn't fork experimental), as I can't write on it directly. I added the changes made to similarity_search module on the new forked branch and made the PR from this one, thus the difference from the experimental branch :/

MatthewMiddlehurst:
Not an issue with the way you did it I'm pretty sure, I think I can push to both and make sure it is up to date 🙂

TonyBagnall:
my branch was indicative only, happy for my branch to be killed.

MatthewMiddlehurst:
Not saying that any specific person should do this, but I think the new module proposal would benefit from a short notebook explaining the use case and using the new interface .

baraline:
Don't worry, I planned to add this along with more module documentation. Would you prefer to close this PR now or when I've done all the TODOs and docs ?

TonyBagnall (approved):

MatthewMiddlehurst:
See https://github.com/aeon-toolkit/aeon/pull/724 for the PR. If you want to do more work this week, I can give you write access to the repo if that makes it easier. Let me know on slack.

PR #757, opened 2023-09-22
[BUG] try fix max for rng in ShapeletTransform

TonyBagnall:
#### Reference Issues/PRs
see #711 

I have no idea why ST intermittently fails on basic motions test. This is shot in the dark to see if maybe the np max function or mod operator for max values is sometimes non-deterministic. I cannot recreate locally so need to just try out on CI


TonyBagnall:
nope

PR #758, opened 2023-09-22
[BUG] Expected results rework and Shapelet Transform bugfix

### LINKED ISSUE #711: [BUG] Intermittent test fail from shapelet transform

MatthewMiddlehurst:
Fixes #711.

Multiple changes in this PR made during the search for the ST bug.
- Moves the RandomState object generation for multiple estimators which use joblib outside of the threaded function. This reduces the amount of lines used for each and is cleaner IMO.
- Reworks the methods for generating expected results and the tests which use them (as a result of the above). The UnitTest check now uses less data, and the different tests for different datasets have been merged into one.
- Adds an expected results test in line for the classification and regression ones.
- Regenerates the expected results where relevant.
- Rounds information gain to 8 decimals in the shapelet transform (the actual bugfix) and changes shapelet sorting so shorter shapelets are kept during information gain ties.

This PR was initially hunting for the bugfix, and regenerating expected results while doing that was very annoying so i made it easier 🙂.

After 10 reruns, the ST bug did not reappear (other failures happened, but not related to this).

I can split this PR up on request, since some of the changes are not really related.


TonyBagnall:
is this not now completely unseeded though?

MatthewMiddlehurst:
Instead of creating the `RandomState` inside the joblib thread/function, this creates it outside and passes it as a parameter. 

The number generated by rng in `check_random_state(rng.randint(np.iinfo(np.int32).max))` to create the new seeded `RandomState` should always be the same if rng is seeded, or random if it is not.

...Of course this changes the rng previously used, so all the old results have to be updated 😢

MatthewMiddlehurst:
also this could just not be the issue and its still broken, but this method may be better either way

TonyBagnall:
> also this could just not be the issue and its still broken, but this method may be better either way

I think it is better, I think the other way was to avoid seeding the shapelet in the same way, but it seems a bit mad to create a new one for every shapelet.

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -325,7 +329,7 @@ def _fit(self, X, y=None):
             for s in class_shapelets
             if s[0] > 0
         ]
-        self.shapelets.sort(reverse=True, key=lambda s: (s[0], s[1], s[2], s[3], s[4]))
+        self.shapelets.sort(reverse=True, key=lambda s: (s[0], -s[1], s[2], s[3], s[4]))
```
TonyBagnall, line 332: this changes ST to favour short shapelets when IG is tied, which is sensible (adding to remind me!)

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -437,7 +438,7 @@ def _extract_random_shapelet(self, X, y, i, shapelets, max_shapelets_per_class):
             worst_quality,
         )
 
-        return quality, length, position, dim, inst_idx, cls_idx
+        return np.round(quality, 8), length, position, dim, inst_idx, cls_idx
```
TonyBagnall, line 441: this may well have been cause of the bug

TonyBagnall (approved):

TonyBagnall:
one more repeat for good luck

PR #762, opened 2023-09-24
[MNT] Kmeans refactor

chrisholder:

This PR refactors The TimeSeriesKmeans clusterer to be consistent with TimeSeriesKMedoids. In addition it removes the complexity with inheritance. Finally is gets rid of the 'metric' directory in the clustering directory.

I have adapted the same tests to work with the new EmptyClusterError. This version is equivalent to the previous version (when the new error isn't raised). 
I have additionally added new tests for initilisation.

#### Any other comments?

Some parameters have been changed in this PR:

metric -> distance for TimeSeriesKmeans class
metric -> distance for elastic_barycenter_average function
the string 'dba' for the averaging string call has been removed and now uses 'ba'

In addition a new error has been introduced: EmptyClusterError. This error occurs when an empty cluster is found. In the new version it will end that iteration rather than trying to continue on.

I did a speed comparison between the old and new and it is essentially the same, but the new code is much more maintainable and readable.

##### For all contributions


TonyBagnall (approved):
there is a lot here, the cosmetic changes such as refactoring metric are fine, but there are some changes I cant easily assess. We trust you though Chris!

PR #786, opened 2023-09-30
[BUG] TDE and boxplot bugfixes

MatthewMiddlehurst:
A few bugfixes and actions maintenance to help with bugfixing:

- Uses the `_typed_dict` method from SFA for handling disabled numba in TDE also
- Changes the parameter used in the boxplot function from `lw` to `linewidth`
- Introduces labels to run the full test suite in PRs and the code coverage test
- Removes the `test_typed_dict_disabled_numba` test from #753. I think this is sufficiently handled with the coverage/periodic actions (which include both with and without numba) and the current SFA test functions (which test both True and False for the typed dict). This is more me not trusting Python with handling the setting of environment variables while pytest is doing threading stuff than disliking the test 🙂 

@TonyBagnall @patrickzib feel free to make suggestions on the bugfixes and test removal. I can add the test back if we just want the bugfixes.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2189,7 +2189,7 @@
       "contributions": [
         "code"
       ]
-    },
+    }
   ],
   "commitType": "docs"
 }
--- a/.github/workflows/pr_pytest.yml
+++ b/.github/workflows/pr_pytest.yml
@@ -17,8 +17,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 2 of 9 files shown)


TonyBagnall:
it passes all tests, so seems to have fixed the boxplot bug, best way to test the non numba tests is put it in and run the overnighters

TonyBagnall (dismissed):
the typed dict solution looks quite brittle and hard to maintain, perhaps revisit at a later date, it would be good to get the overnighters working.

TonyBagnall (approved):
lgtm

PR #801, opened 2023-10-07
[MNT] Comment out `mypy` pre-commit hook and run pre-commit on all files. Remove individual copyright notices

MatthewMiddlehurst:
This PR comments out the `mypy` hook. It may be a little too disruptive currently, my bad for not properly checking.

Runs pre-commit on all files. The majority of this is just removing `# -*- coding: utf-8 -*-` from everything.

Removes individual copyright notices at the top of files. This is currently very inconsistent in the actual message displayed and which files they are present in. I am dubious of the actual usefulness of such a notice. If we want to reintroduce this, we should do it properly.


TonyBagnall:
692 files!

MatthewMiddlehurst:
The remaining pre-commit failures and file name and documentation related. Would rather keep this PR simpler given the massive amount of file changes and fix those elsewhere.

TonyBagnall (approved):

PR #709, opened 2023-08-29
[ENH] BaseCollectionTransformer inherits from BaseCollectionEstimator and BaseTransformer

TonyBagnall:
#### Reference Issues/PRs
dependent on #650 
closes #100, #149 
See #42 #166, #267

#### What does this implement/fix? Explain your changes.

this brings BaseCollectionTransformer into the same structure as BaseClassifier, BaseRegressor, BaseClusterer in making it a BaseCollectionEstimator. All this really does is apply the same tags and conversion code to these related Transformers. It removes a lot of complex code from base transformer, facilitates unequal length using the same mechanism as the others and removes multiple calls to the datatypes modules which repeatedly scan the data for no reason. 

Note this still inherits from BaseTransformer, but whereas before a few lines of conversion code were used, now uses none of its functionality. Its still possible to convert with the existing wrappers.

This is a necessary condition for tidying up the shapelets and rocket transformers. 

Anomalies caused by lack of tags and mismatch in names. Added capabilities:multivariate tag where appropriate, will go through all the huge number of BaseTransformer tags and see which are needed by collections at a later date

#### Does your contribution introduce a new dependency? If yes, which one?

No

#### Did you add any tests for the change?

No test added, but one changed. The test 

```python
@pytest.mark.parametrize("dtype", ["pd.Series"])
def test_collection_transformer_invalid_input(dtype):
```
was altered to remover the pd.DataFrame option. This is because pd.DataFrame of floats resolves to pd-wide format for collections, which has always been allowed. I am open to changing this, but that is a more wide reaching change.

PR #804, opened 2023-10-09
[BUG] Collection transformer wrapper bugfixing and `PlateauFinder` refactor

MatthewMiddlehurst:
#709 made some changes to the collection wrapper due to test failures. Fixes those.

Moves `PlateauFinder` to its own file.


aeon/transformations/segment.py
```diff
@@ -0,0 +1,91 @@
+"""Transformations for segmenting time series."""
```
TonyBagnall, line 1: looks like it can move to segmentation to me (in a different PR)

TonyBagnall (dismissed):
We can deprecate and move PlateauFinder once BaseSegmenter is in

MatthewMiddlehurst:
Agreed, reverted the changes so this is just the wrapper.

aeon/transformations/collection/_collection_wrapper.py
```diff
@@ -47,7 +50,7 @@ class CollectionToSeriesWrapper(BaseTransformer):
     def __init__(
         self,
         transformer,
-        _output_convert=None,
+        _output_convert="auto",
```
TonyBagnall, line 53: I think its better not to do the auto conversion myself. I think at least it should be a parameter so the user can turn it off, you can default to auto if you want

TonyBagnall (approved):

PR #820, opened 2023-10-12
[ENH] Adapt ShapeletTransform to work with unequal length

TonyBagnall:
this adapts the RandomShapeletTransform and the ShapeletTransformClassifier to accept unequal length series through np-list. This is simply done by avoiding calls to shape and using len instead. There is however a numba issue with using "reflective" lists. It is warning about deprecation and suggesting using using TypedList. However, there were issues with this when we turn off numba for overnight tests, I would appreciate any advice on how to handle this.


baraline:
> this adapts the RandomShapeletTransform and the ShapeletTransformClassifier to accept unequal length series through np-list. This is simply done by avoiding calls to shape and using len instead. There is however a numba issue with using "reflective" lists. It is warning about deprecation and suggesting using using TypedList. However, there were issues with this when we turn off numba for overnight tests, I would appreciate any advice on how to handle this.

One workaround to work with uneven length series in numba is to create a numpy array of the maximum length within the inputs, and then have another array of integers which store the true length of the series. Then, when making operations on a sample, you simply call `X[i_sample, :, :length[i_sample]]`. It is a bit of a ductape fix, but it should work for the overnight tests also.

TonyBagnall:
> > this adapts the RandomShapeletTransform and the ShapeletTransformClassifier to accept unequal length series through np-list. This is simply done by avoiding calls to shape and using len instead. There is however a numba issue with using "reflective" lists. It is warning about deprecation and suggesting using using TypedList. However, there were issues with this when we turn off numba for overnight tests, I would appreciate any advice on how to handle this.
> 
> One workaround to work with uneven length series in numba is to create a numpy array `X` of the maximum length within the inputs, and then have another array of integers `length` which store the true length of the series. Then, when making operations on a sample, you simply call `X[i_sample, :, :length[i_sample]]`. It is a bit of a duct tape fix, but it should work for the overnight tests also.

thats quite involved and will require either a definition of a bespoke ArrayList like data structure or a lot of internal changes. The list solution here requires three changes, and I doubt there is much performance hit, I can experiment to see. 

This solution will be fine with the overnights, but might start failing with numba updates ate some point, so may need to resolve the typed list through some ugly switching

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -355,6 +365,14 @@ def _transform(self, X, y=None):
         """
         output = np.zeros((len(X), len(self.shapelets)))
 
+        for i in range(0, len(X)):
+            if X[i].shape[1] < self.min_series_length_:
```
baraline, line 369: Shouldn't this be `if X[i].shape[1] < self.min_shapelet_length_:` rather than `self.min_series_length_` ?

baraline (commented):
Other than the question on the check in the transform, looks good to me !

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: Actually needs to be less than max shapelet length I think

baraline: True, otherwise we can't use all shapelets, my bad !

baraline (approved):
The check for variable length and the tests are good now !

MatthewMiddlehurst (approved):

PR #825, opened 2023-10-13
[ENH] Introduce the "Tabular" output type tag and convert remaining Transformers to CollectionTransformers

TonyBagnall:
#### Reference Issues/PRs

part of #42 
merged with #721 

Updates collection transformers to use BaseCollection (#721) and updates the tags. 

recently refactored input and output type tags for Transformers to be ""input_data_type" and ""output_data_type". Now looking at their values. Currently they are

```python
    (
        "input_data_type",
        "transformer",
        ("list", ["Series", "Panel", "Collection"]),
        "The abstract data type of the transformer input X",
    ),
    (
        "output_data_type",
        "transformer",
        ("list", ["Series", "Primitives", "Panel"]),
        "The input abstract data type of the transformer output, the transformed X",
    ),
```

The primary role of this PR is to stop using "Primitives". It is meant to indicate tabular output suitable for sklearn (rows are instances, columns are attributes where order of attributes has no significance for learning). It is replaced with Tabular (with Primitives left for legacy for now). Also used the introduced "Collection" tag in collection transformers. I wonder why these tags are all upper case though? Any thoughts on refactoring to lower? Would have to do them all at once otherwise it would look insane

```python 
     (
        "input_data_type",
        "transformer",
        ("list", ["Series", "Collection", "Panel"]),
        "The input abstract data type of the transformer, input X. Series "
        "indicates a single series input, Collection indicates a collection of "
        "time series. Panel is a legacy term and equivalent to Collection.",
    ),
    (
        "output_data_type",
        "transformer",
        ("list", ["Tabular", "Series", "Collection", "Primitives", "Panel"]),
        "The output abstract data type of the transformer output, the transformed X. "
        "Tabular indicates 2D output where rows are cases and unordered attributes are "
        "columns. Series indicates a single series output and collection indicates "
        "output is a collection of time series.  Primitives is a legacy term for "
        "Tabular and Panel for Collection.",
    ),
```


aeon/transformations/collection/convolution_based/_multirocket.py
```diff
@@ -139,7 +138,7 @@ def _transform(self, X, y=None):
         pandas DataFrame, transformed features
         """
         X = X.astype(np.float64)
-        X = convert(X, from_type="numpy3D", to_type="numpyflat", as_scitype="Panel")
```
MatthewMiddlehurst, line 142: Is this the current functionality? If so ok. Thought that was changed, but maybe just for collections.

TonyBagnall: yes, convert broadcasts multivariate series onto 2D by concatenating channels. The new converters no longer do that, since we thought it a bit weird. 
``python
def from_numpy3d_to_numpyflat(obj, store=None):
    if not isinstance(obj, np.ndarray) or len(obj.shape) != 3:
        raise TypeError("obj must be a 3D numpy.ndarray")

    shape = obj.shape

    # store second dimension shape/length if we want to restore
    if isinstance(store, dict):
        store["numpy_second_dim"] = shape[1]

    obj_in_2D = obj.reshape(shape[0], shape[1] * shape[2])

    return obj_in_2D
```

MatthewMiddlehurst (approved):

PR #830, opened 2023-10-13
[MNT] Add new periodic tests and tidy workflows

MatthewMiddlehurst:
Adds a `check-manifest` test to the nightly tests. This will check if the manifest is up-to-date rather than doing it right before a release. The MANIFEST.in file determines which files are included when we make a release.

Adds a full `pre-commit` run. This should be fast, and will catch any breakages due to updates or similar.

Enabled the use of the `full pre-commit` label to run the pre-commit on all files instead of changed ones in a PR.

### DIFF
--- a/.github/workflows/periodic_tests.yml
+++ b/.github/workflows/periodic_tests.yml
@@ -10,6 +10,38 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check-manifest:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3
+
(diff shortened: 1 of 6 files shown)


TonyBagnall:
there are a few pre-commit fails I couldn't work out, will this catch them/fail because of them?

MatthewMiddlehurst:
Yes, it would break if these are currently failures. That sort of intentional though.

Also we may have to put down my boy sweep-ai if he does that again. Will let the devs know

TonyBagnall (approved):

PR #832, opened 2023-10-13
[MNT] Add auto label and welcoming comment workflow for PRs 

MatthewMiddlehurst:
This PR adds a workflow and scripts to automatically label PRs based on the title and contents changed. This is similar to how `scikit-learn` does some of their labelling. 

[MNT] in the title should add the maintenance labels, [MNT,GOV] maintenance and governance lablels, etc.

Changes to the classification package should add classification, transformations the transformations label, examples the examples label etc. If this would add more than 3 labels, none are added to reduce bloat.

The actions will (should) post the changes made and a welcoming comment on the PR when it is opened.

I have made two uncontroversial (I hope) doc string changes to demonstrate this.

### DIFF
--- a/.github/workflows/pr_opened.yml
+++ b/.github/workflows/pr_opened.yml
@@ -1,10 +1,7 @@
 name: PR Opened
 on:
-  pull_request:
-    branches:
-      - main
-#  pull_request_target:
-#    types: [opened]
+  pull_request_target:
+    types: [opened]
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


MatthewMiddlehurst:
See above for the action taken for this PR. Removing less relevant labels now.

TonyBagnall (approved):

MatthewMiddlehurst:
I'm actually not quite sure if this will trigger correctly. Lets hope so though 🙂

PR #838, opened 2023-10-13
[ENH] change collection conversions from convert_to to convert_collection

TonyBagnall:
#### Reference Issues/PRs

part of #42

part of the policy of isolating the datatypes module, switch from the datatypes function convert_to to the convert_collection function in utils.


aeon/datasets/_dataframe_loaders.py
```diff
@@ -995,16 +994,4 @@ def load_tsf_to_dataframe(
...
-                loaded_data.index.nlevels == 2
-                and return_type not in MTYPE_LIST_HIERARCHICAL
-            ):
-                loaded_data = convert(
-                    loaded_data, from_type="pd-multiindex", to_type=return_type
-                )
-            else:
-                loaded_data = convert(
-                    loaded_data, from_type="pd_multiindex_hier", to_type=return_type
-                )
```
MatthewMiddlehurst, line 1008: Same as above

aeon/datasets/_data_loaders.py
```diff
@@ -27,6 +26,7 @@
     "load_regression",
     "download_all_regression",
 ]
+from aeon.utils.validation.collection import convert_collection
```
MatthewMiddlehurst, line 29: This should go with the other imports

aeon/datasets/_data_loaders.py
```diff
@@ -679,18 +663,6 @@ def load_tsf_to_dataframe(
...
-                loaded_data.index.nlevels == 2
-                and return_type not in MTYPE_LIST_HIERARCHICAL
-            ):
-                loaded_data = convert(
-                    loaded_data, from_type="pd-multiindex", to_type=return_type
-                )
-            else:
-                loaded_data = convert(
-                    loaded_data, from_type="pd_multiindex_hier", to_type=return_type
-                )
```
MatthewMiddlehurst, line 692: Why is this gone? Don't think we should touch unless we have a replacement.

aeon/utils/plotting.py
```diff
@@ -67,8 +66,6 @@ def plot_series(
         check_y(y, allow_index_names=True)
 
     series = list(series)
-    series = [convert_to(y, "pd.Series", "Series") for y in series]
```
MatthewMiddlehurst, line 70: Does this still work with non-series input? Not sure if this is tested.

aeon/utils/plotting.py
```diff
@@ -320,7 +317,6 @@ def plot_correlations(
     from statsmodels.graphics.tsaplots import plot_acf, plot_pacf
 
     series = check_y(series)
-    series = convert_to(series, "pd.Series", "Series")
```
MatthewMiddlehurst, line 323: Same as above

aeon/datasets/_data_loaders.py
TonyBagnall: it seems pointless to me, it is only there to support dask format which I dont think we really support any more, at least I dont. But fine if you want it in I'll put it back

aeon/datasets/_dataframe_loaders.py
TonyBagnall: reverted, its yours now :)

aeon/datasets/_data_loaders.py
MatthewMiddlehurst: it can convert to any type used in forecasting i thought?

aeon/utils/plotting.py
TonyBagnall: non-series input? The input is 
    series : pd.Series or iterable of pd.Series
        One or more time series.
so not sure why you need to convert it to pd.Series, its already a documented input. I'll just put it back if you want, its just a plotting util

MatthewMiddlehurst: non-pd.Series. If it states it only takes that then fine, but I dont really see the reason to remove (currently).

TonyBagnall: I've reverted it, changed because I am trying to reduce uses of datatypes module, but I have reverted it, its not a big thing. The whole plotting etc needs tidying up, its all over the place atm

aeon/datasets/_data_loaders.py
TonyBagnall: I dont know, I try avoid that code. I have reverted it anyway

MatthewMiddlehurst (approved):

PR #839, opened 2023-10-14
[ENH] Make MiniRocketMultivariateVariable conform to API

TonyBagnall:

this makes two changes to the MiniRocketMultivariateVariable transformer: 

1. it changes the internal type to np-list rather than df-list
2. it makes it return an array rather than a dataframe

I've checked and the results are identical pre and post change. The wider issue of what to do with the rocket transformers is deferred. I dont think we need this class and can adjust MiniRocket to work with uni/multi and unequal length, but thats a much bigger job. Maybe @angus924 has some thoughts?


hadifawaz1999 (approved):
LGTM!

PR #849, opened 2023-10-18
[MNT] Numba workflow caching

chrisholder:

This PR attempts to speed up the CI by caching the numba compiled functions. Currently a significant portion of the test run time is taken up with just compiling the numba functions. 

Before this PR the tests to run on all OS's around 25 minutes (with a fair amount of variance) to run (not including the installation of the dependencies). Now it takes around 12 minutes (with a low amount of vairance).

### How the caching works

To make this PR work I've had to 'monkey patch' numba. This monkey patch is only enabled if the environment variable CICD_RUNNING is set to 1 as seen in aeon/testing/test_config.py: 
```
if os.environ.get("CICD_RUNNING") == "1":
    import aeon.testing.utils._cicd_numba_caching  # noqa: F401
```

The monkey patch is needed due to the way in which numba determines if a cache is valid.

When determining if a cache is valid numba looks at 3 things:
1) Signature of the function (function name, parameters etc) is the same
2) If the hardware/OS is the same determined by looking at:
    - Operating system running
    - CPU name
    - CPU compile flags
3) The file meta data which includes
    - The time the file was last edited 
    - The number of lines in the file 

If all of these are true then a cache is considered 'valid'

While many of the numba cache'd files don't change between different PRs because a new CI environment is spun up everytime the tests are run this means that:
1) each files timestamp is changed to when the enviroment was created which is newer than the cache's date (as it is cloned in fresh).
2) the hardware and CPU compile flags changes between environments (as not all github CICD environments are the same).

As a result restoring the cache normally would mean the cache is always invalid as it failed step 2 and step 3 listed above.

To fix this I have wrote a 'monkey patch' that determines step 3 and set environment variables to ensure step 2 is valid. 

As I had to rewrite the logic of how a files is determined if it has been changed, I used git diff inside _cicd_numba_caching.py line 22
```
    result = subprocess.run(
        ["git", "diff", "--name-only", "origin/main"],
        check=True,
        capture_output=True,
        text=True,  # Makes the output text instead of bytes
    )
```
This checks if the file has changed. I do not check as fine grained as numba and simply say if the file has been changed all the numba functions inside the file are invalid. While this means some cache's will be invalidated when they don't need to be, it does mean there are no chances that a wrong cached version is used when it shouldn't be.

To fix step 2 I manually set the CPU features and CPU name inside the CICD environment variable for each run.

### How is a cache invalidated
A cache is ignored if any of the above are true. In addition the cache is recreated every day when the periodic tests are ran. This ensures that the cache has the most up to date version of the numba functions.

A cache can be seen in the /actions/cache page on github. The caches are labelled as shown:
```numba-pytest-Linux-3.10-23/02/2024```
The key denotes the test being run (pytest), the OS it is run on (Linux) the version of python (3.10) and the date created (23/02/2024). As the periodic tests run a 01:30 am this means there is a gap where there would be no cache. To fix this if the current date cache is not available then the previous dates cache is used i.e.:
```numba-pytest-Linux-3.10-22/02/2024```
If the previous dates cache is not available then no cache is used and the ci runs normally.

##### For all contributions
Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage)
bot to do this for you.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG] or [GOV] indicating
whether the PR topic is related to enhancement, maintenance, documentation, bugs or
governance.

- [x] Optionally, I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS)
to receive notifications about future changes to these files.

### CODE REVIEW

**Comment by chrisholder** [comment]
https://github.com/aeon-toolkit/aeon/actions/caches

File: .github/workflows/pr_examples.yml [hunk 1]
```diff
@@ -20,6 +20,11 @@ concurrency:
 jobs:
   run-notebook-examples:
     runs-on: ubuntu-22.04
+    env:
+      NUMBA_CACHE_DIR: ${{ github.workspace }}/.numba_cache
+      NUMBA_CPU_NAME: generic
+      NUMBA_CPU_FEATURES: +64bit,+adx,+aes,+avx,+avx2,-avx512bf16,-avx512bitalg,-avx512bw,-avx512cd,-avx512dq,-avx512er,-avx512f,-avx512ifma,-avx512pf,-avx512vbmi,-avx512vbmi2,-avx512vl,-avx512vnni,-avx512vpopcntdq,+bmi,+bmi2,-cldemote,-clflushopt,-clwb,-clzero,+cmov,+cx16,+cx8,-enqcmd,+f16c,+fma,-fma4,+fsgsbase,+fxsr,-gfni,+invpcid,-lwp,+lzcnt,+mmx,+movbe,-movdir64b,-movdiri,-mwaitx,+pclmul,-pconfig,-pku,+popcnt,-prefetchwt1,+prfchw,-ptwrite,-rdpid,+rdrnd,+rdseed,+rtm,+sahf,-sgx,-sha,-shstk,+sse,+sse2,+sse3,+sse4.1,+sse4.2,-sse4a,+ssse3,-tbm,-vaes,-vpclmulqdq,-waitpkg,-wbnoinvd,-xop,+xsave,-xsavec,+xsaveopt,-xsaves
```
MatthewMiddlehurst, line 26: Could you expand on what these bits are doing?

.github/workflows/pr_examples.yml
```diff
@@ -28,6 +33,23 @@ jobs:
         with:
           python-version: "3.10"
 
+      # Set CPU Features
+      - name: Set CPU Features
+        run: echo "NUMBA_CPU_FEATURES=${{ join(matrix.cpu-features, ' ') }}" >> $GITHUB_ENV
+        shell: bash
```
MatthewMiddlehurst, line 39: Same as above, is this overwriting the var above or just adding to it?

Also does this workflow have a matrix parameter? `join(matrix.cpu-features, ' ')`

.github/workflows/pr_examples.yml
```diff
@@ -28,6 +33,23 @@ jobs:
...
+
+      - name: Set daily date for cache invalidation
+        id: daily-date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d' --date=@$(( $(date +%s) / (3 * 24 * 60 * 60) * (3 * 24 * 60 * 60) )))"
+        shell: bash
```
MatthewMiddlehurst, line 44: This bit has a deprecation warning
image

aeon/testing/utils/_cicd_numba_caching.py
```diff
@@ -0,0 +1,79 @@
...
+    clean_files = []
+
+    for temp in files:
+        if temp.endswith(".py"):
+            clean_files.append((temp.split("/")[-1]).strip(".py"))
+
+    return clean_files
+
+
+CHANGED_FILES = get_invalid_numba_files()
```
MatthewMiddlehurst, line 42: I assume `custom_load_index` is called repeatedly and this saves time

.github/workflows/pr_pytest.yml
```diff
@@ -95,11 +174,9 @@ jobs:
         run: python -m pip list
 
       - name: Run tests
-        # run the full test suit if a PR has the 'full pytest actions' label
         run: python -m pytest --prtesting ${{ github.event_name != 'pull_request' || !contains(github.event.pull_request.labels.*.name, 'full pytest actions') }}
 
   codecov:
-    # run the code coverage job if a PR has the 'codecov actions' label
```
MatthewMiddlehurst, line 102: Any reason why or a mistake?

.github/workflows/pr_examples.yml
```diff
@@ -28,6 +33,23 @@ jobs:
...
+
+      # Python environment cache
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: .numba_cache
+          key: ${{ runner.os }}-pip-numba-3.10-examples-${{ hashFiles('pyproject.toml') }}-${{ steps.daily-date.outputs.date }}
```
MatthewMiddlehurst, line 51: The OS is variable but Python version is fixes to 3.10. How does it invalidate after 3 days using the date, does the key have specific formatting or is that done elsewhere.

MatthewMiddlehurst (commented):
Very nice stuff, I have some questions about each of the workflow steps.

.github/workflows/pr_examples.yml
chrisholder: So NUMBA_CACHE_DIR is just a environment variable you can set to say where you want your cache'd numba functions to be stored on your device. Between operating systems the structure of the container running the workflow can be different so I use github.workspace variable to point to the appropriate location on each. This is manually set so on each operating system so I know exactly where the cache'd functions are (and therefore can save/load previous cache's ).

NUMBA_CPU_NAME and NUMBA_CPU_FEATURES is the CPU type and features the numba compiler should optimise for. Normally numba auotmatically detects this. However, because each workflow environment can have different hardware this means that each environment could have a different CPU type/features. As a result if the name or any features are different the cache's are automatically invalidated. To stop the cache's always being invalid if the hardware changes I manually set all the features the CPU at minimum must have and in addition set the name to 'generic' so no specifc CPU optimsations are made by numba. While this does slow down the numba functions in theory because the tests are so small and fast to run these optimisations don't make any difference to the runtime.

.github/workflows/pr_examples.yml
chrisholder: Ah i forgot to remove this line I had this while running tests on which CPU features crashed the CI. Ill remove this.

.github/workflows/pr_examples.yml
chrisholder: Good spot Ill update this

.github/workflows/pr_pytest.yml
chrisholder: Mistakenly removed when merging

aeon/testing/utils/_cicd_numba_caching.py
chrisholder: So everytime a numba function is called it will use this function to check for a valid cache. This adds the custom logic to allow the cache to be restored in the workflow.

MatthewMiddlehurst: I meant `CHANGED_FILES = get_invalid_numba_files()` specifically, rather than being called in the function itself

.github/workflows/pr_examples.yml
chrisholder: This variable on the end here: ${{ steps.daily-date.outputs.date }}. In the previous step this is calculated:
```
      - name: Set daily date for cache invalidation
        id: daily-date
        run: echo "::set-output name=date::$(date +'%Y-%m-%d' --date=@$(( $(date +%s) / (3 * 24 * 60 * 60) * (3 * 24 * 60 * 60) )))"
        shell: bash
```

This will produce a unix date time stamp that will remain the same for 3 days. This means every 3 days the key will change. If the key changes then a cache is 'invalidated' because it is never called and a new one will be created for the next 3 days.

I did use chat gpt to generate this and I don't understand all of the details on how this works fully but it seems logical how it did it. This was the explanation:

```
The command date +%s gets the current date in seconds since the UNIX epoch.
Dividing this number by (3 * 24 * 60 * 60) converts the seconds into days and divides by 3 to group the days into 3-day periods. Multiplying back by (3 * 24 * 60 * 60) converts the period back into seconds, ensuring we're working with whole days.
The --date=@seconds option of the date command allows you to create a date from a given number of seconds since the UNIX epoch, effectively normalizing the date to the start of the current 3-day period.
```

aeon/testing/utils/_cicd_numba_caching.py
chrisholder: Ah this just produces a list of files that anything has been changed in. This only gets called once at the start and just stores the file names that have been changed in the PR. If a numba function exists in any of these files their cache is auto invalidated.

.github/workflows/pr_examples.yml
MatthewMiddlehurst: Ah ok so this should work for all the different OS/Python configs then?

.github/workflows/pr_examples.yml
MatthewMiddlehurst: This still seems to exist in some places, it was in all the workflows (unless it should be there for those).

.github/workflows/pr_examples.yml
MatthewMiddlehurst: So the cache is not valid for 3 days since creation, the key just changes every 3 days regardless or what is in there?

.github/workflows/pr_examples.yml
chrisholder: Yes should work on all OS's (maybe not the new mac M1 runner I haven't tested yet) and python versions

.github/workflows/pr_examples.yml
chrisholder: Yea I thought I could just remove it but I actually think I need this! Ill add a better comment on the line.

.github/workflows/pr_examples.yml
MatthewMiddlehurst: so essentially every 3 days someone gets the unlucky PR which takes 20 minutes longer? 🙂

MatthewMiddlehurst: `${{ runner.os }}-pip-numba-3.10-no-soft-deps-${{ hashFiles('pyproject.toml') }}-${{ steps.daily-date.outputs.date }}`

1. is it fine to share this between all Python versions? it is always 3.10. It changes for OS i see already.
2. no-soft-deps is used twice, but do they need to change for examples, no-soft-deps and regular testing either way? Will be the same functions surely
3. I think `pyproject.toml` version changes may be the wrong way to go about this. I think it would be better to check the `numba` and `numpy` version from the install somehow. The current method will rebuild for irrelevant changes and not catch patch version i.e. x.x.0 to x.x.1.

MatthewMiddlehurst: also i think when `pyproject.toml` has been changes it will still cache, but just use a different cache

chrisholder: Yes every 3 days someones PR will take the same amount of time the current CI takes.

I have just refactored how this date is set so we can easily adjust the value. I am happy to set this threshhold much higher (once a month or something). The reason I was thinking to set it a bit lower is because if someone makes a big PR that changes a huge amount of files, all the numba functions that were in these files will need to be recompiled every PR until there is a new cache created. 3 days just seemed like a good amount of time. 

In response to your second question
1. I actually don't know if the python version is relevant? I will try without the python version and see what happens.
2. Yes I have just seen this I thought they would compile the same functions (but they dont). They now have seperate names.
3. I agree. I orginally had the pyproject.toml in the key because I was going to cache all the dependencies (for some reason it takes longer to restore a cached dependency than just installing. This is something I want to investigate further in the future). I can remove the pyproject.toml key dependency.

In regards to the key name I am probably using the word 'invalidate' incorrect for the cache. The cache is not 'deleted' the key changes which just creates a new cache.

MatthewMiddlehurst: >  I actually don't know if the python version is relevant? I will try without the python version and see what happens.

Since previously it was always set to 3.10 it seems to have not caused a crash or anything, but I'm not sure if all of the cached functions would be valid.

> Yes I have just seen this I thought they would compile the same functions (but they dont). They now have seperate names.

They run different tests, but I wouldn't think that the numba functions compiled in test-examples would break in no-soft-deps or the general pytest ones. Does it not work if you just remove the names entirely?

> I agree. I orginally had the pyproject.toml in the key because I was going to cache all the dependencies (for some reason it takes longer to restore a cached dependency than just installing. This is something I want to investigate further in the future). I can remove the pyproject.toml key dependency.

Its probably better to not have it, but i think using `numba` and `numpy` versions to change cache would be safer maybe. Perhaps it does not matter if my above branch comment is true.

chrisholder: I wasn't aware that you couldn't share between branches. This is going to require me to trigger a action whenever a brach is merged into main. Im going to hop into a private repository to do this and test it. I will shoot you a ping once I have this all worked out.

MatthewMiddlehurst: It should be rerun on main when a PR is merged i think? Will leave it with you now. Feel free to send me a message if you want some help.

chrisholder:
I've updated the PR body above but just a summary of the changes I've made:

1) Cache's are now ONLY CREATED on periodic tests run. This means the caches are generated once a day and will be on main branch so every other branch can use it.
2) Cache keys now contain the date they were created. If the current dates cache is unavailable the previous days cache is used. If the previous and current date cache unavaliable no cache is used.
3) Retries were put in to the monkey patch as it was 'randomly' failing to 'git fetch'.

chrisholder:
The PR seems to be failing because it doesn't like the file path to the custom action? Not sure why it's failing here and not in the original PR. For now going to revert the original structure.

MatthewMiddlehurst (approved):
LGTM. Thanks for implementing my suggestions, lets give it a go and monitor.

.github/workflows/periodic_tests.yml
```diff
@@ -60,6 +67,13 @@ jobs:
         run: build_tools/run_examples.sh
         shell: bash
 
+      - uses: actions/cache/save@v3
+        id: save-cache
```
MatthewMiddlehurst, line 71: You gave these ids but they are never used elsewhere i think?

.github/workflows/pr_pytest.yml
```diff
@@ -24,9 +24,16 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-python@v5
+        id: setup-python
```
MatthewMiddlehurst, line 27: Same as above here

.github/workflows/pr_pytest.yml
```diff
@@ -78,12 +91,19 @@ jobs:
         python-version: [ "3.8", "3.10", "3.12" ]
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout main
```
MatthewMiddlehurst, line 94: Not against names but this is the only checkout that has one now 🙂

MatthewMiddlehurst (commented):
These are just very minor style things if you want to change them.

chrisholder:
I removed all the ids for consistency

chrisholder:
I added names to all actions (in addition added some to the other action files for consistency)

.github/workflows/periodic_tests.yml
```diff
@@ -138,13 +205,21 @@ jobs:
       - name: Run tests
         run: python -m pytest
 
+      - uses: actions/cache/save@v3
```
MatthewMiddlehurst, line 208: missing name here

.github/workflows/pr_precommit.yml
```diff
@@ -41,4 +44,5 @@ jobs:
           extra_args: --files ${{ steps.changed-files.outputs.all_changed_files }}
 
       - if: ${{ failure() && github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
+        name: Pre-commit failed
```
MatthewMiddlehurst, line 47: bit misleading, the step pushes pre-commit fixes

chrisholder: Good spot i'll update

.github/workflows/periodic_tests.yml
chrisholder: Knew i'd miss one! Ill add it great spot

MatthewMiddlehurst (approved):
thanks

PR #857, opened 2023-10-22
[ENH] Shapelet transform changes

TonyBagnall:
I inadvertently reverted the changes to STC from #820 when sorting out the precommit  errors. This puts them back in, see


baraline (approved):
Lgtm !

PR #884, opened 2023-11-01
[BUG] Ignore extra whitespace in loader

TonyBagnall:
currently, if there are extra spaces between tag values in the data file, it causes errors in the loading. This simply strips out any unnecessary white space. 

Also removed a dataset list which is not correct with data on tsc.com


dguijo (commented):
Thanks Tony, for your work. 

I think it was me who added `whole_set`. I couldn't find a way to collate the 142 datasets used in the bakeoff redux. As you well know, they are the 112 univariate datasets without missing and unequal + 30 multivariate datasets of the 33. Maybe we can create a new list with the 30 datasets that I suppose are equal-length and non-missing. Am I correct? 

Thank you Tony!

TonyBagnall:
> Thanks Tony, for your work.
> 
> I think it was me who added `whole_set`. I couldn't find a way to collate the 142 datasets used in the bakeoff redux. As you well know, they are the 112 univariate datasets without missing and unequal + 30 multivariate datasets of the 33. Maybe we can create a new list with the 30 datasets that I suppose are equal-length and non-missing. Am I correct?
> 
> Thank you Tony!

whole_set contains datasets that are not yet explicitly on TSC such as AsphaltPavementTypeUniEq so it fails the loading. I think it better to keep lists like this in your own branch, and only include lists where we can download the data directly from one of the websites

dguijo:
> > Thanks Tony, for your work.
> > I think it was me who added `whole_set`. I couldn't find a way to collate the 142 datasets used in the bakeoff redux. As you well know, they are the 112 univariate datasets without missing and unequal + 30 multivariate datasets of the 33. Maybe we can create a new list with the 30 datasets that I suppose are equal-length and non-missing. Am I correct?
> > Thank you Tony!
> 
> whole_set contains datasets that are not yet explicitly on TSC such as AsphaltPavementTypeUniEq so it fails the loading. I think it better to keep lists like this in your own branch, and only include lists where we can download the data directly from one of the websites

Great, I see what you mean. Thanks Tony!

dguijo (approved):

MatthewMiddlehurst (approved):

Issue #590, opened 2023-07-21
[BUG] there is no data folder with data , but it is used in code


Sandy4321:
### Describe the bug

there is no data folder with data , but it is used in code 
https://github.com/aeon-toolkit/aeon/blob/main/examples/datasets/data_files_structures.ipynb

This can be demonstrated using the Arrow Head problem that is included in aeon under aeon/datasets/data

![image](https://github.com/aeon-toolkit/aeon/assets/11426119/5aa06060-f04c-40ed-bb85-175bad71758d)

### Steps/Code to reproduce the bug

This can be demonstrated using the Arrow Head problem that is included in aeon under aeon/datasets/data

### Expected results

data folder with data , but it is used in code

by the way do you have code for reading data from web 
if yes 
what data sets are supported 

### Actual results

This can be demonstrated using the Arrow Head problem that is included in aeon under aeon/datasets/data

TonyBagnall:
hi, the datasets provided are all in the directory aeon/datasets/data

![image](https://github.com/aeon-toolkit/aeon/assets/9594042/346a0913-0f84-4d08-bf05-992aaca9f64b)

yes you can download directly from our website timeseriesclassification.com with something like this

```python
from aeon.datasets import load_classification
X, y, meta_data = load_classification("Yoga")
print(" Shape of X = ", X.shape)
print(" Meta data = ", meta_data)
```
lists of datsets are in aeon/datasets/tcs_data_lists.py (tser for regression, tsf for forecasting). This loads the UCR/UEA/Monash archives of data

TonyBagnall:
hi @Sandy4321 are  you still having problems with this? If so, happy to help, but if not, could you close the issue? thanks

Sandy4321:
according to 
Time Series (Extrinsic) Regression
https://github.com/aeon-toolkit/aeon/blob/main/examples/datasets/load_data_from_web.ipynb

I try to download by code
import pickle

from aeon.datasets import  load_regression 
from aeon.datasets.dataset_collections import list_available_tser_datasets

print(list_available_tser_datasets())
print('number of data sets ', len(   list_available_tser_datasets()  ))

for data_set_name in  list_available_tser_datasets():
    print('\nstart download', data_set_name)
    X, y,  metadata = load_regression(data_set_name)
    with open(data_set_name + '.pickle', 'wb') as handle:           
        pickle.dump([X , y,  metadata], handle)        
    print(X.shape)
    print('\n finish download', data_set_name)

[...]

PR #887, opened 2023-11-02
[ENH] Add timestamped data loading for collections

### LINKED ISSUE #479: [BUG] Data loading ts data with timestamps
### Describe the bug

the data loaders for ts files dont work with time stamps. The classification archive has no datasets with this characteristic, so never been an issue. However, some of the monash regression problems have time stamps, and these crash the loader.

Solution can be adapted from here 
https://github.com/ChangWeiTan/TS-Extrinsic-Regression/blob/master/utils/data_loader.py#L36

this only works once #478 is in

```python
from aeon.datasets import load_regression
X, y, meta = load_regression("AppliancesEnergy")
print(" Type of X = ", type(X))
print("Meta = ", meta)

```

### Expected results

data loads into a numpy array with correct meta data

### Actual results

```python-traceback
  File "C:\Code\aeon\aeon\local\debug.py", line 21, in <module>
    X, y, meta = load_regression("AppliancesEnergy")
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 1199, in load_regression
    return _load_saved_dataset(
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 301, in _load_saved_dataset
    X_train, y_train, meta_data = load_from_tsfile(abspath, return_meta_data=True)
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 228, in load_from_tsfile
    data, y, meta_data = _load_data(file, meta_data)
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 166, in _load_data
    data_series = [float(x) for x in data_series]
  File "C:\Code\aeon\aeon\datasets\_data_loaders.py", line 166, in <listcomp>
    data_series = [float(x) for x in data_series]
ValueError: could not convert string to float: '(2016-02-28 17'
```

### Versions

_No response_

TonyBagnall:
fixes #479 
related to #590 

this allows for the loading of timestamped classification and regression problems. Disclaimer, we never use timestamped data, so this solution is just a patch really, but better than not dealing with it at all, given several [tser](http://tseregression.org/) have timestamps. Following caveats

1. The time stamps are not stored, but simply stripped out. Currently, all our timestamped data at [tser](http://tseregression.org/) are equally spaced, so we can strip out and store in numpy. We can adapt later, but timestamps means pd.Series, which complicates things.
2. The dealing of timestamps is fragile: it uses a simple regular expression to strip out the timestamp and recover the data, which must be in double format. Surely could be done better, but works existing data.

I have also removed PGDalia from the list tser_all: this dataset has different length series for channels in the same time series, and we do not support this format.


aeon/datasets/_data_loaders.py
```diff
@@ -133,13 +147,19 @@ def _load_data(file, meta_data, replace_missing_vals_with="NaN"):
...
         line = line.replace("?", replace_missing_vals_with)
-        channels = line.split(":")
+        if meta_data["timestamps"]:
+            channels = _get_channel_strings(line, target, replace_missing_vals_with)
+        else:
+            channels = line.split(":")
         n_cases += 1
         current_channels = len(channels)
-        if meta_data["classlabel"] or meta_data["targetlabel"]:
+        if target:
```
MatthewMiddlehurst, line 162: There is no target=False so it just would not exist if there is no label I think?

aeon/datasets/_single_problem_loaders.py
```diff
@@ -1,5 +1,19 @@
...
+import numpy as np
+import pandas as pd
+
+from aeon.datasets._data_loaders import _load_saved_dataset, _load_tsc_dataset
+from aeon.datasets._dataframe_loaders import load_tsf_to_dataframe
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+DIRNAME = "data"
+MODULE = os.path.dirname(__file__)
+
```
MatthewMiddlehurst, line 16: I think the current convention is that these go under the author list

aeon/datasets/tests/test_data_loaders.py
```diff
@@ -208,6 +220,10 @@ def test_load_provided_dataset(return_X_y, return_type):
     # Check whether object is same mtype or not, via bool
 
 
+@pytest.mark.skipif(
+    PR_TESTING,
+    reason="Only run on overnights because of intermittent fail for read/write",
+)
```
MatthewMiddlehurst, line 226: Is this needed for every test? I thought the only failures were loading from webpages

TonyBagnall: I think so, reading and writing have sometimes cause intermittent fails in the past and I think overnight testing is sufficient. Although oc I dont really mind!

aeon/datasets/_single_problem_loaders.py
TonyBagnall: the constants DIRNAME/MODULE?

aeon/datasets/_data_loaders.py
TonyBagnall: I changed this because I was going to use it elsewhere then didnt. I'll revert.

aeon/datasets/_data_loaders.py
```diff
@@ -133,10 +147,17 @@ def _load_data(file, meta_data, replace_missing_vals_with="NaN"):
...
+        line = line.replace("nan", replace_missing_vals_with)
         line = line.replace("?", replace_missing_vals_with)
-        channels = line.split(":")
+        if meta_data["timestamps"]:
+            channels = _get_channel_strings(line, target, replace_missing_vals_with)
+        else:
+            channels = line.split(":")
         n_cases += 1
         current_channels = len(channels)
         if meta_data["classlabel"] or meta_data["targetlabel"]:
```
MatthewMiddlehurst, line 163: line 142 can also be target here i think

MatthewMiddlehurst (approved):

PR #899, opened 2023-11-09
[ENH] remove unnecessary conversion calls in plotting

TonyBagnall:
in the function utils/plotting we have public methods:

plot_series
plotr_interval
plot_lags
plot_correlations
plot_windows

where there are doc strings, they all type the series input to pd.Series. Internally, they then call complex functions to convert the series to pd.Series. This is unnecessary and adds dependencies on data types code which we are trying to isolate.


TonyBagnall:
ah didnt allow for pandas input instead of series, not as easy a win as I thought

PR #890, opened 2023-11-04
[ENH] Results loaders for published results

TonyBagnall:
in order to clear up a long standing ambiguity, this PR hard codes loading published results, developed in conjunction with 
https://timeseriesclassification.com/results.php

also fixes #905 
this is bespoke code, intentionally. These legacy results need to be captured, and I think the best way is to do so explicitly, rather than have general loaders, which are there for dynamically evolving results. 

there is some extra testing, but the basic addition is the function get_bake_off_2017_results, get_bake_off_2021_results and get_bake_off_2023_results, which loads the original bake off results, and the addition of a list for the 2015 dataset list of 85 datasets. Not strictly necessary, but useful think.

Added a notebook to demonstrate all this and excluded them from testing because of dependency on tsc.com

PR #902, opened 2023-11-13
CD plot function has been simplified, fixing some errors. 

dguijo:

A few errors have been detected in the way the cliques are built. Hence, we have modified it and simplified it to be in accordance to the literature.

Now, CDs show wilcoxon signed ranks using one-sided tests.
 

Nope

Thanks!


TonyBagnall:
I will add a notebook to reproduce existing, might merge #890 into this? @dguijo what do you think? Can then review once we can recreate

dguijo:
> I will add a notebook to reproduce existing, might merge #890 into this? @dguijo what do you think? Can then review once we can recreate

Yes, might be better.

TonyBagnall (approved):
Core logic from @dguijo is now correct, my contributions were purely cosmetic

PR #915, opened 2023-11-21
[BUG] shapelet transform square root

### LINKED ISSUE #862: [BUG] Inconsistency in online_shapelet_distance
### Describe the bug

raised in the old place, this seems like a genuine bug that will alter results. Thank you @margauxgeuzaine by proxy

https://github.com/sktime/sktime/issues/5480

"There is an inconsistency in sktime/transformations/panel/_shapelet_transform_numba.py, between line 24 where "std = (sum2 - mean * mean * length) / length" and line 56 where "std = math.sqrt((sums2[n] - mean * mean * length) / length)".

Because of that, some time series are treated differently than others regarding normalization. The former are actually divided by the variance, and the latter are divided by the standard deviation. The second way of doing is supposed to be the right one."

_No response_

### Expected results

add sqrt to stdv calculation, will alter reference results and will need to assess if it makes any sig difference overall *may make it better :)

### Actual results

doesnt take sqrt

### Versions

_No response_

TonyBagnall:

Fixes #862 

#### Any other comments?

Added a sqrt to STC and updated the reference results. Ran both with an without, and there is no significant difference in performance. Average accuracy on main is 84.51, with sqrt its 84.30.

image

Also removed a duplicate parameter, had both of these
self.n_shapelet_samples = n_shapelet_samples
self._n_shapelet_samples = n_shapelet_samples

### DIFF
--- a/aeon/benchmarking/_critical_difference.py
+++ b/aeon/benchmarking/_critical_difference.py
@@ -290,7 +290,7 @@ def plot_critical_difference(
     >>> from aeon.benchmarking import plot_critical_difference
     >>> from aeon.benchmarking.results_loaders import get_estimator_results_as_array
     >>> methods = ["IT", "WEASEL-Dilation", "HIVECOTE2", "FreshPRINCE"]
-    >>> results = get_estimator_results_as_array(estimators=methods)
+    >>> results = get_estimator_results_as_array(estimators=methods) # doctest: +SKIP
     >>> plot = plot_critical_difference(results[0], methods, alpha=0.1)\
         # doctest: +SKIP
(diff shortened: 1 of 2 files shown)


MatthewMiddlehurst (approved):
Slightly surprising it makes such a difference, but better to have it correct either way.

TonyBagnall:
> Slightly surprising it makes such a difference, but better to have it correct either way.

I'm running it over 30 resamples, as I was also a bit surprised.

TonyBagnall:
benefit of doing resamples: 
image
no difference between them

PR #928, opened 2023-11-27
[BUG] Check if multiindex store unequal length series

### LINKED ISSUE #816: [BUG] Unequal length multiindex not correctly identified in converters
### Describe the bug

the function

utils.validation.convert_collection._equal_length

tests whether a collection is equal length to properly manage conversions. 
```python

def _equal_length(X, input_type):
    """Test if X contains equal length time series.

    Assumes input_type is a valid type (COLLECTIONS_DATA_TYPES).

    Parameters
    ----------
    X : data structure.
    input_type : string, one of COLLECTIONS_DATA_TYPES

    Returns
    -------
    boolean: True if all series in X are equal length, False otherwise

    Raises
    ------
    ValueError if input_type not in COLLECTIONS_DATA_TYPES.

    Example
    -------
    >>> _equal_length( np.zeros(shape=(10, 3, 20)), "numpy3D")
    True
    """
    always_equal = {"numpy3D", "numpyflat", "pd-wide"}
    if input_type in always_equal:
        return True
    # np-list are shape (n_channels, n_timepoints)
    if input_type == "np-list":
        first = X[0].shape[1]
        for i in range(1, len(X)):
            if X[i].shape[1] != first:
                return False
        return True
    # df-list are shape (n_timepoints, n_channels)
    if input_type == "df-list":
        first = X[0].shape[0]
        for i in range(1, len(X)):
            if X[i].shape[0] != first:
                return False
        return True
    if input_type == "nested_univ":  # Nested univariate or hierachical
        return _nested_univ_is_equal(X)
    if input_type == "pd-multiindex":
        # TEMPORARY: WORK OUT HOW TO TEST IF pd.MULTINDEX are equal length
        return True
    raise ValueError(f" unknown input type {input_type}")
    return False
```

Need to figure out how to test this for pd.MultiIndex 

_No response_

### Expected results

test should return false if unequal

### Actual results

alkways returns true

### Versions

_No response_

TonyBagnall:
#### Reference Issues/PRs

Fixes #816

Test to see if "pd.multiindex" format is storing unequal length series. pd.multiindex actually just add NaN for unequal length, so it was a toss up between enforcing equal length on pd.multiindex or testing for nans. You can do this

```python
import pandas as pd
cols = ["instances", "timepoints"] + [f"var_{i}" for i in range(2)]
Xlist = [
    pd.DataFrame([[0, 0, 1, 4], [0, 1, 2], [0, 2, 3, 6]], columns=cols),
    pd.DataFrame([[1, 0, 1, 4], [1, 1, 2], [1, 2, 3, 6]], columns=cols),
    pd.DataFrame([[2, 0, 1, 42], [2, 1, 2], [2, 2, 3, 6]], columns=cols),
]
X = pd.concat(Xlist)
X = X.set_index(["instances", "timepoints"])
from aeon.utils.validation._convert_collection import _equal_length
print(X)
```
which stores as 
```python
0         0               1    4.0
          1               2    NaN
          2               3    6.0
1         0               1    4.0
          1               2    NaN
          2               3    6.0
2         0               1   42.0
          1               2    NaN
          2               3    6.0
```
There is an edge case where we would call a collection with missing in the middle unequal, but its obscure, no one really is going to store collections in such an inefficient data structure, and this is better than the current solution, which is to do nothing.


hadifawaz1999 (approved):
lgtm

PR #933, opened 2023-11-27
[ENH] Remove the TimeBinner transformer

### LINKED ISSUE #898: [ENH] Remove the transformer TimeBinner

TimeBinner assumes the input is in nested_univ format, which is no longer supported. The Transformer  is not used anywhere in aeon, and if the functionality is required, it would be better done with numpy3D,. Conversion is non trivial though, because it uses the pandas datastructure pd.IntervalIndex. I dont know the original contributor.

```python
class TimeBinner(BaseCollectionTransformer):
    """
    Turns time series collections data into tabular data based on intervals.

    This estimator converts nested pandas dataframe containing
    time-series/panel data with numpy arrays or pandas Series in
    dataframe cells into a tabular numpy array.
    The primitives are calculated based on Intervals defined
    by the IntervalIndex and aggregated by aggfunc.

    This is useful for transforming time-series/panel data
    into a format that is accepted by standard validation learning
    algorithms (as in sklearn).

    Parameters
    ----------
    idx : pd.IntervalIndex
        IntervalIndex defining intervals considered by aggfunc.
    aggfunc : callable
        Function used to aggregate the values in intervals.
        Should have signature 1D -> float and defaults
        to mean if None.
    """

    _tags = {
        "fit_is_empty": True,
        "output_data_type": "Tabular",
        "instancewise": True,
        "X_inner_type": ["nested_univ"],
        "y_inner_type": "None",
    }

    def __init__(self, idx, aggfunc=None):
        assert isinstance(
            idx, pd.IntervalIndex
        ), "idx should be of type pd.IntervalIndex"
        self.aggfunc = aggfunc
        if self.aggfunc is None:
            self._aggfunc = np.mean
            warnings.warn("No aggfunc was passed, defaulting to mean", stacklevel=2)
        else:
            assert callable(aggfunc), (
                "aggfunc should be callable with" "signature 1D -> float"
            )
            if aggfunc.__name__ == "<lambda>":
                warnings.warn(
                    "Save and load will not work with lambda functions", stacklevel=2
                )
            self._aggfunc = self.aggfunc
        self.idx = idx

        super(TimeBinner, self).__init__()

    def _transform(self, X, y=None):
        """Transform X and return a transformed version.

        private _transform containing core logic, called from transform

        Parameters
        ----------
        X : Series or Panel of mtype X_inner_type
            if X_inner_type is list, _transform must support all types in it
            Data to be transformed
        y : Series or Panel of mtype y_inner_type, default=None
            Additional data, e.g., labels for transformation

        Returns
        -------
        transformed version of X
        """
        idx = pd.cut(X.iloc[0, 0].index, bins=self.idx, include_lowest=True)
        Xt = X.applymap(lambda x: x.groupby(idx).apply(self._aggfunc))
        Xt = convert_to(Xt, to_type="numpyflat", as_scitype="Panel")
        return Xt
```

convert to using numpy3D

_No response_

### Additional context

_No response_

TonyBagnall:
Fixes #898 

This PR removes the TimeBinner collection transformer, see #898 for reasons.

To recap: this is the last estimator to use the nested_univ internal type, which we have deprecated. It is not used elsewhere in the package and would more naturally be a series transformer with a pd.Series, which then could be applied over all instances in the style of those transformers. I would be happy to help with that if ther e is any demand for this transformer


TonyBagnall:
another reason to remove this is that #941 has highlighted that this class flattens multivariate with the implicit conversion routine we have scrapped where "numpyflat" means "concatenate all channels into one channel". Tabularizer should be used instead, or just reshape it!

MatthewMiddlehurst (approved):

MatthewMiddlehurst (approved):

PR #936, opened 2023-11-27
[ENH] base deep classifier one hot encoder

TonyBagnall:
OneHotEncoder is generating a deprecation warning from sklearn : the parameter sparse is becoming sparse_output. This pr tests the sckit version and passes sparse_output as an argument. It also removes the argument auto, since that is the default value anyway.


TonyBagnall:
subsumed into #937

PR #941, opened 2023-11-28
[ENH] Refactor numpyflat to be numpy2D

### LINKED ISSUE #815: [ENH] Refactor "numpyflat" to be "numpy2D"

For collections, currently 3D numpy are called "numpy3D" and 2D numpy are called "numpyflat". This was because "numpyflat" was also overloaded to mean "concatenate multivariate". This no longer happens and the name is I feel confusing

refactor numpyflat to be numpy2D. Just need to be careful of the Series name numpy2D

_No response_

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs

fixes #815 

Its a good idea imo to rename this, as it no longer means "concatenated multivariate", where it used to be use for a embedded hidden transformation. 

Further reason for this refactor: there is an irreconcilable difference with how we treat 2D numpy in different parts of the toolkit. if we have either
a) a multivariate series with d channels and m timepoints or
b) a collection of univariate n series with m timepoints
stored in a numpy array, do we treat it as

(d,m)/(n,m) or (m,d)/(m,n) 

3D numpy are universally considered to be (n,d,m) and classifiers, clusterers and regressors treat (n,m) as n univariate series. 
However, forecasting and series transformers all treat this  as (m,d)/(m,n) as is usual in that field. It is indicated by type "np.ndarray". 

So, obvious solution would be to have an axis argument, and writing from scratch, thats what we would do. However, there is a mountain of legacy code in the transformers, that involves all kind of auto conversion and transformers that are meant to work with both series and collections. Thus we have this situation, for example

```python 
from aeon.transformations.series.matrix_profile import MatrixProfileTransformer
import numpy as np
# either five series length 10 or 10 series length five
X = np.random.rand(5, 10)
mp2 = MatrixProfileTransformer(window_length=3) #Will shorten series by  2
X2 = mp2.fit_transform(X)
print("Shape  = ",X2.shape) # 2D treated as (m,d) output shape (3, 10)
# either five series with three channels of length 10 or what?
X = np.random.rand(5, 3, 10)
X2 = mp2.fit_transform(X)
print("Shape  = ",X2.shape) #3D treated as (n,d,m) (5, 3, 8)
```
where the same transformer treats 2D treated as (m,d) but a 3D treated as (n,d,m).


aeon/classification/compose/_pipeline.py
```diff
@@ -444,27 +444,12 @@ def __rmul__(self, other):
...
-            raise TypeError(
-                f"unexpected X output type "
-                f'in tag "output_data_type", found "{output_type}", '
-                'expected one of "Primitives" or "Series"'
-            )
-
+        """Convert X to 2D numpy required by sklearn."""
+        Xt = convert_collection(X, "numpy3D")
+        if Xt.ndim == 2:
+            return X
```
MatthewMiddlehurst, line 450: Will this ever be true if its converted to numpy3D?

aeon/datasets/tests/test_data_loaders.py
```diff
@@ -137,9 +137,9 @@ def test_load_fails():
 def test__alias_datatype_check():
     """Test the alias check"""
     assert _alias_datatype_check("FOO") == "FOO"
-    assert _alias_datatype_check("np2d") == "numpyflat"
-    assert _alias_datatype_check("numpy2d") == "numpyflat"
-    assert _alias_datatype_check("numpy2D") == "numpyflat"
+    assert _alias_datatype_check("np2d") == "numpy2D"
+    assert _alias_datatype_check("numpy2d") == "numpy2D"
+    assert _alias_datatype_check("numpy2D") == "numpy2D"
```
MatthewMiddlehurst, line 142: would swap this function argument to `numpyflat`

aeon/datatypes/_panel/_convert.py
```diff
@@ -1112,7 +1112,7 @@ def from_dflist_to_nested_adp(obj, store=None):
 convert_dict[("df-list", "nested_univ", "Panel")] = from_dflist_to_nested_adp
 
 
-def from_numpy3d_to_numpyflat(obj, store=None):
+def from_numpy3d_to_numpy2D(obj, store=None):
```
MatthewMiddlehurst, line 1115: naming is a bit inconsistent i.e. 3d and 2D

This happens elsewhere in the file.

aeon/datatypes/_panel/_registry.py
```diff
@@ -21,9 +21,9 @@
         "3D np.ndarray of format (n_cases, n_channels, n_timepoints)",
     ),
     (
-        "numpyflat",
+        "numpy2D",
         "Panel",
-        "2D np.ndarray of format (n_cases, n_channels*n_timepoints)",
+        "2D np.ndarray of format (n_cases, n_channels)",
```
MatthewMiddlehurst, line 26: `(n_cases, n_timepoints)`?

aeon/datatypes/tests/test_lookup.py
```diff
@@ -29,6 +30,9 @@ def test_mtype_to_scitype(mtype, scitype):
     AssertionError mtype_to_scitype does not convert mtype to scitype
     Exception if any is raised by mtype_to_scitype
     """
+    if mtype in AMBIGUOUS_MTYPES:
+        return None
```
MatthewMiddlehurst, line 34: Why are the test changes required? I think this may be a consequence of not changing the lists above maybe.

aeon/regression/compose/_pipeline.py
```diff
@@ -438,27 +438,12 @@ def __rmul__(self, other):
...
-            raise TypeError(
-                f"unexpected X output type "
-                f'in tag "output_data_type", found "{output_type}", '
-                'expected one of "Primitives" or "Series"'
-            )
-
+        """Convert X to 2D numpy required by sklearn."""
+        Xt = convert_collection(X, "numpy3D")
+        if Xt.ndim == 2:
+            return X
```
MatthewMiddlehurst, line 444: same as classification

aeon/utils/validation/_convert_collection.py
```diff
@@ -166,16 +167,16 @@ def _from_numpy3d_to_pd_wide(X):
         raise TypeError(numpy3D_error)
     if X.shape[1] != 1:
         raise TypeError("Cannot convert multivariate series to pd-wide")
-    X_flat = _from_numpy3d_to_numpyflat(X)
+    X_flat = _from_numpy3d_to_numpy2D(X)
```
MatthewMiddlehurst, line 170: 3d vs 2D

MatthewMiddlehurst (changes requested):
Fine with this in principle, should be careful not to break whats there though. The comments alluding to a `numpy2D` 'Table' type are a little concerning, don't know if that causes conflicts.

aeon/classification/compose/_pipeline.py
```diff
@@ -313,7 +313,7 @@ class SklearnClassifierPipeline(_HeterogenousMetaEstimator, BaseClassifier):
         sequentially, with `trafo[i]` receiving the output of `trafo[i-1]`,
         and then running `clf.fit` with `X` the output of `trafo[N]` converted to numpy,
         and `y` identical with the input to `self.fit`.
-        `X` is converted to `numpyflat` mtype if `X` is of `Panel` type;
+        `X` is converted to `numpy2D` mtype if `X` is of `Panel` type;
         `X` is converted to `numpy2D` mtype if `X` is of `Table` type.
```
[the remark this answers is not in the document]
TonyBagnall: numpy2D doesnt currently exist as a type afaik

aeon/classification/compose/_pipeline.py
TonyBagnall: no, it can go

aeon/datatypes/_panel/_convert.py
TonyBagnall: this is in datatypes, not fixing this myself

aeon/datatypes/tests/test_lookup.py
TonyBagnall: because numpyflat is explicitly listed as an ambiguous type and are meant to be excluded from checking. 

# mtypes to exclude in checking since they are ambiguous and rare
AMBIGUOUS_MTYPES = ["numpy2D", "alignment_loc"]

you cannot convert to different types unless you fix what it actually means for series. the method ``mtype_to_scitype`` confuses np.ndarray with numpyflat I think. mtype_to_scitype is not ever called with collections.

TonyBagnall:
> Fine with this in principle, should be careful not to break whats there though. The comments alluding to a `numpy2D` 'Table' type are a little concerning, don't know if that causes conflicts.

which conflicts? numpyflat was only used in places where it meant concatenate dimensions. Everything about datatypes is confusing ....

MatthewMiddlehurst:
there is also a "table" type with the same name 
image

TonyBagnall:
> there is also a "table" type with the same name

ah didnt realise that, I guess thats whats causing the ambiguity in the scitype conversion. I have never looked at the Table type. I dont really want to work with the whole scheme tbh, but I will take a look.

MatthewMiddlehurst (approved):

Issue #803, opened 2023-10-08
[ENH] Refactor EAgglo as a segmentation algorithm


TonyBagnall:
### Describe the feature or idea you want to propose

class EAgglo is in the annotation package and extends BaseTransformer. Its description is

"E-Agglo: agglomerative clustering algorithm"

and transform returns

        cluster
            numeric representation of cluster membership for each row of X.

so to me it looks like it should be in the clustering module, possibly in sub package for agglomerative. Any thoughts? @KatieBuc @patrickzib?

### Describe your proposed solution

refactor as a clusterer rather than transformer, move to clustering module

TonyBagnall:
looked at the paper and it is in fact a segmentation algorithm that uses clustering and it will go in the segmentation package

TonyBagnall:
can close this, its now in the segmentation module and deprecated in annotation

```python
class EAggloSegmenter(BaseSegmenter):
```

Issue #722, opened 2023-09-05
[MNT] Split the annotation module into different packages


hadifawaz1999:
### Describe the issue

The `annotation` module is currently a mix of different tasks including anomaly detection, segmentation, motif discovery, etc.

After a discussion on 05/09/23, attending developers thought it would be better to split this module up into different packages, i.e. `anomaly_detection`, `segmentation`.

This issue is here for further discussion on the topic, feel free to give your thoughts.

### Suggest a potential alternative/fix

Create new packages for each task, and migrate algorithms currently in `annotation` over time.

TonyBagnall:
see #592 

I think we should split this into segmentation and anomaly detection, and we should do it before reinstating the deprecation policy. It would also be much easier to have coherent base classes

segmentation
           ClaspSegmentation: Segmentation (BaseSeriesAnnotator)
           GGS: Segmentation (no base class)
           IGTS: Segmentation (no base class)
           InformationGainSegmentation: Segmentation (SegmentationMixin)

anomaly detection:
          STRAY: robust anomaly detection in data streams with concept drift. **(BaseTransformer)**
          PyODAnnotator. "Transformer that applies outlier detector from pyOD." (BaseSeriesAnnotator)


not sure about:

HMM: "Implements a simple HMM fitted with Viterbi algorithm." (BaseSeriesAnnotator). 
EAAgglo: "is a non-parametric clustering approach for multivariate timeseries" Should then be in clustering, but this class extends BaseTransformer, which seems a bit odd. Name not exactly informative

TonyBagnall:
Think we can close this now and open more specific issues if required

Issue #699, opened 2023-08-27
[BUG] GreedyGaussianSegmentation and InformationGainSegmentation fail tests


TonyBagnall:
### Describe the bug

GreedyGaussianSegmentation and InformationGainSegmentation fail the test_all_estimator tests "test_create_instance" and "test_inheritance" 

### Steps/Code to reproduce the bug

Remove

    # GGS inherits from BaseEstimator which breaks this test
    "GreedyGaussianSegmentation": ["test_inheritance", "test_create_test_instance"],
    "InformationGainSegmentation": [
        "test_inheritance",
        "test_create_test_instance",
    ],
from test/_config.py

then run test_all_estimators 

comment from the bad place was # GGS inherits from BaseEstimator which breaks this test

### Expected results

pass the tests for at least find out if they should pass the tests

### Actual results

```python_traceback
aeon\tests\test_all_estimators.py:807 (TestAllObjects.test_inheritance[GreedyGaussianSegmentation])
0 != 1

Expected :1
Actual   :0
<Click to see difference>

self = <aeon.tests.test_all_estimators.TestAllObjects object at 0x0000028827AD0100>
estimator_class = <class 'aeon.annotation.ggs.GreedyGaussianSegmentation'>

    def test_inheritance(self, estimator_class):
        """Check that estimator inherits from BaseObject and/or BaseEstimator."""
        assert issubclass(
            estimator_class, BaseObject
        ), f"object {estimator_class} is not a sub-class of BaseObject."
    
        if hasattr(estimator_class, "fit"):
            assert issubclass(estimator_class, BaseEstimator), (
                f"estimator: {estimator_class} has fit method, but"
                f"is not a sub-class of BaseEstimator."
            )
    
        # Usually estimators inherit only from one BaseEstimator type, but in some cases
        # they may be predictor and transformer at the same time (e.g. pipelines)
        n_base_types = sum(
            issubclass(estimator_class, cls) for cls in VALID_ESTIMATOR_BASE_TYPES
        )
    
>       assert 2 >= n_base_types >= 1
E       assert 0 >= 1

aeon\tests\test_all_estimators.py:710 (TestAllObjects.test_create_test_instance[GreedyGaussianSegmentation])
self = <aeon.tests.test_all_estimators.TestAllObjects object at 0x0000028827CEF640>
estimator_class = <class 'aeon.annotation.ggs.GreedyGaussianSegmentation'>

    def test_create_test_instance(self, estimator_class):
[...]

Issue #592, opened 2023-07-23
[ENH] Annotation module


TonyBagnall:
### Describe the feature or idea you want to propose

Im just doing some docstring/notebook tidying and having a little look around annotation. Personally, I think I would split segmentation, change point detection and anomaly detection into separate modules. I believe it would improve exposure and be more logical. You can call them both annotation, but then you can call classification/regression/forecasting prediction if you want, doesnt make it a good idea. Any thoughts? I'm just scoping it for now, most of this is just for reference

### Describe your proposed solution

ClaspSegmentation: Segmentation (BaseSeriesAnnotator)
GGS: Segmentation (no base class)
EAAgglo:   "is a non-parametric clustering approach for multivariate timeseries" Should then be in clustering, but this class extends BaseTransformer, which seems a bit odd. Name not exactly informatibe
HMM: "Implements a simple HMM fitted with Viterbi algorithm." (BaseSeriesAnnotator). Seems like a full implementation
IGTS: Segmentation (no base class)
InformationGainSegmentation: Segmentation (SegmentationMixin)
STRAY:  robust anomaly detection in data streams with concept drift. (BaseTransformer)

then we have in sub packages
PyODAnnotator. "Transformer that applies outlier detector from pyOD." (BaseSeriesAnnotator)

in hmm_learn package which wraps https://github.com/hmmlearn/hmmlearn 
BaseHMMLearn (BaseSeriesAnnotator): Base class for all HMM wrappers, handles required overlap between packages.
GaussianHMM: "Hidden Markov Model with Gaussian emissions."
GMMHMM: "Hidden Markov Model with Gaussian mixture emissions"
PoissonHMM: Hidden Markov Model with Poisson emissions

there is also a base and a plotting subpackage.

Seems the easiest rationalisation would be to move evertything that is obviously segmentation into its own package, perhaps initially in annotations/segmentation.

TonyBagnall:
Add to this
"Heteregeneous Intrinsic Dimensionality Algorithm (Hidalgo) model" which is in

from aeon.transformations.series.hidalgo import Hidalgo

TonyBagnall:
there are still issues with the annotation module, but the issue is different to the one described above, which has been resolved, so closing this

PR #946, opened 2023-11-30
[ENH] Introducing BaseSegmenter for segmentation module

### LINKED ISSUE #802: [ENH] BaseSegmenter class

after the split of annotation module and the introduction of the experimental segmentation one, it would be good to settle on a standard interface for segmentation through the introduction of a BaseSegmenter. Currently, CLASP extends BaseSeriesAnnotator, GGS extends BaseEstimator and Hidalgo is in transformers.

need to see if we can work with standard fit/predict, take conversions up to base class etc

_No response_

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs
Fixes #802 #803
See #722 #699 #592 
    Segmenters take a single time series of size $m$ and return a segmentation.
    Series can be multivariate, with $d$ dimensions or channels.

    Design issues
    1. input and internal data format
    Univariate series:
        Numpy array, shape $(m,)$, $(m, 1)$ or $(1, m)$
            if self has no multivariate capability, all converted to 1D numpy $(m,)$
            if self has multivariate capability, converted to 2D numpy $(m,1)$ or $(1,
            m)$ depending on axis
        pandas DataFrame single column shape $(m,1)$, $(1,m)$ pandas Series shape $(m,)$
            if self has no multivariate capability, all converted to Series $(m,)$
            if self has multivariate capability, all converted to Pandas DataFrame
            shape $(m,1)$, $(1,m)$ depending on axis

    Multivariate series:
        Numpy array, shape $(m,d)$ or $(d,m)$
        pandas DataFrame $(m,d)$ or $(d,m)$

    2. Conversion and axis resolution for multivariate

    Conversion between numpy and pandas is handled by the base class. Sub classses
    can assume the data is in the correct format (determined by
    ``"X_inner_type"``, one of ``VALID_INNER_TYPES)`` and represented with the expected
    axis.

    Multivariate series are segmented along an axis determined by ``self.axis``. Axis
    plays two roles:
    1) the axis the segmenter expects the data to be in for its internal methods
    ``_fit`` and ``_predict``: 0 means each column is a time series channel $(m,d)$,
    1 means each row is a time series channel, sometimes called wide
    format, shape $(d,m)$. This should be set for a given child class through the
    BaseSegmenter constructor.

    2) The optional ``axis`` argument passed to the base class ``fit`` and ``predict``
    methods. If the data ``axis`` is different to the ``axis`` expected (i.e. value
    stored in ``self.axis``, then it is htransposed in this base class if self has
    multivariate capability.

    3. Segmentation representation

    Given a time series of 10 points with two change points found in position 4 and 8
    (lets index from 1 for clarity)

    The segmentation can be output in two forms:
    a) A list of change points: output example [4,8].
        This dense representation is the default behaviour, as it is the minimal
        representation.
    b) A list of integers of length n indicating the segment of each time point:
        output [0,0,0,1,1,1,1,2,2,2] or output [0,0,0,1,1,1,1,0,0,0]
        This sparse representation can be used to indicate shared segments (indicating
        segment 1 is somehow the same (perhaps in generative process) as segment 3.

    #TODO Add a tag to indicate which type of output
    Multivariate series are always segmented at the same points. If independent
    segmentation is required, fit a different segmenter to each channel.


ermshaua:
Hi Tony.

Here are my thoughts on the BaseSegmenter.

1. Use fit/predict (current) or fit/transform

I think one can argue for one or the other. Personally, I prefer fit/predict, because I see segmentation as an annotation task, that "adds change points to the series" instead of "transforming it to another (collection) of series". This preference goes hand in hand with my thoughts on point 2. However, I also don't feel strongly.

2. Return array of same length (current) i.e. sparse format, change points (dense format) make it a class parameter (annotation model)

In the ClaSP implementation, "sparse" currently returns change points, while "dense" returns segment intervals. Both options make sense, personally I prefer just a numpy array of change points. There is a third option, however, which is a series of segment labels. GGS uses this representation. This cannot be returned by all segmentation algorithms, because ClaSP, for instance, only provides you the change points, not the segment labels (which may re-occur throughout the segmentation). 

So I think it would be a good choice to have a format parameter in the BaseSegmenter, where the user can choose one of the three options (if available). Default value should be the numpy array of change points. In the literature, sometimes methods define the first or last data point as a change point, I find this not to be necessary.

4. Applicable to series or to both series and collections?

Segmentation is typically defined as a series problem, not a collection of series task. Therefore, I would prefer the BaseSegmenter to be applicable to series. Currently, the user can provide different series in the fit and predict functions. To my understanding, this makes not too much sense. I'd rather prefer just a fit_predict function or have the series in the predict method optional and note that it is ignored.

Cheers, Arik

TonyBagnall:
thanks Arik, so

aeon/segmentation/base.py
```diff
@@ -0,0 +1,188 @@
...
+# allowed types for transformers - Series and Panel
+ALLOWED_INPUT_TYPES = [
+    "ndarray",
+    "Series",
+    "DataFrame",
+]
+
+
+class BaseSegmenter(BaseEstimator, ABC):
+    """Docstring.
```
baraline, line 55: I guess the "Docstring."  should be replaced ?

aeon/segmentation/tests/test_base.py
```diff
@@ -0,0 +1,128 @@
...
+testy = np.random.random(10)
+uni = []
+uni.append(testy)
+uni.append(pd.Series(testy))
+uni.append(pd.DataFrame(testy))
+# uni.append(np.random.random((10, 1)))
+testy = np.random.random((4, 10))
+multi = []
+multi.append(testy)
+multi.append(pd.DataFrame(testy))
```
baraline, line 64: Is this supposed to be here and not inside a test function ?

baraline (commented):
Didn't find much to say on the segmentation logic as I only know the basics, apart from these minor issue the rest seems good to me !

aeon/segmentation/tests/test_base.py
TonyBagnall: its used in two tests so made it global. Just doing a bit of test based development.

docs/api_reference/segmentation.rst
```diff
@@ -13,6 +13,13 @@ contains algorithms and tools for time series segmentation.
...
+    BaseSegmenter
+    ClaSPSegmentater
+    InformationGainSegmenter
+    GreedyGaussianSegmenter
+    GreedyGaussianSegmenter
+    InformationGainSegmenter
+    DummySegmenter
+    EAggloSegmenter
+    HMMSegmenter
+    HidalgoSegmenter
```
MatthewMiddlehurst, line 25: There is some duplication here.

MatthewMiddlehurst (commented):
I don't have much to comment on segmentation really, but had a quick look through and looks fine in principle. 

In the interest of moving things forward I would say @ anyone who may be interested to give comments on the changes and we can put it in after a bit if everyone is ok/no response.

It is all experimental, so can change easily.

docs/api_reference/segmentation.rst
TonyBagnall: good spot!

docs/api_reference/annotation.rst
```diff
@@ -7,15 +7,6 @@ The :mod:`aeon.annotation` module contains algorithms and tools
 for time series annotation. This no longer includes anomaly/outlier
 detection and time series segmentation, which are now in their own module.
```
baraline, line 8: Minor point, but a `:mod:` link to the new module would be nice

baraline (dismissed):
Couldn't find much to say on this PR apart from a very minor comment in the docs, LGTM !

aeon/segmentation/base.py
```diff
@@ -0,0 +1,296 @@
...
+    1) the axis the segmenter expects the data to be in for its internal methods
+    ``_fit`` and ``_predict``: 0 means each column is a time series channel `(m,d)`,
+    1 means each row is a time series channel, sometimes called wide
+    format, shape `(d,m)`. This should be set for a given child class through the
+    BaseSegmenter constructor.
+
+
+    2) The optional ``axis`` argument passed to the base class ``fit`` and ``predict``
+    methods. If the data ``axis`` is different to the ``axis`` expected (i.e. value
+    stored in ``self.axis``, then it is htransposed in this base class if self has
```
lmmentel, line 67:
```suggestion
    stored in ``self.axis``, then it is transposed in this base class if self has
```

aeon/segmentation/base.py
```diff
@@ -0,0 +1,296 @@
...
+    segmentation is required, fit a different segmenter to each channel.
+
+    Parameters
+    ----------
+    n_segments : int, default = 2
+        Number of segments to split the time series into. If None, then the number of
+        segments needs to be found in fit.
+    axis : int, default = 0
+        Axis along which to segment if passed a multivariate series (2D input). If axis
+        is 0, it is assumed each row is a time series and each column is a channel.
```
lmmentel, line 95: Not sure what the difference here is between "time series" and "channel", maybe "time point" or "time instance" would be better?

lmmentel (commented):
Looks good :raised_hands:  Very nice that all segmentation algos now live in one place :)

aeon/segmentation/_dummy.py
```diff
@@ -0,0 +1,54 @@
...
+        ----------
+        X : 2D np.ndarray
+            Time series of shape `(n_channels, series_length)`
+        y : np.ndarray or None, default = None
+
+        Returns
+        -------
+        self : reference to self.
+        """
+        self.n_segments_ = self.n_segments
```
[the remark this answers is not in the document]
TonyBagnall: think I got a bit mixed up here with number of segments in the base class. I say it defaults to 2 but then default to None. I think default to 2 is reasonable. I will rework it.

aeon/segmentation/base.py
TonyBagnall: agreed

aeon/segmentation/_dummy.py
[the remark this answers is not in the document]
TonyBagnall: reworked the n_segments to default to 2

TonyBagnall:
hopefully addressed comments, thanks for the input. I'd appreciate it if someone could read the docstrings, the whole axis thing does my head in, and it was incorrect before

lmmentel (approved):

Issue #920, opened 2023-11-23
[BUG] Benchmarking `plot_scatter` produces strange figures with small amounts of data


MatthewMiddlehurst:
### Describe the bug

Running the `plot_scatter` function with 3 datasets, I get this output:

![output](https://github.com/aeon-toolkit/aeon/assets/25731235/6264b895-9b78-4262-83d7-cb422d8a2be8)

### Steps/Code to reproduce the bug

I can provide the input used and a pickled version of the figure if requested, it is part of a larger evaluation framework.

### Expected results

The figure is drawn without errors

### Actual results

There are three issues which I can see:
- The text labels are in the centre instead of the corners and are overlapping with the plotted points
- The points where ROCKET wins are orange instead of green, and draws are blue instead of orange. This may be because TSF wins on 0 datasets.
- The axis labels go up to 1.02, which is impossible for accuracy. This may be less of a bug as the metric won't always be accuracy, but it may be good allow users to select bounds.

Halzorg:
I would like to take a look at this issue and attempt to fix it. Could you please assign it to me?

MatthewMiddlehurst:
Done, feel free to ask any questions here or on Slack.

TonyBagnall:
hi, I think this is fixed now @MatthewMiddlehurst? I was just drawing some with 3 data and got this 
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/83a286a6-0c99-4ae5-9e94-4edcc4d74077)

PR #950, opened 2023-11-30
[BUG, ENH] Fixes several issues and enhances plotting module

dguijo:

- Fixes an division-by-0 error in the boxplot function when there is a method obtaining 0 for all datasets.
- Adds the average value per method to the scatter. Avg value is added to the axis label.
- Adds wilcoxon or ttest p-value to the suptitle of the scatter if the parameter statistic_test is set. Adding this feature also solves third issue of #920 as a new parameter is introduced with the metric.
- Fixes text going beyond axis lines. (Fixes first issue of #920)

Nope

##### For all contributions
- [X] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring or governance.


hadifawaz1999 (commented):
Thanks @dguijo for these fixes, can you show some plotting examples of these use cases like the zero case for box plot, and an example of where the pvalue is shown etc. and just put them in a comment on the PR ? would be great so we can keep track of everything

dguijo:
Hello, @hadifawaz1999 thank you! sure, I can add some plots:

The first one shows an example for comparing accuracies of HC2 vs IT, whereas the second one shows the same results but in terms of error (1-acc). Wilcoxon test is used for the first one, whereas ttest is used for second one. As can be observed, the avg value for each approach is now represented with a line, and the specific value is shown along with the name in the axis. 
image

image

The division-by-0 issue is only raised when there is an approach achieving always "0" values. This figure is very uninformative as it does not make any sense to have these results and compare them. In any case, this is the example:

image

hadifawaz1999:
Thanks @dguijo , i am just wondering about the labeling "pvalue ... for HC2" what would happen if the pvalue is higher than 0.05 ? in that case its not "in favor" of any of the classifiers in the pair, so what will the code output ?

hadifawaz1999:
> Thanks @dguijo , i am just wondering about the labeling "pvalue ... for HC2" what would happen if the pvalue is higher than 0.05 ? in that case its not "in favor" of any of the classifiers in the pair, so what will the code output ?

I might be saying nonsense though @dguijo because am used to using two-sided so :p

TonyBagnall:
well p-values have nothing to do with one sided or two sided, god knows why scipy doubles the p-value. The correct correction for one vs two sided is to alpha, the critical level and the critical region. If there is room I would put

Wilcoxon test for equality of median, p-value = 0.01
paired t-test for equality of means, p-value = 0.005

then always show the value that is less than 0.5. 

But I would also put the better algorithm in the top?

dguijo:
> well p-values have nothing to do with one sided or two sided, god knows why scipy doubles the p-value. The correct correction for one vs two sided is to alpha, the critical level and the critical region. If there is room I would put
> 
> Wilcoxon test for equality of median, p-value = 0.01 paired t-test for equality of means, p-value = 0.005
> 
> then always show the value that is less than 0.5.
> 
> But I would also put the better algorithm in the top? We do one sided test because when forming cliques we have already ordered them by rank. Testing alternative including B rank is better than A makes no sense if we already know A average rank is better than B.

I can add the best algorithm in the top (the best is the one achieving the best average value for the evaluation metric, or rank, which will coincide).

I can also add two lines, one for equality of medians and one for equality of means. I wonder if you want wilcoxon for testing equality of medians and ttest for equality of means or the above was just an example.

TonyBagnall:
> > well p-values have nothing to do with one sided or two sided, god knows why scipy doubles the p-value. The correct correction for one vs two sided is to alpha, the critical level and the critical region. If there is room I would put
> > Wilcoxon test for equality of median, p-value = 0.01 paired t-test for equality of means, p-value = 0.005
> > then always show the value that is less than 0.5.
> > But I would also put the better algorithm in the top? We do one sided test because when forming cliques we have already ordered them by rank. Testing alternative including B rank is better than A makes no sense if we already know A average rank is better than B.
> 
> I can add the best algorithm in the top (the best is the one achieving the best average value for the evaluation metric, or rank, which will coincide).
> 
> I can also add two lines, one for equality of medians and one for equality of means. I wonder if you want wilcoxon for testing equality of medians and ttest for equality of means or the above was just an example.

not sure if it will be too busy? Go with what you think best

TonyBagnall:
scatter plots: 
```python
cls = ["HC2", "FreshPRINCE", "InceptionT", "WEASEL-D"]

data = univariate_equal_length
res = get_estimator_results_as_array(
    estimators=cls, datasets=data, path=data_path, include_missing=True
)
res2 = res[:, [3, 0]]
fig = plot_scatter(res2, cls[3], cls[0], metric="accuracy", statistic_test="ttest", title="test")
fig.show()
fig.savefig("C:/Temp/scatter.png")
```
image

I think by default the best algorithms should be the y-axis and just lose the "for HC2". Also should say paired t-test, thats assuming you are doing a paired t-test rather than the standard t-test (should be paired)

TonyBagnall:
For CDs it cuts off the names
```python
data = univariate_equal_length
res = get_estimator_results_as_array(estimators=cls, datasets=data, path=data_path)
fig = plot_critical_difference(res,cls)
fig.show()
fig.savefig("C:/Temp/cd1.png")
fig.savefig("C:/Temp/cd2.pdf")
```

image

TonyBagnall:
```python
from random import uniform
from matplotlib.figure import Figure
y = np.array([uniform(0.0, 10.0) for _ in range(100)])
y_pred = y + np.array([uniform(-1, 1) for _ in range(100)])

method = "new method"
dataset = "toy"

fig = plot_scatter_predictions(y, y_pred, method, dataset)
fig.show()
fig.savefig("C:/Temp/scatter_preds.png")
```
image
looks good, but I dont think it should have a title by default. You generally cant put titles in figures for publication.

dguijo:
> They seem to be cutting off the name

I think this is a matplotlib issue :thinking: 

Using tight layout fixes this issue:

`plot.savefig("strip.png", bbox_inches="tight")`

TonyBagnall:
> > scale is different to violin, should probably be the same?
> 
> I think the scale differs because the param `outliers` is set to False. Could you please check? Thanks!

yep, you are correct
image

TonyBagnall:
looks good now, although not sure if these warnings are of note?

image

MatthewMiddlehurst (commented):
Good work, I don't have much to comment beyond what has already been discussed. It would be good to get this into the release.

This is slightly unrelated, but If you make further changes line 73 of the plot test `_check_soft_dependencies("matplotlib")` is redundant I think.

dguijo:
> For CDs it cuts off the names
> 
> ```python
> data = univariate_equal_length
> res = get_estimator_results_as_array(estimators=cls, datasets=data, path=data_path)
> fig = plot_critical_difference(res,cls)
> fig.show()
> fig.savefig("C:/Temp/cd1.png")
> fig.savefig("C:/Temp/cd2.pdf")
> ```
> 
> image

I'm going to leave this issue for the next PR, the one with new graphs. This is something related with the way of building the graph and can be solved for now with `plot.savefig(f"CD.pdf", bbox_inches="tight")`.

PR #957, opened 2023-12-02
[ENH, BUG, DOC] Refactor distance profiles, introduce STD threshold for numba

baraline:
#### Reference Issues/PRs

Fixes #870 #952 #875

#### What does this implement/fix? Explain your changes.

- Refactor the naive distance profile to accept any distance from the distance module or any user specified distance function.
- To run functions from the distance module with arguments other than the default, a function to copy and recompile a numba function with new default argument was created in utils/numba/general.py.
- Introduce a STD threshold for function using numba to normalize series to solve #952 
- Modify the docs for the similarity search module to fix #875

#### Does your contribution introduce a new dependency? If yes, which one?

No.

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [X] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring or governance.

##### For new estimators and functions
- [X] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).

##### For developers with write access
- [X] Optionally, I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #958, opened 2023-12-03
[ENH] ADD SAST transformer and SASTClassifier

frankl1:

Add implementation of the SAST ans SASTClassifier:
- SAST: a subsequence based transformation for time series
- SASTClassifier: as helper classifier to build a time series classifier based on the SAST transformation
[Associated paper](https://www.sciencedirect.com/science/article/abs/pii/S003132032300818X)

matplotlib

##### For all contributions
- [X] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring or governance.


TonyBagnall:
awesome, thanks @frankl1, we will take a look next week.

frankl1:
Hi all,

I could not figure out the cause of the failing test. It doesn't seem to be related to my changes. Since I couldn't manage to reproduce it locally, I wonder if there is a way to get more information on the error from the job log. 

Thanks

frankl1:
Thanks for the reply @TonyBagnall. It is my pleasure to contribute to this great project and I plan to contribute more in the future. I haven't joined the Slack, will catch up.

TonyBagnall (commented):
I'd rather not introduce a whole new category, could this go in the shapelet package?

frankl1:
Hi @TonyBagnall ,

Thanks for reviewing the PR.

I don't mind using the shapelet package for this since the method is able to learn shapelets.

aeon/classification/subsequence_based/_sast_classifier.py
```diff
@@ -0,0 +1,204 @@
...
+            znorm_kernel = (kernel - kernel.mean()) / (kernel.std() + 1e-8)
+            d_best = np.inf
+            for i in range(ts.size - kernel.size):
+                s = ts[i : i + kernel.size]
+                s = (s - s.mean()) / (s.std() + 1e-8)
+                d = np.sum((s - znorm_kernel) ** 2)
+                if d < d_best:
+                    d_best = d
+                    start_pos = i
+            axes[f].plot(range(start_pos, start_pos + kernel.size), kernel, linewidth=5)
```
baraline, line 200: I find this is a bit confusing for visualization purposes. 
If you compute a z-normalized distance but plot the non normalized kernel, it could lead to confusing plots. I would recommend scaling the kernel by inverting the normalization as:
`kernel = (kernel * std_best_match) + mean_best_match`
This would allow the "kernel" to be scaled to the best match on which it is positioned.

aeon/transformations/collection/subsequence_based/_sast.py
```diff
@@ -0,0 +1,209 @@
+import multiprocessing
+
+import numpy as np
+from numba import get_num_threads, njit, prange, set_num_threads
+
+from aeon.transformations.collection import BaseCollectionTransformer
+
+
+@njit(fastmath=True)
+def _znormalize_array(arr):
```
baraline, line 10: It would be better if you used the function [z_normalise_series](https://github.com/aeon-toolkit/aeon/blob/72afafad3c1d0d045d57f82a22f362937cd37533/aeon/utils/numba/general.py#L182) that already exists in aeon, a known bug exists for the std check but it should be fixed with #957.

aeon/transformations/collection/subsequence_based/_sast.py
```diff
@@ -0,0 +1,209 @@
...
+
+@njit(fastmath=False)
+def _apply_kernel(ts, arr):
+    d_best = np.inf  # sdist
+    m = ts.shape[0]
+    kernel = arr[~np.isnan(arr)]  # ignore nan
+
+    kernel_len = kernel.shape[0]
+    for i in range(m - kernel_len + 1):
+        d = np.sum((_znormalize_array(ts[i : i + kernel_len]) - kernel) ** 2)
```
baraline, line 25: Totally optional, you could precompute the means and std more efficiently for the whole series before applying your kernel by using the [sliding windows approach](https://github.com/aeon-toolkit/aeon/blob/72afafad3c1d0d045d57f82a22f362937cd37533/aeon/utils/numba/general.py#L397) instead of recomputing them here.

aeon/classification/subsequence_based/_sast_classifier.py
```diff
@@ -0,0 +1,204 @@
...
+        self.seed = seed
+
+        self.classifier = classifier
+
+    def _fit(self, X, y):
+        """Fit SASTClassifier to the training data.
+
+        Parameters
+        ----------
+        X : float[:,:,:]
```
baraline, line 89: It would be nice if we could use a somewhat similar documentation format as (at least) other shapelet methods.
For example, in STC : 
`X: np.ndarray shape (n_time_series, n_channels, n_timepoints)`

aeon/classification/subsequence_based/_sast_classifier.py
```diff
@@ -0,0 +1,204 @@
...
+        If None, will be infered during fit as np.arange(3, X.shape[1])
+    stride : int, default = 1
+        the stride used when generating subsquences
+    nb_inst_per_class : int default = 1
+        the number of reference time series to select per class
+    seed : int, default = None
+        the seed of the random generator
+    classifier : sklearn compatible classifier, default = None
+        if None, a RidgeClassifierCV(alphas=np.logspace(-3, 3, 10)) is used.
+    n_jobs : int, default -1
```
baraline, line 36: That is more of an open question, but shouldn't the default value for n_jobs parameters be 1 ?

aeon/transformations/collection/subsequence_based/_sast.py
```diff
@@ -0,0 +1,209 @@
...
+            containing the time series
+        y : ignored argument for interface compatibility
+
+        Return
+        ------
+        np.ndarray [n_time_series, n_subsequences], the transformed data
+        """
+        X_ = np.reshape(X, (X.shape[0], X.shape[-1]))
+
+        prev_threads = get_num_threads()
```
baraline, line 198: Do you mind explaining why you need this ? I know some behaviour with numba can be hard to grasp and I can't see why you would need this

aeon/transformations/collection/subsequence_based/_sast.py
```diff
@@ -0,0 +1,209 @@
...
+
+        if self.n_jobs < 1 or self.n_jobs > multiprocessing.cpu_count():
```
baraline, line 200: You might want to use the [check_n_jobs](https://github.com/aeon-toolkit/aeon/blob/72afafad3c1d0d045d57f82a22f362937cd37533/aeon/utils/validation/__init__.py#L137) function here

aeon/transformations/collection/subsequence_based/_sast.py
```diff
@@ -0,0 +1,209 @@
...
+            time_series_length) containing the time series
+        y : Any[:]
+            an array of shape (n_time_series,), containing
+            the class label of each time series in X
+
+        Return
+        ------
+        self
+        """
+        X_ = np.reshape(X, (X.shape[0], X.shape[-1]))
```
baraline, line 131: Couldn't this simply be `X_ = X[:, 0]` ?
Note that neither this nor the original create a copy of X

aeon/classification/subsequence_based/_sast_classifier.py
frankl1: Hi @baraline ,

Thanks for getting involved in this.

I am doing exactly what you are suggesting. ```kernel``` is not normalized, the normalized version is saved in the variable ```znorm_kernel``` so I don't have to invert the normalization.

I hope this clarifies things

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: I agree, I did not know about that one, thanks

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: I am tempted, but I am wondering how I could handle the 0 std using this suggestion in such a way that any change in the ```z_normalise_series``` is propagated here as well. Any recommendation?

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: Sure, it is because I wrote the SAST code to take a 2d array as input, but the input in Aeon classifiers always includes the channel dimension. This is just a workaround, I have plans to make SAST multivariate compatible and line 196 will be removed at that time.

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: Yes, it could.

Note that from the doc, the original could create a copy 
image

aeon/transformations/collection/subsequence_based/_sast.py
baraline: Sorry for the misunderstanding, I was referring to the `prev_threads  = get_num_threads()` part not the reshaping

aeon/transformations/collection/subsequence_based/_sast.py
baraline: Then I guess If you need a copy of it, just explicitly call numpy copy, as the "if possible" is kinda vague ...
If you don't make direct modifications to the input, you can spare yourself doing the copy anyway. 

Don't hesitate to resolve the conversation if you have fixed the issues in your commits !

aeon/classification/subsequence_based/_sast_classifier.py
baraline: I think it's more for test and debugging purpose, as the error logs from a parallel code can sometime be a bit messy. Also tests runtime may sometime be faster with 1 job than more as they use very small datasets, where the overhead of parallel jobs can not be worth. Thanks !

aeon/transformations/collection/subsequence_based/_sast.py
baraline: #957 will also introduce new normalization function that take mean and std as input like this :
``` python 
def z_normalize_series_with_mean_std(
    X: np.ndarray, series_mean: float, series_std: float
):
    """
    Numba series normalization function for a 1d numpy array with mean and std.
    Parameters
    ----------
    X : 1d numpy array
        A 1d numpy array of values
    series_mean : float
        The mean of the series
    series_std : float
        The standard deviation of the series
    Returns
    -------
    arr :  1d numpy array
        The normalized series
    """
    if series_std > AEON_NUMBA_STD_THRESHOLD:
        arr = (X - series_mean) / series_std
    else:
        arr = X - series_mean
    return arr
 ```
 
I would suggest you keep the code like this for now, and we can take a look together at how we can improve it when #957 will be merged !

frankl1: Alright, sounds good

aeon/classification/subsequence_based/_sast_classifier.py
frankl1: I haven't experienced that with Numba yet when parallelizing Python for loops, but I get your point.

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: Oh no worries, 

It is used to set number of threads back to the initial value after applying the kernels. 
image

aeon/transformations/collection/subsequence_based/_sast.py
frankl1: I don't modify the input, so we are good :) , thanks

aeon/classification/subsequence_based/_sast_classifier.py
baraline: We can leave it as is for now, we can talk about it later if you want !

TonyBagnall (approved):
LGTM, thanks

aeon/transformations/collection/subsequence_based/_sast.py
MatthewMiddlehurst: Hi, sorry for the delay. I am the one who originally added similar code to ROCKET. I think returning external variables like this back to the original state is good practice IMO, alternatively adding documentation that it will change but I think the current method is fine.

We discussed this in the last dev meeting, so resolving.

MatthewMiddlehurst:
Thanks for the contribution @frankl1, I didnt see anything holding this back from going in on my look through. I think some tests would be good to add, but won't push for them in this PR. 

Are there any more changes you want to make for this? If not, you can ping @baraline to re-review and we can move it forward I think.

Sorry for the delay!

baraline (approved):

frankl1:
Hi,

I am sorry for the late reply, I was on vacation. There are no changes I would like to add to this @MatthewMiddlehurst. I will work on the multivariate version and make a new PR. 

Thanks for merging with the main branch.

Regard

Issue #974, opened 2023-12-05
[ENH] Locations of remaining uses of datatypes


TonyBagnall:
### Describe the feature or idea you want to propose

Update 6/2/24
forecasting
<img width="436" alt="image" src="https://github.com/aeon-toolkit/aeon/assets/9594042/bc6bae89-328d-4d4c-b649-1f12d0e258db">

testing
<img width="431" alt="image" src="https://github.com/aeon-toolkit/aeon/assets/9594042/edc65f97-9ef0-4d2e-8c50-3b81799f0389">

transformers
<img width="431" alt="image" src="https://github.com/aeon-toolkit/aeon/assets/9594042/f42cf71c-d717-48fe-9530-e400383f41db">

utils
<img width="431" alt="image" src="https://github.com/aeon-toolkit/aeon/assets/9594042/e7cbbe17-e2b6-45e8-b690-3a4375c840b5">

in the continued efforts to isolate the horrible datatypes module, just dumping this here for future reference

### previous
forecasting
39 matches in 24 files
performance_metrics
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/f1e40889-24ce-4e48-a0a3-175827c73356)
transformations
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/e1fbac70-bf64-4a90-8848-a21b1e24dbf8)

TonyBagnall:
switching now to replacing function by function, starting with the check functions. Locations of check_is_mtype, check_is_scitype. Both are used usually to extract all meta data ujnnecessarily and are really just a very simple type check.

check_is_mtype
-[ ] BaseTransformer; line 1023: used to check if valid series input and whether univariate
-[ ] test_all_forecasters: line 333: check if proba, line 639, check if multiindex, 
- [ ] test_base in forecasting. Five uses to do a simple type check

check_is_scitype
-[ ] BaseTransformer: twice used just to check validity
-[ ] test_base (transformer) used 16 times to simply check input is an allowed series data structure, 
-[ ] BaseForecaster: used twice in checking y lines 1336 and 1380
-[ ] _split: BaseSplitter: used to check valid input type line 529
-[ ] _functions: evaluate() : used to check valid input twice.

all other uses are isolated to datatypes. Once utils validation/conversion is tidied up, should be easy to replace these calls.

TonyBagnall:
personal crusade continues: Conversion utils are more complex because they do all sorts of other things such as store variables and

## convert
[...]

PR #987, opened 2023-12-11
[ENH] Tidy up and restructure load to nested univariate

TonyBagnall:
Part of #974 

This simplifies and tidies up the file datasets._dataframe_loaders.py. Context is this is very old code to load data into the now defunct nested univ format. It is not used anywhere and has under 50% coverage. Changes:

1. Remove load from long format to nested univariate. This function makes a lot of assumptions about the input format and has no facility to allow class values. I see no point in keeping it
2. Load the tsv, ts and arff load to dataframe functions simply call their counterparts in _data_loaders, then convert to nested_univ. Removes a lot of code and makes it all more readable and robust. It does mean reducing some of the options in the arguments, but these were not tested anywhere and seemed pointless to me. 
3. Tidies up a weird potential circular import I left in when writing convert_collection
4. Removes a function load_tsf_to_dataframe from _data_loaders that was a direct copy of the function in _dataframe_loaders
5. Remove the duplication of having load_from_tsf_file and load_tsf_to_dataframe, because they were essentially doing the same thing. Just have load_from_tsf_file
6. Adds some tests for _data_loaders
7. Make separate test files for _data_generators and forecasting loading, to make things clearer 
8. Adjusted tsf single problem loader to return either default format or the hierarchical long format.

Two issues for me need resolving, possibly after this PR
1. _data_generators contains a function _convert_tsf_to_hierarchical which should be elsewhere really
2. Im still unclear as to the best format for returning time series for forecasting, need to explore a little more


aeon/datasets/_dataframe_loaders.py
```diff
@@ -615,110 +61,27 @@ def load_from_arff_to_dataframe(
     return_separate_X_and_y=True,
```
MatthewMiddlehurst, line 61: Unused parameters again

aeon/datasets/tests/test_dataframe_loaders.py
```diff
@@ -16,10 +23,357 @@
...
+        "contain_equal_length": False,
+    }
+
+    df, metadata = load_tsf_to_dataframe(data_path, return_type=return_type)
+    assert isinstance(df, pd.DataFrame)
+    assert isinstance(metadata, dict)
+    assert_frame_equal(df, output_df, check_dtype=False)
+    assert metadata == expected_metadata
+    # default_tsf"
+    #    assert check_is_mtype(obj=df, mtype=return_type)
```
MatthewMiddlehurst, line 373: strange comment can probably go.

Im amazed this dosent fail given the conversion bits are removed, does the function before already set it to these types?

TonyBagnall: the only conversion removed is from pd_multiindex_hier to pd_multiindex (_convert_tsf_to_hierarchical returns  pd_multiindex_hier) and the distinction between the two is not actually tested. If anyone ever wanted such a distinction we could put it in, but I dont see the need, I think let users sort their datatypes out really!

aeon/datasets/_data_loaders.py
```diff
@@ -11,14 +11,17 @@
...
+import aeon
 from aeon.datasets.dataset_collections import (
     list_downloaded_tsc_tsr_datasets,
     list_downloaded_tsf_datasets,
 )
-from aeon.datatypes import MTYPE_LIST_HIERARCHICAL, convert
 from aeon.utils.validation.collection import convert_collection
 
+DIRNAME = "data"
+MODULE = os.path.join(os.path.dirname(aeon.__file__), "datasets")
```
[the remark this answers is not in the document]
TonyBagnall: too tedious to tidy up this legacy code imo

MatthewMiddlehurst (approved):

Issue #833, opened 2023-10-13
[ENH] Rationalise matrix profile transformers


TonyBagnall:
### Describe the feature or idea you want to propose

there are two transformers that implement the MatrixProfile. The one in transformations.collections is a bespoke implementation, the one in transformations.series uses (an incorrectly handled) soft dependency. There can be only one

### Describe your proposed solution

we see if they are indeed the same, profile one and then assess alternatives for including MP as a transform.

TonyBagnall:
@aeon-actions-bot assign @CodeFor2001

TonyBagnall:
The first thing to do is to look at the two transformers. We have a series transformer MatrixProfileSeriesTransformer that transforms a single series using stumpy and our own collection transformer MatrixProfile that transforms a collection by applying our own implementation of a single series transformer in a loop. 

We only need one of these, and we should only have a series transformer that can be used with broadcast. it would be nice to keep our own version but it needs to be maintainable, logically equivalent and computationally efficient. The initial questions are

1. are they logically equivalent?
2. what is the relative run time?
3. could we speed our own version up with numba?

MatthewMiddlehurst:
Unless someone has write access they need to post on the thread first before they can be assigned.

PR #996, opened 2023-12-15
[ENH] New base classes: base series estimator , segmenters and base series transformers

### LINKED ISSUE #1063: [BUG] Segmentation method to_classification and to_clustering do not work as expected
### Describe the bug

these two methods in BaseSegmenter are meant to convert change points  into 0/1 series where change points occur , comment says 
        For example change points [2, 8] for a time series of length 10
        would result in: [0, 0, 1, 0, 0, 0, 0, 0, 1, 0].

but I cant see how it can be correct without being told what the length is, and output is not as expected. 

the actual code is simply
```python
        return np.bincount(change_points[1:-1], minlength=change_points[-1])
```
similar for to_clustering. Also not keen on the function names and they should be static methods. I think these came from clasp? @patrickzib ring any bells?

calling it
```python
import numpy as np
from aeon.segmentation import ClaSPSegmenter
bs = ClaSPSegmenter()
labels=bs.to_classification([2, 8])
```

### Expected results

 [0, 0, 1, 0, 0, 0, 0, 0, 1, 0].

### Actual results

[0 0 0 0 0 0 0 0]

### Versions

_No response_

TonyBagnall:
Part of #833 
See #537 
Fixes #1063 
introduces a `BaseSeriesEstimator` along the lines of `BaseCollectionEstimator`. `BaseSeriesEstimator` provides common checking and conversion tools. This was developed as successor to #946 and `BaseSegmenter` will also extend `BaseSeriesEstimator` when its finished.

The motivation to do this is to side step all the excessive checking, conversion, so called "vectorisation" under the combined series and collection transformers, and to sort out the problem with different expectations as to the shape - `(m,d) vs (d,m)`

series must be of type `np.ndarray, pd.Series, pd.DataFrame`. The major feature is that BaseSeriesEstimator has an axis. This means an estimator can be designed to treat a single series as either ``(n_timepoints, n_channels)`` with `axis = 0` or (n_channels,n_timepoints) with `axis = 1`.

The other point to note is that the tag `"capability:multivariate"` influences the conversion. If the tag is false, data passed in shape `(n_timepoints,1)` or `(1,n_timepoints)` is squeezed to be `(n_timepoints,)`. If the tag is true, data passed as  `(n_timepoints,)` is expanded to be `(n_timepoints,1)` or `(1,n_timepoints)` dependent on axis.

Matrix Profile converted to this format as first example. soon to be followed by a collection container. Need to put back the originals for deprecation


aeon/base/_base_series.py
```diff
@@ -0,0 +1,92 @@
+"""Base class for estimators that fit single (possibly multivariate) time series."""
+
+import numpy as np
+import pandas as pd
+
+from aeon.base._base import BaseEstimator
+from aeon.utils.validation._dependencies import _check_estimator_deps
+
+# allowed input and internal data types for Segmenters
```
MatthewMiddlehurst, line 9: Segmenters

aeon/base/_base_series.py
```diff
@@ -0,0 +1,92 @@
...
+        "X_inner_type": "ndarray",
+        "capability:multithreading": False,
+    }
+
+    def __init__(self, axis=0):
+        self.axis = axis
+        super(BaseSeriesEstimator, self).__init__()
+        _check_estimator_deps(self)
+
+    def _check_input_series(self, X):
```
MatthewMiddlehurst, line 42: I think we should try to align to the other base class when it comes to naming/functionality where possible. Would this be the equivalent of `_check_X` in the collection one?

aeon/transformations/series/base.py
```diff
@@ -0,0 +1,268 @@
...
+            )
+
+        # check whether is fitted
+        self.check_is_fitted()
+
+        self._check_input_series(X)
+        self._check_capabilities(X, axis)
+        X = self._convert_series(X, axis)
+        return self._inverse_transform(X=X, y=y)
+
```
MatthewMiddlehurst, line 187: I think some of these transformers make use of `def update`, which is missing atm.

aeon/transformations/series/_matrix_profile.py
```diff
@@ -0,0 +1,71 @@
+"""Implements matrix profile transformation."""
+
+__author__ = ["mloning", "TonyBagnall"]
+__all__ = ["MatrixProfileSeriesTransformer"]
+
+from aeon.transformations.series.base import BaseSeriesTransformer
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
+class MatrixProfileSeriesTransformer(BaseSeriesTransformer):
```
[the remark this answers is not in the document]
TonyBagnall: you mean call it MatrixProfileTransformer?

aeon/base/_base_series.py
TonyBagnall: yeah I guess

aeon/transformations/series/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests for series transformations."""
```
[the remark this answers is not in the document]
TonyBagnall: thats true, can have a dummy inheritance as with base collection trans

aeon/transformations/series/base.py
TonyBagnall: I'm not a fan of update, and so far only MatrixProfile is there, which does not use it. I say we wait for update until we want to port in an estimator that actually needs it.

TonyBagnall:
revert to draft until #946 is in, can address tests and inheritance then

aeon/transformations/series/base.py
TonyBagnall: added update

aeon/base/_base_series.py
```diff
@@ -0,0 +1,253 @@
...
+            metadata["missing_values"] = np.isnan(X).any()
+        elif isinstance(X, pd.Series):
+            metadata["missing_values"] = X.isna().any()
+        elif isinstance(X, pd.DataFrame):
+            metadata["missing_values"] = X.isna().any().any()
+        allow_multivariate = self.get_tag("capability:multivariate")
+        allow_univariate = self.get_tag("capability:univariate")
+        allow_missing = self.get_tag("capability:missing_values")
+        if metadata["missing_values"] and not allow_missing:
+            raise ValueError(" Missing values not supported")
```
baraline, line 136: Not very important, but spaces at the start of exception message here and below.

aeon/segmentation/tests/test_base.py
```diff
@@ -1,138 +1,47 @@
...
+    # assert labels not None
+
+
+#    assert labels == np.array([0, 0, 1, 0, 0, 0, 0, 0, 1, 0])
+
+
+def test_to_clusters():
+    pass
+    # labels = BaseSegmenter.to_classification(None, [2, 8])
+    # assert labels not None
```
baraline, line 47: Are they expected to stay like this, or did you forget to remove the comments ?

aeon/transformations/series/base.py
```diff
@@ -0,0 +1,301 @@
...
+        model attributes (ending in "_") : dependent on estimator
+
+        Parameters
+        ----------
+        X : Input data
+            Time series to fit transform to, of type ``np.ndarray``, ``pd.Series``
+            ``pd.DataFrame``.
+        y : Target variable, default=None
+            Additional data, e.g., labels for transformation
+
```
baraline, line 53: axis parameter do not appear in the docs here despite being in method definition

baraline (commented):
The application of axis and its role to convert the time series format is clear to me, only minor documentation points identified, the rest LGTM !

aeon/transformations/series/base.py
TonyBagnall: good spot

aeon/segmentation/tests/test_base.py
TonyBagnall: forgot about them! :)

TonyBagnall: ah put them in the wrong place. Ive made them class methods

aeon/base/_base_series.py
```diff
@@ -0,0 +1,253 @@
...
+        if type(y) not in VALID_INPUT_TYPES:
+            raise ValueError(
+                f"Error in input type for y: it should be one of "
+                f"{VALID_INPUT_TYPES}, saw {type(y)}"
+            )
+        if isinstance(y, np.ndarray):
+            # Check valid shape
+            if y.ndim > 1:
+                raise ValueError(
+                    "Error in input type for y: y input as np.ndarray " "should be 1D"
```
MatthewMiddlehurst, line 157: strings here are a bit weird `np.ndarray " "should`

aeon/transformations/matrix_profile.py
```diff
@@ -3,9 +3,19 @@
...
 
+from deprecated.sphinx import deprecated
+
 from aeon.transformations.base import BaseTransformer
 
 
+# TODO: remove in v0.8.0
+@deprecated(
+    version="0.6.0",
+    reason="MatrixProfileSeriesTransformer will be removed from the base directory of "
```
MatthewMiddlehurst, line 14: wrong name here

aeon/segmentation/_random.py
```diff
@@ -47,4 +49,20 @@ def _fit(self, X, y=None):
...
+        Parameters
+        ----------
+        parameter_set : str, default="default"
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class.
+        """
+        return {}
```
MatthewMiddlehurst, line 68: Why can't we use the series class default here, is the "axis=0" going to break stuff?

MatthewMiddlehurst (commented):
I can't really say I have considered everything, there is a lot going on in this PR.

One of my concerns is this will eventually all clash with forecasting design, where they usually do i.e. `fit(y, X=None)`.

TonyBagnall:
> I can't really say I have considered everything, there is a lot going on in this PR.
> 
> One of my concerns is this will eventually all clash with forecasting design, where they usually do i.e. `fit(y, X=None)`.

I dont intend making forecasters base series estimators series estimators any time soon, the interface is completely different, X and y dont mean the same things and they have a forecasting horizon parameter.

TonyBagnall:
> > I can't really say I have considered everything, there is a lot going on in this PR.
> > One of my concerns is this will eventually all clash with forecasting design, where they usually do i.e. `fit(y, X=None)`.
> 
> I dont intend making forecasters base series estimators series estimators any time soon, the interface is completely different, X and y dont mean the same things and they have a forecasting horizon parameter.

but there is no reason not to have axis with forecasting, I would in any case have a BaseForecaster, where you could hard code axis to 1

aeon/segmentation/_random.py
TonyBagnall: can use base class get_test_params, axis is not a parameter, it is a hard coded class value

aeon/base/_base_series.py
```diff
@@ -0,0 +1,253 @@
...
+            capability, i.e.``self.get_tag(
+            ""capability:multivariate") == False``, all are converted to 1D
+            numpy `(m,)`
+            if ``self`` has multivariate capability, converted to 2D numpy `(m,1)` or
+            `(1, m)` depending on axis
+            pandas DataFrame or Series:
+            DataFrame single column shape `(m,1)`, `(1,m)` or Series shape `(m,)`
+            if ``self`` has no multivariate capability, all converted to Series `(m,)`
+            if ``self`` has multivariate capability, all converted to Pandas DataFrame
+            shape `(m,1)`, `(1,m)` depending on axis
```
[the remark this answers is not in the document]
TonyBagnall: looks better now, esp as I have also put it in the API :)

baraline (approved):
Didn't pick any problem, and considering the discussion about forecasting, I think we can move on with this PR for segmentation and transformation.

Maybe for a future PR, but one weird edge case that might be worth testing is the case where a different axis is given as input during fit and transform. From what I understand of the code, it shouldn't be a problem as you check for the axis defined in `__init__` against the one given as parameter in `_convert_X`, but better safe than sorry I guess ?

Great work !

MatthewMiddlehurst (approved):
Yep think this is fine for now. Best to use it and build from there.

Issue #140, opened 2023-03-02
[DOC] Add plot_windows to docs and provide docstring example


aiwalter:
function can be found here:
https://github.com/scikit-time/scikit-time/blob/15a3faf2a203b22f6e0b536a5e2d99352c83ac7a/sktime/utils/plotting.py#L387

vermaaatul07:
I am working on this , assign it to me please.

TonyBagnall:
I think this issue is stale post visualisation and KWF, closing it, feel free to bring it back if Im wrong

PR #1002, opened 2023-12-18
[ENH] Introducing visualisation module

### LINKED ISSUE #794: [ENH] Where to locate plotting functionality?

during #782 I came across series plotting utils in annotation. We have other plotting functionality for both series and results in various locations. I think it could be tidied up. 

any thoughts?

Possible solutions
1. Move all plotting to utils.
2. Move all to benchmarking
3. Introduce a visualisation module, where the results plotting (benchmarking) and series plotting (annotation and utils)

feel free to add more.

_No response_

### Additional context

_No response_

TonyBagnall:
Fixes #794 
See #140 #451 #592 

this centralises the plotting functionality  from annotation, benchmarking, clustering and utils into a new module, visualisation. 

I've called it visualisation rather than plotting because we can include other tools for explainability such as feature relevance, but dont mind if its plotting instead. 

Still to do is to centralise the examples, sort out naming for plot series and standardise the function signatures and naming


dguijo (commented):
Thank you for your work, Tony!

I think is a good idea to unify all plotting stuff in one module with different sub-modules depending on the functionality. 

I have a question regarding methods. Do all these plot_ functions return the same matplotlib object? or just plot whatever and do not return anything. 
I think we should keep the same practice for all these functions, and in my opinion just return the plot without plotting anything (if so, maybe the function names should replace "plot" by "create" or other more appropriate word, which I leave for native people :))

Thank you again, good work!

TonyBagnall:
> Thank you for your work, Tony!
> 
> I think is a good idea to unify all plotting stuff in one module with different sub-modules depending on the functionality.
> 
> I have a question regarding methods. Do all these plot_ functions return the same matplotlib object? or just plot whatever and do not return anything. I think we should keep the same practice for all these functions, and in my opinion just return the plot without plotting anything (if so, maybe the function names should replace "plot" by "create" or other more appropriate word, which I leave for native people :))
> 
> Thank you again, good work!

hi David, I dont really know tbh, just want to get it all in one place then we can look more closely at how they work

dguijo (dismissed):
LGTM. Good work, Tony. We leave the decision about how to organise plot functions for the next PR :)

dguijo (approved):
LGTM, Thanks Tony!

PR #1006, opened 2023-12-20
[ENH] New regression problems

TonyBagnall:
Code to load all 63 regression problems from TSC.com. To note

1. Data is by default loaded from tsc.com. If that fails and it is one of the original 19, it then tries monash. tsc.com is I think more reliable than monash and we have not yet hosted all 63 with them
2. There is a mechanism for dealing with missing or unequal length. It is in my opinion the least bad way of doing this. There are now flags for load_regression that allow the user to specify the original data or to load data sets with missing values removed and/or standardised to equal length in the way we have done in our experiments. This makes it much easier to reproduce experiments, and there has been confusion in the past about the algorithms for interpolatation and making equal length

added tests for this on overnights. I will do results loaders in a separate PR


dguijo (approved):
LGTM. Thank you Tony for your work.  :smile:

PR #1009, opened 2023-12-21
[ENH, DOC] Add Euclidean distance with convolution for similarity search

baraline:

- Add computation of Euclidean and squared Euclidean distance with dot product computed by convolution. 
- Add tests to check the consistency of the results between naive and speedup version of distance profile computations
- Fix speed-up argument usage for BaseSimilaritySearch class, add a function to return all available speed-ups.
- Improve docstring and variable names consistency overall in the module
- Add the option to apply the exclusion zone to the prediction of the best matches. For example, in the case of TopKSimilaritySearch, it means that when k>1, the returned best matches will be spaced by at least `query_length//exclusion_factor` timestamps when selected from the same sample. This avoids cases where the best matches are direct neighbours.
- Update the associated notebooks, compare the performance of naive and convolution version, show how to use the new `apply_exclusion_to_result` parameter of predict.

No additional dependency.

##### For all contributions


aeon/similarity_search/distance_profiles/_commons.py
```diff
@@ -43,20 +43,20 @@ def fft_sliding_dot_product(X, q):
...
     -------
-    output : shape=(n_features, n_timestamps - (length - 1))
+    output : shape=(n_features, series_length - length + 1)
         Sliding dot product between q and X.
     """
-    n_features, n_timestamps = X.shape[0]
+    n_features, n_timestamps = X.shape
     length = q.shape[1]
-    out = np.zeros((n_features, n_timestamps - (length - 1)))
+    out = np.zeros((n_features, n_timestamps - length + 1))
```
TonyBagnall, line 59: seem to be mixing series_length and n_timestamps. In the other modules we use n_timepoints (and n_channels if multivariate)

TonyBagnall (changes requested):
all looks good, but we should consider n_timepoints vs series_length. I think I would prefer package wide standardisation, but I dont really mind series_length. There is I think still one instance of n_timestamps though :)

aeon/similarity_search/distance_profiles/_commons.py
baraline: Thanks ! I thought the standard in aeon was series_length and not n_timepoints, am I mistaking ?
Seems I focused on docstring more than variable names, will make some standardisation

baraline: Should be all fixed now !

TonyBagnall (approved):
LGTM

PR #1010, opened 2023-12-22
[ENH] Improvements to scatterplot when lower is better

dguijo:
This PR enhances scatterplots when using performance metrics to be minimised.

For example, in this case we are comparing the results of DrCIF and InceptionE using the relative RMSE (to minimise):
image
Now, the x-axis is in the top and the y-axis in the right. The reason for these changes are to improve clarification. If they were in their original position, if may be a little bit confusing as reader can interpret that the relative RMSE for DrCIF is in the bottom axis whereas those for Inception are in the left, being the way around. 

The standard case where a maximisation measure is used works as expected. This is an example:
image
As can be observed, the axis are now where they are expected given that it can not be object of confusion. Ties are shown, though they only exists for accuracy = 1, so can't be easily seen. 

Note that the winning method is shown in green colour (x-axis), whereas losing one is in blue colour and in the y-axis.

Also, now #920 is finished. 
If we run the scatter with three datasets, the result is the following:
image


aeon/benchmarking/_critical_difference.py
```diff
@@ -339,7 +347,7 @@ def plot_critical_difference(
     # Step 3 : check whether Friedman test is significant
     p_value_friedman = _check_friedman(ranks)
     # Step 4: If Friedman test is significant find cliques
-    if p_value_friedman < alpha:
+    if p_value_friedman < alpha or True:
```
TonyBagnall, line 350: this will always execute?

aeon/benchmarking/results_plotting.py
```diff
@@ -277,11 +276,11 @@ def plot_scatter(
     min_value = max(results.min() * 0.97, 0)
     max_value = results.max() * 1.03
 
-    if metric == "accuracy":
-        max_value = min(max_value, 1)
+    if metric in ["accuracy", "R2", "CCR", "AUC", "F1", "Kappa", "AUPRC"]:
+        max_value = min(max_value, 1.001)
```
TonyBagnall, line 280: not sure what this is for?

TonyBagnall (changes requested):
not sure about that if statement

aeon/benchmarking/results_plotting.py
dguijo: points plotted in (1, 1) can't be shown if the axis is limited to 1. They are difficult to be seen. 
Here they can be seen. 
image

But here is difficult:
image

aeon/benchmarking/_critical_difference.py
dguijo: LOL, I just wanted to make sure significant results were achieved by Friedman for some testing I was doing at that moment. Thank you Tony!

TonyBagnall (approved):

PR #1014, opened 2023-12-29
[BUG] fix tag errors, refactor comments np.array -> np.ndarray

TonyBagnall:
whilst removing datatypes references, I noticed a tag in y_type  

```python
    (
        "y_inner_type",
        ["forecaster", "transformer"],
        (
            "list",
            [
                "pd.Series",
                "pd.DataFrame",
                "np.array",
                "nested_univ",
                "pd-multiindex",
                "numpy3D",
                "df-list",
            ],
        ),
```
this is not a valid type even in the complete madness of datatypes, since this

```
x = mtype_to_scitype("np.array")
```
gives an error ValueError: np.array is not a supported mtype

Also noticed many forecasters assign     "pd_multiindex_hier" to y_inner_type, so I have also added that to the tag registry. 

I then look further and see many references in comments to np.array, which as I understand it is a function to create an array, whereas the type should be np.ndarray, so I have refactored them,.

X_inner_types also had np.array tag, and also was listed for "forecaster" only, when in truth this tag is used in all collection estimators


aeon/classification/_delegate.py
```diff
@@ -47,7 +47,7 @@ def _fit(self, X, y):
 
         Parameters
         ----------
-        X : 3D np.array
+        X : 3D np.ndarray
```
chrisholder, line 50: Added shape of np.ndarray in other docstring maybe add in all these ones with just 3D np.ndarray for consistency

aeon/classification/_dummy.py
```diff
@@ -83,7 +83,7 @@ def _fit(self, X, y):
 
         Parameters
         ----------
-        X : 3D np.array of shape = [n_instances, n_channels, series_length]
+        X : 3D np.ndarray of shape = [n_instances, n_channels, series_length]
         y : array-like, shape = [n_instances] - the class labels
```
chrisholder, line 87: Inconsistent docstring  3D np.ndarray of shape = [n_instances, n_channels, series_length].

Make it consistent with classification:

X : 3D np.array
    Input data, any number of channels, equal length series of shape ``(
    n_instances, n_channels, n_timepoints)``
    or 2D np.array (univariate, equal length series) of shape
    ``(n_instances, n_timepoints)``

aeon/classification/base.py
```diff
@@ -227,7 +227,7 @@ def score(self, X, y) -> float:
 
         Parameters
         ----------
-        X : 3D np.array (any number of channels, equal length series)
+        X : 3D np.ndarray (any number of channels, equal length series)
```
chrisholder, line 230: Inconsistent docstring   X : 3D np.ndarray (any number of channels, equal length series)

Make it consistent with classification:

X : 3D np.array
    Input data, any number of channels, equal length series of shape ``(
    n_instances, n_channels, n_timepoints)``
    or 2D np.array (univariate, equal length series) of shape
    ``(n_instances, n_timepoints)``

chrisholder (changes requested):
Need to make the docstring consistent with the other modules.

aeon/classification/_delegate.py
TonyBagnall: I think it should be 
X : np.ndarray
  then desciption. 
  I think its meant to be just the type after the name

PR #1023, opened 2024-01-06
[ENH] `.arff` writer function

MatthewMiddlehurst:
Adds a limited function to write univariate classification type arff files. Moves some private functions about to group related functions together.


aeon/datasets/_data_writers.py
```diff
@@ -154,6 +155,172 @@ def _write_data_to_tsfile(
...
+    suffix="",
+):
+    """Write an aeon collection of time series to text file in .arff format.
+
+    Only compatible for classification-like problems with univariate equal
+    length time series currently.
+
+    Parameters
+    ----------
+    X : np.ndarray (n_cases, n_channels, series_length)
```
TonyBagnall, line 263: series_length ->n_timepoints

aeon/datasets/_data_writers.py
```diff
@@ -154,6 +155,172 @@ def _write_data_to_tsfile(
     file.close()
 
 
+def _write_dataframe_to_tsfile(
```
TonyBagnall, line 158: this is I think for "nested_univ" dataframes, which could be misleading. It will fail if passed a normal dataframe. I dont really think we need this function, given nothing is stored in nested dataframes any more, but if you want it call it write_nested_univ or something?

aeon/datasets/tests/test_data_writers.py
```diff
@@ -125,18 +141,36 @@ def test_write_dataframe_to_ts(tsfile_writer):
...
-    ):
-        _write_data_to_tsfile(X, "temp", "temp", y=y)
+def test__write_header():
+    with tempfile.TemporaryDirectory() as tmp:
+        problem_name = "header.csv"
+        with pytest.raises(
+            ValueError, match="Cannot have class_labels true for a regression problem"
+        ):
+            _write_header(tmp, problem_name, class_labels=True, regression=True)
+    _write_header(
```
TonyBagnall, line 151: shouldnt this be inside the with tempfile statement?

TonyBagnall (approved):

aeon/datasets/_data_writers.py
MatthewMiddlehurst: This is not a new function, just moved from the bottom closer to the functions which actually use it.

aeon/datasets/tests/test_data_writers.py
MatthewMiddlehurst: Probably yeah, just a moved function. Can fix it if you want

TonyBagnall: ah just looks weird on github

TonyBagnall (approved):

Issue #975, opened 2023-12-06
[ENH] Get Type Information from the Datasets


PaulRabich:
### Describe the feature or idea you want to propose

On the webside each dataset has an attribute called ``type``. This attribute is not accessable via the API. 

### Describe your proposed solution

Add the information in the meta-data block, or have a seperate dict with the information.

TonyBagnall:
hi @PaulRabich proposed a simple solution here, you can just pull down the characteristics from https://timeseriesclassification.com/aeon-toolkit/metadata.csv or use a function in aeon to do it. Happy to adjust this to suite other use cases if required.

PR #1025, opened 2024-01-08
[ENH] Retrieve metadata for datasets from tsc.com

TonyBagnall:
see #975 

Simple function to load the meta data for the data hosted on timeseriesclassification.com, which I have put in a csv. Happy to add anything. Note I have changed some of the classifications of type from MOTION to HAR, since it is more accurate. I will no doubt live to regret that, but its more correct that way.


MatthewMiddlehurst (commented):
function is fine, but the name could maybe be a bit more specific i.e. `get_archive_meta_data` or similar.

Did you update the dataset page categories as well as the csv ones? not really relevant to this but just wondering

TonyBagnall:
> function is fine, but the name could maybe be a bit more specific i.e. `get_archive_meta_data` or similar.

done
 
> Did you update the dataset page categories as well as the csv ones? not really relevant to this but just wondering

not yet, waiting for this

MatthewMiddlehurst (approved):
Will approve because its not a blocker, but I think the name could still be confused with the .ts metadata stuff. A bit on the user to read the docs though.

PR #1026, opened 2024-01-08
[ENH] Import forecasters through init

TonyBagnall:
adds init import to the root forecasting module. No other changes, so no need for deprecation, but next stage would logically be to make the files private.


TonyBagnall:
theta transformer breaks imports

PR #1033, opened 2024-01-11
[ENH] scatter plots enhancement. 

### LINKED ISSUE #1030: [ENH] Scatter plots

the new scatter plots are great
image
we were just talking about them and thought it might be good if the dotted line represented the median not the mean, so we can show half above the line half below it. What do you think @dguijo ?

change dotted line to show median not mean

_No response_

### Additional context

_No response_

dguijo:
Fixes #1030

### DIFF
--- a/aeon/visualisation/results_plotting.py
+++ b/aeon/visualisation/results_plotting.py
@@ -333,7 +333,7 @@ def plot_scatter(
         zorder=2,
     )
 
-    # Draw the average value per method as a dashed line from 0 to the median value.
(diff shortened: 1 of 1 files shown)


dguijo:
Some examples:

Higher is better and no statistical tests. 
scatter

Higher is better and statistical tests. 
scatter

Lower is better and no statistical tests.
scatter

Lower is better and statistical tests.
scatter


btw, the warning with the tight layout only happens with `statistical_tests=True`. Though it is not important imo.

TonyBagnall (approved):
looks awesome

PR #1034, opened 2024-01-11
[ENH] Load the regression results from tsc.com

TonyBagnall:
see #451 
provides capacity to load results of 19 regressors on 63 tser datasets. Currently regression and classification results are in different formats. This will change when we have finished the classification redux, but for now they are parameterised


aeon/benchmarking/results_loaders.py
```diff
@@ -122,14 +150,43 @@ def get_available_estimators(task="classification") -> pd.DataFrame:
...
+    path = f"{path}/{task}/{measure}/"
+    all_results = {}
+    for cls in estimators:
+        alias_cls = estimator_alias(cls)
+        url = path + alias_cls + suffix
+        try:
+            data = pd.read_csv(url)
+        except Exception:
+            raise ValueError(
+                f"Cannot connect to {url} website down or results not " f"present"
```
dguijo, line 166: This string can be simplified:
`f"Cannot connect to {url} website down or results not present"`

aeon/benchmarking/results_loaders.py
```diff
@@ -122,14 +150,43 @@ def get_available_estimators(task="classification") -> pd.DataFrame:
...
+                    cls_results[problem] = results[pos][0]
+                else:
+                    cls_results[problem] = results[pos]
+        all_results[cls] = cls_results
+    return all_results
+
+
 def get_estimator_results(
     estimators: list,
     datasets=UCR,
```
dguijo, line 185: I've seen you test if the measure and task are valid, but I'm not sure you also test the datasets.  (maybe in _load_results()?)

dguijo (changes requested):
Thank you, Tony. Good work :) Some minor minor changes 

PS. I'll check the python notebooks in the reviewNB.

aeon/benchmarking/results_loaders.py
```diff
@@ -46,6 +63,7 @@
     "TSF": {"tsf", "TimeSeriesForest"},
```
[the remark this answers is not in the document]
TonyBagnall: no we should put them in, just were not part of the previous version so I missed them. thanks

aeon/benchmarking/results_loaders.py
TonyBagnall: cant really test this in aeon, since we cannot guarantee all datasets are available to all algorithms. By datasets where results not present are by default ignored if loaded into a dictionary
image

but have the option of NaNs in load to array
image

aeon/benchmarking/results_loaders.py
```diff
@@ -46,6 +63,14 @@
...
     "TSFresh": {"tsfresh", "TSFreshClassifier"},
     "WEASEL-Dilation": {"WEASEL", "WEASEL-D", "Weasel-D", "WEASEL2"},
+    "1NN-DTW": {
+        "1NNDTW",
+        "1nn-dtw",
+        "KNeighborsTimeSeriesRegressor",
+        "KNeighborsTimeSeriesClassifier",
+    },
+    "5NN-DTW": {"1NNDTW", "1nn-dtw"},
```
dguijo, line 72: I think it should be `"5NN-DTW": {"5NNDTW", "5nn-dtw"}`,

Also, I think the 1nn for the Euclidean Distance is still missing :) 

    "1NN-ED": {
        "1NNED",
        "1nn-ed",
    },
    "5NN-ED": {"5NNED", "1nn-ed"},

dguijo (commented):
minor change :) thanks Tony!

aeon/benchmarking/results_loaders.py
TonyBagnall: good spot!

dguijo (approved):
LGTM! Good work, Tony! Thanks!

PR #1038, opened 2024-01-12
[ENH] Remove the PODS adaptor

### LINKED ISSUE #997: [ENH] Remove PODSAnnotator  wrapper

In annotation we have PyODAnnotator which is a wrapper for the pyod package.

https://pyod.readthedocs.io/en/latest/

so I think just remove this and the associated soft dependency. My reasoning

1. The above link says use TODS for time series
https://github.com/datamllab/tods
2. It does not check soft deps properly.
3. It is BaseSeriesAnnotator, which is on its way out

as a minimum, it should move to anomaly_detection package and not be in its own "adaptor" directory. I would prefer just to remove it, then use TODS when we get round to anomaly detection. 

_No response_

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs
Fixes  #997 

this is untested, not set up for soft dependency checking and not time series specific. If we want it in we should use TODS and put it in anomaly detection


TonyBagnall:
just looking and none of the original authors are involved with aeon. @lmmentel @patrickzib most involved in annotation, any thoughts on this?

lmmentel (dismissed):
Makes sense to me. Seems like interfacing TODS is a better fit than pyod and is a separate chunk of work.

patrickzib (dismissed):

lmmentel (approved):

TonyBagnall:
oops, forgot to wait to see it passes, will monitor, it was just updated from main, not changed

PR #1042, opened 2024-01-12
[ENH] Tidy up the utils directory

TonyBagnall:
see #1094 

there is a lot of old legacy code clutering up utils, this tidies it up

1. slope_and_trend.py: two private methods, slope not used anywhere, trend only used in theta forecaster. Removes slope, moved private method and test to theta.py
2. profiling.py: very simplistic method for time profiling a classifier. Not used or useful. Removed
3. datetime.py: method get_time_index is only used in transformations.collections.segmenter, but _coerce_duration_to_int is used all over, leave this one
4. estimator_checks.py: single method called check_estimator that runs tests on a single estimator. Quite useful, but should probably be in the testing module. Leave for now
5. mflow_aeon.py. "The ``mlflow_aeon`` module provides an MLflow API for ``aeon`` forecasters." Just going to leave this here
6. multiindex.py. Single function flatten_multiindex used in transformations and datatypes. leave
7. sampling.py. stratified_resample used in benchmarking/experiments. Move there
8. seasonality.py. autocorrelation_seasonality_test, used only in transformations.detrend.ConditionalDeasonalizer. Can move ther
9. sklearn.py. Tests if a sklearn estimator. USed in the overloaded operators. Not going to mess with that atm
10. stats.py. Weighting functions used in performance functions and forecasting ensemble. Rename?


MatthewMiddlehurst:
are there mote changes to come or just these for now?

TonyBagnall:
just these for now, lost the will to go further

MatthewMiddlehurst (approved):
You may want to add the remaining points to an issue so they are not lost. Will revisit if you want to add more.

MatthewMiddlehurst (approved):

PR #1043, opened 2024-01-13
[MNT] Add `pytest` plugin to retry failed tests

MatthewMiddlehurst:
Somewhat of a reoccurring issue with our test suite is instability. Some of these are legitimate failures with sporadically failing tests, but a majority are just nodes crashing and causing the whole set to fail.

This would make the runs a lot more stable, but the downside of the above is it would rerun some of the legitimate failures, and they may get lost instead of being noticed and fixed. It's a bit hacky.

The main cause of these crashes from what I have gathered is `pytest-xdist` which runs the tests in parallel, and removing it would make everything a lot slower. One of the chosen plugins features is to be able to recover from hard crashed when `pytest-xdist` is installed.

One potential mitigator would be to automatically parse the output logs and create issues for re-runs which aren't crashes, but that's just an idea, and I'm not sure how much effort this would take or if it is possible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ dev = [
     "pytest-randomly",
     "pytest-timeout",
     "pytest-xdist",
+    "pytest-rerunfailures",
     "wheel",
     "boto3",  # mlflow related
     "botocore",  # mlflow related
@@ -173,6 +174,7 @@ addopts = '''
     --timeout 600
     --showlocals
     --numprocesses auto
+    --reruns 2
 '''
 filterwarnings = '''
     ignore::UserWarning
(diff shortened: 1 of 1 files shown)


TonyBagnall:
does it retry just once? Dont want it to keep rechecking when the fail is genuine

MatthewMiddlehurst:
The time spent re-running legitimate failures would be relatively small IMO unless the test takes multiple minutes.

Thats not so relevant anymore though, I played around with this on `tsml-eval` and you can configure it to only work with specific failures. I have added an option to limit this to the `xdist` failures we usually see and verified it works in similar scenarios.

MatthewMiddlehurst:
The plugin cant record the traceback/output for restarts currently, so this is the best option IMO.

TonyBagnall (approved):

PR #1045, opened 2024-01-13
[ENH] Reorganise mock estimators in utils

### LINKED ISSUE #1046: [ENH] Tidy up the estimators used to test interface compliance

we use dummy or mock estimators in testing in several modules. Forecasting estimators are currently in util.estimators. Classifiers are defined in the testing routines. In the spirit of unification, these should be in one place and adopt standard naming 

refactor MockForecasters and _TestingClassifiers to the new testing module, then repeat for regression and clustering 

_No response_

### Additional context

_No response_

TonyBagnall:
Fixes #1046 
this PR centralises some code with similar logic and purpose used in testing to the new testing package. 

It does the following:

1. Moves a function "construct_dispatch" from its own file to the single place it is used (_reduce.py)
2. Combines the mock forecaster base and implementation files into a single file, _mock_forecasters dont think they need to be separate
3. Creates a new package in testing called mock_estimators, moving mock_forecasters there
4. Moves the testing classes in classification base class testing into a file mock_classifiers in the same directory.
5. Removes the now empty utils.estimators package


TonyBagnall:
@ltsaprounis what do you think?

ltsaprounis (approved):
LGTM! 

I'm not a big fan of the MockEstimator name (although admittedly I was the one who made the first one 🤣 ) and I think DummyEstimator and then something like LoggingMixIn would be better, but as long as we all know what it means I don't really mind.

Issue #1016, opened 2023-12-30
[ENH] Centralise data generation code


TonyBagnall:
### Describe the feature or idea you want to propose

Simulated data is created at various locations in the toolkit. I think it should be centralised. This is a precursor to porting in data simulators for different types of classification problems that is currently in the java tsml package.

### Describe your proposed solution

I think it should all be refactored to a single package in utils called data_generation. 

Locations

utils
datatypes
annotation
clustering

TonyBagnall:
fixed by #1105

PR #1047, opened 2024-01-14
[REF] Refactor `utils._testing` to `testing.utils`

MatthewMiddlehurst:
These are currently a bunch of private utilities made for testing. This refactor centralises them in the `testing` directory. I imagine a lot of this is going to change eventually, but good to have it all in one place for when it does.


TonyBagnall (approved):
as you say, this will need further e.g. #1016 but it makes sense to do this first

TonyBagnall (approved):

PR #1050, opened 2024-01-15
[ENH,REF] Refactor `visualisation` and add temporal importance curves function

MatthewMiddlehurst:
Refactors the `visualisation` package into `series`, `results` and `estimator` categories. Splits the different `results` functions into separate files based on the plot type. Moves a function for plotting collections from the clustering file to the `series` package.

Adds a function for plotting temporal importance curves, previously removed from classification.

This all needs better documentation and testing, but would prefer to leave that for another PR.


TonyBagnall:
I was about to ask about testing, but then saw #1051

TonyBagnall (approved):

TonyBagnall:
did you change the API?

MatthewMiddlehurst:
The imports should still be the same, but i just noticed it would not matter because there is no vis page currently on the API 🙂. I can get to that, currently doing the rounds in the module

Issue #1052, opened 2024-01-15
[BUG] REDCOMETS crashes on UWaveZ 


TonyBagnall:
### Describe the bug

REDCOMET classifier crashes on the one UWave dataset, but not on others. Some form of sampling issue maybe? @zy18811  any ideas?

### Steps/Code to reproduce the bug

```python
from aeon.classification.dictionary_based import REDCOMETS
from aeon.datasets import load_classification
trainX, trainy = load_classification("UWaveGestureLibraryZ", split="TRAIN")
testX, testy = load_classification("UWaveGestureLibraryZ", split="TEST")

red = REDCOMETS()
red.fit(trainX, trainy)
acc = red.score(testX,testy)
print("REDCOMETS accuracy: ", acc)
```

### Expected results

prints accuracy

### Actual results

```python-traceback
2024-01-15 14:38:54.269164: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX AVX2
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
1/1 [==============================] - 0s 361ms/step
1/1 [==============================] - 0s 332ms/step
1/1 [==============================] - 0s 330ms/step
1/1 [==============================] - 0s 340ms/step
WARNING:tensorflow:5 out of the last 5 calls to <function Model.make_predict_function.<locals>.predict_function at 0x00000252D72ADE50> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has reduce_retracing=True option that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/guide/function#controlling_retracing and https://www.tensorflow.org/api_docs/python/tf/function for  more details.
1/1 [==============================] - 0s 327ms/step
Traceback (most recent call last):
  File "C:\Code\aeon\venv\lib\site-packages\IPython\core\interactiveshell.py", line 3460, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-dde97c65104f>", line 1, in <module>
[...]

PR #1057, opened 2024-01-15
[BUG] Fixed missing z-normalisation for SFA in RED CoMETS

zy18811:
#### Reference Issues/PRs
Stops problem in #1052 affecting _REDCOMETS()_ but does not fix the root issue. 

RED CoMETS should z-normalise before applying both SAX and SFA but the current implementation only normalises for SAX. This PR changes how the RED CoMETS implementation does z-normalisation so that it is correctly applied for both. 

No

##### For all contributions

### DIFF
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -17,6 +17,7 @@ from sklearn.ensemble import RandomForestClassifier
 from sklearn.model_selection import cross_val_score
 from sklearn.neighbors import NearestNeighbors
 from sklearn.utils import check_random_state
+from sklearn.preprocessing import scale
 
 from aeon.classification.base import BaseClassifier
 from aeon.transformations.collection.dictionary_based import SAX, SFA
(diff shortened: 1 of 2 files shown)


aeon/classification/dictionary_based/_redcomets.py
```diff
@@ -397,14 +400,15 @@ def _predict_proba_unvivariate(self, X) -> np.ndarray:
 
         Parameters
         ----------
-        X : 3D np.ndarray of shape = [n_instances, n_channels, series_length]
+        X : 2D np.ndarray of shape = [n_instances, series_length]
```
TonyBagnall, line 403: soz to be that guy, but our standard format for docs is 
X : np.ndarray
            2D np.ndarray of shape ``(n_instances, n_timepoints)``

trying to conform across the package

TonyBagnall: second line should be indented, not sure why it isnt in the comment

TonyBagnall (changes requested):
I class this as a bugfix, so no need to deprecate imo

aeon/classification/dictionary_based/_redcomets.py
zy18811: think I've now updated all of the np.ndarrays in docstrings in latest commit

TonyBagnall: great, thanks

TonyBagnall (approved):

PR #1070, opened 2024-01-18
[MNT] Set upper bound for `tensorflow`

MatthewMiddlehurst:
CI is currently failing with `tensorflow` related errors. Sets an upper bound for the dependency.


hadifawaz1999 (approved):
thanks !

PR #1051, opened 2024-01-15
[ENH] Testing and basic notebooks for series and learning task specific plotting functions

MatthewMiddlehurst:
Relies on #1050 

More visualisation refactoring.

Adds testing and basic notebooks for functions in the `series` and `learning_task` directories. 

Reworks the function in the collections plotting file and adds a new one to plot by class value.

PR #1071, opened 2024-01-18
[ENH] Testing for estimator specific plotting functions

MatthewMiddlehurst:
Relies on #1051

Adds tests and performs a general clean-up for estimator specific visualisation functions. Adds a function used to generate temporal importance curves to the interval forest base class.


TonyBagnall (approved):
Probably easiest (for me) if I read through these notebooks once they are in. I think Ali has quite a lot to add to vis :)

PR #1018, opened 2023-12-30
[ENH] Part 1 of removing functors from performance metrics

TonyBagnall:
See #1017 #769 

This switches all uses of the performance metric objects for function calls.  The context here is I do not think anyone involved with aeon is willing to maintain this code. We have to own this code and shed our over engineered legacy. I would rather strip it back and make it more maintainable, then react to any user requests/issues as they happen. Bu

The main argument against this is the removal of tags from the functions. These are the tags for the standard performance metrics
    _tags = {
        "requires-y-train": False,
        "requires-y-pred-benchmark": True,
        "univariate-only": False,
    }
and probabilistic ones have these
    _tags = {
        "y_input_type_pred": "pred_quantiles",
        "lower_is_better": True,
    }

I think we can handle these without 2000+ lines of pointless code.

Issue #1017, opened 2023-12-30
[ENH] Remove functors from performance metrics


TonyBagnall:
### Describe the feature or idea you want to propose

currently, aeon.performance_metrics contains functions and functor wrappers for each function. The functors file (aeon.performance_metric.forecasting._classes) is 2300+ lines long. Each object is callable through operator overloading and used as a Callable. It seems to be completely pointless and used in two places in the code base. Just another example of excessive use of OO.

see #769 

### Describe your proposed solution

I think we should strip it out to reduce the bloat and also further isolate the datatypes module. The only complexity in doing so is the probabilistic measures do not have an equivalent function call. 

I think do it in three stages
1. Replace all object calls with function calls, check nothing breaks through full CI
2. Remove all subclasses of _BaseProbaForecastingErrorMetric, and adjust testing
3. Implement the probabilistic subclasses of _BaseProbaForecastingErrorMetric (PinballLoss, EmpiricalCoverage and ConstraintViolation) as functions.

TonyBagnall:
closing as they are all deprecated

Issue #769, opened 2023-09-26
[BUG] Warning in performance_metrics/forecasting/_classes


TonyBagnall:
### Describe the bug

just looking through warnings, we currently generate 127. 72 of them are internal from performance_metrics: 

2023-09-26T13:43:12.0629830Z /Users/runner/work/aeon/aeon/aeon/performance_metrics/forecasting/_classes.py:421: UserWarning: y_pred and y_true do not have the same column index. This may indicate incorrect objects passed to the metric. Indices of y_true will be used for y_pred.

any ideas? I have no idea about this module. 

### Steps/Code to reproduce the bug

run tests

### Expected results

no warnings

### Actual results

72 warnings+

TonyBagnall:
its from here in

    def _check_consistent_input(self, y_true, y_pred, multioutput, multilevel):

```
        if not same_rows:
            warn(
                "y_pred and y_true do not have the same row index. "
                "This may indicate incorrect objects passed to the metric. "
                "Indices of y_true will be used for y_pred."
            )
            y_pred_orig = y_pred_orig.copy()
            if isinstance(y_pred_orig, VectorizedDF):
                y_pred_orig.X.index = y_true.index
            else:
                y_pred_orig.index = y_true.index
        if not same_cols:
            warn(
                "y_pred and y_true do not have the same column index. "
                "This may indicate incorrect objects passed to the metric. "
                "Indices of y_true will be used for y_pred."
            )
            y_pred_orig = y_pred_orig.copy()
            if isinstance(y_pred_orig, VectorizedDF):
                y_pred_orig.X.columns = y_true.columns
            else:
                y_pred_orig.columns = y_true.columns
```

TonyBagnall:
Note this only appears on MacOS
https://github.com/aeon-toolkit/aeon/actions/runs/6956935941/job/18929266765?pr=916
not, for example, here 
https://github.com/aeon-toolkit/aeon/actions/runs/6956935941/job/18929268554?pr=916
hence I cant currently debug this, its very minor though

TonyBagnall:
these have been deprecated in #1081  so this is no longer an issue

PR #1081, opened 2024-01-19
[ENH] Deprecate functors from performance metrics and replace with function calls

TonyBagnall:
(subsumes #1018 which got hopelessly mangled in refactor)

see #1017 #769 

This switches all uses of the performance metric objects for function calls.  The context here is I do not think anyone involved with aeon is willing to maintain this code. We have to own this code and shed our over engineered legacy. I would rather strip it back and make it more maintainable, then react to any user requests/issues as they happen. Bu

The main argument against this is the removal of tags from the functions. These are the tags for the standard performance metrics
    _tags = {
        "requires-y-train": False,
        "requires-y-pred-benchmark": True,
        "univariate-only": False,
    }
and probabilistic ones have these
    _tags = {
        "y_input_type_pred": "pred_quantiles",
        "lower_is_better": True,
    }

I think we can handle these without 2000+ lines of pointless code. 
Deprecates all the class wrappers for performance metric functions


MatthewMiddlehurst (approved):
Reapproving, but I suggest we wait to merge until just before the next release to give time for comment (same with #968).

PR #1087, opened 2024-01-21
[ENH] Remove tensor flow probability, replace with scipy

### LINKED ISSUE #1077: [MNT] `tensorflow-probability` as a dependency
### Describe the issue

`tensorflow-probability` as a dependency is currently only used in a single place in the forecasting package, to simulate a normal distribution. This seems rather unnecessary at first glance, and alternatives should be investigated.

Replace the current usage and `tensorflow-probability` as a dependency.

### Additional context

_No response_

TonyBagnall:
#### Reference Issues/PRs
Fixes #1077 

switches the probability distribution of the base forecaster from tensorflow to scipy and removes the soft dep. 

Also added a test function for test_format_moving_cutoff_predictions for coverage 

the only real problem is the implicit assumption in MLFlow that the object returned by predict_proba has a method quantiles. This hard coded undocumented assumption is in my mind a bad thing. If you want quantiles call predict_quantiles. scipy uses ppf to find a single quantile, and it returns values for every predicted value in the forecasting horizon. tfp behaves differently.

Now we could wrap the returned object in an interface with set methods, or just implement a normal distribution object, but I dont think there is a desire or capacity to do this at this moment in time. Hence I have somewhat bodged ml-flow (see #968) by extracting the same info from scipy normal distribution using ppf that the tfp object returned using quantile. If anyone wants to maintain this or take on making predict_proba a bit more standardised, that would be fine with me.


ltsaprounis (dismissed):
LGTM!

Issue #942, opened 2023-11-29
[DOC] Add examples to TimeSeriesKernelKMeans and TimeSeriesKShapes


TonyBagnall:
### Describe the issue linked to the documentation

these two clusterers lack examples in the doc strings, would be good to have

MatthewMiddlehurst:
I think this is resolved @TonyBagnall?

PR #1091, opened 2024-01-23
[DOC] Add Examples to doc strings for `TimeSeriesKernelKMeans` and `TimeSeriesKShapes`

andregdmitri:
#### Reference Issues/PRs
Solves #942 

Added doc string example for:
-  `TimeSeriesKernelKMeans`
- `TimeSeriesKShapes`

### DIFF
--- a/aeon/classification/dictionary_based/_boss.py
+++ b/aeon/classification/dictionary_based/_boss.py
@@ -687,6 +687,8 @@ class IndividualBOSS(BaseClassifier):
         return new_boss
 
     def _clean(self):
+        if self._transformer is None:
+            return
         self._transformer.words = None
         self._transformer.save_words = False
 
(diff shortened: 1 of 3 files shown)


MatthewMiddlehurst:
Thanks for the PR. The examples seem to be failing the tests. You can run them locally with `pytest`, it will be a lot faster if you isolate the test collection to those files 🙂. Alternatively, you can check on the Checks tab here.

Just taking a quick look at the output, it appears that both of these estimators are not actually in the `clustering` `__init__.py`, which is causing an import failure. I'm not sure whether this is intentional or not, do you @chrisholder? If it's an accident, then I think they are both only accessible through the private file, and should be added to the `__init__.py`.

andregdmitri:
Thanks for the tips, I didn't expect it to fail the tests since I only changed the docstring comments. Now I understand that it also checks  the docstring code that has `>>>`. As you said `TimeSeriesKernelKMeans` and `TimeSeriesKShapes` probably should be in `__init__.py` for `clustring`. I messed up the PR by deleting the branch and will create a new one, sorry about that!

PR #1093, opened 2024-01-23
[ENH] Update `scikit-learn` version

MatthewMiddlehurst:
Updates the sklearn version bound to allow version 1.4.0.


TonyBagnall (approved):
We should have a more automated/documented method of checking deprecations, scikit creates quite a few warnings

PR #1097, opened 2024-01-25
[DOC, REF] Add `TimeSeriesKernelKMeans` and `TimeSeriesKShapes` to `clustering/__init__.py` and also adds examples to dosctring

andregdmitri:
#### Reference Issues/PRs
Solves #942. See also #1091.

Adds `TimeSeriesKernelKMeans` and `TimeSeriesKShapes` to `clustering/__init__.py`.

Adds doc string example for:
- `TimeSeriesKernelKMeans` in file `_kernel_k_means`
- `TimeSeriesKShapes` in file `_k_shapes`

### DIFF
--- a/aeon/clustering/_k_shapes.py
+++ b/aeon/clustering/_k_shapes.py
@@ -45,6 +45,19 @@ class TimeSeriesKShapes(BaseClusterer):
         the sample weights if provided.
     n_iter_: int
         Number of iterations run.
+
+    Examples
+    --------
+    >>> from aeon.clustering import TimeSeriesKShapes
+    >>> from aeon.datasets import load_basic_motions
+    >>> # Load data
+    >>> X_train, y_train = load_basic_motions(split="TRAIN")[0:10]
+    >>> X_test, y_test = load_basic_motions(split="TEST")[0:10]
+    >>> # Example of KShapes clustering
+    >>> ks = TimeSeriesKShapes(n_clusters=3, random_state=1)
+    >>> ks.fit(X_train)
+    TimeSeriesKShapes(n_clusters=3, random_state=1)
+    >>> preds = ks.predict(X_test)
     """
 
     _tags = {
--- a/aeon/clustering/_k_shapes.py
+++ b/aeon/clustering/_k_shapes.py
@@ -54,10 +54,10 @@ class TimeSeriesKShapes(BaseClusterer):
     >>> X_train, y_train = load_basic_motions(split="TRAIN")[0:10]
     >>> X_test, y_test = load_basic_motions(split="TEST")[0:10]
     >>> # Example of KShapes clustering
-    >>> ks = TimeSeriesKShapes(n_clusters=3, random_state=1)
-    >>> ks.fit(X_train)
+    >>> ks = TimeSeriesKShapes(n_clusters=3, random_state=1)  # doctest: +SKIP
+    >>> ks.fit(X_train)  # doctest: +SKIP
     TimeSeriesKShapes(n_clusters=3, random_state=1)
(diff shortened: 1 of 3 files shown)


TonyBagnall:
ah unfortunately there is a problem doing the init for estimators with soft dependencies. iirc if we import in init it means everything in the package then requires the soft dependency to run. Am I remembering correctly @chrisholder @MatthewMiddlehurst ?

aeon/clustering/_k_shapes.py
```diff
@@ -45,6 +45,19 @@ class TimeSeriesKShapes(BaseClusterer):
...
+    >>> from aeon.clustering import TimeSeriesKShapes
+    >>> from aeon.datasets import load_basic_motions
+    >>> # Load data
+    >>> X_train, y_train = load_basic_motions(split="TRAIN")[0:10]
+    >>> X_test, y_test = load_basic_motions(split="TEST")[0:10]
+    >>> # Example of KShapes clustering
+    >>> ks = TimeSeriesKShapes(n_clusters=3, random_state=1)
+    >>> ks.fit(X_train)
+    TimeSeriesKShapes(n_clusters=3, random_state=1)
+    >>> preds = ks.predict(X_test)
```
MatthewMiddlehurst, line 60:
```suggestion
    >>> ks = TimeSeriesKShapes(n_clusters=3, random_state=1)  # doctest: +SKIP
    >>> ks.fit(X_train)  # doctest: +SKIP
    TimeSeriesKShapes(n_clusters=3, random_state=1)
    >>> preds = ks.predict(X_test)  # doctest: +SKIP
```

aeon/clustering/_kernel_k_means.py
```diff
@@ -68,6 +68,19 @@ class TimeSeriesKernelKMeans(BaseClusterer):
...
+    >>> from aeon.clustering import TimeSeriesKernelKMeans
+    >>> from aeon.datasets import load_basic_motions
+    >>> # Load data
+    >>> X_train, y_train = load_basic_motions(split="TRAIN")[0:10]
+    >>> X_test, y_test = load_basic_motions(split="TEST")[0:10]
+    >>> # Example of KernelKMeans Clustering
+    >>> kkm = TimeSeriesKernelKMeans(n_clusters=3, kernel='rbf', random_state=1)
+    >>> kkm.fit(X_train)
+    TimeSeriesKernelKMeans(kernel='rbf', n_clusters=3, random_state=1)
+    >>> preds = kkm.predict(X_test)
```
MatthewMiddlehurst, line 83:
```suggestion
    >>> kkm = TimeSeriesKernelKMeans(n_clusters=3, kernel='rbf', random_state=1)  # doctest: +SKIP
    >>> kkm.fit(X_train)  # doctest: +SKIP
    TimeSeriesKernelKMeans(kernel='rbf', n_clusters=3, random_state=1)
    >>> preds = kkm.predict(X_test)  # doctest: +SKIP
```

MatthewMiddlehurst (changes requested):
That shouldn't be the case, it should only import on the estimator init/function calls. But yes, if it has a soft dependency I think the current workaround for examples is to explicitly skip the portions which use them. See the below suggestions. 

The examples are good, thanks for adding them!

andregdmitri:
I implemented the changes suggested by @MatthewMiddlehurst (thanks!), but I also had to remove `random_state=1`  from `TimeSeriesKernelKMeans.fit( )` so that the Linter would accept the line lenght.

I hope everything is in order now

chrisholder (approved):
Great change, this wasn't initially in init because the tests were crashing for soft depends. Think we since fixed the bug that caused that and forgot to re add it in. Great change!

PR #1101, opened 2024-01-25
[ENH] Remove conversion decorators from SignaturesClassifier

TonyBagnall:
on a hunt to remove redundant code, I found that the private fit/predict/predict_proba were decorated with a converter that switched X into "nested_univ" format. This would then get converted back into standard numpy by the transformers, so its pointless. Removed the decorators and related function.


MatthewMiddlehurst (approved):
Yeah this should be (and it seems is by the lack of failing tests) handled by the base class.

PR #1105, opened 2024-01-25
[ENH] Centralise testing data generation

TonyBagnall:
#### Reference Issues/PRs
see #1016 

It has long bugged me that the test generation is scattered, duplicated and does not conform to a standard. This is a large refactor with zero change of functionality. Once this is done, we can remove redundancy and standardise naming, but I think first just gather it all in the same place. It will no doubt take an age to resolve pycharms awful refactoring


MatthewMiddlehurst (commented):
The `aeon.datasets` ones should probably be deprecated.

If the new file/directory is public that will also need deprecation in the future if you make changes after we make another release.

TonyBagnall:
> The `aeon.datasets` ones should probably be deprecated.
> 
> If the new file/directory is public that will also need deprecation in the future if you make changes after we make another release.

you think dataset changes for test data generation needs deprecation? Seems excessive to me but ok

MatthewMiddlehurst:
For those methods yeah. Everything else is fine. You can add `testing` to the `deprecation.md` list if you want to avoid this for future changes.

Looks like some random dependency is breaking, in the dev dependency in pyproject set `moto` to `<5.0.0.` or just comment out the test, `mlflow` related.

MatthewMiddlehurst (dismissed):

MatthewMiddlehurst (approved):

Issue #1108, opened 2024-01-26
[ENH] Deprecate Filter in transformers


TonyBagnall:
### Describe the feature or idea you want to propose

The Filter class in transform was ported in from aeon-neuro, where it should have stayed. 

### Describe your proposed solution

Deprecate and then link back to neuro in docs more explicitly once it has been worked up a bit

Issue #1086, opened 2024-01-21
[MNT] Review all_extras for soft dependencies.


TonyBagnall:
### Describe the feature or idea you want to propose

I suspect we dont need all of these. 

1. mne is a tool for modelling neuro data, see #1108, #1110
2. h5py was in old benchmarking, can I think go
3. Is scikit_posthocs used anywhere?
4. tensorflow-probability, see #1077 
5. xarray and dask are only used in the datatypes to convert back and forward. Cant see them actually used by an estimator

there are several other single use softdeps,   
all_extras = [
    "cloudpickle",
    "dask",
    "esig>=0.9.7,<0.9.8.3; python_version < '3.11'",
    "filterpy>=1.4.5",
    "h5py",
    "hmmlearn>=0.2.7",
    "gluonts>=0.12.4",
    "keras-self-attention",
    "kotsu>=0.3.1",
    "matplotlib>=3.3.2",
    "mne",
    "pmdarima>=1.8.0",
    "prophet>=1.1.0",
    "scikit_posthocs>=0.6.5",
    "seaborn>=0.11.0",
    "statsforecast>=0.5.2",
    "plotly-resampler>=0.9.0", # from statsforecast, needed for pandas2
    "statsmodels>=0.12.1",
    "stumpy>=1.5.1",
    "tbats>=1.1.0",
    "tensorflow<2.16.0",
    "tensorflow-probability",
    "tsfresh>=0.20.0",
    "tslearn>=0.5.2",
    "xarray",
    "mlflow<2.4.0", # see https://github.com/mlflow/mlflow/issues/8629
    "imbalanced-learn",
]
dl = [
    "keras-self-attention",
    "tensorflow<2.16.0",
    "tensorflow-probability",
]

### Describe your proposed solution

remove unnecessary soft dependencies, reduce code bloat

### Describe alternatives you've considered, if relevant

leave it all in there

TonyBagnall:
closing this as all_extras has moved on, still dont think we need "h5py". Will revisit after forecasting goes

PR #1110, opened 2024-01-26
[ENH] Deprecate Filter transformer

TonyBagnall:
part of #1086 
see #1108 
Filter was ported in from aeon-neuro for unknown reasons, and introduces a soft dependency on mne which we dont really need.


MatthewMiddlehurst:
Fine with this. Deprecation message was left and needs updating.

hadifawaz1999 (approved):
lgtm !

MatthewMiddlehurst (approved):

aeon/transformations/filter.py
```diff
@@ -4,10 +4,18 @@
...
 from aeon.transformations.base import BaseTransformer
 
 
+# TODO: remove in v0.8.0
+@deprecated(
+    version="0.6.0",
+    reason="EAgglo will be removed from annotation in v0.8.0, it has been replaced by "
+    "EAggloSegmenter in the segmentation module.",
+    category=FutureWarning,
+)
```
MatthewMiddlehurst, line 18: Message needs to be changed.

MatthewMiddlehurst (changes requested):
Oops above was an accident.

aeon/transformations/filter.py
TonyBagnall: huh, thought Id done that

MatthewMiddlehurst (approved):

PR #1114, opened 2024-01-26
[MNT] Legacy deprecation

TonyBagnall:
implemented a TODO deprecation from the bad place


ltsaprounis:
> implemented a TODO deprecation from the bad place

Very few people have returned from the bad place. We admire you for your bravery.

PR #1107, opened 2024-01-25
[ENH] Classification `_fit_predict` mockup

MatthewMiddlehurst:
For comparison with #1104
See #1102 

Adds a `_fit_predict` and `_fit_predict_proba` method to the classifier base class. This is intended to replace the `_get_train_probs` used currently in some classifiers.

PR #1116, opened 2024-01-28
[ENH] `BaseClassifier` testing for `fit_predict`/`fit_predict_proba` and classifier conversions to use said methods

MatthewMiddlehurst:
Relies on #1107

1. Fixes a bug in `BaseClassifier` for `fit_predict`.
2. Adds `fit_predict` and similar methods to the classifier testing. Tests both the base class methods, and individual classifiers which sets `capability:train_estimate=True`.
3. Updates the `capability:contractable=True` testing to have more strict requirements while doing the above.
4. Updates current classifiers with `capability:train_estimate=True` to use the new `fit_predict` methods.

Unfortunately, I can't really do 2 without 4, so this got a little larger than I hoped. Updates to previous results should be the result of a swap to random tie breaking rather than the previous method in HC.


TonyBagnall (approved):
lgtm

TonyBagnall (changes requested):
still draft and failing a test

aeon/classification/base.py
```diff
@@ -259,6 +263,11 @@ def fit_predict_proba(self, X, y) -> np.ndarray:
 
         Default behaviour is to estimate probabilities using 10x cross-validation.
         Bespoke behaviour implemented through overriding method _fit_predict_proba.
+        Classifiers which override _fit_predict_proba will have the
+        ``capability:train_estimate`` tag set to True.
+
+        This is very unlikely to be equivalent to calling
+        ``.fit(X, y).predict_proba(X)``.
```
TonyBagnall, line 270: not keen on this comment. I think something more explicit:

"fit_predict produces prediction estimates using just the train data. By default, this is through 10x cross validation, although some estimators may utilise specialist techniques such as bagging. Generally, this will not be the same as fitting on the whole train data then making train predictions. To do this, you should call
fit(X,y).predict(X)"

MatthewMiddlehurst: updated

TonyBagnall (approved):

PR #1122, opened 2024-01-31
[ENH] Some fixes network module

hadifawaz1999:
- Normally code coverage should be 100% for the networks module
-  Fixed bugs when customizing some of the networks
- Added a `plot_network` functionality, its dependent on `pydot` so i added a catching error for it, i would be happy to add it as a soft dependency as well, any comments are welcome


TonyBagnall:
thanks Ali, looks good. Two things
1. I think pydot should be a soft dependency
2. Would it make sense to the plotting functionality to the visualisation module as a function, then just pass the network? You can then remove the need for the soft dep in the base class, as its not really needed to build the network

hadifawaz1999:
i would put it as in the visluazation model yes, thought about it at first, makes more sense once someone else proposed it as well :) thanks @TonyBagnall

TonyBagnall (approved):
LGTM

PR #1145, opened 2024-02-07
[MNT] Fixes deprecation warnings for save_transformed_data in interval based classifiers

TonyBagnall:
sets         save_transformed_data="deprecated" in the interval based classifiers to stop getting warnings from the base class


MatthewMiddlehurst (approved):
Pushed some docstring fixes. Was my mistake originally

Issue #1142, opened 2024-02-07
[ENH] Deprecate datatypes endgame


TonyBagnall:
Ongoing isolation of the datatypes module, almost there.  See #42 #974. This issue is to track whats left that needs doing. This is a summary issue updated on 16/4/24, once #1433 gets merged, there are the following imports outside of the datatypes module.

- [ ] forecasting.base.base.py: convert_to
There are five calls to convert_to, 
             one in predict, update_predict_single,
             two in _check_X_y, one with store_behaviour="reset",
             one in _predict_moving_cutoff, seems simple
            they use the odd parameters of store and store behaviour
- [ ] transformations.base.py: convert_to, check_is_scitype
- [x] transformations.tests.test_base.py: mtype_to_scitype #1434 
- [x] forecasting.tests.test_interval_wrappers: convert_to #1435 
- [x] performance_metrics.forecasting.probabilisic._classes.py: convert_to, check_is_scitype and _VectorizedDF see #1436 

and then have to deal with
- [ ] forecasting.base.base.py: VectorizedDF
- [ ] transformations.base.py: _VectorizedDF

TonyBagnall:
closing this issue, since all datatypes modules are isolated to forecasting and BaseTransformer, both of which will be removed soon

PR #1160, opened 2024-02-10
[ENH] Purge check_is_mtype

TonyBagnall:
see #42 #974 #1142

Replaces the function check_is_mtype with same functionality in the utils directory. 

Only locations were in forecasting and transformations, mostly in the testing.


ltsaprounis (approved):
_vectorization_ changed to _broadcasting_ made me very very happy

PR #1175, opened 2024-02-13
[ENH] Remove use of PinballLoss object

TonyBagnall:
part of #1017 

This removes the two uses of PinballLoss in forecasting, both of which are not needed. One was an example, the other in testing. Also removes a couple of references to loss objects in the comments.


ltsaprounis (approved):
🤠

PR #1180, opened 2024-02-13
[ENH] RIST classifier and regressor

MatthewMiddlehurst:
Implements `RISTClassifier` and `RISTRegressor`, inheriting from a `BaseRIST` class which holds the shared functionality. From [this publication](https://link.springer.com/chapter/10.1007/978-3-031-49896-1_8) of mine.


TonyBagnall (changes requested):
could you run it with codcov and check everything is covered with tests, and if not add them? Thanks

TonyBagnall (approved):
LGTM

Issue #1189, opened 2024-02-16
[MNT] Review `test_sklearn_compatibility.py` tests


MatthewMiddlehurst:
### Describe the issue

I'm not quite sure if some of the tests in `test_sklearn_compatibility.py` make sense currently. I'm fairly certain for the KNN and SVM one using a different set would break it completely at least, as it seems to be just creating a distance matrix and using that to build the estimator.

### Suggest a potential alternative/fix

Review the testing in the file to see if they make sense to keep, expand on them if the functionality is useful.

TonyBagnall:
fixed by #1193 (I think, feel free to reopen @MatthewMiddlehurst )

PR #1193, opened 2024-02-17
[BUG] make metric precomputed in distances sklearn tests

TonyBagnall:
#### Reference Issues/PRs
See #1189 (fixes it I think, @MatthewMiddlehurst)

Sets kernel to precomputed in SVM distance test for pairwise.


MatthewMiddlehurst:
I think the KNN test also has the same issue.

It would be good to test predict with a different dataset and n_cases IMO.

TonyBagnall:
> I think the KNN test also has the same issue.
> 
> It would be good to test predict with a different dataset and n_cases IMO.

changed them round, take a look

MatthewMiddlehurst (approved):
LGTM

PR #1199, opened 2024-02-17
[ENH] Added rocket on gpu

hadifawaz1999:
Added a tensorflow implementation of ROCKET

Added 2 tests with the transformer

### DIFF
--- a/aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
+++ b/aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
@@ -201,9 +201,7 @@ class ROCKETGPU(BaseROCKETGPU):
                     padding=self._list_of_paddings[f],
                 )
 
-                _output_convolution = _output_convolution.numpy().reshape(
-                    (len(batch_indices), len(_output_convolution[0]))
-                )
+                _output_convolution = np.squeeze(_output_convolution.numpy(), axis=-1)
                 _output_convolution += self._list_of_biases[f]
 
                 _ppv = self._get_ppv(x=_output_convolution)
@@ -216,13 +214,37 @@ class ROCKETGPU(BaseROCKETGPU):
                     )
                 )
 
-            output_features.append(np.concatenate(output_features_filter, axis=0))
+            output_features.append(
+                np.expand_dims(np.concatenate(output_features_filter, axis=0), axis=0)
+            )
 
-        output_rocket = (
-            np.concatenate(output_features, axis=0)
-            .reshape(self.n_filters, len(X), 2)
-            .transpose([1, 0, 2])
+        output_rocket = np.concatenate(output_features, axis=0).swapaxes(0, 1)
+        output_rocket = output_rocket.reshape(
+            (output_rocket.shape[0], output_rocket.shape[1] * output_rocket.shape[2])
         )
-        output_rocket = output_rocket.reshape((len(X), self.n_filters * 2))
 
         return output_rocket
+
+    @classmethod
+    def get_test_params(cls, parameter_set="default"):
+        """Return testing parameter settings for the estimator.
+
+        Parameters
+        ----------
+        parameter_set : str, default="default"
+            Name of the set of test parameters to return, for use in tests. If no
+            special parameters are defined for a value, will return `"default"` set.
+
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params = {
+            "n_filters": 5,
+        }
+        return params
--- a/aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
+++ b/aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
@@ -227,7 +227,7 @@ class ROCKETGPU(BaseROCKETGPU):
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):
-        """Return testing parameter settings for the estimator.
+        """Return testing parameter settings for the transformer.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
```diff
@@ -0,0 +1,230 @@
...
+        self.use_dilation = use_dilation
+        self.bias_range = bias_range
+        self.batch_size = batch_size
+        self.random_state = random_state
+
+    def _define_parameters(self):
+        """Define the parameters of ROCKET."""
+        import numpy as np
+
+        np.random.seed(self.random_state)
```
baraline, line 76: Wouldn't it be better to use the new [NumPy generator API](https://numpy.org/doc/stable/reference/random/generator.html) for random here ? Results would surely be the same, but we might avoid some deprecation down the line.

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
```diff
@@ -0,0 +1,230 @@
...
+            collection of time series to transform.
+        y : ignored argument for interface compatibility.
+
+        Returns
+        -------
+        output_rocket : np.ndarray [n_instances, n_filters * 2]
+            transformed features.
+        """
+        import numpy as np
+        import tensorflow as tf
```
baraline, line 184: Any particular reason for the imports to be defined here ? (Not familiar with potential impact of tf import)

baraline (commented):
Apart from these two minor comments, everything LTGM !

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: Yes, its kinda a convention we're following, we did the same for deep learners where we import tensorflow only in methods that use it, so not globally because you can create an instance of the class without importing it, but need it only for model creation inside the instance and then model training. Same thing for rocket-gpu

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: Makes sense actually, thanks for the comment ! will update this

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
```diff
@@ -0,0 +1,228 @@
...
+                    input=X[batch_indices],
+                    stride=1,
+                    filters=self._list_of_kernels[f],
+                    dilations=self._list_of_dilations[f],
+                    padding=self._list_of_paddings[f],
+                )
+
+                _output_convolution = _output_convolution.numpy().reshape(
+                    (len(batch_indices), len(_output_convolution[0]))
+                )
```
baraline, line 206: Careful using numpy reshape, it can sometime lead to undesired change in the array structure. Check if you don't need to use swapaxes instead. For example :
```python
import numpy as np

a = np.array([[1,2,3],[4,5,6]])
print(a.shape)
print(a)
b = a.reshape(3,2)
print(b.shape)
print(b)
c = np.swapaxes(a, 0, 1)
print(c.shape)
print(c)
```
Will give 
```python
(2, 3)
[[1 2 3]
 [4 5 6]]
(3, 2)
[[1 2]
 [3 4]
 [5 6]]
(3, 2)
[[1 4]
 [2 5]
 [3 6]]
```

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
```diff
@@ -0,0 +1,228 @@
...
+                        (np.expand_dims(_ppv, axis=-1), np.expand_dims(_max, axis=-1)),
+                        axis=1,
+                    )
+                )
+
+            output_features.append(np.concatenate(output_features_filter, axis=0))
+
+        output_rocket = (
+            np.concatenate(output_features, axis=0)
+            .reshape(self.n_filters, len(X), 2)
```
baraline, line 223: Same comment on reshape here. Does it do what it should ?

baraline (commented):
Some possible very sneaky bugs, it may not be, but It can be a pain to debug if not checked early on.

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: thanks i actually forgot about this annoying difference :) in this case it wont matter it actually should squeeze axis, its basically transforming from (B, T, 1) to (B, T) so even using np.squeeze is better

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: for this one however yes i think it might be problematic, not sure, but to not risk it will change the code. Thanks @baraline !

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: fixed the issue @baraline

aeon/transformations/collection/convolution_based/rocketGPU/_rocket_gpu.py
hadifawaz1999: fixed this @baraline

baraline (commented):
Potential issues with reshape seem to be fixed ! 
It may a good idea to include a consistency test between CPU & GPU versions (i.e. test that a same-seeded transform output the same results for both versions with an assert_array_almost_equal).
If you feel like this fits into another PR, we can close this one and raise an issue to work on it later.

hadifawaz1999:
> Potential issues with reshape seem to be fixed ! It may a good idea to include a consistency test between CPU & GPU versions (i.e. test that a same-seeded transform output the same results for both versions with an assert_array_almost_equal). If you feel like this fits into another PR, we can close this one and raise an issue to work on it later.

Happy to do both, but maybe putting it as a feature request can be a good first issue for new contributors.

baraline (approved):

PR #1200, opened 2024-02-18
[ENH] Changes to `BaseClassifier` and `BaseRegressor`

### LINKED ISSUE #1178: [ENH] replace get_train_probs for regressors

classification has replace `get_train_preds` with `fit_predict`, would be good to do the same with regression (I think its only implemented for FreshPrinceRegressor).

redesign regressors for fit_predict

_No response_

### Additional context

_No response_

MatthewMiddlehurst:
Closes #1178

Updates `BaseClassifier` and `BaseRegressor`. For the most part this is just adding docs and updating regressor functionality to be similar to that in classification. This probably requires additional tests to be added, but for now it just includes the base class changes.

`BaseClassifier`:
- docstring updates
- adds defaults to certain attributes in `__init__`
- adds `_check_y` to score (without updating `n_classes` etc.)
- removes (IMO) redundant `get_test_params` overload
- changes `n_instances` to `n_cases`

`BaseRegressor`:
- docstring updates to fit the style of `BaseClassifier`
- adds defaults to certain attributes in `__init__`
- deprecates the * overload similar to classification, pipeline to follow
- adds `fit_predict` and associated methods (i.e. `_fit_setup`) similar to classification
- adds `_check_y` to score
- adds to `_check_y` i.e. using `type_of_target(y)`
- changes `n_instances` to `n_cases`
- swaps `score` metric to MSE

Can add or remove from this PR if its too big.


aeon/classification/base.py
```diff
@@ -507,9 +502,51 @@ def _predict_proba(self, X) -> np.ndarray:
...
+            if ``self.get_tag("X_inner_type")`` equals "np-list":
+                list of 2D np.ndarray of shape ``(n_instances)``
+        y : np.ndarray
+            1D np.array of float or str, of shape ``(n_instances)`` - class labels
+            (ground truth) for fitting indices corresponding to instance indices in X.
+
+        Returns
+        -------
+        probabilities: np.ndarray
+            2D array of shape ``(n_cases, n_classes)`` - predicted class probabilities
```
hadifawaz1999, line 545: I think here it should be ``n_instances, n_classes`` to stay consistent with the rest of the doc, same goes for the predict_proba function, unless it was put like this on purpose

hadifawaz1999 (changes requested):
LGTM, just that one comment that actually is valid for many places in the docstrings, unless am mistaken, let me know

aeon/classification/base.py
```diff
@@ -493,10 +487,11 @@ def _predict_proba(self, X) -> np.ndarray:
 
         Returns
         -------
-        y : 2D array of shape [n_instances, n_classes] - predicted class probabilities
-            1st dimension indices correspond to instance indices in X
-            2nd dimension indices correspond to possible labels (integers)
-            (i, j)-th entry is predictive probability that i-th instance is of class j
+        probabilities: np.ndarray
```
dguijo, line 490: sorry for being picky, but there should be an space between probabilities and :

aeon/regression/base.py
```diff
@@ -21,14 +21,20 @@ class name: BaseRegressor
     "BaseRegressor",
 ]
 __author__ = ["MatthewMiddlehurst", "TonyBagnll", "mloning", "fkiraly"]
```
dguijo, line 23: TonyBagnall, though I think we are getting rid of all of this `__author__` stuff.

aeon/regression/base.py
```diff
@@ -141,49 +175,121 @@ def predict(self, X) -> np.ndarray:
         Parameters
         ----------
         X : np.ndarray
```
dguijo, line 177: I was going to leave you a message in the BaseClassifier. There you used "Train data", and here you use the type np.ndarray. 

I'm not sure about the best way, but maybe saying something like `X : np.ndarray or list` may be better. 

Not sure if there is a better nomenclature for when you have two possible options.

aeon/regression/base.py
```diff
@@ -141,49 +175,121 @@ def predict(self, X) -> np.ndarray:
...
+        # this should happen last
+        self._is_fitted = True
+        return y_pred
+
     def score(self, X, y) -> float:
         """Scores predicted labels against ground truth labels on X.
 
         Parameters
         ----------
         X : np.ndarray
```
dguijo, line 264: same here.

dguijo: Not going to write more comments related to this issue to avoid overloading this PR :)

aeon/regression/base.py
```diff
@@ -141,49 +175,121 @@ def predict(self, X) -> np.ndarray:
...
+            R-squared score of predict(X) vs y
         """
-        from sklearn.metrics import r2_score
-
         self.check_is_fitted()
-        if isinstance(y, pd.Series):
-            y = pd.Series.to_numpy(y)
-        y = y.astype("float")
+        y = self._check_y(y, len(X))
         return r2_score(y, self.predict(X))
```
dguijo, line 293: I'm not sure if it is just me or not. But I don't like R2 as the score default measure for regression. I prefer MSE, RMSE or MAE, which are more descriptive imho. Not sure if you aim to change this in this PR though.

dguijo (commented):
Thank you Matt for this PR. Good job. 

Just some minor comments. Thank you!

aeon/classification/base.py
MatthewMiddlehurst: Good spot. This is kinda inconsistent everywhere, but I have swapped it to `n_cases` for the base classes. 

In the same file there is the method `get_n_cases` and the metadata dict use `n_cases` as a key, so it would be odd to deviate from that IMO. Would have to decide which one to use in a larger conversation.

aeon/regression/base.py
MatthewMiddlehurst: yeah gonna leave it for now to avoid conflicts 🙂

aeon/regression/base.py
MatthewMiddlehurst: Included `or list`. Will keep the private methods as they are, as the internal datatype can be other stuff.

aeon/regression/base.py
MatthewMiddlehurst: swapped to MSE

dguijo (approved):
LGTM! Thank you Matt! Good work :)

hadifawaz1999 (approved):
LGTM !

TonyBagnall (approved):

TonyBagnall:
looks good, we should start populating the maintainers maybe though?

MatthewMiddlehurst:
Agreed, but wanted to do it in a separate PR

Issue #1095, opened 2024-01-24
[MNT,DOC] Updates to `CODEOWNERS` and file maintenance


MatthewMiddlehurst:
### Describe the issue

We currently have a `CODEOWNERS` file and `__author__`  at the top of individual python files to denote who has worked on and maintains bits of code. However, `CODEOWNERS` has functional limitations based on the listed users write access to the repository and over time (especially with the fork) the `__author__` list has become outdated and unreliable for maintenance.

A while back in Slack I mentioned replacing the `__author__` list at the top of the file with `__maintainer__`, the contents of which can change as needed and is a reliable source for who to contact if necessary regarding a file. I do not really see the loss of the `__author__` list as major, as they can get very bloated and [commit history](https://github.com/aeon-toolkit/aeon/commits/main/aeon/distances/_distance.py) (very easily available through GitHub) is a much better method for tracking contributions.

### Suggest a potential alternative/fix

Replace cases of `__author__`  with blank `__maintainer__` lists and begin to populate these over time.

### Additional context

Currently, the `CODEOWNERS` file has broken entries as a result of its usage as a maintainer tracker. Keeping maintainers listed elsewhere will allow us to fix these and use the file for its intended purpose (sending out review requests for people who want/need them).

TonyBagnall:
I agree I think, would we also get rid of the CODEOWNERS? I would like to do so. I wonder if there is a tool for collating commit history, auto generate class vs contributor record? Just a random thought

PR #1206, opened 2024-02-19
[MNT,DOC] Remove `__author__` lines for `__maintainer__`

MatthewMiddlehurst:
Implements #1095. See issue for reasoning.

Big annoying PR that replaces `__author__` with blank `__maintainer__` lines.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -52,9 +52,10 @@ Please go through the checklist below. Please feel free to remove points if they
 
 ##### For new estimators and functions
 - [ ] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
+- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed.
 
 ##### For developers with write access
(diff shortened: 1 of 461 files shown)


dguijo:
Hello Matt! Thank you for this PR, I also think it's very needed for organisation. 
However, I think that maybe, as an starting point we can keep the names included in `__author__`, at least from those files that have changed since the split. If the file hasn't changed, we can include an empty list. 

What do you think? 

Thank you again, Matt! Good work.

MatthewMiddlehurst:
It is going to be difficult to determine what file has changed since to split, or how significant a change must be to change the list (since every files has changes most likely due to refactors etc.).

I think a clean slate is easier IMO, but open to suggestions. Showing you have worked on the project is easier (and probably shows more valuable information) through GitHub.

dguijo:
I agree with you. I proposed including some names just not to have the blank page syndrome or the name of the same people in almost all files.

Either way is good IMO. 
Thanks!

baraline (commented):
I agree with the comments of Matthew, if people want to show their contributions, GitHub history will do the trick. We will populate the names as we work through files in future PRs and thanks to it people will know who to contact for questions about specific files.

baraline (approved):

TonyBagnall (approved):

MatthewMiddlehurst:
Will merge, probably some stragglers with the current PRs but can do them later.

Feel free to fill in files as you wish.

PR #1207, opened 2024-02-19
[ENH] Hydra and MultiRocketHydra for regression

MatthewMiddlehurst:
Implements `HydraRegressor` and `MultiRocketHydraRegressor` for regression. There is no publication for these, but the transforms for both are unsupervised. Not much change from the classification version. Tidies up the rocket classes for both classification and regression.


TonyBagnall (approved):
missed this, sorry

PR #1210, opened 2024-02-20
[ENH] Introducing inverse_transform into StandardScaler

dguijo:

Includes the possibility to perform the inverse transform to obtain the original representation. 

Nope

### DIFF
--- a/aeon/transformations/collection/scaler.py
+++ b/aeon/transformations/collection/scaler.py
@@ -1,6 +1,6 @@
 """Wrapper for sklearn StandardScaler."""
 
-__author__ = ["TonyBagnall"]
+__author__ = ["TonyBagnall", "dguijo"]
 __all__ = ["TimeSeriesScaler"]
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


hadifawaz1999 (commented):
Thanks @dguijo ! looks good but could you maybe change the existing example while at the same time continue the example to use the inverse transform ? for instance the example used for now is odd, input 0s and output 0s, maybe better to use a real example ?

aeon/transformations/collection/scaler.py
```diff
@@ -74,35 +79,75 @@ class TimeSeriesScaler(BaseCollectionTransformer):
...
+    def __init__(
+        self, copy=True, with_mean=True, with_std=True, inverse_transform_needed=False
+    ):
         self.with_mean = with_mean
         self.with_std = with_std
         self.copy = copy
         self.scaler = StandardScaler(copy=copy, with_mean=with_mean, with_std=with_std)
+        self.inverse_transform_needed = inverse_transform_needed
+        if self.inverse_transform_needed:
+            self._tags["capability:inverse_transform"] = True
```
MatthewMiddlehurst, line 91: Could you add a check for this in the test? I don't think this is setting correctly to be honest. See https://github.com/aeon-toolkit/aeon/blob/main/aeon/classification/compose/_pipeline.py for a (more complicated) version of dynamic tags

MatthewMiddlehurst (commented):
Given that you must first transform the data to be able to do the inverse transform, is there a case where this will be useful (as you already have the data).

Also I'm surprised this didnt break the testing, I thought transform was supposed to keep the same state. Saving the objects is better suited to fit.

dguijo:
Hello, Matt:
There is some point in having the inverse transform, but maybe not following the way this function is designed ATM. 

Imagine that you have created new synthetic patterns from the training set when it was already standarised and you want to get the original representation of these synthetic patterns to do some plotting or explainability, or whatever. For this,  you need to have the inverse transform. 

However, this can't be achieved now as this transformer is being performed ATM, as it fits a StandardScaler per time series (n_channels x n_timepoints) instead of doing it per channel (n_timeseries x n_timepoints). 

It can be used in time series reconstruction (get the original representation of the reconstructed time series), time series averaging (get the original representation of the averaged time series), and so on. I must admit it is not a typical behaviour in standard machine learning tasks, but I find it useful :smile:.

So, for me, I think we have to give a thought to the current design, and think if the other one (one StandardScaler per channel) is equally useful and then: 1) have both of them, 2) decide which to have.

Thanks!

MatthewMiddlehurst:
I think to do this properly you would have to introduce a `_fit` method and change the `fit_is_empy` tag (perhaps dynamically like you are with inverse transform?).

TonyBagnall:
@dguijo closing this if its ok, it will now be covered here https://github.com/aeon-toolkit/aeon/issues/2075

PR #1231, opened 2024-02-23
[ENH] Add implementation for sbd distance (from k-Shape)

CodeLionX:

Introduces the shape-based distance (SBD) from the k-Shape paper (https://doi.org/10.1145/2949741.2949758) to aeon.

- new `sbd_distance` and `sbd_pairwise_distance` functions with documentation
- updated distance dispatcher functions
- Numba-optimized code

> [!WARNING]
> The current implemenetation computes independent distances for multivariate TS (mean over all channels). 

I compared the implementation of SBD using FFT in Numba's object mode and `np.convolve`:

- time series length < 100: `np.convolve` is faster than FFT
- time series length ~= 100: both are equally fast
- time series length > 100: FFT is significantly faster than `np.convolve` (the speed-up scales with the TS length)

➡️ **always use FFT**

```python
for i in range(1, 6):
    ts1 = np.random.random(10**i)
    ts2 = np.random.random(10**i)
    print(f"With time series size: {10**i}: FFT vs. convolve:")
    %timeit sbd_distance_correlate(ts1, ts2, True)
    %timeit sbd_distance_conv(ts1, ts2, True)
```

```
With time series size: 10: FFT vs. convolve:
90.4 µs ± 5.99 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
5.6 µs ± 519 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

With time series size: 100: FFT vs. convolve:
94.5 µs ± 11.9 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
64.4 µs ± 6.17 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

With time series size: 1000: FFT vs. convolve:
130 µs ± 706 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
1.3 ms ± 9.04 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

With time series size: 10000: FFT vs. convolve:
659 µs ± 4.93 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
89 ms ± 260 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

With time series size: 100000: FFT vs. convolve:
20.8 ms ± 383 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
31.2 s ± 6.54 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

No.

##### For all contributions


aeon/distances/_distance.py
```diff
@@ -222,6 +223,8 @@ def distance(
             kwargs.get("itakura_max_slope"),
             kwargs.get("warp_penalty", 1.0),
         )
+    elif metric == "sbd":
+        return sbd_distance(x, y)
```
chrisholder, line 227: You probably want to pass through the standardize argument:
`return sbd_distance(x, y, kwargs.get('standardize', True)`

aeon/distances/_distance.py
```diff
@@ -385,6 +388,8 @@ def pairwise_distance(
             kwargs.get("itakura_max_slope"),
             kwargs.get("warp_penalty", 1.0),
         )
+    elif metric == "sbd":
+        return sbd_pairwise_distance(x, y)
```
chrisholder, line 392: Again you probably want to pass through the standardize argument:
`return sbd_distance(x, y, kwargs.get('standardize', True)`

chrisholder (changes requested):
Looks great! Great code nicely formatted, fantastic docs and great job figuring the tests out! Two very minor changes just to pass through arguments if used with the utility functions.

aeon/distances/_distance.py
CodeLionX: You are absolutely right. I completely missed that, thanks!

chrisholder (approved):

PR #1232, opened 2024-02-23
[MNT] Suggest changes to #849

MatthewMiddlehurst:
Suggested changes to #849


chrisholder:
I tried using the same keys in a private repo. I get this error:
```
Failed to save: Unable to reserve cache with key numba-test-no-softdeps-full-3.10, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/main, Key: numba-test-no-softdeps-full-3.10, Version: 352f012c1201288c23a77d969304cb2712bf5b5fc77ca03590316fcbf4a85c66
Warning: Cache save failed.'
```
Guess we do need separate keys which is annoying.

PR #1244, opened 2024-02-24
[ENH] BroadcasterTransformer to broadcast a series transformer over a collection

TonyBagnall:
Currently, BaseTransformers can handle both single series and collections by internally "Vectorize" (i.e. broadcast) across channels and instances depending on  the input data. This is legacy code which to me very confusing, involves huge amounts of redundant checks and is not maintainable by the current team. We should imo move away from this model.

This is a quick mock up of my favoured solution, for comment. I've not tested it properly yet, putting it up before I go too far down the line. Basically have a BroadcastTransformer that we pass a BaseSeriesTransformer which is applied over all instances in a collection.

Main points
1. It takes all its tags from the series transformer. These should then be checked in the base class 
2. It does not broadcast over channels. If the SeriesTransformer cannot handle multivariate, nor can the Broadcaster.
3. If fit is empty is true in the SeriesTransformer, then Broadcaster only uses a single instance of the SeriesTransformer. If fit is empty is false, it clones it n_cases times and stores the transformers
4. it calls the private fit/transform of the SeriesTransformer. The assumption is that if it passes checks in the BaseCollectionTransformer, and the tags are taken from SeriesTransformer, then this should be fine. I think this should be true. If its  not, we should make it so.


baraline:
A drawback of our current approach is that a transformer that needs to be fitted could only be broadcasted to a set of samples from the same size. For example if we had `X.shape = (5, 1, 100)` used for fit and then transform, a set `X2.shape = (10, 1, 100)` could not be only transformed. 
But I don't know if this would ever happen considering we are broadcasting a single series transformer, which by essence (if fit is not empty) fits and transform one series.

baraline:
Here is a draft for using joblib, need to add some tests and sort out tags for input format, as for now the example using MatrixProfileSeriesTransformer fails because it didn't convert the series to univariate as expected

baraline:
To-dos left for this PR before it is ready for review:

- [ ] Add new broadcaster transformer to the docs
- [ ] Add an example notebook on how to use the broadcaster for different use cases

aeon/transformations/collection/_broadcaster.py
```diff
@@ -0,0 +1,267 @@
...
+        ----------
+        n_cases : int
+            Number of samples to transform. Used to limit the number of jobs of the
+            broadcaster, which don't need to be higher than the number of instances,
+            as it will loop on them.
+
+        Raises
+        ------
+        ValueError
+            DESCRIPTION.
```
MatthewMiddlehurst, line 111: Needs to be filled or removed.

aeon/transformations/collection/_broadcaster.py
```diff
@@ -0,0 +1,267 @@
...
+        -------
+        Xt : np.ndarray
+            The transformed collection of time series.
+
+        """
+        n_samples = len(X)
+        if y is None:
+            y = [None] * n_samples
+
+        n_jobs_joblib, _ = self._check_n_jobs_broadcast(n_samples)
```
MatthewMiddlehurst, line 193: Is the second half of this output used anywhere?

aeon/transformations/series/_dummy.py
```diff
@@ -0,0 +1,165 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class.
+        """
+        return {}
+
+
+class DummySeriesTransformer_no_fit(BaseSeriesTransformer):
```
MatthewMiddlehurst, line 99: Even though it is for testing use camel case 🙂

aeon/transformations/series/tests/test_matrix_profile.py
```diff
@@ -28,3 +29,22 @@ def test_matrix_profile():
...
+        [
+            [584.0, -11.0, 23.0, 79.0, 1001.0, 0.0, -19.0],
+            [584.0, -11.0, 23.0, 79.0, 1001.0, 0.0, -19.0],
+        ]
+    )
+    mp = MatrixProfileSeriesTransformer(window_length=3)
+    try:
+        mp.fit_transform(series)
+    except ValueError:
+        assert True
```
MatthewMiddlehurst, line 50: These seem like base class tests rather than matrix profile

MatthewMiddlehurst (commented):
So is the essentially intended to be the parallel of https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/_collection_wrapper.py?

I think we should attempt to unify design a bit for these, the collection version currently does not use the series transformer base class. Don't have to change the collection version in this PR of course.

I prefer `CollectionToSeriesWrapper` or similar to the current name.

MatthewMiddlehurst:
included the coverage tag, just to make sure most of it is happening correctly

baraline:
Thanks for reviewing ! Concerning your questions:

> So is the essentially intended to be the parallel of https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/_collection_wrapper.py?

Isn't `_collection_wrapper.py` a Collection to Series while the introduction of this PR is a Series to Collection ? (In the Collection to Series cases, there is no need for parallel I think ?)

> I think we should attempt to unify design a bit for these, the collection version currently does not use the series transformer base class. Don't have to change the collection version in this PR of course.
> I prefer `CollectionToSeriesWrapper` or similar to the current name.

I agree, both Collection to Series and Series to Collection transformers should be close to each other. 
So It would be `SeriesToCollectionWrapper` and `CollectionToSeriesWrapper` (for `_collection_wrapper.py`). As the second falls under deprecation policy, I'll raise an issue and edit it in another PR after the requested changes.

aeon/transformations/collection/_broadcaster.py
```diff
@@ -0,0 +1,267 @@
...
+        transformer: BaseSeriesTransformer,
+        n_jobs: int = 1,
+        joblib_backend: str = "loky",
+    ) -> None:
+        self.transformer = transformer
+        self.n_jobs = n_jobs
+        self.joblib_backend = joblib_backend
+        super().__init__()
+        # Setting tags before __init__() cause them to be overwriten.
+        _tags = {key: transformer.get_tags()[key] for key in self._tags_to_inherit}
```
MatthewMiddlehurst, line 86: Are there tags we dont want to copy? The collection version just does a full clone. perhaps this is better.

MatthewMiddlehurst (commented):
> Isn't _collection_wrapper.py a Collection to Series while the introduction of this PR is a Series to Collection ? (In the Collection to Series cases, there is no need for parallel I think ?)

sorry, i meant that they are essentially opposites 🙂.

aeon/transformations/collection/_broadcaster.py
```diff
@@ -0,0 +1,267 @@
...
+        "fit_is_empty",
+        "X-y-must-have-same-index",
+        "skip-inverse-transform",
+    ]
+
+    def __init__(
+        self,
+        transformer: BaseSeriesTransformer,
+        n_jobs: int = 1,
+        joblib_backend: str = "loky",
```
MatthewMiddlehurst, line 79: Usually we default to `None`, with the parallel `prefer="threads"` (i.e. https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/shapelet_based/_shapelet_transform.py#L380). `"loky"` has some history of being unreliable and crashing in our usage.

baraline: Fine by me for loky, but using threading by default will hurt the performance of some transformer that do no release the GIL, hence the default to process over threads. I guess the better way to deal with this would be to introduce another tag for algorithm that do release the GIL for their computation heavy part. Do we want to go in this direction ?

aeon/transformations/collection/_broadcaster.py
baraline: I've tried doing a full clone in a first version, it leads to some bugs on the input checks, I can find some logs if needed.

aeon/transformations/collection/_broadcaster.py
baraline: It should be passed to the transformer as its `n_jobs` parameters, I seem to have forgotten to include it.

aeon/transformations/series/tests/test_matrix_profile.py
baraline: Indeed, It is more fitted as base class test as it only checks input conversion, I'll move it.

aeon/transformations/collection/_broadcaster.py
MatthewMiddlehurst: Some logs would be good yeah if you have them

aeon/transformations/collection/_broadcaster.py
MatthewMiddlehurst: I am not fully familiar with how each threading methods works to be honest, but we arrived at the current default a fix to commonly occurring crashes rather than efficiency. I think @TonyBagnall did a bit of investigation, but that may have been a while ago.

For now I would stick with the `"threads"` default and have a parameter available to change it i.e. `joblib_backend=None`. I think the best way to handle multithreading and joblib is a larger discussion.

baraline:
Some notes on the updates :

- Removed axis from Matrix Profile as it is univariate
- Added data generation for 1D/2D series in testing utils
- Added tests for BaseSeriesTransformer
- Removed axis as input parameter for Dummies, this parameter should stay fixed as it represent the expected shape for the implemented methods
- Modified dummy to affect random value per channel to test axis conversion
- Renamed Broadcaster to SeriesToCollectionWrapper

aeon/transformations/collection/_broadcaster.py
baraline: If we simply copy the tags from a `DummySeriesTransformer` we have the `SeriesToCollectionWrapper` with the following tags:
```python
 {
    "input_data_type": "Series",
    "output_data_type": "Series",
    "transform_labels": "None",
    "instancewise": True,
    "univariate-only": False,
    "X_inner_type": "ndarray",
    "y_inner_type": "None",
    "requires_y": False,
    "enforce_index_type": None,
    "fit_is_empty": False,
    "X-y-must-have-same-index": False,
    "transform-returns-same-time-index": False,
    "skip-inverse-transform": False,
    "capability:inverse_transform": True,
    "capability:unequal_length": True,
    "capability:unequal_length:removes": False,
    "capability:missing_values": False,
    "capability:missing_values:removes": False,
    "python_version": None,
    "remember_data": False,
    "capability:multithreading": False,
    "capability:univariate": True,
}
```
"input_data_type" and  "output_data_type" already conflict as they should be "Collection" (should be inherited from BaseCollectionTransformer). I also get convertion errors as :
```python
TypeError: Attempting to convert from numpy3D to ndarray but this is not a valid conversion. See aeon.utils.conversion.COLLECTIONS_DATA_TYPE for the list of valid collections
```

baraline: If we copy the tags before `super().__init__()`, we get the correct "input_data_type" and "output_data_type", but then we get other errors as we override tags from the transformer
```python
NotImplementedError: <class 'aeon.transformations.collection._series_wrapper.SeriesToCollectionWrapper'> does not implement inverse_transform
```

aeon/transformations/collection/_series_wrapper.py
```diff
@@ -0,0 +1,346 @@
...
+        n_jobs : int
+            Number of jobs to affect to transformer objects
+
+        Returns
+        -------
+        None.
+
+        """
+        if hasattr(self.transformer, "n_jobs"):
+            self.transformer.n_jobs = n_jobs
```
MatthewMiddlehurst, line 157: I dont think we should edit the attributes of the original transformer if possible. Could we clone it first?

aeon/transformations/collection/_series_wrapper.py
```diff
@@ -0,0 +1,346 @@
...
+    joblib_backend : str, optional
+        Joblib backend to use. The default is "loky".
+
+    Examples
+    --------
+    >>> from aeon.transformations.collection import SeriesToCollectionWrapper
+    >>> from aeon.transformations.series import DummySeriesTransformer
+    >>> from aeon.datasets import load_unit_test
+    >>> X, y = load_unit_test()
+    >>> transformer = SeriesToCollectionWrapper(DummySeriesTransformer())
```
MatthewMiddlehurst, line 50: I think it would be better to use an actual transformer for a docs example if possible

docs/api_reference/transformations.rst
```diff
@@ -91,6 +91,20 @@ Sklearn and pandas adapters
...
+These transformers take a series transformer, which only handles single time series,
+and adapt it to handle collections of time series.
+
+.. currentmodule:: aeon.transformations.collection
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BroadcastTransformer
```
MatthewMiddlehurst, line 106: needs to be updated

MatthewMiddlehurst (commented):
Not sure about the changes to the base class. Think it may need a bit more discussion. For single series y could be forecasting targets or labels for anomaly detection?

aeon/transformations/collection/_series_wrapper.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        n_samples = len(X)
+        if y is None:
+            y = [None] * n_samples
+
+        n_jobs_joblib, n_jobs_transformer = self._check_n_jobs_wrapper(n_samples)
+        self._set_n_jobs_transformer(n_jobs_transformer)
+
+        self.series_transformers = Parallel(
+            n_jobs=n_jobs_joblib, backend=self.joblib_backend
```
MatthewMiddlehurst, line 190: should have `prefer="threads"` (same for others)

aeon/transformations/collection/_series_wrapper.py
```diff
@@ -0,0 +1,346 @@
...
+
+    Parameters
+    ----------
+    transformer : BaseSeriesTransformer
+        The single series transformer to warp accross the collection.
+    n_jobs : int, optional
+        Number of jobs to used. This will be used call the methods of the
+        SeriesTransformer in parallel. The default is 1.
+    joblib_backend : str, optional
+        Joblib backend to use. The default is "loky".
```
MatthewMiddlehurst, line 42: needs to be updated, would default to None. See https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/shapelet_based/_shapelet_transform.py#L74 for some documentation on this and `n_jobs`.

aeon/transformations/collection/_series_wrapper.py
```diff
@@ -0,0 +1,346 @@
...
+            y = [None] * n_samples
+
+        n_jobs_joblib, n_jobs_transformer = self._check_n_jobs_wrapper(n_samples)
+        self._set_n_jobs_transformer(n_jobs_transformer)
+
+        if self.get_tag("fit_is_empty"):
+            # Not Cloning transformer for joblib parallel with threading
+            # might cause some non-efficient parallelism as we call from
+            # the same object. But cloning remove deep params such as
+            # _is_fitted. Processes/Loky would copy it by default.
```
MatthewMiddlehurst, line 299: Does this matter since fit is empty?

MatthewMiddlehurst (commented):
Not read the notebook yet, will revisit.

How does this deal with random number generation with the cloning, does it just copy the seed set by the user (if any)?

aeon/transformations/collection/_series_wrapper.py
baraline: It was simply a matter of correctly parallelizing the transformer. My reasoning was that if the transform function is called from the same object (not cloned), in a threading context, it would lead to threads mostly waiting on the GIL as the same object is used.  I'll do a quick benchmark.

TonyBagnall:
> A drawback of our current approach is that a transformer that needs to be fitted could only be broadcasted to a set of samples from the same size. For example if we had `X1.shape = (5, 1, 100)` used for fit and then transform, a set `X2.shape = (10, 1, 100)` could not be transformed, as we stored the 5 transformer fitted on `X1`. But I don't know if this would ever happen considering we are broadcasting a single series transformer, which by essence (if fit is not empty) fits and transform one series.

I think if the transformer has a _fit function, then it needs to have been called on any case passed. I think calling transform on its own is not allowed, since is_fitted will be false

TonyBagnall:
taking another look here, I'll take a look at tags. Firstly though, I'm not sure about the name series_wrapper. To me at least, a wrapper contains a single object, whereas this broadcasts

TonyBagnall:
also, we enforce calling fit by setting is fitted to true in the base class fit not the constructor

TonyBagnall:
so I think we have some issues with threading about cloning/not cloning and the old problem of where to set n_jobs. I propose that we get it in without threading first, then raise an issue and describe the problems @baraline @MatthewMiddlehurst

baraline:
Agreed, then we can benchmark it to select the best solution

TonyBagnall:
I'll push some changes, but they are open for discussion

TonyBagnall:
there is in fact a problem calling _fit, since the X_inner_type of collections and series are fundamentally different, will unfortunately have to call base class fit

aeon/transformations/collection/_series_wrapper.py
TonyBagnall: delaying parallel until base class settles down

aeon/transformations/collection/_series_wrapper.py
TonyBagnall: delay threading

aeon/transformations/collection/_series_wrapper.py
TonyBagnall: still todo

docs/api_reference/transformations.rst
TonyBagnall: done

aeon/transformations/collection/_series_wrapper.py
TonyBagnall: delay threading

TonyBagnall:
replaced with #1632

PR #1247, opened 2024-02-25
[ENH] Learning Shapelet Classifier

### LINKED ISSUE #1246: [ENH] Implement Learning Shapelet algorithm for Time Series Classification

Learning shapelet is a TSC algorithm with several advantages over the Shapelet Transform algorithm. Currently, aeon doesn't have its implementation but it would be great to have one.

https://www.ismll.uni-hildesheim.de/pub/pdfs/grabocka2014e-kdd.pdf

As discussed on Slack, we can wrap the TSLearn implementation for now and then improve its performance.

_No response_

### Additional context

_No response_

itsdivya1309:
#### Reference Issues/PRs
Fixes #1246 

##### For all contributions


MatthewMiddlehurst:
Wrappers from other packages need some extra considerations. See `MrSQMClassifier` for an example.

aeon/benchmarking/results_loaders.py
```diff
@@ -58,6 +58,11 @@
         "InceptionTimeClassifier",
         "InceptionTimeRegressor",
     },
+    "LearningShapelet": {
+        "LerningShapeletClassifier",
+        "LearningShapelet",
+        "ls",
+    },
```
MatthewMiddlehurst, line 65: We do not have any results uploaded currently, so this does not have to be included.

MatthewMiddlehurst (commented):
Some of these errors are quite strange, let me know if you need a hand.

I have updated the branch to the latest version of main since we were having some CI issues.

MatthewMiddlehurst:
Ah we currently dont support TensorFlow for version 3.8 due to some deprecations and 3.12 because of its recency. Did not think that this would impact tslearn though?

itsdivya1309:
> Some of these errors are quite strange, let me know if you need a hand.
> 
> I have updated the branch to the latest version of main since we were having some CI issues.

Sure, I'll try to figure it out.

MatthewMiddlehurst:
You may have to add it as an additional dependency to the classifier for now in tags. I think SignatureClassifier has an example of a dependency that does not accept all python version

itsdivya1309:
@MatthewMiddlehurst I think the examples are failing because we don't have any results for LearningShapelet. How can I resolve it?

MatthewMiddlehurst:
I think you are correct, we may have to change the notebook. Any thoughts @TonyBagnall?

TonyBagnall:
yep, I have hard coded it to the ones we have, I can run the LS after this is in and add the results. Hopefully merging once this is in will fix the above

https://github.com/aeon-toolkit/aeon/pull/1331

TonyBagnall:
hi @itsdivya1309 if you update from main now it should be fine, thanks for doing this. I havent tried running LS for some time, it will be interesting to see how it performs

TonyBagnall (commented):
sorry for how long this has taken. Things you should add 
get_test_params(
which sets the parameters to be really small, to test interface compliance. look at , for example. RDSTClassifier. 

Things that could be done in a subsequent PR: You might want to add a test function to test the locate function and also to add expected results in the test function.

itsdivya1309:
> sorry for how long this has taken. Things you should add get_test_params( which sets the parameters to be really small, to test interface compliance. look at , for example. RDSTClassifier.
> 
> Things that could be done in a subsequent PR: You might want to add a test function to test the locate function and also to add expected results in the test function.

Thanks for looking into this, I'll add the get_test_params()

aeon/classification/shapelet_based/_ls.py
```diff
@@ -0,0 +1,195 @@
...
+        X_transformed = np.transpose(X, (0, 2, 1))
+    elif X.ndim == 2:
+        X_transformed = np.transpose(X)
+    return X_transformed
+
+
+class LearningShapeletClassifier(BaseClassifier):
+    """
+    Learning Shapelet.
+
```
MatthewMiddlehurst, line 24: Could you add a bit more documentation here above the parameters, at least mentioning that this is a wrapped for the `tslearn` package.

aeon/classification/shapelet_based/_ls.py
```diff
@@ -0,0 +1,195 @@
...
+    Parameters
+    ----------
+    n_shapelets_per_size: dict (default: None)
+        Dictionary giving, for each shapelet size (key),
+        the number of such shapelets to be trained (value).
+        If None, `grabocka_params_to_shapelet_size_dict` is used and the
+        size used to compute is that of the shortest time series passed at fit
+        time.a
```
MatthewMiddlehurst, line 32: Trailing "a" at the end, typo?

aeon/classification/shapelet_based/_ls.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict, default={}
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`.
+        """
+        return {"n_shapelets_per_size": None, "shapelet_length": 0.15}
```
MatthewMiddlehurst, line 195: Generally we want to set these to be fast running parameters. No need to set them if they are the default. I would change `max_iter` to something like 50.

MatthewMiddlehurst (changes requested):
Looking good, very close to done I feel. Thanks for working on this. Few documentation and testing requests.

aeon/classification/shapelet_based/_ls.py
itsdivya1309: Yes :)

aeon/classification/shapelet_based/_ls.py
itsdivya1309: Well `max_iter` does have a default value of 10000, but that would indeed make it slow. So I'll set it to 50.

MatthewMiddlehurst (approved):
Good work, thanks for implementing my requests.

Don't mind adding to the documentation in a separate PR, up to others.

PR #1269, opened 2024-03-01
[MNT] Release 0.7.1

MatthewMiddlehurst:
Release 0.7.1


hadifawaz1999:
Could we hold this off to wait for #1271 ?

MatthewMiddlehurst:
I was planning on doing this tomorrow, so unless it is going to be quick it would be better if it can wait.

hadifawaz1999:
> I was planning on doing this tomorrow, so unless it is going to be quick it would be better if it can wait.

I insisted if we can wait for the PR because its a bug, which means no user can correctly use the random_state, but happy to let it wait if its not possible to get it in before the release

baraline (approved):
Not much to say, LGTM !

PR #1271, opened 2024-03-02
[BUG] Fix random state for deep learning models in classification/regression and clustering

hadifawaz1999:
random state for deep learners was wrongly used, it should set the seed in tf-keras not tf
added a test that runs over all deep learners to make sure the weights are used after training a deep model twice with the same seed


MatthewMiddlehurst:
Please use for informative titles 🙂. Are these all deterministic now with the random state set? If so you should remove the non-deterministic tag from the base class.

hadifawaz1999:
> Please use for informative titles 🙂. Are these all deterministic now with the random state set? If so you should remove the non-deterministic tag from the base class.

by default its always will be non-deterministic, its only when you set the random_state parameter to a specific value then it will be deterministic

The functionality's idea did not chance, its still the same as before, it was just wrongly used in the case we set a random_state, but by default its always None

So the tag should still be there

MatthewMiddlehurst:
The tag is not for the algorithm it is for the code. Unless even with the `random_state` set it will still produce a different result every time, it should not be tagged as such.

hadifawaz1999:
> The tag is not for the algorithm it is for the code. Unless even with the `random_state` set it will still produce a different result every time, it should not be tagged as such.

I think in aeon we had 2 different understanding of the tag then, which is still odd for me because i cant think of an example where no reproducibility is possible, so not sure what "non-determinstic" means now, i always thought it was for models that need random state to be reproducible, and deterministic means models that dont need random state to be reproducible, such as KNN

anyway, tried to remove the tag but tests wont go through cz it supposes that the models can be pickled, which is not the case for deep learners

hadifawaz1999:
@MatthewMiddlehurst let me know of the changes done, i added a detailed explanation of the random state parameter in the docstrings

aeon/classification/deep_learning/_encoder.py
```diff
@@ -60,8 +58,10 @@ class EncoderClassifier(BaseDeepClassifier):
         The name of the file of the last model, if
         save_last_model is set to False, this parameter
         is discarded.
-    random_state : int, default = 0
-        Seed to any needed random actions.
+    random_state : int, default = None
+        Seed to any needed random actions. On CPU its guaranteed to end up with
+            same outcome, but on GPU no given the nature of numerical executions
+            being non deterministic on GPU.
```
MatthewMiddlehurst, line 64: Something like this would better conform to other estimators. Not all of them have this description, but would be good to start standardising.

```
    random_state : int, RandomState instance or None, default=None
        If `int`, random_state is the seed used by the random number generator;
        If `RandomState` instance, random_state is the random number generator;
        If `None`, the random number generator is the `RandomState` instance used
        by `np.random`.
        Seeded random number generation can only be guarenteed on CPU processing,
        GPU processing will be non-deterministic.
```

aeon/classification/deep_learning/_encoder.py
```diff
@@ -167,12 +167,13 @@ def build_model(self, input_shape, n_classes, **kwargs):
...
 
-        tf.random.set_seed(self.random_state)
-
         if self.metrics is None:
             metrics = ["accuracy"]
         else:
             metrics = self.metrics
+
+        if self.random_state is not None:
+            tf.keras.utils.set_random_seed(self.random_state)
```
MatthewMiddlehurst, line 176: Does the function only accept an int? May be worth keeping `rng = check_random_state(self.random_state)`.

hadifawaz1999: yes only int and None, if we use the check function, it will return a random generator, do we re-generated a random state int using this rng you mean ?

MatthewMiddlehurst: yeah thats what i had in mind

hadifawaz1999: done

aeon/classification/deep_learning/_encoder.py
hadifawaz1999: done

TonyBagnall (approved):
LGTM

MatthewMiddlehurst (approved):
there are some small things but i wont delay this any longer

PR #1197, opened 2024-02-17
[DOC] Updates to conform to `ruff` `pydocstyle` hook

AnonymousCodes911:
#### Reference Issues/PRs

from: #1191 , See also: #1055 

#### What does this implement/fix? 
`type()` can't be used to check whether an object is of specific type, thus it was replaced by `isinstance()`

#### Does your contribution introduce a new dependency? If yes, which one?
No

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).

##### For developers with write access
- [ ] Optionally, I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

Issue #1191, opened 2024-02-16
[DOC] Add comments/basic docstrings to testing functions


MatthewMiddlehurst:
### Describe the issue linked to the documentation

[pydocstyle](https://github.com/pycqa/pydocstyle) is deprecated, the recommended replacement is [ruff](https://github.com/astral-sh/ruff).

A lot of testing functions are currently missing the new `pre-commit` requirements introduced by this change. 

### Suggest a potential alternative/fix

Add comments describing the test functions. These can be basic and do not have to all be done in the same PR. 

See #1055 for current PR swapping the `pre-commit` hook and the failing `pre-commit` action on the PR ([this](https://github.com/aeon-toolkit/aeon/actions/runs/7996469823/job/21838996920?pr=1055#step:6:172) at the time of posting) for functions which require fixing.

AnonymousCodes911:
@MatthewMiddlehurst 
 I ran `ruff` and found some inconsistencies with the  PEP 257 docstring conventions, should i raise a PR while addressing the same?

MatthewMiddlehurst:
That would be great. I already have done some fixes in #1055, I will put them in as a separate PR right now so there is no overlap.

AnonymousCodes911:
sure thing!

PR #1055, opened 2024-01-15
[MNT] Replace `pydocstyle` with `ruff`

MatthewMiddlehurst:
[pydocstyle](https://github.com/pycqa/pydocstyle) is deprecated, the recommended replacement is [ruff](https://github.com/astral-sh/ruff).

PR #1275, opened 2024-03-04
[REF] Refactoring some bits of code

AnonymousCodes911:
-`type()` cant be used to check the specific type, thus we use `isinstance()` to check if an object is of a particular type. -`RandomForestCLassifier` and `SummaryTransformer` were not called but imported -Undefined name `l` was replaced by labels

#### Reference Issues/PRs
from: https://github.com/aeon-toolkit/aeon/issues/1191 , See also: https://github.com/aeon-toolkit/aeon/pull/1055
closes: #1197


MatthewMiddlehurst (approved):
lgtm

Issue #1238, opened 2024-02-24
[ENH] Checklist for deprecation in V0.8.0


TonyBagnall:
### Describe the feature or idea you want to propose

todo list for deprecation prior to 0.8.0

### Describe your proposed solution

list

- [x] remove the support_probabilities parameter in WEASELV2. @patrickzib 
- [x] remove plot_series from utils/plotting.py
- [x] remove remove 'save_transformed_data' and 'transformed_data_' in BaseIntervalForest, FreshPrince, STC, Arsenal see #1332 1332
- [x] remove the save_train_predictions parameter in BOSSEnsemble, CBOSS, TemporalDictionaryEnsemble, OrdinalTDE #1333
- [x] remove MatrixProfileClassifier #1333
- [x] refactor uses of get_examples, leave version in datatypes just for use in datatypes.
- [x] swap the position of the classifier and transformers argument in ClassifierPipeline
- [x] EAgglo class in annotation #1306 
- [x] HMM class in annotation #1306 
- [x] remove overloaded operators in BaseClassifier #1265 
- [x] remove overloaded operators in ClassifierPipeline #1265 
- [x] remove BaseForecastingErrorMetric and all subclasses  #1261 
- [x] remove Filter from transformations #1306 
- [x] remove Hidalgo in transformations #1306 
- [x] remove MatrixProfileTransformer from transformations #1306 
- [x] remove mlflow_aeon.py #1285 

going to roll this over
- [x] refactor datatypes._adaptors to utils.conversion

TonyBagnall:
done

PR #401, opened 2023-05-09
[ENH] Classifier pipeline rework, remove operator overload and delegate class

MatthewMiddlehurst:
#### Reference Issues/PRs

See #78 
Resolves #128 

#### What does this implement/fix? Explain your changes.

Reworks the `ClassifierPipeline` class, allowing sklearn transformers and non-list transformer input. Removes usage of `TransformerPipeline` and uses the default methods from `_HeterogenousMetaEstimator`.

This PR deprecates the operator overloading in the `BaseClassifier` class to create pipelines, which IMO does not need to exist and is not intuitive. See #78. 

This PR also removes the unused and private `_DelegatedClassifier` class.

PR #1279, opened 2024-03-06
[ENH] Update regression pipeline

MatthewMiddlehurst:
- Updates the regression module pipeline similar to the classification update recently #401. Deprecated the sklearn pipeline and gives a warning to swap the two parameters.
- Merges the main functionality from both pipelines into `BaseCollectionPipeline`. This could also be used for clustering.
- Some deprecation items from #1238, parameters are swapped and overload is removed. Full deprecation is in #1265.
- Updates the testing for both classification and regression pipelines


TonyBagnall (dismissed):
lgtm

TonyBagnall (approved):

PR #1280, opened 2024-03-06
[ENH] Remove regression delegate file

MatthewMiddlehurst:
Part of #578 (already closed)

Removes the delegate class and file from regression. This has already been removed from classification. The class and file is unused anywhere.


baraline (approved):
The change is straightforward and in line with #578, LGTM !

PR #1281, opened 2024-03-06
[ENH] Tidy dummy estimators for classification and regression

MatthewMiddlehurst:
todo


baraline (commented):
LGTM ! Not related to the changes, but there is for both regressor and classifier `n_instances` instead of `n_cases` in the docstring of the fit methods. Do you want to change them now ?

Edit: NVM it is what is creating the conflict with main probably.

TonyBagnall (approved):
lgtm

Issue #1227, opened 2024-02-23
[BUG] aeon.distances.pairwise_distance does fail for unequal-length time series datasets


CodeLionX:
### Describe the bug

All pairwise distance functions raise an error when presented with a dataset containing unequal-length time series, such as `"japanese_vowels"`. The implementations support unequal-length time series but the wrapping function assumes a single `np.ndarray` as input. However unequal-length datasets have `List[np.ndarray]`-type.

### Steps/Code to reproduce the bug

```python
import aeon
from aeon.datasets import load_japanese_vowels
from aeon.distances import pairwise_distance, get_distance_function_names

print("aeon version", aeon.__version__)
X, y = load_japanese_vowels(split="train")
print("Type of input:", type(X), "shapes=")
print([x.shape for x in X])

error = None
for distance in get_distance_function_names():
    try:
        distance_matrix = pairwise_distance(X, metric=distance)
    except Exception as e:
        error = e
        print(f"Distance '{distance}' failed to process varying-length time series!")

if error is not None:
    print("Last traceback:")
    print(error)
```

### Expected results

All distances functions produce a distance matrix instead of raising an exception.

### Actual results

```
aeon version 0.7.0
Type of input: <class 'list'> shapes=
[...]

PR #1287, opened 2024-03-06
[ENH] SBD and MSM: Support pairwise distance calculation for unequal length time series

CodeLionX:
#### Reference Issues/PRs

First part of resolving issue #1227

Enhances `sbd_pairwise_distance` and `msm_pairwise_distance` to support varying-length time series datasets:

- the `*_pairwise_distance`-functions are not JIT-compiled anymore and just dispatcher functions
- I have a single specialized JITed function for `List[np.ndarray]` (varying-length and equal-length) and inputs are transformed to this type before entering Numba-compiled code.
- I use `numba.typed.List` ([ref](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#typed-list))
- I have new tests for varying-length time series datasets that run conditionally when a function uses `List[np.ndarray]` in its parameter annotations for `X` or `y`.

The SBD implementation for multivariate time series uses the independent algorithm. IMHO, using this way to compute the distance between two multivariate time series with different number of channels is not defined. Thus, I added descriptions to the pydoc. There is no check in the code because of performance.

I have not yet updated the global dispatcher functions (`aeon.distances.pairwise_distance` and `aeon.distances.distance`) because not all distances support the new types yet. I will update the other measure in additional PRs.

No.

##### For all contributions


baraline:
Hi thanks for taking the issue ! Got a quick look at the PR, some of the functions you defined to discriminate between the typed list and array case seem to just exist because of the typing difference in the function arguments, for example : 

```
 def _sbd_pairwise_distance_list( 
     x: NumbaList[np.ndarray], y: NumbaList[np.ndarray], standardize: bool 
 ) -> np.ndarray: 
     n_instances = len(x) 
     m_instances = len(y) 
     distances = np.zeros((n_instances, m_instances))
... 
```
And

CodeLionX:
> Hi thanks for taking the issue ! Got a quick look at the PR, some of the functions you defined to discriminate between the typed list and array case seem to just exist because of the typing difference in the function arguments, [...]
> Only differs by using X. shape[0] instead of len(X), while we could just use len(X). Is there a reason why ? [...]

Yes. That is the only way I could find to teach Numba to compile successfully. Unfortunately, Numba does not like the Python built-in lists and cannot always infer the types of the elements. Additionally, it does not play nice with `isinstance()`-checks within the compiled code when we use `Union[np.ndarray, List[np.ndarray]]` as type for `X`.

Trust me, I first tried

```python
  def _sbd_pairwise_distance_single( 
      x: Union[np.ndarray, List[np.ndarray]], standardize: bool
  ) -> np.ndarray
    n_instances = len(x)
    distances = np.zeros((n_instances, n_instances))

    for i in range(n_instances):
        for j in range(i + 1, n_instances):
            ts1, ts2 = _reshape_pairwise_single(x[i], x[j])  # also with union types for x and y and more complex checking logic
            distances[i, j] = sbd_distance(ts1, ts2, standardize)
            distances[j, i] = distances[i, j]

    return distances
```

If you find a way to get it compiled, please tell me!

baraline:
We stumbled on similar issues with converting some uneven length transformation, but we did found a way to do it in the end. I'll try to do the same for a distance tomorrow and keep you updated.

baraline:
Ok, so we can indeed do it with pure numba, but this involves a lot of modification of the existing code and a restriction of the allowed input shape to 2D/3D for pairwise distance. For example with this function: 

```python
def _msm_from_multiple_to_multiple_distance(
    x: Union[np.ndarray, List[np.ndarray]],
    y: Union[np.ndarray, List[np.ndarray]],
    window: float,
    independent: bool,
    c: float,
    itakura_max_slope: float,
    id_timestamp: int,
) -> np.ndarray:
    n_instances = len(x)
    m_instances = len(y)
    distances = np.zeros((n_instances, m_instances))

    prev_x_shape = -1
    prev_y_shape = -1
    for i in range(n_instances):
        for j in range(m_instances):
            x_size = x[i].shape[id_timestamp]
            y_size = y[j].shape[id_timestamp]
            if x_size != prev_x_shape or y_size != prev_y_shape:
                bounding_matrix = create_bounding_matrix(
                    x_size, y_size, window, itakura_max_slope
                )
                prev_x_shape = x_size
                prev_y_shape = y_size
            if id_timestamp == 0:
                distances[i, j] = _msm_distance(
                    x[i][np.newaxis, :],
                    y[j][np.newaxis, :],
                    bounding_matrix,
                    independent,
                    c,
                )
            else:
                distances[i, j] = _msm_distance(
                    x[i], y[j], bounding_matrix, independent, c
                )
    return distances
```

This involve never calling reshape or any specific numpy function on the collection itself, which modify quite a lot of things and make them unecessarly complicated compared to having a python warper handling some of the input reshaping. This would also need to be tested for performance.

The issue is quite complex is this regard, @chrisholder might have some ideas about this ?

CodeLionX:
Wow, this looks really complicated compared to the existing code and sacrifices a lot of readability. IMHO, only a significant speedup would make a case for this implementation.

What is the focus in Aeon? Maintainability or speed?

I doubt that restricting the input to 2D/3D arrays is a problem at all, though. We currently already check for the shapes of the incoming data in the reshaping-helper methods.

I had another thought: We could also just use the implementations for `List[np.ndarray]` and convert the incoming ndarrays to lists before entering the Numba-code. Within Numba, we iterate over the individual TS anyway. So there would be only a small performance-impact for the Numba-code and just a one-time overhead for the conversion of `np.ndarray` to `List[np.ndarray]` in the dispatcher-method. This means, we can get rid of all the code duplication of my proposed solution (maintainability 🔝), but we need to benchmark the performance.
What do you think about that?

MatthewMiddlehurst:
Hi, we talked about this a bit in a meeting and basically came to the same conclusion as your post above. We dont think the pairwise functions being numba compiled is important, same with accepting 1d input. I think always converting to a list internally is the best solution.

We have a utility function which would be good to use for this (https://github.com/aeon-toolkit/aeon/blob/main/aeon/utils/conversion/_convert_collection.py#L543). This would let you convert from any accepted collection of series datatype to `np-list`.

It would be nice to benchmark performance for equal length series. From what I have gathered speed has been a priority for the module so far, but I would be willing to sacrifice a bit of that (if within reason) for this additional functionality and maintainability.

CodeLionX:
Switch to draft because I have to do some benchmarks first.

CodeLionX:
I performed the requested benchmarks and changed the code in this PR accordingly. The new implementation (red line) is almost the same compared to the current `main` (blue line) but supports variable-length time series. The performance-difference is barely noticeable (in the plots).

The current implementation has the following characteristics:

- Input: either 3D-, 2D-, or 1D-`np.ndarray`s (equal length) or lists of 2D- or 1D-`np.ndarray`s (variable length)
- `<name>_pairwise_distance` is pure Python and performs all checking and reshaping
  - there is a single, simple function `_convert_to_list(X)` that returns a `numba.typed.List` of 2D-`np.ndarray`s
  - conversion is done independent for `x` and `y`
- the internal implementation uses `numba.typed.List`s of `np.ndarray`s independent of the input

Functional changes:

- I had to change `aeon/distances/_distance.py#_custom_func_pairwise`, `aeon/distances/_distance.py#_custom_func_pairwise`, and `aeon/distances/_distance.py#_custom_from_multiple_to_multiple_distance` to make the tests successful (might need another refactoring after all distances have been changed)
- The following documentation example changed the result shape. IMO, the old result was incorrect:

  ```diff
      >>> X = np.array([[[1, 2, 3]],[[4, 5, 6]], [[7, 8, 9]]])
      >>> y_univariate = np.array([[11, 12, 13],[14, 15, 16], [17, 18, 19]])
      >>> msm_pairwise_distance(X, y_univariate)
  -    array([[16.],
  -           [13.],
  -           [10.]])
  +    array([[16., 19., 22.],
  +           [13., 16., 19.],
  +           [10., 13., 16.]])
  ```

  The same applies to the corresponding test in `sbd_pairwise_distance`.

- The reshaping logic and corresponding tests were inconsistent. I.a. when both inputs are 2D, the inputs are assumed to be `(n_cases, n_instances)`:

  https://github.com/aeon-toolkit/aeon/blob/3189363da57fd4f756db69624c0f7085f56c929d/aeon/distances/_utils.py#L35-L39

  However, when one input is 3D and one input is 2D, the 2D-input is assumed to be `(n_channels, n_instances)`:

  https://github.com/aeon-toolkit/aeon/blob/3189363da57fd4f756db69624c0f7085f56c929d/aeon/distances/_utils.py#L47-L53

  Now, we always assume 2D-input to be `(n_cases, n_timesteps)`

- We need to compute the bounding matrix conditionally because time series could have different lengths:

  https://github.com/CodeLionX/aeon/blob/6581292313a6e95559f3a0b80334358d4e884045/aeon/distances/_msm.py#L447-L459

---

## Benchmark Details

Benchmark was performed on a Linux server with 10 cores and 64GB memory using `timeit` and warming up the JIT before running the measurements. You can find the benchmark code here: https://github.com/CodeLionX/aeon/tree/bench/numba-distances/numba-benchmark

I compared the following implementations:

- msm_pairwise_distance_main: implementation from the `main`-branch (no support for variable-length TS)
- msm_pairwise_distance_two_func: first draft of this PR using two independent functions for equal-length and variable-length TS
- msm_pairwise_distance_only_list: converting everything to `List[np.ndarray]` and using the default implementation (reshaping within loops)
- msm_pairwise_distance_custom_only_list: reshaping outside of loops and conditional bounding-matrix-computation
- msm_pairwise_distance_custom_unjit: reshaping in pure python before calling any jited function

scale-collection-size-equal-length
scale-collection-size-variable-length
scale-time-series-length

aeon/distances/_msm.py
```diff
@@ -152,9 +153,9 @@ def msm_cost_matrix(
     distance matrix between :math:`X` and one or more series stored in :math:`y`. If
     :math:`y` is 1 dimensional, we assume it is a single univariate series and the
     distance matrix returned is shape ``(n_cases,1)``. If it is 2D, we assume it
-    is a collection of univariate series with shape ``(m_cases, m_timepoints)``
-    and the distance ``(n_cases,m_cases)``. If it is 3 dimensional,
-    it is assumed to be shape ``(m_cases, m_channels, m_timepoints)``.
+    is a collection of univariate series with shape ``(m_instances, m_timepoints)``
+    and the distance ``(n_cases,m_instances)``. If it is 3 dimensional,
```
chrisholder, line 157: Use m_cases instead of instances

aeon/distances/_msm.py
```diff
@@ -357,12 +357,12 @@ def msm_pairwise_distance(
...
     ----------
-    X : np.ndarray
+    X : np.ndarray or List of np.ndarray
         A collection of time series instances  of shape ``(n_cases, n_timepoints)``
         or ``(n_cases, n_channels, n_timepoints)``.
-    y : np.ndarray or None, default=None
+    y : np.ndarray or List of np.ndarray or None, default=None
         A single series or a collection of time series of shape ``(m_timepoints,)`` or
-        ``(m_cases, m_timepoints)`` or ``(m_cases, m_channels, m_timepoints)``.
+        ``(m_instances, m_timepoints)`` or ``(m_instances, m_channels, m_timepoints)``.
```
chrisholder, line 365: use m_cases

aeon/distances/_sbd.py
```diff
@@ -96,12 +97,14 @@ def sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool = True) -> floa
...
     if x.ndim == 2 and y.ndim == 2:
-        if x.shape[0] == y.shape[0] == 1:
+        if x.shape[0] == 1 and y.shape[0] == 1:
             _x = x.ravel()
             _y = y.ravel()
             return _univariate_sbd_distance(_x, _y, standardize)
         else:
-            # independent
+
+            # if x.shape[0] == y.shape[0]:
```
chrisholder, line 106: Remove?

aeon/distances/_sbd.py
```diff
@@ -111,21 +114,27 @@ def sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool = True) -> floa
...
     ----------
-    x : np.ndarray
+    x : np.ndarray or List of np.ndarray
         A collection of time series instances  of shape ``(n_cases, n_timepoints)``
         or ``(n_cases, n_channels, n_timepoints)``.
-    y : np.ndarray or None, default=None
+    y : np.ndarray or List of np.ndarray or None, default=None
         A single series or a collection of time series of shape ``(m_timepoints,)`` or
-        ``(m_cases, m_timepoints)`` or ``(m_cases, m_channels, m_timepoints)``.
+        ``(m_instances, m_timepoints)`` or ``(m_instances, m_channels, m_timepoints)``.
```
chrisholder, line 137: Uses cases intead of instances

aeon/distances/_utils.py
```diff
@@ -17,8 +18,8 @@ def reshape_pairwise_to_multiple(
         n_timepoints) or
             (n_cases, n_timepoints) or (n_timepoints,).
     y : np.ndarray
-        One or more time series of shape (m_cases, m_channels, m_timepoints) or
-            (m_cases, m_timepoints) or (m_timepoints,)
+        One or more time series of shape (m_instances, m_channels, m_timepoints) or
+            (m_instances, m_timepoints) or (m_timepoints,)
```
chrisholder, line 22: cases instead instances

aeon/distances/tests/test_pairwise.py
```diff
@@ -387,74 +406,75 @@ def test_new_single_to_multiple_distances(dist):
...
-    # # (n_cases, 1, m_timepoints)
-    # _validate_single_to_multiple_result(
-    #     make_series(5, 1, return_numpy=True, random_state=1),
-    #     make_example_unequal_length(5, 1, random_state=2, return_y=False),
-    #     dist["name"],
-    #     dist["distance"],
-    #     dist["pairwise_distance"],
-    # )
-    #
+    # TODO: this is not possible: a 2D-array is interpreted as (n_cases, n_timepoints)
```
chrisholder, line 409: Remove todo

aeon/distances/tests/test_pairwise.py
```diff
@@ -387,74 +406,75 @@ def test_new_single_to_multiple_distances(dist):
...
+            make_example_unequal_length(5, 1, random_state=2, return_y=False),
+            dist["name"],
+            dist["distance"],
+            dist["pairwise_distance"],
+        )
+
+        # # Test passing a singular multivariate time series of shape
+        # # (n_channels, n_timepoints) compared to a collection of unequal length
+        # # multivariate time series of shape (n_cases, n_channels, m_timepoints)
+        # _validate_single_to_multiple_result(
```
chrisholder, line 470: Remove

chrisholder (changes requested):

chrisholder:
PR is awesome and the extra benchmarking work is excellent! Great addition and excited to extend it to all the distances we have!

MatthewMiddlehurst:
Will take a full look soon, but I agree this is a lot of good work and I appreciate the in depth benchmarking.

chrisholder (approved):

MatthewMiddlehurst (approved):

PR #1290, opened 2024-03-09
[ENH] Replace calls to check_is_scitype that return metadata with validate_input

TonyBagnall:
replaces the calls to check_is_scitype that return meta data with a function called validate_input in the utils.validation package. 

Part of #42 #974 #1142

I've run it with the full tests


aeon/transformations/tests/test_base.py
```diff
@@ -3,6 +3,7 @@
 Each test covers a "decision path" in the base class boilerplate,
     with a focus on frequently breaking paths in base class refactor and bugfixing.
 The path taken depends on tags of a given transformer, and input data type.
+Concrete transformer classes from aeon are imported to cover
```
MatthewMiddlehurst, line 6: This seems to be a duplicate

aeon/utils/validation/_input.py
```diff
@@ -0,0 +1,93 @@
...
+        True if input is either a single series, a collection or a hierarchy.
+    """
+    if is_hierarchical(X) or is_collection(X) or is_single_series(X):
+        return True
+    return False
+
+
+def validate_input(X):
+    """Validate input.
+
```
MatthewMiddlehurst, line 42: Quite a lot going on inside the function, would be nice to have a bit more description in the docstring.

MatthewMiddlehurst (approved):
I cannot see any major which is off with the changes, just another part of the big refactor. Good work 🙂.

It would be good if someone else could have a look through before it goes in, since this is unfamiliar code.

PR #1292, opened 2024-03-09
[MNT] Changing deep learning save format from `.hdf5` to `.keras`

hadifawaz1999:
given the new deprecation of `.hdf5` format in tensorflow-keras, this PR uses the new `.keras` format


TonyBagnall (dismissed):

MatthewMiddlehurst:
Looks like one was missed in the notebooks 🙂

As we talked about on Slack, if you are aware of the versions which break with `.keras`, it would be good to add a bound in `pyproject.toml` to stop them from being installed.

hadifawaz1999:
> Looks like one was missed in the notebooks 🙂
> 
> As we talked about on Slack, if you are aware of the versions which break with `.keras`, it would be good to add a bound in `pyproject.toml` to stop them from being installed.

Thanks forgot about that line in the notebook, should be good now.
I checked and if my search is good it is tf.2.12 so i added a lower bound, was based on [this](https://blog.tensorflow.org/2023/03/whats-new-in-tensorflow-212.html)

TonyBagnall (approved):

PR #1298, opened 2024-03-11
[ENH] Replace mtype to scitype function

TonyBagnall:
replaces the calls to mytpe_to_scitype with a a function called abstract_types in the utils.validation package.

Part of #42 #974 #1142 

this is a minimal change to get it done, I'm not fixed in this way of doing things but would like to isolate datatypes more.


aeon/utils/validation/_input.py
```diff
@@ -18,6 +18,50 @@
 )
 from aeon.utils.validation.series import is_hierarchical, is_single_series
 
+# String identifiers for the different types of time series data
+COLLECTIONS = ["numpy3D", "nested_univ", "pd-multiindex"]
+SERIES = ["pd.Series", "pd.DataFrame", "np.ndarray"]
+HIERARCHICAL = ["pd_multiindex_hier"]
```
MatthewMiddlehurst, line 24: Seems limited?

aeon/forecasting/base/_base.py
```diff
@@ -1314,8 +1314,8 @@ def _check_missing(metadata, obj_name):
         # retrieve supported mtypes
         y_inner_type = _coerce_to_list(self.get_tag("y_inner_type"))
         X_inner_type = _coerce_to_list(self.get_tag("X_inner_type"))
-        y_inner_abstract_type = mtype_to_scitype(y_inner_type, return_unique=True)
-        X_inner_abstract_type = mtype_to_scitype(X_inner_type, return_unique=True)
+        y_inner_abstract_type = abstract_types(y_inner_type)
+        X_inner_abstract_type = abstract_types(X_inner_type)
```
MatthewMiddlehurst, line 1318: What was the previous output vs new? looks like it may have consequences for all forecasters if it is different i.e.

```
req_vec_because_rows = y_type not in y_inner_abstract_type
req_vec_because_cols = (
   self.get_tag("y_input_type") == "univariate"
   and not y_metadata["is_univariate"]
)
requires_vectorization = req_vec_because_rows or req_vec_because_cols
```

MatthewMiddlehurst (commented):
Im wary about this one. Seems like it could have hidden consequences given the differences between the new and old.

aeon/utils/validation/_input.py
TonyBagnall: You want to support other types? These are the core types

MatthewMiddlehurst: The contents of the lists seem to be missing bits though? i.e. `np-list` for one

TonyBagnall: irrc this is mirroring the existing datatypes functionality which does not support np-list, its really just trying to stop using datatypes. Ill have a look

aeon/forecasting/base/_base.py
TonyBagnall: so abstract_types returns a list of either "Panel", "Series", "Hierarchical" depending on the inner types. This is essentially what mtype_to_scitype does, with lots of nonsense. This is only meant to work with Panel, Series, and Hierarchical according to the documents, since        
 y_inner : Series, Panel, or Hierarchical object

MatthewMiddlehurst (approved):
Failures seem unrelated

MatthewMiddlehurst (approved):

PR #1308, opened 2024-03-12
Revert "[ENH] Refactor dask converters"

chrisholder:
Reverts aeon-toolkit/aeon#1302


MatthewMiddlehurst:
should be fixed by #1309, do not think a revert is required

Issue #1094, opened 2024-01-23
[ENH] Tidy up the utils directory


TonyBagnall:
### Describe the feature or idea you want to propose

there is a lot of old legacy code clutering up utils, this tidies it up. Some done by #1042

### Describe your proposed solution

- [x] slope_and_trend.py: two private methods, slope not used anywhere, trend only used in theta forecaster. Removes slope, moved private method and test to theta.py
- [x] profiling.py: very simplistic method for time profiling a classifier. Not used or useful. Removed
datetime.py: method get_time_index is only used in transformations.collections.segmenter, but _coerce_duration_to_int is used all over, leave this one
- [x] estimator_checks.py: single method called check_estimator that runs tests on a single estimator. Quite useful, but should probably be in the testing module. #1313  
- [x] mflow_aeon.py. "The mlflow_aeon module provides an MLflow API for aeon forecasters." See #968 #1285 
- [x] multiindex.py. Single function flatten_multiindex used in transformations and datatypes. leave
- [x] sampling.py. stratified_resample used in benchmarking/experiments. Move there
- [x] seasonality.py. autocorrelation_seasonality_test, used only in transformations.detrend.ConditionalDeasonalizer. Can move ther
- [x] sklearn.py. Tests if a sklearn estimator. USed in the overloaded operators. Not going to mess with that atm
- [x] stats.py. Weighting functions used in performance functions and forecasting ensemble. Rename?

itsdivya1309:
Hi!
I would like to work on the sampling.py and seasonality.py files.

TonyBagnall:
all done

PR #1312, opened 2024-03-14
[MNT] Move estimator_checks to the Testing Module

aadya940:
As per #1094 , Sub-Issue 3.

Move estimator_checks.py to testing module.
Change necessary import paths.


TonyBagnall (approved):
LGTM, thanks for this

MatthewMiddlehurst (changes requested):
This should be deprecated first. You can keep the changes, but include the old file as well with a warning (the developer guide has instructions on how to do so).

MatthewMiddlehurst (approved):
Thanks, LGTM

Issue #1170, opened 2024-02-12
[ENH] Add more clustering algorithms to base deep clusterer


hadifawaz1999:
### Describe the feature or idea you want to propose

For now deep clustering models handle kmeans clustering algorithm in the latent space only

### Describe your proposed solution

It would be better to add more capability in the base model, such as kshape or kmedoids

PR #1318, opened 2024-03-17
[ENH] Modified `BaseDeepClusterer`

AnonymousCodes911:
fixes: #1170 
added `KShape` and `KMedoids` to `BaseDeepClusterer`

### DIFF
--- a/aeon/clustering/deep_learning/base.py
+++ b/aeon/clustering/deep_learning/base.py
@@ -17,8 +17,13 @@ class BaseDeepClusterer(BaseClusterer, ABC):
     ----------
     n_clusters : int, default=None
         Number of clusters for the deep learning model.
-    clustering_algorithm : str, default="kmeans"
+    clustering_algorithm : str, {'kmeans', 'kshape', 'kmedoids'},
+        default="kmeans"
         The clustering algorithm used in the latent space.
+        Options include:
+        'kmeans' for K-means clustering,
+        'kshape' for KShape clustering,
+        'kmedoids' for KMedoids clustering.
     clustering_params : dict, default=None
         Dictionary containing the parameters of the clustering algorithm chosen.
     batch_size : int, default = 40
--- a/aeon/clustering/deep_learning/base.py
+++ b/aeon/clustering/deep_learning/base.py
@@ -5,6 +5,8 @@ __maintainer__ = []
 from abc import ABC, abstractmethod
 
 from aeon.clustering._k_means import TimeSeriesKMeans
+from aeon.clustering._k_medoids import TimeSeriesKMedoids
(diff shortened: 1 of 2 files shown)


aeon/clustering/deep_learning/base.py
```diff
@@ -116,6 +124,16 @@ def _fit_clustering(self, X):
                 self.clusterer = TimeSeriesKMeans(
                     n_clusters=self.n_clusters, **clustering_params_
                 )
+        elif self.clustering_algorithm == "kshape":
+            self.clusterer = TimeSeriesKShapes(
+                n_clusters=self.n_clusters, **clustering_params_
+            )
```
MatthewMiddlehurst, line 130: The new `elif` and the above code should be similar, if the `if len(clustering_params_.keys()) == 0:` line is not needed it should be removed.

aeon/clustering/deep_learning/base.py
```diff
@@ -116,6 +124,16 @@ def _fit_clustering(self, X):
...
+        elif self.clustering_algorithm == "kmedoids":
+            self.clusterer = TimeSeriesKMedoids(
+                n_clusters=self.n_clusters, **clustering_params_
+            )
+        else:
+            raise ValueError(f"Invalid algorithm: {self.clustering_algorithm}")
```
MatthewMiddlehurst, line 136: could you add the parameter name in the error somewhere, i.e. "Invalid input for clustering_algorithm"

aeon/clustering/deep_learning/base.py
```diff
@@ -15,8 +17,13 @@ class BaseDeepClusterer(BaseClusterer, ABC):
     ----------
     n_clusters : int, default=None
         Number of clusters for the deep learning model.
-    clustering_algorithm : str, default="kmeans"
+    clustering_algorithm : str, {'kmeans', 'kshape', 'kmedoids'},
+        default="kmeans"
         The clustering algorithm used in the latent space.
+        Options include:
+        'kmeans' for K-means clustering,
```
MatthewMiddlehurst, line 24: `K-means`  to `KMeans` if we want to keep the style consistent

MatthewMiddlehurst (commented):
Could you add a test in `test_deep_clusterer_base.py` for your new options? The input can be a parameter for the test.

aeon/clustering/deep_learning/base.py
AnonymousCodes911: cleaned the code!

MatthewMiddlehurst:
The test should be new, not an edit the the current test.

hadifawaz1999 (approved):
lgtm

PR #1323, opened 2024-03-18
[ENH] Capability to use shape-dtw on precomputed transformation

hadifawaz1999:
The PR adds a capability to use shape-dtw on pre-computed transformation.

This is for 2 reasons:

1- suppose someone wants to use a transformation we dont have in aeon
2- this is a pre-refactor to fix the ShapeDTWClassifier because it assumes that shapeDTW(x,y)=DTW(f(x),f(y)) which is wrong and this capability will make the refactor much easier


baraline (commented):
I got a couple questions (that might be due to me not being familiar with this piece of code): 

- Couldn't you replace the inputs (`x,y`) by their transformed counterparts (what you call `transformed_x`) and keep the boolean `transformation_precomputed` to flag that these inputs (`x,y`) are already transformed instead of introducing  `transformed_x, transformed_y` ? (Maybe a numba problem ?)

- Can you handle the case where only one of the two inputs is precomputed ? It doesn't seem the case if I understood the code correctly, but doesn't require much change to support it. (Don't know if this can be useful)

Other than that, new functionality is clear and tested against default behaviour so LGTM.

hadifawaz1999:
> I got a couple questions (that might be due to me not being familiar with this piece of code):
> 
>     * Couldn't you replace the inputs (`x,y`) by their transformed counterparts (what you call `transformed_x`) and keep the boolean `transformation_precomputed` to flag that these inputs (`x,y`) are already transformed instead of introducing  `transformed_x, transformed_y` ? (Maybe a numba problem ?)
> 
>     * Can you handle the case where only one of the two inputs is precomputed ? It doesn't seem the case if I understood the code correctly, but doesn't require much change to support it. (Don't know if this can be useful)
> 
> 
> Other than that, new functionality is clear and tested against default behaviour so LGTM.

thanks for the suggestions @baraline , but actually both cannot be added for the following reason:

When calculating shapeDTW, the first step is always to do the transformation, the 2nd step is to find the dtw path between transformed series, and the 3rd step, which is the most the important one, is we transfer the path back to the original series and loop over the distance matrix between the raw series and find the new dtw value on raw series.

So every time you need the raw and transformed series, cannot be one of the other.

baraline (approved):
Ok thanks I understand, in that case I would agree on the PR. As I'm not as familiar as I would want to with the distance module, i'd rather have another review from @chrisholder or @TonyBagnall, to make sure I didn't overlook something 🙃

aeon/distances/_shape_dtw.py
```diff
@@ -119,6 +119,9 @@ def shape_dtw_distance(
     descriptor: str = "identity",
     reach: int = 30,
     itakura_max_slope: float = None,
+    transformation_precomputed: bool = False,
+    transformed_x: np.ndarray = None,
```
chrisholder, line 123: Optional[np.ndarray]

aeon/distances/_shape_dtw.py
```diff
@@ -217,13 +232,27 @@ def _shape_dtw_distance(
...
 ) -> float:
-    new_x = _transform_subsequences(x=x, descriptor=descriptor, reach=reach)
-    new_y = _transform_subsequences(x=y, descriptor=descriptor, reach=reach)
-
-    shape_dtw_cost_mat = _dtw_cost_matrix(
-        x=new_x, y=new_y, bounding_matrix=bounding_matrix
-    )
+    # for compilation purposes
+    if transformed_x is None:
+        transformed_x = x.copy()
```
chrisholder, line 241: Why the copy are you mutating the array in the function somewhere?

aeon/distances/_shape_dtw.py
```diff
@@ -264,6 +293,9 @@ def shape_dtw_cost_matrix(
     descriptor: str = "identity",
     reach: int = 30,
     itakura_max_slope: float = None,
+    transformation_precomputed: bool = False,
+    transformed_x: np.ndarray = None,
```
chrisholder, line 297: Optional[np.ndarray]

aeon/distances/tests/test_shape_dtw.py
```diff
@@ -0,0 +1,34 @@
...
+    shape_dtw_precomputed_transformation = shape_dtw_distance(
+        x=X[0],
+        y=X[1],
+        reach=10,
+        transformation_precomputed=True,
+        transformed_x=transformed_x,
+        transformed_y=transformed_y,
+    )
+
+    assert shape_dtw_default == shape_dtw_precomputed_transformation
```
chrisholder, line 34: Not a huge fan of having this as a separate file can you either add it to the test_numba_distance_parameters. Ideally if you can add it to the actual dist_param dict and add the results:

```
DIST_PARAMS = {
    "dtw": BASIC_BOUNDING_PARAMS,
    "erp": BASIC_BOUNDING_PARAMS + [{"g": 0.5}, {"g": None}],
    "edr": BASIC_BOUNDING_PARAMS + [{"epsilon": 0.5}],
    "lcss": BASIC_BOUNDING_PARAMS + [{"epsilon": 0.5}],
    "ddtw": BASIC_BOUNDING_PARAMS,
    "wdtw": BASIC_BOUNDING_PARAMS + [{"g": 1.0}],
    "wddtw": BASIC_BOUNDING_PARAMS + [{"g": 1.0}],
    "twe": BASIC_BOUNDING_PARAMS + [{"lmbda": 0.5}, {"nu": 0.9}],
    "msm": BASIC_BOUNDING_PARAMS + [{"independent": False}, {"c": 0.1}],
    "adtw": BASIC_BOUNDING_PARAMS + [{"warp_penalty": 5.0}],
    "minkowski": [{"p": 1.0}, {"p": 2.0}],
    "sbd": [{"standardize": False}],
}
```
If you need help with doing this message me on slack it's a bit annoying to do.

chrisholder (changes requested):

aeon/distances/_shape_dtw.py
hadifawaz1999: done, thanks for the tip !

aeon/distances/_shape_dtw.py
hadifawaz1999: done

aeon/distances/_shape_dtw.py
hadifawaz1999: makes sense not to copy, given its only to avoid numba fails

aeon/distances/tests/test_shape_dtw.py
hadifawaz1999: it would not work actually
because its not only the flag that is tested, if the flag is true both transformations np.ndarrays should be added as well in the function
so i dont think it is compatible with your setup, in your case its really just a parameter, in my case its a whole array that depends on the input

chrisholder (approved):
Changes have been addressed great new addition!

chrisholder (approved):
I caused some merge issues so fixed them and looks good

PR #1327, opened 2024-03-19
[MNT] Add new dependency: typing-extensions

chrisholder:

This PR proposes to add a new dependency typing-extensions. 

This PR adds typing-extensions: https://pypi.org/project/typing-extensions/.

"The typing_extensions module serves two related purposes:

Enable use of new type system features on older Python versions. For example, typing.TypeGuard is new in Python 3.10, but typing_extensions allows users on previous Python versions to use it too.
Enable experimentation with new type system PEPs before they are accepted and added to the typing module.
typing_extensions is treated specially by static type checkers such as mypy and pyright. Objects defined in typing_extensions are treated the same way as equivalent forms in typing."

This dependency is maintained by the Python core team and can be found at: https://github.com/python/typing_extensions

#### Any other comments?
The reason I would like to add this as a core dependency is it will allow us to type our modules much better and will be able to use all the typing features that aren't avaliable in earlier versions of python.

##### For all contributions

- [x] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,12 +6,8 @@ build-backend = "setuptools.build_meta"
 name = "aeon"
 version = "0.7.1"
 description = "A toolkit for conducting machine learning tasks with time series data"
-authors = [
-    {name = "aeon developers", email = "contact@aeon-toolkit.org"},
-]
-maintainers = [
-    {name = "aeon developers", email = "contact@aeon-toolkit.org"},
-]
+authors = [{ name = "aeon developers", email = "contact@aeon-toolkit.org" }]
+maintainers = [{ name = "aeon developers", email = "contact@aeon-toolkit.org" }]
 readme = "README.md"
 keywords = [
     "data-science",
@@ -53,6 +49,7 @@ dependencies = [
     "pandas>=1.5.3,<2.1.0",
     "scikit-learn>=1.0.0,<1.5.0",
     "scipy>=1.2.0,<1.13.0",
+    "typing-extensions>=4.10.0",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -95,8 +92,8 @@ dl = [
     "tensorflow-addons; python_version < '3.12'",
 ]
 unstable_extras = [
-    "mrsqm>=0.0.1,<0.1.0 ; platform_system == 'Darwin'",  # requires gcc and fftw to be installed for Windows and some other OS (see http://www.fftw.org/index.html)
-    "pycatch22<=0.4.3",  # known to fail installation on some setups
+    "mrsqm>=0.0.1,<0.1.0 ; platform_system == 'Darwin'", # requires gcc and fftw to be installed for Windows and some other OS (see http://www.fftw.org/index.html)
+    "pycatch22<=0.4.3",                                  # known to fail installation on some setups
```
MatthewMiddlehurst, line 96: not sure why we have gained some white space here

pyproject.toml
```diff
@@ -53,6 +49,7 @@ dependencies = [
     "pandas>=1.5.3,<2.1.0",
     "scikit-learn>=1.0.0,<1.5.0",
     "scipy>=1.2.0,<1.13.0",
+    "typing-extensions>=4.10.0",
```
MatthewMiddlehurst, line 52: This bound seems unnecessarily tight unless there are features in this version which are required. There would only be a single version available for install currently I think?

MatthewMiddlehurst (commented):
If this is just porting new type hint functionality to older versions as extras I'm fine with it. It should not impact how anyone who installs has to interact with their code.

Will give more time for others to comment.

pyproject.toml
chrisholder: I think maybe changing it to 4.0 and above since i believe thats for python 3.12 features

pyproject.toml
chrisholder: I had some weird auto formatting going on will revert all these whitespaces changes

pyproject.toml
chrisholder: I had a look through their release notes to find the lowest patch I think is appropriate to support everything we may want and landed on 4.6.0 and above

chrisholder:
I also should note the addition to the .gitignore is intentional and is a config for a static type checker I run locally that can now make use of the new types

MatthewMiddlehurst (approved):
LGTM. Would wait a bit for others to give input though.

TonyBagnall (approved):
I'm fine with this

baraline (approved):
More typing features without Python version restriction is great. The downside of having an additional dependency, which is well maintained and active, does not seem like a real problem.
A nice addition may be to add some example in the developer documentation on how we want to use these features in aeon. But that might be redundant with the documentation of the typing modules.

Issue #1172, opened 2024-02-12
[ENH] BaseCollectionTransformers fit_is_empty tag


TonyBagnall:
### Describe the feature or idea you want to propose

the BaseCollectionTransformer by default has tags 
```python

    # tag values specific to CollectionTransformers
    _tags = {
        "input_data_type": "Collection",
        "output_data_type": "Collection",
        "fit_is_empty": False,
        "requires_y": False,
        "capability:inverse_transform": False,
    }
```
if a subclass does not implement fit, it should set the tag
```pyth
    _tags = {
        "fit_is_empty": True,
    }
```
thus overriding the base class tag. By doing this, less checking is done in base class fit, and its easier to search for transformers. Some do not reset this tag. 

### Describe your proposed solution

add this tag to any base collection transformers that do not implement `_fit`

quick look, may be more:
- [ ] PeriodogramTransformer
- [ ] TimeSeriesScaler
- [ ] ARCoefficientTransformer

but there may be more. Just see how its done in, for example HOG1DTransformer

PR #1266, opened 2024-03-01
Tidy up the utils directory

itsdivya1309:
#### Reference Issues/PRs
Fixes parts of #1094 

#### What does this implement/fix? Explain your changes.

- [x] sampling.py. stratified_resample used in benchmarking/experiments. Move there
- [x] seasonality.py. autocorrelation_seasonality_test, used only in transformations.detrend.ConditionalDeasonalizer. Can move ther

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed.

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #1328, opened 2024-03-21
[ENH] Enhancing `BaseCollectionTransformers`+Tidying `Utils` dir

AnonymousCodes911:
Fixes : #1172 
added `"fit_is_empty": True` to
* ar_coeffient.py
* periodogram.py
* TimeSeriesScaler

Part of : #1094 
- renamed `stats.py` from aeon/utils


MatthewMiddlehurst:
Like #1266, any public `stats.py` methods will need to be deprecated first.

MatthewMiddlehurst:
Are there any other transformers which do not have `_fit` but are missing the tag? We should check before closing #1172.

AnonymousCodes911:
> Like #1266, any public `stats.py` methods will need to be deprecated first.

should it be done in a separate PR or should I try implementing it here only?

AnonymousCodes911:
> Are there any other transformers which do not have `_fit` but are missing the tag? We should check before closing #1172.

@MatthewMiddlehurst, I think now it covers the entire `aeon/transformations`, please review.

AnonymousCodes911:
In `subset.py`>`IndexSubset` we have the  ` "remember_data": True,` tag so according to the tests it should have the `"fit_is_empty":False`, but currently _fit is not implemented in it, so should we use something like?

```python
_fit(self, X, y=None):   
    self._X = X.copy()    
    return self
```

MatthewMiddlehurst:
You can undo the change to `aeon/utils/stats.py`, the deprecation only applies to public methods.

Please isolate the changes to `collections/` for now and don't edit the transformer base class. You do not need to add the tag if it matches the default `"fit_is_empty": False,`.

AnonymousCodes911:
@MatthewMiddlehurst are the changes in line with the requirements?

TonyBagnall (approved):
This all looks good to me, thanks

MatthewMiddlehurst (approved):

PR #1332, opened 2024-03-23
[DEP]  remove deprecated 'save_transformed_data' and 'transformed_data_' in BaseIntervalForest, FreshPrince, STC,

TonyBagnall:
Deprecation task #1238 

Not sure about whether I can just remove the None returned by _fit_estimator, will see if it is ok as is first.


MatthewMiddlehurst:
@TonyBagnall dont think i should review this 🙂

chrisholder (approved):
LGTM

PR #1339, opened 2024-03-23
[ENH] Subgradient barycentre averaging

chrisholder:
#### Reference Issues/PRs
This PR relies on PR #1326 and #1327

This PR implements a different way to compute a barycentre using stochastic subgradient. More details of which can be found in the paper: https://www.sciencedirect.com/science/article/abs/pii/S0031320317303163. arxiv version: https://arxiv.org/abs/1701.06393

The high level overview of subgradient barycentre averaging is instead of doing a exhaustive optimisation by looping through every time series in the collection and updating the barycentre gradually, the subgradient approach randomises the collection order and uses a much more aggressive optimisation allowing it to converge much quicker. This does lead to a reduction in the quality of barycentre produced but is significantly faster. 

I will be running experiments in the near future to measure this computation time vs quality of barycentre average produced.

The subgradient supports all the same parameters as the original barycentre average including support for all the same distances.

In addition this PR cleans up the barycentre module and adds a utility function that allows for either the original implementation of the subgradient approach to be easily callable. In addition more tests have been added both for subgradient and the original barycentre average.

I have also ensured the results are the same as tslearn (for dtw as that is the only distance it supports) for correctness.

#### Any other comments?
I did some benchmarking comparing the implementation to tslearn (only for dtw as tslearn only supports dtw) and these were the results for large a collection of time series of shape: n_cases=500, n_channels=1, n_timestamps=500:
```
  Aeon petitjean average time: 20.97852702140808 seconds
  Aeon Subgraident average time: 1.4203758716583252 seconds
  Tslearn petitjean average time: 51.35894193649292 seconds
  Tslearn Subgradient average time: 5.135100698471069 seconds
```

Each approach was run 10 times and the average time was computed.

##### For all contributions

- [x] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed.


chrisholder:
In addition to these change I have also added a new parameter to barycentre averaging called 'init_barycentre' which is the initial barycentre that is refined. Previously it was using the medoids that yields the 'best' results however, was very computationally expensive for minor performance gains. As such I have now added 'mean', 'random' and passing a np.ndarray. 

The 'mean' option takes the mean average of the collection as the initial centre. This is actually recommend by the original paper and is signficantly computationally cheaper than the medoids so has now been set as the new default. 

'random'  selects a random time series as the initial

Finally you can also pass a np.ndarray which is the initial time series for anyone who wants to control the init further.

TonyBagnall:
fantastic work, but have you done any sanity checks for correctness against tsdlearn @chrisholder ? Just one a few, does it produce sensible clusters, as accurate as tslearn etc?

aeon/clustering/_k_means.py
```diff
@@ -8,8 +8,8 @@
 from numpy.random import RandomState
 from sklearn.utils import check_random_state
 
-from aeon.clustering.averaging import _resolve_average_callable
-from aeon.clustering.averaging._barycenter_averaging import VALID_BA_METRICS
```
TonyBagnall, line 12: can you do this through init? how come it is  in from aeon.clustering.averaging._ba_utils import VALID_BA_METRICS
also?

aeon/clustering/averaging/_ba_petitjean.py
```diff
@@ -0,0 +1,147 @@
...
+import numpy as np
+from numba import njit
+
+from aeon.clustering.averaging._ba_utils import (
+    _get_alignment_path,
+    _get_init_barycenter,
+)
+
+
+def petitjean_barycenter_average(
```
TonyBagnall, line 12: I'm not sure I like the naming, firstly I would avoid naming after author unless its standard and secondly, center is spelt incorrectly :) Can this not just be barycentre_average ro barycentre_average_full?

TonyBagnall (dismissed):

aeon/clustering/averaging/_ba_petitjean.py
chrisholder: I think the name petitjean is the only real way to distinguish the difference between the two since it is his method of estimating a barycentre. We also had a conversation about centre vs center and I decided for the package I was going to use center so kwargs and parameters to other packages (like sklearn) is consistent and not confusing.

chrisholder:
I have checked equivalence to tslearn using dtw subgradient. I also compared the averages produced visually which as you can see they seem very sensible and close to what DBA/MBA produces:
image
(DBA/MBA is petitjean BA and SSG is stochastic sub gradient BA)

TonyBagnall (approved):
great stuff

PR #1343, opened 2024-03-26
[ENH] Anomaly detection base class and MERLIN detector

MatthewMiddlehurst:
#### Reference Issues/PRs

Part of #722

This PR implements a base class for anomaly detectors extending from `BaseSeriesEstimator` (can provide more info regarding this base class if requested).  Also implements a basic version of the `MERLIN` algorithm as a test case. Includes testing for the base class and classes which extend from it.

Currently, only takes a single series, either univariate or multivariate. Anomaly detectors return a 1D numpy array which contains booleans, ints for floats. This can be either a true/false for each time point in the input series being an anomaly, or an "anomaly score" for each time point.

`_fit` is optional based on the `fit_is_empty` tag, as well as `y` input for `fit` using the `requires_y` tag.

No

##### For all contributions

- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed.


aeon/anomaly_detection/base.py
```diff
@@ -0,0 +1,263 @@
...
+                    "Error in input type for y: y input as np.ndarray should be 1D."
+                    + req_msg
+                )
+
+            # check valid dtype
+            fail = False
+            if issubclass(y.dtype.type, np.integer):
+                new_y = y.astype(bool)
+                fail = not np.array_equal(y, new_y)
+            elif not issubclass(y.dtype.type, numpy.bool_):
```
CodeLionX, line 223:
```suggestion
            elif not issubclass(y.dtype.type, np.bool_):
```

aeon/anomaly_detection/base.py
```diff
@@ -0,0 +1,263 @@
+"""Abstract base class for time series anomaly detectors."""
+
+__maintainer__ = ["MatthewMiddlehurst"]
+__all__ = ["BaseAnomalyDetector"]
+
+from abc import ABC, abstractmethod
+from typing import final
+
+import numpy
```
CodeLionX, line 9: Should be unused because you also import `import numpy as np`.

aeon/anomaly_detection/_merlin.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+
+        for i in range(X.shape[0] - self.min_length + 1):
+            if std(X[i : i + self.min_length]) > AEON_NUMBA_STD_THRESHOLD:
+                warnings.warn(
+                    "There is region close to constant that will cause the results "
+                    "will be unstable. It is suggested to delete the constant region "
+                    "or try again with a longer min_length.",
+                    stacklevel=2,
+                )
```
CodeLionX, line 76: Nice 👍🏼  We experienced this _close to constant problem_ in some of our experiments as well. This is an annoyance of many matrix profile-based algorithms.

aeon/anomaly_detection/_merlin.py
```diff
@@ -0,0 +1,184 @@
...
+
+        anomalies = np.zeros(X.shape[0], dtype=bool)
+        for i in indicies:
+            if i != np.nan:
+                anomalies[i] = True
+
+        return anomalies
+
+    @staticmethod
+    def _drag(X, length, discord_range):
```
CodeLionX, line 118: Should we JIT this as well? Looks quite compute heavy.

CodeLionX (approved):
Nice work 👍🏼 I think the proposed API sounds reasonable, and we can accommodate many existing detectors with it.

I did not check the correctness of MERLIN.

aeon/anomaly_detection/_merlin.py
MatthewMiddlehurst: done, not many changes needed.

aeon/anomaly_detection/_merlin.py
MatthewMiddlehurst: Seems a bit sensitive given it does not seem like an uncommon occurrence, but guess as long as theres a warning should be fine 🙂.

MatthewMiddlehurst:
> I did not check the correctness of MERLIN.

I'm still not 100% either, but think it is close. The original code does some strange stuff and uses MATLAB data structures which were a bit difficult to understand. Only had a quick go through really. Will have to thoroughly test at some point, but just an example for the API for now.

CodeLionX (approved):

CodeLionX (approved):

Issue #1349, opened 2024-03-27
[DOC] Correcting the documentation of the `random_state` parameter


hadifawaz1999:
### Describe the issue linked to the documentation

Given the problem solved in #1271 we should make sure in all modules if there exists a miss-explanation of the `random_state` parameter in the docstrings

hx4nn:
hey @AnonymousCodes911, can you assign this to me..

AnonymousCodes911:
Hi @hx4nn, I am already in the process of doing this 😅

hx4nn:
sure 👍

hadifawaz1999:
i think we can close this now and re-open if we find something again or disagree

PR #1350, opened 2024-03-27
[DOC] Fixing `random_state` in docstrings

AnonymousCodes911:
fixes: #1349 
adds a general docstring for `random_state`

### DIFF
--- a/aeon/base/estimator/hybrid/base_rist.py
+++ b/aeon/base/estimator/hybrid/base_rist.py
@@ -80,8 +80,6 @@ class BaseRIST(metaclass=ABCMeta):
         If `RandomState` instance, random_state is the random number generator;
         If `None`, the random number generator is the `RandomState` instance used
         by `np.random`.
-        Seeded random number generation can only be guaranteed on CPU processing,
-        GPU processing will be non-deterministic.
     n_jobs : int, default=1
         The number of jobs to run in parallel for both `fit` and `predict`.
         ``-1`` means using all processors.
--- a/aeon/base/estimator/hybrid/base_rist.py
+++ b/aeon/base/estimator/hybrid/base_rist.py
@@ -80,6 +80,8 @@ class BaseRIST(metaclass=ABCMeta):
         If `RandomState` instance, random_state is the random number generator;
         If `None`, the random number generator is the `RandomState` instance used
         by `np.random`.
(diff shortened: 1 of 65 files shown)


hadifawaz1999:
I think there was a miss understanding, the idea is not to use the same documentation we used in #1271 , the idea is to check if there is somewhere where the documentation needs more information. For example in the case of deep learning it needed to be clarified about the GPU CPU thing in #1271  but thats not the case everywhere

AnonymousCodes911:
> I think there was a miss understanding, the idea is not to use the same documentation we used in #1271 , the idea is to check if there is somewhere where the documentation needs more information. For example in the case of deep learning it needed to be clarified about the GPU CPU thing in #1271 but thats not the case everywhere

Understood!, will update according to the individual models! also, I think @MatthewMiddlehurst  has given general docstrings to some, should I follow the same suite and add specifics to the modules which need them? 

general docstrings: 
```
    random_state : int, RandomState instance or None, default=None
        If `int`, random_state is the seed used by the random number generator;
        If `RandomState` instance, random_state is the random number generator;
        If `None`, the random number generator is the `RandomState` instance used
        by `np.random`.
```

hadifawaz1999:
> > I think there was a miss understanding, the idea is not to use the same documentation we used in #1271 , the idea is to check if there is somewhere where the documentation needs more information. For example in the case of deep learning it needed to be clarified about the GPU CPU thing in #1271 but thats not the case everywhere
> 
> Understood!, will update according to the individual models! also, I think @MatthewMiddlehurst has given general docstrings to some, should I follow the same suite and add specifics to the modules which need them?
> 
> general docstrings:
> 
> ```
>     random_state : int, RandomState instance or None, default=None
>         If `int`, random_state is the seed used by the random number generator;
>         If `RandomState` instance, random_state is the random number generator;
>         If `None`, the random number generator is the `RandomState` instance used
>         by `np.random`.
> ```

yeah i think this part should be the same everywhere (if the code does that, if not than not) correct me if am wrong @MatthewMiddlehurst

MatthewMiddlehurst:
Be careful, this may not be appropriate for every `random_state`. Generally if it goes through `check_random_state` it should be fine.

AnonymousCodes911:
will be done in #1417

PR #1352, opened 2024-03-27
[MNT] CI add pip dependency caching

chrisholder:
This PR adds further caching to the CI allowing all dependencies and wheels to be cache'd and restored locally greatly speeding up the CI.

One of the main bottle necks for the CI currently is installing dependencies and building wheels. Currently every time the CI is ran dependencies are installed from pip and built. In the 'perfect' scenario this is fine and installs and building of wheels takes around 5 minutes or less. However, due to external factors such as network issues, a bad CI node and other factors sometimes this install and building step can take upwards of 20 minutes. Normally in every CI run, one runner takes 15+ minutes to install and build the dependencies. As the CI is only as fast as its slowest runner this presents a bottleneck.

This PR caches the dependencies and the wheels built each time a PR is merged to main (if pyproject.toml is updated). As a result the CI can the restore the dependencies and wheel on all branches and runs skipping the download and build phase. This reduces the time to install down to around 3 minutes in the best case and in the worse case (due to bad runner) 7 minutes. This greatly speeds up the CI as the slowest runner should be much faster.

##### For all contributions

### DIFF
--- a/.github/actions/numba_cache/action.yml
+++ b/.github/actions/numba_cache/action.yml
@@ -44,12 +44,6 @@ runs:
       run: echo "CURRENT_DATE=$(date +'%d/%m/%Y')" >> $GITHUB_ENV
       shell: bash
 
-      # GNU tar on windows runs much faster than the default BSD tar
-    - name: Use GNU tar instead BSD tar if Windows
-      if: ${{ inputs.runner_os == 'Windows' }}
-      shell: cmd
-      run: echo C:\Program Files\Git\usr\bin>> $GITHUB_PATH
-
     # Restore cache if restore_cache is true
     - name: Restore cache
       uses: actions/cache/restore@v3
--- a/.github/actions/pip_cache/action.yml
+++ b/.github/actions/pip_cache/action.yml
@@ -21,12 +21,6 @@ runs:
       run: echo "CURRENT_DATE=$(date +'%d/%m/%Y')" >> $GITHUB_ENV
       shell: bash
 
-      # GNU tar on windows runs much faster than the default BSD tar
-    - name: Use GNU tar instead BSD tar if Windows
-      if: ${{ inputs.runner_os == 'Windows' }}
-      shell: cmd
-      run: echo C:\Program Files\Git\usr\bin>> $GITHUB_PATH
-
       # For some reason on windows I/O on the C drive is really slow. To get around this
       # we set the python, and the cache directories to the D drive which doesn't
       # have slow I/O.


TonyBagnall:
is this sped up? 
image

chrisholder:
The caching doesn't exist until merged onto main (as the cache's need to be made on main)

chrisholder:
This is the usage on this branch:
image

Overall the caching saves a hour over all tests

.github/workflows/pr_pytest.yml
```diff
@@ -104,6 +104,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Set optimal environment variables for pip cache and restore
+        uses: ./.github/actions/pip_cache
+        with:
+          runner_os: ${{ runner.os }}
+          python_version: ${{ matrix.python-version }}
+          restore_cache: "true"
```
MatthewMiddlehurst, line 112: Any reason for these two jobs only?

.github/actions/pip_cache/action.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: Set Python environment for linux
+      if: ${{ inputs.runner_os == 'Linux' }}
+      run: echo "PIP_CACHE_DIR=~/.cache/pip" >> $GITHUB_ENV
+      shell: bash
+
+    - name: Set Python environment for macos
+      if: ${{ inputs.runner_os == 'macOS' }}
+      run: echo "PIP_CACHE_DIR=~/Library/Caches/pip" >> $GITHUB_ENV
+      shell: bash
```
MatthewMiddlehurst, line 54: Can you format this a bit like the release if/else? https://github.com/aeon-toolkit/aeon/blob/main/.github/workflows/release.yml#L69

Does not really matter I guess, just style.

.github/workflows/update_pip_cache.yml
```diff
@@ -0,0 +1,79 @@
...
+          path: ${{ env.PIP_CACHE_DIR }}
+          key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashfiles('**/pyproject.toml') }}
+
+      - name: Install CPU version of pytorch on linux
+        uses: nick-fields/retry@v3
+        if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Linux'
+        with:
+          timeout_minutes: 30
+          max_attempts: 3
+          command: python -m pip install torch --index-url https://download.pytorch.org/whl/cpu
```
MatthewMiddlehurst, line 51: Why only Linux, are there no benefits for other OS?

.github/workflows/update_pip_cache.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Install aeon and dependencies
+        uses: nick-fields/retry@v3
+        if: steps.cache.outputs.cache-hit != 'true'
+        with:
+          timeout_minutes: 30
+          max_attempts: 3
+          command: python -m pip install .[all_extras,dev]
+
+      # Install notebook if on linux 3.10 in addition
+      - name: Install notebook if on linux 3.10
```
MatthewMiddlehurst, line 62: This could potentially cause issues maybe? Are `python -m pip install .[all_extras,dev] python -m pip install .[binder]` and `python -m pip install .[all_extras,binder,dev]` going to produce the same results always?

.github/workflows/update_pip_cache.yml
```diff
@@ -0,0 +1,79 @@
+name: Update pip cache on merge to main
+on:
+  push:
+    branches:
+      - main
```
MatthewMiddlehurst, line 5: Maybe set it so this can be run manually.

MatthewMiddlehurst (changes requested):
> This PR caches the dependencies and the wheels built each time a PR is merged to main (if pyproject.toml is updated).

The majority of dependencies (even more for patch versions) will install new versions without any update to `pyproject.toml`. Our main indication that there is a breakage on fresh installations is the general CI failing due to an incompatible dependency update. 

This change currently could potentially delay a response. Even if we catch this on the periodic tests and make an update, there is a potential for a mismatch of dependency versions used in that test and PR tests. Someone could probably pick at some security issues on being behind on versions, but I'm not going to do that.

tl;dr: I think we should update the cache more frequently 🙂.

MatthewMiddlehurst:
Even with the comments in the mega-review, I really like the idea of this and hope it works out. Thanks for the work.

.github/actions/numba_cache/action.yml
```diff
@@ -44,12 +44,6 @@ runs:
       run: echo "CURRENT_DATE=$(date +'%d/%m/%Y')" >> $GITHUB_ENV
       shell: bash
 
-      # GNU tar on windows runs much faster than the default BSD tar
-    - name: Use GNU tar instead BSD tar if Windows
-      if: ${{ inputs.runner_os == 'Windows' }}
-      shell: cmd
-      run: echo C:\Program Files\Git\usr\bin>>"%GITHUB_PATH%"
-
```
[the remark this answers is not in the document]
chrisholder: I moved it into the new action

.github/workflows/update_pip_cache.yml
chrisholder: So pytorch by default on macos and windows will only install the CPU version unless you specify otherwise. Linux on the otherhand it installs everything (GPU version). This is why the Linux cache sizes were over 2GBs whereas windows and macos was like that 600 mbs.

.github/workflows/update_pip_cache.yml
chrisholder: Not sure but I can split them their own action to ensure this.

.github/actions/numba_cache/action.yml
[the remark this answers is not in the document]
chrisholder: I have removed this completely now since it doesn't make a difference now we're using the windows D drive (it seems to have fixed all the cache restoring issues)

.github/workflows/pr_examples.yml
```diff
@@ -30,7 +30,14 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Use numba cache to set env variables but not restore cache
+      - name: Set optimal environment variables for pip cache and restore
+        uses: ./.github/actions/pip_cache
+        with:
+          runner_os: "Linux"
```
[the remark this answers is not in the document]
chrisholder: Fixed

.github/actions/pip_cache/action.yml
chrisholder: Done

.github/workflows/update_pip_cache.yml
chrisholder: Fixed

.github/workflows/update_pip_cache.yml
chrisholder: I made it so it runs at the same time as periodic. The reason they are separate actions is because I don't want to use the cpu version of pytorch in periodic just in case something breaks because of using the cpu vs regular install (it should be identical)

PR #1353, opened 2024-03-27
[BUG] Stop joining Nones

TonyBagnall:
fix incorrect usage of os.join("FOO",None) in edge test configuration

### DIFF
--- a/aeon/datasets/_data_loaders.py
+++ b/aeon/datasets/_data_loaders.py
@@ -1302,7 +1302,7 @@ def load_classification(
                 # Check the status code of the response
                 if response.status != 200:
                     raise ValueError(msg)
-            except HTTPError:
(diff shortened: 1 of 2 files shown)


chrisholder (approved):
LGTM

aeon/datasets/_data_loaders.py
```diff
@@ -1299,7 +1302,7 @@ def load_classification(
                 # Check the status code of the response
                 if response.status != 200:
                     raise ValueError(msg)
-            except HTTPError:
+            except (HTTPError, URLError, TimeoutError):
                 raise ValueError(msg)
```
MatthewMiddlehurst, line 1306: I'm not sure this is the best way to handle this, as it's not really a `ValueError`. If this is a testing issue, I think it would be better to use a try/catch and xfail the test instead.

PR #1365, opened 2024-04-03
[ENH] Add Swale Scoring Model

aadya940:
Related to: #425

Implemented as per [this paper](http://dbgroup.eecs.umich.edu/files/sigmod07timeseries.pdf), Section 5.

Couldn't verify the correctness due to a lack of alternate implementations. I shall continue to work on this once we track down an alternate implementation?


MatthewMiddlehurst:
Is this ready to go? There seems to be a conflict currently. People are less likely to look at drafts 🙂.

aadya940:
@MatthewMiddlehurst From my side this looks ready to go. However, I'm not 100% confident about the logic used. We should check this up against an alternate implementation (which I couldn't track down) if something looks wrong to you. :-)

TonyBagnall:
so the implementation is true to the definition in the paper, but I dont think defining this recursively is a good idea. It would be better if it followed the model used by LCSS. @chrisholder and I will take a look and get back to you

chrisholder (dismissed):

MatthewMiddlehurst:
There seems to be a conflict which has to be resolved, but @chrisholder has approved so this should be quick to merge after given him and @TonyBagnall have no issues with it.

MatthewMiddlehurst:
@aadya940 Not quite sure where we left off with this. Seems like it was close?

aadya940:
@MatthewMiddlehurst  Yes, I guess during this PR some testing suite in distances was revamped. Hence, the testing part is left.

TonyBagnall:
closing, two years old and conflicted, feel free to reopen

PR #1366, opened 2024-04-03
[DOCS] Update the mentoring projects

TonyBagnall:
have done some work on the project pages, it would be good if someone else could proof read it and add anything they think would be good.

### DIFF
--- a/docs/mentoring.md
+++ b/docs/mentoring.md
@@ -27,7 +27,9 @@ via Slack if you are interested in any of these projects or have any questions.
 
 We will more widely advertise funding opportunities as and when they become available.
 
-We list projects by time series task
+Most projects can be extended, possibly into a research project that may lead to
+publication. These projects are for anyone, from core devs to those completely new
+to open source. We list projects by time series task
 
 [Classification](#classification)
 1. Optimizing the Shapelet Transform for classification and similarity search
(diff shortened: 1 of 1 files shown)


docs/mentoring.md
```diff
@@ -3,10 +3,11 @@
 
 # Mentoring and Projects
 
-`aeon` runs a range of short projects interacting with the community and the code
+`aeon` runs a range of short to medium duration projects interacting with the community
+and the code
 base. These projects are designed for internships, usage as part of
 undergraduate/postgraduate projects at academic institutions, and as options for
-programs such as [Google Summer of Code (GSoC)](https://summerofcode.withgoogle.com/).
+programs such as [Google Summer of Code (GSoC)](https://github.com/aeon-toolkit/aeon-admin/blob/main/gsoc/gsoc-2024-projects.md).
```
MatthewMiddlehurst, line 10: I would keep this general, GSoC already has its advertisements (which we have to remove now also)

MatthewMiddlehurst (commented):
I would rather remove the ones which are already taken IMO.

TonyBagnall:
> I would rather remove the ones which are already taken IMO.

Which ones are any taken? I was going to have a completed list once we have any, but Im not aware of any that have been assigned yet

MatthewMiddlehurst:
Fair enough, getting ahead of myself a bit there 🙂.

chrisholder (approved):

PR #1369, opened 2024-04-04
[DEP] Make `VectorizedDF` private

MatthewMiddlehurst:
Deprecates the public version and swaps usage to a private version.


TonyBagnall (approved):
a thousand times yes

PR #1370, opened 2024-04-04
[BUG] `LearningShapeletClassifier` fixes

MatthewMiddlehurst:
Fixes and skips broken tests for the `LearningShapeletClassifier`. I believe the extra functions were causing some issues so they have been commented for now. The deterministic test was due to a `keras` issue, I'm not quite sure how to solve that and would rather get the release out.


hadifawaz1999:
@MatthewMiddlehurst do you have a link for the error produced by the fail ? interested in seeing what keras is doing here :)

hadifawaz1999 (approved):

MatthewMiddlehurst:
https://github.com/aeon-toolkit/aeon/actions/runs/8560988997/job/23461264808

PR #1372, opened 2024-04-06
[ENH] BaseSeriesTransformer maintains axis for user determined output with BaseSeriesEstimator function

TonyBagnall:
This introduces a new function in BaseSeriesEstimator called _postprocess_series that checks the axis  passed and is different to self.axis, then the series is transposed back to the axis passed.

It also squeezes the output for univariate only transformers back to being univariate. See #1375 for how this may change subsequently.

this function is called in  BaseSeriesTransformer in transform, fit_transform and inverse_transform  as follows

```python
        if axis is None:
            axis = self.axis
        X = self._preprocess_series(X, axis=axis)
        Xt = self._transform(X)
        return self._postprocess_series(Xt, axis=axis)
```

### DIFF
--- a/aeon/base/_base_series.py
+++ b/aeon/base/_base_series.py
@@ -195,7 +195,7 @@ class BaseSeriesEstimator(BaseEstimator):
                 X = pd.DataFrame(X)
             else:
                 tag = self.get_tag("X_inner_type")
-                raise ValueError(f"Unknown inner type {inner} derived from {tag}")
+                raise ValueError(f"Unsupported inner type {inner} derived from {tag}")
         if axis > 1 or axis < 0:
             raise ValueError("Axis should be 0 or 1")
         if not self.get_tag("capability:multivariate"):
--- a/aeon/base/_base_series.py
+++ b/aeon/base/_base_series.py
@@ -237,8 +237,8 @@ class BaseSeriesEstimator(BaseEstimator):
 
     def _postprocess_series(self, X, axis=None):
         """Postprocess data X to revert to original shape."""
-        # If passed a univariate series, return a univariate series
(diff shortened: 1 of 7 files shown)


baraline (commented):
Didn't pick up anything wrong code wise. Concerning the logic behind the change, it is consistent with the `_preprocess` function we were already using in `BaseSeriesEstimator`, and should make handling the axis problem for both us and the user simpler.

My only concern may be on the deprecation side, as we are modifying the base series transformer class behaviour, shouldn't we use some deprecation notice here ? (@MatthewMiddlehurst)
Once this is question is settled, I would approve this PR.

baraline (dismissed):
After discussion with Tony, I agree that we should make this submodule (Series Transformers) experimental, as, well, we are still experimenting with this new interface for transformer. If people agree, we can move on with this PR. (Do we mention currently experimental module somewhere in the docs ?)

TonyBagnall:
made transformations/series experimental @MatthewMiddlehurst

baraline (approved):
Didn't pick anything wrong in the preprocess/postprocess/axis logic here. Input type checks also look good and well tested. Well done !

MatthewMiddlehurst (approved):
Fine with this. I think we should make the experimental bits more visible, like on the project README. Just a small section listing which modules and the implications of that.

MatthewMiddlehurst:
@TonyBagnall just came to mind, what if the output is not a series but a feature vector?

baraline:
As in a multivariate time series with only one  timestep ?

MatthewMiddlehurst:
As in a list of summary stats i.e. mean, stdev, min etc.

Issue #402, opened 2023-05-09
[ENH] ability to control the `fill_value` in padding transformer


hadifawaz1999:
### Describe the feature or idea you want to propose

For instance the user can choose a method for the padding, by default it should always be zero padding, but the user can either choose different values of filling per sequence (i.e. an array of fill values) or for example to padd with the average of the sequence

### Describe your proposed solution

I think these could be helpful:

- [ ] `fill_value` can be array-like
- [ ] `fill_value` can be string indicating 'avg' 'max', 'min' etc.

MatthewMiddlehurst:
would you consider this resolved with #1373 @hadifawaz1999?

hadifawaz1999:
Oh nice wasn't aware of the PR ! thanks for the mention, closing this now as completed

PR #1373, opened 2024-04-06
[ENH] Add support for array-like and statistical inputs as arguments in fill_value of Padding Transformer

aadya940:
Completes #402

### DIFF
--- a/aeon/transformations/collection/pad.py
+++ b/aeon/transformations/collection/pad.py
@@ -115,6 +115,8 @@ class PaddingTransformer(BaseCollectionTransformer):
                     "The length of fill_value must match the \
                                 length of X if a numpy array is passed as fill_value."
                 )
+        else:
+            self.fill_value = np.zeros(X.shape[0])
 
         return self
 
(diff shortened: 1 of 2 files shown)


TonyBagnall:
hi, thanks for this. What will happen if the series are multivariate, so n_channels >1. Does this pad each channel with (e.g.) an average value over all channels? If so, it would be better to work out the statistic for each channel separately. So if, each ```n_cases``` series is shape ```(n_channels, n_timepoints)``, then you should pad with the channel mean value, not the mean over all channels. Esp relevant if not normalised. For example, if this is (X,Y,Z) motion data, then the mean over all (X,Y,Z) will be very different to the mean over each channel separately.

aadya940:
@TonyBagnall Got your point. Will make the changes ASAP, Thanks. Does the part where `fill_value` is an `np.ndarray` look fine?

aadya940:
@TonyBagnall I hope this is better now. Should we just cast the Numpy Array to the appropriate size in 2D when a 1D Numpy Array is passed or raise an exception as done in my code?

TonyBagnall (approved):
lgtm, thanks

PR #1377, opened 2024-04-07
[ENH] removes the TagAliasMixin from the Base class

TonyBagnall:
the TagAliasMixin is a complex and unmaintained mechanism for deprecating tags bought over from the other place. It is not covered with testing (about 30 lines uncovered), or used anywhere in the toolkit. I dont think it needs deprecating because it is only usable in the toolkit, but I guess we could do so if anyone feels strongly.


MatthewMiddlehurst:
#717 🙂. I'm still in favour of keeping it. If testing/coverage is the issue, it can have its own file which is excluded.

TonyBagnall:
repeating myself, could you sort it out @MatthewMiddlehurst please as you want it?

PR #1378, opened 2024-04-07
[ENH] Deprecate _serialize file and load function

TonyBagnall:
for some reason there is a file called _serialize in the base class directory that simply wraps pickling functionality. I see no point in it. Can deprecate it or keep it if anyone wants to. I think the example should change if we keep it, it uses a CNNClassifier and I think this is not the best way to save/load keras estimators?


hadifawaz1999:
> for some reason there is a file called _serialize in the base class directory that simply wraps pickling functionality. I see no point in it. Can deprecate it or keep it if anyone wants to. I think the example should change if we keep it, it uses a CNNClassifier and I think this is not the best way to save/load keras estimators?

Yes i agree, there was even lots of this kind of tests in deep learning module, where someone tried to force this kind of saving on deep learners but never worked, i just deleted them :)

aeon/base/__init__.py
```diff
@@ -6,11 +6,9 @@
     "BaseCollectionEstimator",
     "BaseSeriesEstimator",
     "_HeterogenousMetaEstimator",
-    "load",
```
MatthewMiddlehurst, line 9: This would require deprecation

aeon/testing/test_all_estimators.py
```diff
@@ -1267,9 +1268,8 @@ def test_persistence_via_pickle(
         vanilla_result = scenario.run(estimator, method_sequence=[method_nsc])
 
         # Serialize and deserialize
-        serialized_estimator = estimator.save()
```
MatthewMiddlehurst, line 1270: Would keep this the same unless you are going to remove save also.

TonyBagnall: I would probably remove save, it confusingly returns a tuple, and I think this test is clearer without it. I think a separate test for save is appropriate, if we want to keep it.

MatthewMiddlehurst (approved):

PR #1379, opened 2024-04-07
[MNT] Add Classifiers For Result Comparison

### LINKED ISSUE #1187: [MNT] Add expected results for classifiers and regressors
### Describe the issue

Typically we like to compare the results of our classifiers and regressors against a saved set. Some algorithms in these packages do not do this currently. This does not have to be done in a single PR.

This will most likely involve adding a "results comparison" parameter set for the estimator (i.e. [this](https://www.aeon-toolkit.org/en/stable/developer_guide/dev_installation.html). The parameters chosen should be large enough that the results change when significant changes to the estimator are made, but small enough that the test does not clog up our CI runtime.

See the contents of [testing/expected_results/](https://github.com/aeon-toolkit/aeon/tree/main/aeon/testing/expected_results). Run the missing estimators (not those which are non-deterministic, i.e. deep learners), using the utilities within, and add them to the expected results file.

### Additional context

_No response_

harshithasudhakar:
#### Reference Issues/PRs

Fixes classification portion of #1187 

Added Convolution based, dictionary_based, distance_based, early_classification, feature_based, hybrid, interval_based: 

`HydraClassifier`, `MultiRocketHydraClassifier`, `REDCOMETS, KNeighborsTimeSeriesClassifier, BaseEarlyClassifier, TSFreshClassifier, RISTClassifier, IntervalForestClassifier, QUANTClassifier, RSTSF`

 to `classifier_results_reproduction.py` 
 
 Added `HydraClassifier`, `MultiRocketHydraClassifier` results to `expected_classifier_output.py`


MatthewMiddlehurst:
Some estimators may be non-deterministic FYI, in which case we cannot compare results to them. I don't remember off the top of my head, but the ones which are should have a tag.

MatthewMiddlehurst:
You may have to go into the individual classifier files and add a `"resullts_comparison"` set if the default does not work well.

MatthewMiddlehurst:
I think this is good to go if you remove RIST, will have a closer look then.

harshithasudhakar:
The tag `"capability:multivariate": False` seems to be missing for several classifiers like: `WEASEL, WEASEL_V2, BOSSEnsemble, ShapeDTW..`

harshithasudhakar:
I've added results for all classifiers except: `BaseEarlyClassifier, MrSQMClassifier, ContinuousIntervalTree, RotationForestClassifier, OrdinalTDE: BasicMotions` as I'm encountering some errors. I'll try to figure out the fix, and let you know if I need help with it.

harshithasudhakar:
> I think this is enough for now, if there are some missing it can be dealt with in another PR. The default for the multivariate tag is `False`, but maybe we should change that.

I thought so, but `SASTClassifier` explicitly has the tag set to `false`

TonyBagnall (approved):
LGTM, @MatthewMiddlehurst do you want to comment further?

MatthewMiddlehurst:
Nope, think this is good for now. Edited the comment to not close the issue, as it is classification only.

PR #1380, opened 2024-04-07
[ENH] Autocorrelation SeriesTransformer

TonyBagnall:
introduces an ACF series transformer that will ultimately replace the Transformer and CollectionTransformer versions. Will use this to test the broadcaster


TonyBagnall:
First thing to note is this ACF is NOT the same as statsmodels, which adjusts it all for various things including mean and trend. I dont think you can actually generate the raw ACF from the basic definition from statsmodels, at least not easily

doing some timing tests and results are a bit confusing. Correctness test is just a sanity check, the code is identical really and all ported from our old java. Code is at the end for future reference. 

Five related versions. 
acf1 is the existing BaseTransformer implementation with all the added baggage
acf2 is the new BaseSeriesTransformer with a numba implementation
acf3 is the the current BaseCollectionTransformer
acf4 is the BaseCollectionTransformer wrapped to BaseTransformer
acf5 repeatedly call the statsmodels acf function. Note this is really fast with fft set to true, but that involves operations I dont want to include (I assume it pads somewhere). 

So table here, not worth drawing graph
image


observations:
1. current version (column1) is very slow. I thought this might actually be due to statsmodels, but ...
2. statsmodels version (column5) is very fast. I would use it if you could actually get the acf from it (if you pass it [1,2,3,4,5,6,7,8,9,10] it does not return [1,1,1,1] for lag == 4.)
3. So it must be the weird broadcasting? But then wrapping the collection transformer back into the base transformer makes no speed difference (which makes me wonder if this wrapper is actually using VectorisedDF), in fact the numba version is slower than statsmodels
4. series converter is marginally faster than collection transformer, probably because I am not storing the results

so  really all I can conclude is (if further evidence were needed) this supports the need to remove BaseTransformer from the toolkit. 

```python
from aeon.transformations.acf import AutoCorrelationTransformer as acf1
from aeon.transformations.series._acf import AutoCorrelationTransformer as acf2
from aeon.transformations.collection import AutocorrelationFunctionTransformer as acf3
from aeon.testing.utils.data_gen import make_example_3d_numpy
from aeon.transformations.collection import CollectionToSeriesWrapper
import time
import numpy as np

def correctness_comparison():
    for i in range(100):
        X, _ = make_example_3d_numpy(n_cases=1, n_channels=1, n_timepoints=100+i)
        a2 = acf2(n_lags=int((100+i)/4))
        a3 = acf3(n_lags=int((100+i)/4))
        t1 = a2.fit_transform(X[0])
        t2 = a3.fit_transform(X)
        np.testing.assert_allclose(t1, t2[0])

def timing_comparison():
    print(" first = Existing collection transform with statsmodels")
    print(" second = Series transformer with numpy implementation")
    print(" third = Collection transformer with numpy implementation")
    print(" fourth = Collection transformer wrapped to be a BaseTransformer")
    print(" fifth = direct call to statsmodels")
    for i in range(100,5000,500):
        X, _ = make_example_3d_numpy(n_cases=i,n_channels=1, n_timepoints=i)
        a1= acf1(n_lags=int(i/4))
        start1 = time.time()
        t1=a1.fit_transform(X)
        end1= time.time()
#        print("Shape for a1 = ",t1.shape)
        a2= acf2(n_lags=int(i/4))
        start2 = time.time()
        for j in range(0,i):
            t2 = a2.fit_transform(X[j])
#        print("Shape for a2 = ",t2.shape)

        end2= time.time()
        a3= acf3(n_lags=int(i/4))
        start3 = time.time()
        t3 = a3.fit_transform(X,int(i/4))
        end3= time.time()
#        print("Shape for a3 = ",t3.shape)
        a4 = CollectionToSeriesWrapper(acf3(n_lags=int(i/4)))
        start4 = time.time()
        t4=a4.fit_transform(X)
        end4= time.time()
#        print("Shape for a4 = ",t4.shape)

        from statsmodels.tsa.stattools import acf
        start5 = time.time()
        for j in range(0,i):
            t5 = acf(X[j][0],nlags=int(i/4), fft=False)
        end5= time.time()
#        print("Shape for a5 = ",t5.shape)

        print(i,",",(end1-start1),",",(end2-start2), ", ", (end3-start3), ", ",
              (end4-start4),", ", (end5-start5))
```

TonyBagnall:
update, so the  CollectionToSeriesWrapper does not actually use the VectorizedDF stuff, which is good

MatthewMiddlehurst (approved):

PR #1394, opened 2024-04-09
[ENH] `BaseSeriesEstimator` rework

MatthewMiddlehurst:
This should all be internal, and does not impact any input types by the user. Testing, documentation and the notebook have been updated to match these changes.

- No longer converts `pd.DataFrame` to `pd.Series` if univariate
- Converts `np.ndarray` to a 2d array regardless of whether it's univariate or multivariate 

This should stop the swapping between data types and array shapes depending on input data and tags other than `X_inner_type`. This puts a bit more importance on the axis parameter, as 2 univariate types are now 2d.

- `X_inner_type` can be a list now

This is mostly for compatibility and parity with other usages of the tag. @TonyBagnall has mentioned they don't see a use-case for multiple datatypes, and I don't really either except for OTT optimisation.

- removes default for `axis` parameters and adds an `axis` parameter to `_check_y`

Setting the class and check function `axis` parameters correctly is very important to get correct conversions. I have removed the defaults, so hopefully it is not overlooked anywhere in the future. 

- Parameter for updating `metadata_` in `_preprocess_series`

General usage in this PR: overwrite in `fit` always, only overwrite in `predict`/`transform` if `fit` is empty.

- Moves `_check_y` to segmentation

This is specific to segmentation, labels will be different in other learning tasks i.e. anomaly detection.


aeon/base/_base_series.py
```diff
@@ -94,160 +102,150 @@ def _check_X(self, X):
...
+                if X.ndim > 1:
+                    n_channels = X.shape[0] if axis == 1 else X.shape[1]
+                    if n_channels > 1:
+                        raise ValueError(
+                            "Cannot convert to pd.Series for multivariate data. Found "
+                            f"{n_channels} channels"
+                        )
+
+                X = X.squeeze()
+                X = pd.Series(X)
```
TonyBagnall, line 195: dont need to do this, squeeze makes it a pd.Series if 1 dimensional I think

MatthewMiddlehurst: This is the bit for converting to a pd.Series. May be a better way to do it, but it if you just pass a pd.DataFrame it crashes.

MatthewMiddlehurst:
@TonyBagnall @baraline likely to impact your stuff

aeon/base/_base_series.py
```diff
@@ -61,13 +68,14 @@ class BaseSeriesEstimator(BaseEstimator):
...
-    def __init__(self, axis=0):
+    def __init__(self, axis):
         self.axis = axis
         self.metadata_ = {}  # metadata/properties of data seen in fit
+
         super().__init__()
         _check_estimator_deps(self)
 
-    def _check_X(self, X):
+    def _check_X(self, X, axis):
```
baraline, line 78: Missing docstring for axis parameter

aeon/base/_base_series.py
```diff
@@ -94,160 +102,150 @@ def _check_X(self, X):
...
-            if self.axis != axis:
-                X = X.T
+                raise ValueError(
+                    f"Unknown inner type {inner_type[0]} derived from {tag}"
+                )
+
+        if X.ndim > 1 and self.axis != axis:
+            X = X.T
+        elif X.ndim == 1 and isinstance(X, np.ndarray):
+            X = X.reshape(1, -1) if axis == 1 else X.reshape(1, -1)
```
baraline, line 215: I would prefer to use `X=X[np.newaxis, :]` instead of reshape, but in this case the end result is the same. 
Additionaly both condition are the same, here probably a typo. (One should be the reverse for axis=0 right ?)

baraline (changes requested):
Changes lgtm, just a possible typo in an if condition and some docstring. I think tests cover every case so that's good.

aeon/base/_base_series.py
MatthewMiddlehurst: yeah, looks incorrect, may be a blind spot in testing. Will look at coverage.

aeon/base/_base_series.py
MatthewMiddlehurst: This has a doc string now. Still have to update the main class one at least.

MatthewMiddlehurst:
I have updated the notebook.

baraline (approved):

baraline (approved):
Nice changes for `BaseSeriesEstimator` docs, much clearer now. Otherwise previous issues are resolved and the notebook explain the internal working of preprocess clearly to me. We could add a bit on postprocess too but that can maybe be made as a good first issue ?

MatthewMiddlehurst:
I'm fine with defaulting the multivariate capability to true.

TonyBagnall:
its quite hard to review on files changed tab, so will do it here, re _convert_X. Some minor points

if pd.Series is passed we have this
```
            elif inner_type[0] == "Series":
                if self.get_tag("capability:multivariate"):
                    raise ValueError(
                        "Cannot convert to pd.Series for multivariate capable "
                        "estimators"
                    )
                if X.ndim > 1:
                    n_channels = X.shape[0] if axis == 1 else X.shape[1]
                    if n_channels > 1:
                        raise ValueError(
                            "Cannot convert to pd.Series for multivariate data. Found "
                            f"{n_channels} channels"
                        )
```
1. Why cant we convert a series to multivariate capable? We do with 1D numpy. You can just put it in a DataFrame? It seems inconsistent to me. 
2. If you squeeze a DataFrame with 1D it becomes a pd.Series, so this conversion is redundant
```
                X = X.squeeze()
                X = pd.Series(X)
```
e.g.
```import pandas as pd
X = pd.DataFrame([1,2,3,4,5])
y = X.squeeze()
print(type(y)) 
```
is pd.Series
3. So figured this out, but it seems a bit twisty. 
```
            elif inner_type[0] == "DataFrame":
                # converting a 1d array will create a 2d array in axis 0 format
                transpose = False
                if X.ndim == 1 and axis == 1:
                    transpose = True

                X = pd.DataFrame(X)

                if transpose:
                    X = X.T
```
cant you just have
```
                X = pd.DataFrame(X)
               if X.ndim == 1 and axis == 1:
                    X = X.T
```
it all looks like it could be condensed to fewer lines of code tbh. 

4. Why do we have to call get tag again? Already called it earlier,
```
            else:
                tag = self.get_tag("X_inner_type")
                raise ValueError(
                    f"Unknown inner type {inner_type[0]} derived from {tag}"
                )
```
5. Multiple values of inner_type seem to just be ignored, since you only test inner_type[0], so why have them? I still dont see the use case, and this means any list is ignored I think, which could be very confusing. if my inner type were ["np.ndarray", "pd.Series"], my expectation would be that if either were passed, the base class would not convert them. However, atm, it would I think get converted to np.ndarray?

TonyBagnall (changes requested):
see comments

MatthewMiddlehurst:
> Why cant we convert a series to multivariate capable? We do with 1D numpy. You can just put it in a DataFrame? It seems inconsistent to me.

We can, but that would involve removing `pd.Series` as an inner type altogether essentially. Not opposed.

> If you squeeze a DataFrame with 1D it becomes a pd.Series, so this conversion is redundant

This is also used for `np.ndarray` > `pd.Series`, where squeeze turns a 2D array into a 1D one (it would all go if we do the above).

> So figured this out, but it seems a bit twisty. cant you just have
```
X = pd.DataFrame(X)
if X.ndim == 1 and axis == 1:
    X = X.T
```

`pd.DataFrame` is always `X.ndim==2`, no such thing as a 1D `pd.DataFrame`, that's just a `pd.Series`. No idea why. What I want to do there is check if we have a 1D `np.ndarray` or `pd.Series` and see if we have to transpose the new `pd.DataFrame` (as it will always be `(n_timepoints, n_channels)` if given a 1D input).

> Why do we have to call get tag again? Already called it earlier,

I edit the output of the previous call. Can just use a different variable name for the edited version and ditch this line.

> Multiple values of inner_type seem to just be ignored, since you only test inner_type[0], so why have them? I still dont see the use case, and this means any list is ignored I think, which could be very confusing. if my inner type were ["np.ndarray", "pd.Series"], my expectation would be that if either were passed, the base class would not convert them. However, atm, it would I think get converted to np.ndarray?

This is mostly for compatibility with the tag. It says it (or other usages can) can take an array. The following line does what you suggest `if input not in inner_type:`, if input is already of an accepted type no conversion is done. Though now that I say that, it seems to be skipping some bits we wouldn't want it too currently. `inner_type[0]` is only used if we have to convert to something different.

TonyBagnall (approved):

TonyBagnall:
yeah ok, can put this in if you resolve conflicts and we can discuss 1D options after

MatthewMiddlehurst:
I have merged main into this. Please do double-check the postprocess function, some changes to match the new contents. Also changes to include docs and to testing.

TonyBagnall (approved):

TonyBagnall:
weirdly, code cov is failing with these errors
image

MatthewMiddlehurst:
unrelated

TonyBagnall:
Unrelated codecov fails

PR #1356, opened 2024-03-28
[ENH] Distance module pairwise unequal length support

chrisholder:
#### Reference Issues/PRs
#1287 

#### What does this implement/fix? Explain your changes.
This PR continues the work of @CodeLionX in PR #1287 and extends his work to the whole of the distance module so every distance can support unequal pairwise distances.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #1415, opened 2024-04-11
[ENH] Update MPDist function

### LINKED ISSUE #423: [ENH] update Matrix Profile distance function (MPDist)

the mpdist function is described in
https://link.springer.com/article/10.1007/s10618-020-00695-8

It was implemented some time ago by @ClaudiaSanches.
https://github.com/alan-turing-institute/sktime/blob/main/sktime/distances/mpdist.py

t needs updating to use numba, adapting to conform to the distances module code structure and would benefit from enhanced docstrings. This is part of a wider issue with a need to tidy up/sort out our matrix profile implementations (or import them instead)

_No response_

### Additional context

_No response_

itsdivya1309:
#### Reference Issues/PRs
Fixes #423

1. Updates the code to work with numba
2. Includes `mpdist_pairwise_distance` function
3. Improves docstrings for `mpdist`


CodeLionX:
If you JIT `_mpdist`, please already consider the unequal-length implementation. #1287 and #1356 are good pointers.

aeon/distances/_mpdist.py
```diff
@@ -0,0 +1,372 @@
...
+            2
+            * q_len
+            * (
+                1
+                - (
+                    (dot_prod[i] - q_len * q_mean * t_mean[i])
+                    / (q_len * q_std * t_std[i])
+                )
+            )
+        )
```
TonyBagnall, line 180: I think its easier to read if you split these operations to make this more readable? i.e.
temp =(dot_prod[i] - q_len * q_mean * t_mean[i])/ (q_len * q_std * t_std[i])
d[i] = 2* q_len* (1-temp)
I realise this was not your code, but may as well take the opportunity to tidy it up.

TonyBagnall (approved):
create stuff, one minor change suggested but its not a blocker

aeon/distances/_mpdist.py
```diff
@@ -0,0 +1,363 @@
...
+           [2.82842712, 2.82842712, 2.82842712],
+           [2.82842712, 2.82842712, 2.82842712]])
+
+    >>> X = np.array([[[1, 2, 3]],[[4, 5, 6]], [[7, 8, 9]]])
+    >>> y_univariate = np.array([[22, 18, 12]])
+    >>> mpdist_pairwise_distance(X, y_univariate, m = 2)
+    array([[2.82842712],
+           [2.82842712],
+           [2.82842712]])
+    """
```
itsdivya1309, line 320: Hi @TonyBagnall, I have kept the code logic the same, but I just noticed that there may be an issue with the results it generates. As you can see here in the examples it gives for `mpdist_pairwise_distance`, I doubt this is the desired output.
Also, please look at the `mpdist` example. Is this the way this function is supposed to work?

CodeLionX:
Nice work 👍🏼
However, the `pairwise_distance`-function is not jitted yet and might be slow. @chrisholder and I need to take a look if it is possible to jit it and actually measure if it makes sense.

When the discussion about the correct output is resolved, I'm in favor of merging and having another PR considering the potential performance bottleneck.

aeon/distances/_mpdist.py
itsdivya1309: I have updated the code, but it still produces the same results. I think this is how the MPDist algo performs altogether because even the mpdist gives minimal values.

aeon/distances/_mpdist.py
```diff
@@ -0,0 +1,359 @@
...
+    >>> mpdist(x, y, m)
+    0.05663764013361034
+    """
+    x = np.squeeze(x)
+    y = np.squeeze(y)
+
+    if x.ndim == 1 and y.ndim == 1:
+        return _mpdist(x, y, m)
+
+    raise ValueError("x and y must be a 1D array of shape (n_timepoints,)")
```
CodeLionX, line 65: All distances in aeon support multivariate TS, some natively, others by just summing up the channel-wise distances, e.g.

https://github.com/aeon-toolkit/aeon/blob/957704a45ab130cf19d6302d9e9f6c4e13dc2001/aeon/distances/_manhattan.py#L52-L63

Is there a reason to not do this with MPDist?

CodeLionX (changes requested):

aeon/distances/_mpdist.py
CodeLionX: Sry for requesting more work so late. I have not seen this before. We can also merge this PR first and create a second one for making `mpdist` support multivariate TS. I can imagine that it requires a lot of modification to the different functions.

itsdivya1309: Hi! MPDist was proposed for univariate time series data in the paper. This PR aimed to update the mpdist to work with numba and introduce pairwise distance.
I would like to implement it for multivariate data, but I have my end-semester exams, so I'll not be able to take this up this week :)
We can open another PR for this.

CodeLionX (approved):

aeon/distances/_mpdist.py
CodeLionX: Alright. Please create a new issue for this so that we don't forget it 😉 Thank you for working on this!

itsdivya1309: > Alright. Please create a new issue for this so that we don't forget it 😉 Thank you for working on this!

Sure 👍

TonyBagnall (approved):
LGTM

aeon/distances/_mpdist.py
MatthewMiddlehurst: We still need this issue I think, grateful if you could @itsdivya1309 as I have not been following this closely 🙂

PR #1426, opened 2024-04-15
[DOC] Add estimator overview table with capabilities

### LINKED ISSUE #1242: [DOC] Proposal: list capabilities of individual estimators in their documentation

When browsing the available estimators (classifiers, clusterers, ...) in the aeon documentation or using IntelliSense, I cannot directly see which estimator supports which input types (the source code is of limited use because of class inheritance etc.).

Aeon already stores the capabilities of individual estimators as tags in machine-readable form (see below), and we can list estimators based on their tags using the registry (e.g., with `registry.all_estimators(filter_tags={"capability:unequal_length": True})`).

https://github.com/aeon-toolkit/aeon/blob/620caa54d29481c9c85b06799354e65b9dd28464/aeon/base/_base_collection.py#L30-L37

and

https://github.com/aeon-toolkit/aeon/blob/620caa54d29481c9c85b06799354e65b9dd28464/aeon/base/_base_series.py#L57-L62

Can we use the tags to display the estimator's capabilities also in the documentation?

baraline:
#### Reference Issues/PRs

Fixes #1242 

Add back the estimator overview table from 0.3, but add capabilities of individual estimators to the table.

When this is merged, i'll raise an issue for the sorting by columns / search function that I didn't manage to implement.


docs/conf.py
```diff
@@ -262,7 +262,10 @@ def find_source():
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
-html_css_files = ["css/custom.css"]
+html_css_files = [
+    "css/custom.css",
+]
+html_js_files = []
```
CodeLionX, line 268: unused

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+            data["Family of method"].append("N/A")
+        for capability_name in capabilities_to_include:
+            _val = tag_dict.get(f"capability:{capability_name}")
+            _str = capability_name.replace("_", " ")
+
+            # For case where tag is not included output as not supported.
+            if not _val or _val is None:
+                data[f"Support {_str}"].append("\u274C")
+            else:
+                data[f"Support {_str}"].append("\u2705")
```
CodeLionX, line 384:
```suggestion
            if not _val or _val is None:
                data[f"Supports {_str}"].append("\u274C")
            else:
                data[f"Supports {_str}"].append("\u2705")
```

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+
+        # includes part of class string
+        modpath = str(estimator_class)[8:-2]
+        path_parts = modpath.split(".")
+        # joins strings excluding starting with '_'
+        clean_path = ".".join(list(filter(_does_not_start_with_underscore, path_parts)))
+        # adds html link reference
+        estimator_name_as_link = str(
+            '<a href="https://www.aeon-toolkit.org/en/latest/api_reference'
+            + "/auto_generated/"
```
CodeLionX, line 363:
```suggestion
            '<a href="api_reference/auto_generated/'
```
This always referenced the _latest_ version of the documentation instead of the currently selected/displayed version, e.g. the PR build.

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+            + clean_path
+            + '.html">'
+            + estimator_name
+            + "</a>"
+        )
+        algorithm_type = algorithm_type.split("::")
+        data["Estimator name"].append(estimator_name_as_link)
+        data["Module"].append(algorithm_type[0])
+        if len(algorithm_type) > 1:
+            data["Family of method"].append("/".join(algorithm_type[1:]))
```
CodeLionX, line 373: Personal preference:
```suggestion
            data["Method family"].append("/".join(algorithm_type[1:]))
```

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+    """Make estimator overview table."""
+    import pandas as pd
+
+    from aeon.registry import all_estimators
+
+    def _does_not_start_with_underscore(input_string):
+        return not input_string.startswith("_")
+
+    # Columns for the output table
+    COLNAMES = ["Estimator name", "Module", "Family of method"]
```
CodeLionX, line 337: Personal preference:
```suggestion
    COLNAMES = ["Estimator name", "Module", "Method family"]
```
(also see below)

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+    capabilities_to_include = [
+        "multivariate",
+        "unequal_length",
+        "missing_values",
+    ]
+
+    for capability_name in capabilities_to_include:
+        _str = capability_name.replace("_", " ")
+        COLNAMES.append(f"Support {_str}")
```
CodeLionX, line 346:
```suggestion
        COLNAMES.append(f"Supports {_str}")
```

docs/conf.py
```diff
@@ -314,6 +323,84 @@ def find_source():
...
+            _str = capability_name.replace("_", " ")
+
+            # For case where tag is not included output as not supported.
+            if not _val or _val is None:
+                data[f"Support {_str}"].append("\u274C")
+            else:
+                data[f"Support {_str}"].append("\u2705")
+
+    df = pd.DataFrame.from_dict(data).sort_values(
+        by=["Module", "Family of method", "Estimator name"]
```
CodeLionX, line 387: Personal preference
```suggestion
        by=["Module", "Method family", "Estimator name"]
```

CodeLionX (changes requested):
I like the overview table. It already provides a good overview and the capabilities are displayed as well. For now, one can just use the browsers search functionality to navigate it. 👍🏼 

The generation-function contains quite a lot of code. Should we extract it to a separate file?

There is also `build_tools/generate_developer_tables.py`, which also generates an html table... I don't know if this is similar enough, though.

baraline:
The trick is that it should be generated while building the docs, and I don't know if `build_tools` would be accessible while readthedocs builds. But I agree we should put it in a separate file, maybe create a folder/python script inside `docs/` for script that generate content ?

CodeLionX (approved):

MatthewMiddlehurst (approved):
Looks great. I agree searching can wait with an issue raised. Do we want to close #1242? There is still an issue maybe in not having tags on individual API pages.

baraline:
Ok, a quick try on moving the script somewhere else made a mess, I'll keep it that way for now. I opened #1430 for individual estimators, and I'll create the issue for search/sort. So I guess we can close #1242 for now.

PR #1437, opened 2024-04-18
remove probabilistic metrics

TonyBagnall:
@MatthewMiddlehurst is skeptical that removing these will not break things :)


TonyBagnall:
and he was correct to be so :) Some horrible code that allows the replacement of any raised error with nan confused me, such a hacky design.

PR #1439, opened 2024-04-18
[DEP] deprecate check_scorer

TonyBagnall:
deprecates the function check_scorer. This function is used in some places to set a scoring function through a call to check_scorer(None). This is confusing, and it is much clearer imo to set the scorer explicitly. This PR does that and tags the function for deprecation.


MatthewMiddlehurst (approved):
A simple one and I agree. LGTM

PR #1441, opened 2024-04-19
[DEP] Remove annotation module

TonyBagnall:
this PR completely removes the annotation module and the associated soft dependency for hmmlearn. 

The base class for annotators does not conform to standard classes, and the only algorithms that implement it were simple wrappers for hmmlearn. We could make them SeriesTransformers, but personally I prefer to just remove the whole thing, can always bring back as SeriesTransformers if anyone wants hmm wrappers.  

The reason I am doing this is I am looking to rework check_series in the validation module. This is called in BaseAnnotator, and it seems a waste of effort to rewqork BaseAnnotator when we have split the tasks into anomaly detection and segmentation


lmmentel:
I can have a look over the weekend.

aeon/forecasting/compose/tests/test_bagging.py
```diff
@@ -16,7 +16,7 @@
 
 @pytest.mark.skipif(
     not _check_soft_dependencies("statsmodels", severity="none"),
-    reason="skip test if required soft dependency for hmmlearn not available",
```
lmmentel, line 19: Is this a intentional? On first glance looks like fixing something else?

lmmentel:
Overall, I'm quite supportive of the changes since the "annotation" as a concept or a learning task is quite ambiguous and hasn't been very useful for making progress on implementations. I remember a lot of discussions in the past that were hard to conclude because of the breadth of the term. 

As far as I can see the `hmmlearn` wrappers are thin so we are not loosing a differentiating feature but this affects the value that comes from having HMM estimators being compatible with pipelines and transformations. I haven't really used HMM module with or without the rest of `aeon` but maybe someone else could comment on how easy/difficult it would be to run aeon alongside hmmlearn without the wrapper?  

Given that @miraep8 authored most of the affected code I would be curious to hear her take on this change.

--- 

This is probably off topic but I was surprised to see `hmmlearn` mentioned in forecasting tests. I couldn't find any trace of HMM in the actual tests, was that related to copy-pasting code blocks around?

aeon/forecasting/compose/tests/test_bagging.py
MatthewMiddlehurst: I assume this was a copy-paste error found when looking for `hmmlearn` usages 🙂

MatthewMiddlehurst:
I have a similar opinion, really. I think the module itself is of limited use, it's just this one estimator which does not really fit into segmentation or anomaly detection. I don't think there is a rush to get this in, If anyone is willing to turn these into transformers I think that would be the best compromise for now. If no one volunteers then it is always in the version control and can be re-added.

aeon/forecasting/compose/tests/test_bagging.py
TonyBagnall: thats exactly what it is :) good spot

TonyBagnall: yes, loads of pytest warnings were cut and paste with hmml but had nothing to do with it

TonyBagnall:
> This is probably off topic but I was surprised to see `hmmlearn` mentioned in forecasting tests. I couldn't find any trace of HMM in the actual tests, was that related to copy-pasting code blocks around?

there was some cut and pasting of warnings from hmmlearn, jsut fixed them

TonyBagnall:
I'm fine to keep the hmmlearn wrappers if they move to transformations as series transformers, main reason for this is to get rid of the BaseAnnotator, but personally I dont want to maintain them. If  @miraep8 does want them I'm happy to do the conversion

lmmentel (approved):

PR #1442, opened 2024-04-19
[DEP] simplify check_series/check_X

TonyBagnall:
in a move to try unify the validation of single series, this PR tidies up some of the checking done in forecasting. 

There are two methods currently used: check_X and check_series. check_X calls check_series with these parameters
```
check_series(
        X,
        allow_empty=allow_empty,
        enforce_univariate=enforce_univariate,
        enforce_index_type=enforce_index_type,
        allow_numpy=False,
    )
```
I assume there are two because calling check_X(Z) seems a bit weird. Check_X has extract optional parameters that are never used in the code base. 

This PR 
1. replaces calls to check_X with identical calls to check_series is forecasting and plotting
2. adds numeric data tests taken from BaseSeriesEstimator
3. Removes  the optional  enforce_multivariate argument and allow_None=True which was not currently used anywhere can put it back if anyone wants it, but will need to test it before numeric checks. 
4. calls the same check_series function in BaseSeriesEstimator that is used in forecasting
5. adds and reformats tests for check_series, 
6. deprecates check_X in utils.validation.forecasting.py


aeon/utils/validation/forecasting.py
```diff
@@ -72,12 +72,22 @@ def check_y_X(
...
     return y, X
 
 
+# TODO: remove in v0.10.0
+@deprecated(
+    version="0.9.0",
+    reason=(
+        "check_X in aeon.validation.forecasting.py has been replaced by "
+        "check_series in aeon.validation.series.py, please use check_series "
+        "instead."
```
MatthewMiddlehurst, line 87: include the version its removed in

aeon/utils/validation/series.py
```diff
@@ -51,6 +53,11 @@ def is_single_series(y):
     if isinstance(y, np.ndarray):
         if y.ndim > 2:
             return False
+        if not (
+            issubclass(y.dtype.type, np.integer)
+            or issubclass(y.dtype.type, np.floating)
+        ):
```
MatthewMiddlehurst, line 59: should check dtype for other inputs as well

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
         type of time index
     allow_index_names : bool, default = False
         If False, names of Z.index will be set to None
-    var_name : str, default = "input" - variable name printed in error messages
 
     Returns
     -------
-    Z : pd.Series, pd.DataFrame, np.ndarray, or None
+    X : pd.Series, pd.DataFrame or np.ndarray
         Validated time series - a reference to the input Z
```
MatthewMiddlehurst, line 239: some of these descriptions are still Z

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
-    # check time index if input data is not an NumPy ndarray
-    if not isinstance(Z, np.ndarray):
-        check_time_index(
-            Z.index,
-            allow_empty=allow_empty,
-            enforce_index_type=enforce_index_type,
-            var_name=var_name,
+    if not isinstance(X, VALID_DATA_TYPES):
+        raise TypeError(
+            f"Series must be a one of {VALID_DATA_TYPES}, but found type: {type(X)}"
```
MatthewMiddlehurst, line 258: Seems pretty generic. Liked it better when it specifically referred to the input parameter name i.e. X

same with other errors i.e. one above

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
     # Check if pandas series or numpy array
-    if not allow_numpy:
-        valid_data_types = tuple(
-            filter(lambda x: x is not np.ndarray, VALID_DATA_TYPES)
-        )
-    else:
-        valid_data_types = VALID_DATA_TYPES
-
-    if not isinstance(Z, valid_data_types):
+    if not allow_numpy and isinstance(X, np.ndarray):
```
MatthewMiddlehurst, line 252: Don't mind the parameter, but why only numpy?

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
+    X,
     allow_empty=False,
-    allow_numpy=True,
-    allow_None=True,
+    enforce_univariate=False,
     enforce_index_type=None,
+    allow_numpy=True,
     allow_index_names=False,
-    var_name="input",
 ):
```
MatthewMiddlehurst, line 220: If we are deprecating check_X these should be as well

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
+        raise TypeError(
+            f"Series must be a one of {VALID_DATA_TYPES}, but found type: {type(X)}"
         )
-
-    if not allow_index_names and not isinstance(Z, np.ndarray):
-        Z.index.names = [None for name in Z.index.names]
-
-    return Z
+    if enforce_univariate and not is_univariate_series(X):
+        raise TypeError("Series must be univariate if enforce_univariate is True")
```
MatthewMiddlehurst, line 261: similar to the datatypes, why not multivariate as well. We have the output of `is_univariate_series(X)` already so should be easy

aeon/utils/validation/series.py
```diff
@@ -204,103 +211,81 @@ def _check_is_multivariate(Z, var_name="input"):
...
+    if not allow_empty:
```
MatthewMiddlehurst, line 262: not quite sure what the point is of this

MatthewMiddlehurst (commented):
Not the biggest fan in the sense that it makes the code block we currently have in base series a lot messier. I feel this ties the newer base class to some of the legacy stuff from forecasting.

aeon/utils/validation/series.py
TonyBagnall: I thought we did not have to deprecate optional parameters?

MatthewMiddlehurst: it would be best to. dont really need to keep the internal functionality for them, just the keyword

PR #1443, opened 2024-04-20
[ENH] Add control over the score method of classification and regression

### LINKED ISSUE #1062: [ENH] Adding control over the `score` method of estimators

For now, the base class of each estimator sector has a scoring technique, for instance its R2 score in base regressor, its open to override in inherited classes. I think its best if we add parameter string to the base class score function, defaulted to R2 for example in regression, accuracy in classification etc. and depending on that argument, the score function calculates a metric.

This should be done over all the estimator sectors i.e. classification, regression, forecasting etc. but one by one !

_No response_

### Additional context

_No response_

aadya940:
Closes #1062


baraline:
Nice work, althought I would suggest that rather listing some/all metrics for a task, you could use a sklearn defined method such as [get_scorer](https://github.com/scikit-learn/scikit-learn/blob/8721245511de2f225ff5f9aa5f5fadce663cd4a3/sklearn/metrics/_scorer.py#L359). Then you could do the following to get back the metric function from the scorer 

```python
from sklearn.metrics import get_scorer
get_scorer('accuracy')._score_func(y_true, y_pred)
```

Thought on this @hadifawaz1999 ? ( as you raised the issue )

MatthewMiddlehurst (commented):
Agree with the above, but that could get awkward when it comes to parameterising some i.e. auroc. Maybe a mix?

Tests should include an invalid input for metric.

baraline:
We could do something like this to also accept callable and deal with parameterization ?

```python
def score(self, X, y, metric="accuracy", metrics_params=None):
  ...
  if metrics_params is None:
    metrics_params_ = {}
  # let score func raise error if metrics params is not right format
  if isinstance(metric, str):
    return  get_scorer(metric')._score_func(y_true, y_pred, **metrics_params_)
  elif callable(metric):
    return  metric(y_true, y_pred, **metrics_params_)
  else:
    # throw an error
```

aadya940:
@baraline 
What about the cases where `y_pred` is `predict_proba(X)` instead of `predict(X)`?

baraline:
I'm not sure, but I see your point, it is indeed a problem.

For example [roc_auc_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html) would handle probabilities estimates but not [accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html#sklearn.metrics.accuracy_score) according to docs. 
To summarize:
- If the metric support probabilities, we want to support both probabilities estimate and predicted class as input
- Otherwise, we only accept predicted class as inputs.

This raise the question of how should the user be able to use probabilities estimates over predict class (i.e. score with the output of `predict_probas` instead of `predict`). 

One easy solution is to introduce another parameter in the `score` function, something like a boolean `use_probabilities_estimate` to determine wheter `predict_probas` should be used over `predict`. Then let sklearn raise the error if the metric does not support probabilities estimate.

I'm pretty sure there might be a better way to do this though...

aadya940:
I'm not completely sure but we could do something along the following lines?

```
scorer = get_scorer(metric)
try:
        return scorer._score_func(y_true, predict(X))
    except:
        try:
            return scorer._score_func(y_true, predict_proba(X))
        except:
            # throw an exception
```

aadya940:
I think, for now, wouldn't supporting a few scoring methods be a good idea? Since we already claim that aeon is sklearn compatible so it wouldn't be much of an issue if someone wants to use another scoring method.

MatthewMiddlehurst:
I think the string and callable input with `use_probabilities_estimate` or however you name it is fine. Changes to base classes should be more strictly curated IMO so better to get it correct now.

aadya940:
> I think the string and callable input with `use_probabilities_estimate` or however you name it is fine. Changes to base classes should be more strictly curated IMO so better to get it correct now.

Sounds Good! I'll change it accordingly.

aeon/classification/base.py
```diff
@@ -368,7 +378,26 @@ def score(self, X, y) -> float:
...
-        return accuracy_score(y, self.predict(X), normalize=True)
+        _metric_params = metric_params
+        if metric_params is None:
+            _metric_params = {}
+        if isinstance(metric, str):
+            __names = get_scorer_names()
+            if metric not in __names:
+                raise ValueError(
+                    """metric name incompatible with `sklearn.get_scorer`. Use
+                                 a valid metric name instead."""
```
baraline, line 389: I would give a more expressive error message, something along the line of :
```python
raise ValueError(
    f"Metric {metric} is incompatible with `sklearn.metrics.get_scorer` function."
    "Valid list of metrics can be obtained using the `sklearn.metrics.get_scorer_names` function"
)
```

aeon/classification/base.py
```diff
@@ -368,7 +378,26 @@ def score(self, X, y) -> float:
...
+            scorer = get_scorer(metric)
+            if use_proba:
+                return scorer._score_func(y, self.predict_proba(X), **_metric_params)
+            return scorer._score_func(y, self.predict(X), **_metric_params)
+        elif callable(metric):
+            if use_proba:
+                return metric(y, self.predict_proba(X), **_metric_params)
+            return metric(y, self.predict(X), **_metric_params)
+        else:
+            raise ValueError("Enter a valid metric format.")
```
baraline, line 400: Again the same here, something along the lines of 
```python
raise ValueError(
    "The metric parameter should be either a string or a callable"
     f", but got {metric} of type {type(metric)}"
)
```

aeon/regression/base.py
```diff
@@ -244,7 +250,22 @@ def score(self, X, y) -> float:
...
+        _metric_params = metric_params
+        if metric_params is None:
+            _metric_params = {}
+        if isinstance(metric, str):
+            __names = get_scorer_names()
+            if metric not in __names:
+                raise ValueError(
+                    """metric name incompatible with `sklearn.get_scorer`. Use
+                                 a valid metric name instead."""
+                )
```
baraline, line 262: Same comment as in classification base

aeon/regression/base.py
```diff
@@ -244,7 +250,22 @@ def score(self, X, y) -> float:
...
+            scorer = get_scorer(metric)
+            return scorer._score_func(y, self.predict(X), **_metric_params)
+        elif callable(metric):
+            return metric(y, self.predict(X), **_metric_params)
+        else:
+            raise ValueError("Enter a valid metric format.")
```
baraline, line 268: Same comment as in classification base

baraline (changes requested):
Some minor changes for error message. The rest looks good, test are added for both unknown metrics and callable inputs, so LGTM !

aadya940:
I've done the needed changes. @baraline Thankyou for reviewing.

baraline (dismissed):
Thanks, the rest LGTM. As this is touching base class, I'd wait for another review to merge it, nice work !

aeon/regression/base.py
```diff
@@ -236,6 +236,12 @@ def score(self, X, y) -> float:
         y : np.ndarray
             1D np.array of float, of shape ``(n_cases)`` - regression targets
             (ground truth) for fitting indices corresponding to instance indices in X.
+        metric : Union[str, callable], default="accuracy",
```
MatthewMiddlehurst, line 239: incorrect default here

MatthewMiddlehurst (changes requested):
Good changes IMO, spotted one doc error. Any thoughts @TonyBagnall?

aeon/regression/base.py
aadya940: Sorry for the oversight :-)

hadifawaz1999 (approved):
lgtm !

baraline (approved):
LGTM, didn't catch the documentation issue.

MatthewMiddlehurst (approved):

PR #1445, opened 2024-04-20
[BUG] Fix RDST doctest failure & numba deprecration warning

### LINKED ISSUE #1444: [BUG] RDST transformer sporadic test failures.
### Describe the bug

The RDST Regressor sometime fails tests (see [this](https://github.com/aeon-toolkit/aeon/actions/runs/8753180014/job/24025718908?pr=1441)) on the docstring example.

```python

from aeon.regression.shapelet_based import RDSTRegressor
from aeon.datasets import load_covid_3month
X_train, y_train = load_covid_3month(split="train", return_X_y=True)
X_test, y_test = load_covid_3month(split="test", return_X_y=True)
for i in range(10000):
    RDSTRegressor(max_shapelets=10).fit(X_train, y_train)
```

### Expected results

No error should be raised.

### Actual results

```python-traceback

RuntimeError: No shapelets were extracted during the fit method with the specified parameters.

```

### Versions

_No response_

baraline:
#### Reference Issues/PRs

Fixes #1444 

Fix sporadic doctest failure. At the end of shapelet extraction, a check is performed to remove shapelets which values were full of 0. This should happen if a shapelet is pruned during the extraction process, the value of 0 is due to the initialization of the array of shapelet values as full of 0 (`np.zeros`). But as the covid dataset has long successive 0 values, it was possible that they were all pruned with bad luck, causing the failure.

The test is now based on `np.inf` value, and the array initialized as full of `np.inf` to avoid any problem related to input values.

### DIFF
--- a/aeon/classification/shapelet_based/_rdst.py
+++ b/aeon/classification/shapelet_based/_rdst.py
@@ -4,7 +4,7 @@ A Random Dilated Shapelet Transform classifier pipeline that simply performs a r
 shapelet dilated transform and build (by default) a ridge classifier on the output.
 """
 
-__maintainer__ = []
+__maintainer__ = ["baraline"]
 __all__ = ["RDSTClassifier"]
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


TonyBagnall (approved):
lgtm

PR #1453, opened 2024-04-22
[BUG] Fix MultiRocketMultivariate failures

### LINKED ISSUE #1278: [BUG] Fix `MultiRocketMultivariate` failures
### Describe the bug

`test_multirocket_multivariate_on_basic_motions` occasionally fails due to a difference in accuracy despite a `random_state` being set.

`MultiRocketHydraRegressor` fails `test_fit_deterministic` with multivariate data, the `HydraRegressor` does not, so I assume this is the multirocket side.

Run `MultiRocketHydraRegressor` `fit` and `predict` twice and check results.

### Expected results

Tests pass.

### Actual results

Results are different causing test failures.

### Versions

_No response_

baraline:
#### Reference Issues/PRs

Fixes #1278

Random state was not set in outside of numba function, which, as both need to be set (see [docs](https://numba.readthedocs.io/en/stable/reference/numpysupported.html#initialization)). This caused difference in accuracy during some tests.

#### Any other comments?

Using the new numpy random generator would bump numba version to `0.56` from `0.55`, so I stick with the old numpy random seed.

### DIFF
--- a/aeon/transformations/collection/convolution_based/_multirocket_multivariate.py
+++ b/aeon/transformations/collection/convolution_based/_multirocket_multivariate.py
@@ -112,6 +112,9 @@ class MultiRocketMultivariate(BaseCollectionTransformer):
         -------
         self
         """
+        if self.random_state is not None:
+            np.random.seed(self.random_state)
+
         if self.normalise:
             X = (X - X.mean(axis=-1, keepdims=True)) / (
                 X.std(axis=-1, keepdims=True) + 1e-8


hadifawaz1999:
nice ! thanks for confirming i was thinking it was only me not seeing the random state being set somewhere :) do u think its worth it if in this PR we do the change to the rng of numpy in rocket ? or do we keep it on a separate issue and put it as good first issue ?

hadifawaz1999 (approved):

MatthewMiddlehurst (approved):
LGTM. Good catch. Would create an issue if there are concerns with the bound so it does not get lost.

PR #1459, opened 2024-04-22
[MNT] Adds an Issue/PR Assigning Bot

AnonymousCodes911:
As per the discussion with @MatthewMiddlehurst, 

It implements an Issue/PR assigning bot to eliminate the chances of people working on the same Issue.

#### Any other comments?
This would look something like : 
image


AnonymousCodes911:
@chrisholder, do you think we could add more interactivity to this?

MatthewMiddlehurst:
Could you run through the code and what each bit is doing? Not very reader friendly currently for anyone who may not know much about bash 🙂.

AnonymousCodes911:
@MatthewMiddlehurst, added the comments to make it more readable!

MatthewMiddlehurst:
Three issues I have with this really at the moment.

1. It is rather rigid, requiring a specific phrase and formatting
2. The code does not really match the style of other actions, it would be better as a bash script (or preferably a python file using pygithub).
3. There is nothing to indicate that this is available, I don't expect new users (who are most likely to not have write access) to know this is available. Probably best place is the contributor guide.

AnonymousCodes911:
Hi @MatthewMiddlehurst !, so answer your questions: 
> 1. It is rather rigid, requiring a specific phrase and formatting

 We can make it more generic to include flexibility in the phrases. 
by making it accept phrases like : 
```python
"Aeon-Assign bot", "assign this to", "please assign"
```
>2. The code does not really match the style of other actions, it would be better as a bash script (or preferably a python file using pygithub).

I'll opt for using PyGithub


> 3. There is nothing to indicate that this is available, I don't expect new users (who are most likely to not have write access) to know this is available. Probably best place is the contributor guide.

We can add it to the contributor's guide or describe a simple bot, that would comment on the Issues/PRs stating how we can trigger this bot.

Do you think this approach is fine?

AnonymousCodes911:
@MatthewMiddlehurst please review.

MatthewMiddlehurst:
We are mostly volunteers and there is a lot of other work we do, so try to be a bit patient and we will get around to it 🙂

AnonymousCodes911:
I understand, I apologize if it felt rushed, I just meant it is up for review at your convenience.

MatthewMiddlehurst (commented):
Would it be possible to only run this if the message starts with @aeon-actions-bot?

AnonymousCodes911:
> Would it be possible to only run this if the message starts with @aeon-actions-bot? If so the comment text only needs to contain "assign" for it to run, probably.

Even now, if the text only contains " assign " it would still Assign the PR, I just added the other phrases as contingencies to keep the bot flexible.

AnonymousCodes911:
> I would look at the other workflows i.e. the PR comment one to see the token generator. That will allow the bot to assign.

@MatthewMiddlehurst, 
I think when the task is as trivial as just assigning the PR, we can directly use the personal access token (PAT) stored in the repository, and we would not need the app token(did you mean [this?](https://github.com/aeon-toolkit/aeon/blob/625a1448ecebca89d0312c68dbf3fdd5a96de8e8/.github/workflows/pr_opened.yml#L31))

MatthewMiddlehurst:
Using the app token will make the bot assign rather than github actions. Its cosmetic I guess.

I mean since you are explicitly asking the bot, you can just remove the list and run it on any string which contains "assign" (which would cover "assigning", "assign this to", etc). You could avoid parsing users (any word starting with @) as well, but probably not required.

.github/workflows/self_assign_bot.yml
```diff
@@ -0,0 +1,31 @@
...
+          python-version: '3.x'
+
+      - name: Install PyGithub
+        run: |
+          python -m pip install --upgrade pip
+          pip install PyGithub
+      - name: Set env
+        run: |
+          echo "ISSUE_NUMBER=$(jq --raw-output .issue.number "$GITHUB_EVENT_PATH")" >> $GITHUB_ENV
+          echo "COMMENT_BODY=$(jq --raw-output .comment.body "$GITHUB_EVENT_PATH")" >> $GITHUB_ENV
```
MatthewMiddlehurst, line 26: Thought we may be able to find the comment body via pygithub but you are probably right and this is easier

.github/workflows/self_assign_bot.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
```
MatthewMiddlehurst, line 17: I would set this to 3.10, same as the other actions

.github/workflows/self_assign_bot.yml
```diff
@@ -0,0 +1,31 @@
+name: Self-Assign Bot
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  self-assign:
+    runs-on: ubuntu-latest
```
MatthewMiddlehurst, line 8: go with ubuntu-20.04 like other actions

AnonymousCodes911: Understood!

.github/workflows/self_assign_bot.yml
AnonymousCodes911: done!

MatthewMiddlehurst (approved):
LGTM. I may give it and the docs a few tweaks eventually depending on how the usage goes, but it is good enough to start testing IMO.

Issue #1239, opened 2024-02-24
[ENH] Checklist for deprecation in V0.9.0


TonyBagnall:
### Describe the feature or idea you want to propose

Would be good to actively update this list rather than do it in batch mode

### Describe your proposed solution
- [x] Remove utils.validation.panel.py #1464
- [x] Remove overloaded multiplication in BaseRegressor #1464
- [x] refactor datatypes._adaptors to utils.conversion. Reversed deprecation, #1489
- [x] remove stratified_resample from utils.sampling.py #1464
- [x] remove classification.distance_based.ShapeDTW https://github.com/aeon-toolkit/aeon/pull/1470
- [x] move channel selection to its own package https://github.com/aeon-toolkit/aeon/pull/1466
- [x] remove seasonality.py from utils #1464 
- [x] Regressor pipeline init function #1488 
- [x] Remove SklearnRegressorPipeline #1488

Issue #788, opened 2023-10-01
[MNT] Deprecation of old ShapeDTW classifier


hadifawaz1999:
### Describe the issue

After having ShapeDTW in the distance module (see #754 ) , the old [ShapeDTW classifier](https://github.com/aeon-toolkit/aeon/blob/main/aeon/classification/distance_based/_shape_dtw.py) can be deprecated given its KNN couple with ShapeDTW.

The implementation supposes that shapeDTW with knn is simply knn-dtw on transformed series, this is false.

### Suggest a potential alternative/fix

When deprecation policy is back we can take care of this before releasing v0.9.0

TonyBagnall:
we should also consider whether to deprecate all the associated transformers that are used here and no where else. I have no real strong feeling either way, but we should discuss it

TonyBagnall:
this is done

PR #1470, opened 2024-04-24
[DEP] Remove shapeDTW

TonyBagnall:
Removes the shapeDTW classifier as part of #1239. Apparently its logically incorrect, see #788 @hadifawaz1999


hadifawaz1999:
yes am working now on a new implementation that is correct and more clear, should be ready soon @TonyBagnall

hadifawaz1999:
i think it should be removed from the notebooks as well

MatthewMiddlehurst (commented):
Is this still in the API? Probably should have asked that for the other deprecations 🙂

TonyBagnall:
> Is this still in the API? Probably should have asked that for the other deprecations 🙂

good point, I will check them all on this PR

MatthewMiddlehurst (approved):

PR #1472, opened 2024-04-24
 [ENH] Type hints for primitives and string arguments:Part1

### LINKED ISSUE #1454: [ENH] Type hints for primitives and string arguments

we would like to increase the use of type hints, and a good first issue is to add them for the simplest case when arguments are primitives (int, float, bool) or strings. Pick a package, look through the code and add type hints. Just for primitives for now, more complex types need a bit of thought. Many have npt.ArrayLike types, but we have not decided on standardising on that, so leave these alone. Dont feel you have to do them all, this is a good first issue to get you started. 

for example, the constructor for the anomaly detection algorithm STRAY has these type hints

```
    def __init__(
        self,
        alpha: float = 0.01,
        k: int = 10,
        knn_algorithm: str = "brute",
        p: float = 0.5,
        size_threshold: int = 50,
        outlier_tail: str = "max",
    ):
 ```
but the segmentation algorithm CLASP does not

```
def _is_trivial_match(candidate, change_points, n_timepoints, exclusion_radius=0.05):
    """Check if a candidate change point is in close proximity to other change points.
```

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by nileenagp**
#### Reference Issues/PRs
Fixes #1454

Adding type hints to 2 methods in the script: aeon.classification.distance_based._time_series_neighbors.
Methods are : __init__(),  get_test_params()

None

#### Any other comments?
Type hints are NOT added in other methods since they were not primitive types.

### DIFF
--- a/aeon/classification/distance_based/_time_series_neighbors.py
+++ b/aeon/classification/distance_based/_time_series_neighbors.py
@@ -69,12 +69,12 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
 
     def __init__(
         self,
-        distance="dtw",
-        distance_params=None,
-        n_neighbors=1,
-        weights="uniform",
-        n_jobs=1,
-    ):
+        distance: str = "dtw",
+        distance_params: dict = None,
+        n_neighbors: int = 1,
+        weights: str = "uniform",
+        n_jobs: int = 1,
+    ) -> None:
         self.distance = distance
         self.distance_params = distance_params
         self.n_neighbors = n_neighbors
@@ -214,7 +214,7 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
         return closest_idx, ws
 
     @classmethod
-    def get_test_params(cls, parameter_set="default"):
+    def get_test_params(cls, parameter_set:str ="default") -> list[dict]:
         """Return testing parameter settings for the estimator.
 
         Parameters
--- a/aeon/classification/distance_based/_time_series_neighbors.py
+++ b/aeon/classification/distance_based/_time_series_neighbors.py
@@ -214,7 +214,7 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
         return closest_idx, ws
 
     @classmethod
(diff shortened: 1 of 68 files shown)

### CODE REVIEW

**Comment by nileenagp** [comment]
All the checks are passed successfully. Please review.

**Review by TonyBagnall — APPROVED** [review]
LGTM, thanks for the PR

File: aeon/classification/distance_based/_time_series_neighbors.py [hunk 1]
```diff
@@ -69,12 +71,12 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
...
-        n_neighbors=1,
-        weights="uniform",
-        n_jobs=1,
-    ):
+        distance: str = "dtw",
+        distance_params: dict = None,
+        n_neighbors: int = 1,
+        weights: str = "uniform",
+        n_jobs: int = 1,
+    ) -> None:
```
MatthewMiddlehurst, line 79: Does not seem correct?

nileenagp: I used None here as I have used earlier also, since we don't return any values from init method. While I looked into the appropriate annotation, I can see that in PEP it mentioned as 
"(Note that the return type of __init__ ought to be annotated with -> None. The reason for this is subtle. If __init__ assumed a return annotation of -> None, would that mean that an argument-less, un-annotated __init__ method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that __init__ ought to have a return annotation; the default behavior is thus the same as for other methods.)"

I hope this is fine and I can go ahead with this in other scripts also?

MatthewMiddlehurst: Fair enough, if thats stated as the correct output then so be it. Would not be my choice personally but whatever 🙂.

MatthewMiddlehurst (commented):
The current ones here do not match the docstring currently as well I'm pretty sure.

n_jobs could be  `Optional[int]` (defaults to None)
distance and weights could be `Union[str, Callable]` (accepts functions)

PR #1476, opened 2024-04-26
[MNT] Changes to no soft dependency tests

MatthewMiddlehurst:

Merged the two soft dependency workflow jobs into a single one which runs all tests on the periodic workflow and all tests but the "TestAllEstimator" tests on PRs.

Slims down the module soft dependency check. This did not actually work previously as the module variable was an empty list. Previously the test would check for "soft dependency modules" which rely on soft dependencies, the new test removes this concept. Everything must be isolated with no exceptions (yet) in the new version. Skips modules containing "tests" in the path on PRs.


TonyBagnall (approved):
looks fine to me, maybe @chrisholder could take a look?

PR #1479, opened 2024-04-27
[MNT] Unit testing revamp part 1: `check_estimator`

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #1476 
See [AEP 05](https://github.com/aeon-toolkit/aeon-admin/pull/10/files)

Updates the `check_estimator` method to use a scikit-learn like method for generating estimator tests rather than the current `pytest` fixtures. The old method is still available for now in `check_estimator_legacy`.  `parametrize_with_checks` simply outputs the generated checks as parameters for a `pytest` function.

`_yield_estimator_checks.py` ports tests over from `test_all_estimators`. The data used and the calling of estimator methods (i.e. `fit` and `predict`) has to change due to the detachment from the "scenarios" framework used previously.

Quick run down of contents:
- Has some small changes to base classes mostly regarding meta-data, as the storing of it in class attributes breaks the condition of no changes in `predict`/`transform`.
- Implements `parametrize_with_checks` and `check_estimator` for non-deprecated interfaces. These are only run on test estimators for now.
- Implements the current `BaseEstimator` tests as checks for the above without the "scenario" framework.


aeon/base/_base.py
```diff
@@ -485,46 +485,44 @@ def get_test_params(cls, parameter_set="default"):
...
+    def create_test_instance(cls, parameter_set="default", return_first=True):
         """
         Construct Estimator instance if possible.
 
+        Calls the `get_test_params` method and returns an instance or list of instances
+        using the returned dict or list of dict.
+
         Parameters
         ----------
         parameter_set : str, default="default"
```
baraline, line 497: Missing `cls` docstring

aeon/base/_base_collection.py
```diff
@@ -43,7 +43,7 @@ def __init__(self):
         super().__init__()
         _check_estimator_deps(self)
 
-    def _preprocess_collection(self, X):
+    def _preprocess_collection(self, X, store_metadata=True):
         """Preprocess input X prior to call to fit.
 
         1. Checks the characteristics of X, store metadata, checks self can handle
```
baraline, line 49: Missing `store_metadata`  doc string

aeon/base/_base.py
baraline, line 497: Missing `cls` docstring

aeon/testing/estimator_checking/_estimator_checks.py
```diff
@@ -0,0 +1,337 @@
...
+    independently and report the checks that are failing.
+
+    scikit-learn provides a pytest specific decorator,
+    :func:`~sklearn.utils.estimator_checks.parametrize_with_checks`, making it
+    easier to test multiple estimators.
+
+    Parameters
+    ----------
+    estimator : BaseEstimator instance or classe
+        Estimator instance or class to check.
```
baraline, line 164: Missing other parameter docstring, as this is a core function of the new testing module I feel like we would need the full docstring here

baraline (commented):
Had a quick look and I must say that the testing pipeline seems way clearer to me with this format, from testing data to which test are run for which estimator. 

I'll delve a bit more into it tomorrow, just some typos in found while reading through it.

aeon/base/_base.py
MatthewMiddlehurst: `cls` is just the class, like `self`.

aeon/testing/estimator_checking/_estimator_checks.py
MatthewMiddlehurst: The docstring for both of these methods are actually old, I forgot to update them! Good catch, will get it in next commit.

MatthewMiddlehurst:
Actually, with forecasting more officially deprecated now, I can probably remove the legacy check function. Can keep for now if people disagree.

TonyBagnall:
> Actually, with forecasting more officially deprecated now, I can probably remove the legacy check function. Can keep for now if people disagree.

I vote remove it

aeon/testing/estimator_checking/_estimator_checks.py
```diff
@@ -0,0 +1,411 @@
...
+    estimator: Union[BaseEstimator, Type[BaseEstimator]],
+    raise_exceptions: bool = False,
+    use_first_parameter_set: bool = False,
+    checks_to_run: Union[str, List[str]] = None,
+    checks_to_exclude: Union[str, List[str]] = None,
+    full_checks_to_run: Union[str, List[str]] = None,
+    full_checks_to_exclude: Union[str, List[str]] = None,
+    verbose: bool = False,
+):
+    """Check if estimator adheres to scikit-learn conventions.
```
TonyBagnall, line 143: aeon rather than scikit?

aeon/transformations/series/base.py
```diff
@@ -186,9 +184,7 @@ def inverse_transform(self, X, y=None, axis=1):
 
         # check whether is fitted
         self.check_is_fitted()
-        X = self._preprocess_series(
-            X, axis=axis, store_metadata=self.get_class_tag("fit_is_empty")
-        )
+        X = self._preprocess_series(X, axis=axis, store_metadata=False)
```
TonyBagnall, line 187: if store metadata is now false everywhere, I suggest remove it

TonyBagnall: but its not, its false in some places, need to look closer

aeon/testing/estimator_checking/_estimator_checks.py
MatthewMiddlehurst: good catch 🙂

TonyBagnall (approved):
LGTM, onwards!

baraline (approved):

PR #1483, opened 2024-04-29
[BUG] Set dynamic tag after `__init__` for classifiers and regressors.

MatthewMiddlehurst:
Bug fix for certain estimators based on the current failing coverage tests. Didn't look into why for now, but setting tags dynamically prior to `__init__` will reset them.

### DIFF
--- a/aeon/classification/hybrid/_rist.py
+++ b/aeon/classification/hybrid/_rist.py
@@ -132,7 +132,7 @@ class RISTClassifier(BaseRIST, BaseClassifier):
             n_jobs=n_jobs,
         )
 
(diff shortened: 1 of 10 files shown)


baraline:
I had that happend when working on the broadcaster. It happened when using clone or set_params / reset method I think. 
Is this the kind of thing we are looking for ?

MatthewMiddlehurst:
I'm not sure, does not seem intuitive. Would the tags reset on `fit` also if its `reset`? That does not seem intentional.

TonyBagnall (approved):
good spot, we saw this elsewhere, tags have a bespoke inheritance built in

PR #1486, opened 2024-04-30
[ENH] Channel selection

TonyBagnall:
this PR does the following:

1. adds a base class for channel selection because all channel selectors are going to choose a list of indexes, so transform is always the same. 

2. Adds a random channel selector as we used it in experiments. 

3. Adds a test_all_collection_transformers, with specific test for BaseChannelSelector class


aeon/transformations/collection/channel_selection/_random.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, p=0.4):
+        if p <= 0:
+            raise ValueError("Proportion to select cannot be zero or negative")
+        self.p = p
+        super().__init__()
+
+    def _fit(self, X, y):
+        """Randomly select channels to retain."""
+        to_select = math.ceil(self.p * len(X))
+        self.channels_selected_ = random.sample(list(range(X[0].shape[0])), to_select)
```
baraline, line 40: Wouldn't you want to add some seed parameter for reproducibility? 
Also what happens if p > 1 ? (I'm not sure)

baraline (commented):
Other than the seeding comment, the rest looks good to me

aeon/transformations/collection/channel_selection/_random.py
TonyBagnall: yes good point

MatthewMiddlehurst (approved):

Issue #1446, opened 2024-04-21
[DEP] BaseTransformer list for deprecation to BaseSeriesTransformer 


TonyBagnall:
### Describe the issue linked to the documentation

we are scrapping the whole BaseTransformer with its grossly inefficient VectorizedDF. This is a list of all files and classes to track changes. Some of these we may be able to just remove, some of it seems unnecessary, my thoughts in brackets. 

aeon.transformers:
- [ ] _delegate.py: _DelegatedTransformer (remove)
- [ ] acf.py: AutoCorrelationTransformer, PartialAutoCorrelationTransformer
- [ ] adapt.py: TabularToSeriesAdaptor, PandasTransformAdaptor
- [ ] augmenter,py: WhiteNoiseAugmenter, ReverseAugmenter, InvertAugmenter, RandomSamplesAugmenter (remove?)
- [ ] binning.py: TimeBinAggregate
- [ ] bkfilter.py: BKFilter
- [ ] boxcox.py: BoxCoxTransformer, LogTransformer
- [ ] clasp.py: ClaSPTransformer
- [ ] clear_sky.py: ClearSky
- [ ] compose.py: eek: ColumnwiseTransformer,ColumnConcatenator,FeatureUnion,FitInTransform, Id,InvertTransform,MultiplexTransformer,OptionalPassthrough,TransformerPipeline,YtoX
- [ ] cos.py: CosineTransformer
- [ ] date.py: DateTimeFeatures
- [ ] difference.py: Differencer
- [ ] dobin.py: DOBIN: (move to outlier detection)
- [ ] exponent.py: ExponentTransformer, SqrtTransformer (remove both)
- [ ] feature_selection.py: FeatureSelection
- [ ] fourier.py: FourierFeatures
- [ ] func_transform.py: FunctionTransformer
- [ ] impute.py: Imputer
- [ ] kalman_filter.py: BaseKalmanFilter, KalmanFilterTransformer
- [ ] lag.py: Lag, ReducerTransform
- [ ] outlier_detection.py: HampelFilter (move to outlier detection)
- [ ] pca.py: PCATransformer (remove)
- [ ] scaledlogit.py: ScaledLogitTransformer
- [ ] subset.py: IndexSubset, ColumnSelect (remove)
- [ ] summarize.py: SummaryTransformer, WindowSummarizer
- [ ] theta.py: ThetaLinesTransformer
- [ ] timesince.py: TimeSince

aeon.transformers.bootstrap
[...]

PR #1494, opened 2024-05-05
[DEP] Deprecate simple wrapper transformers

TonyBagnall:
deprecate transformers we dont want to keep 

Removes 
- TimeBinning: not used anywhere, not covered by tests
- IndexSubset and ColumnSelect: simple operations
- CosineTransformer: one line scikit wrapper
- SqrtTransformer: one line numpy wrapper
- Augmenters: adds noise to series, not really needed

part of https://github.com/aeon-toolkit/aeon/issues/1446


MatthewMiddlehurst (approved):
I don't think they have to be used in the package for them to be useful to have. Some of these seem like they could be useful. Still, happy to deprecate them given they are the old format and can be moved over later.

Issue #1243, opened 2024-02-24
[MNT] Update similarity search with new base classes


baraline:
### Describe the issue

As new base classes were introduced by #996 , we should update the similarity search module to use these classes. 
We are still in the context of a single query matching a series or a collection. The case of a collection of queries will come later. 

### Suggest a potential alternative/fix

Redisgn the API to use BaseSeriesEstimators as a base case and warp around it for collections.

TonyBagnall:
definitely, let me know if you want any input. BaseSeriesEstimator is still experimental,  so we can adapt to your use case if necessary.

baraline:
In my mind there is 4 main cases : 
- single query to single series
- single query to collection of series
- collection of query to single series
- collection of query to collection of series

All collection of series warp around the single case. In the case of top K matches, you simply update them iteratively or compute them independently in parallel for all single series. 

I'll try it out and keep you updated, but as the module is more aimed toward data analysis, I think the BaseSeriesEstimator looks fine as it is for now.

baraline:
Pasting the message I made on slack here to keep track of it.

Hi everyone, regarding PR #1310 and the update of the similarity search module, and to prepare for issue #1311 and the future additions, I would like your inputs on the following proposition:

### Module structure :

```
- aeon/
|---- similarity_search/
|-------- BaseSimilaritySearch.py
|-------- query_search/
|------------ BaseQuerySearch.py
|-------- series_search/
|------------ BaseSeriesSearch.py
|-------- index_search/
|------------ BaseIndexSearch.py
```

- Query search : Given a query Q and a series/collection X, evaluate the similarity between Q and each admissible candidate in X.
- Series search : Given a length parameter (for now, we add techniques that don't require it later) do a query search for all admissible queries in a series/collection X. In the naive case, this is simply a broadcasting of query search, but more optimized algorithms exists for this case (e.g. STUMP/STOMP for Euclidean distance)
- Index search : Given a series/collection X and a length parameter (again, for now), build an indexing (e.g.what the Faiss library does) of all admissible candidates in X. Then, this indexing can be used as an estimator to answer query search tasks. This is generally used when you have a frozen historical set and want fast answers for new queries / when the inputs do not fit in memory.

### Expected data and internal input conversion :

We could accept series/collection in numpy and series in pd.Series data as input, but we would ideally convert all of it to numpy collection, and make use of the axis argument we introduced in other modules to avoid the channel problem:

[...]

Issue #1311, opened 2024-03-13
[ENH] Implement SWAMP for DTW similarity search


baraline:
### Describe the feature or idea you want to propose

Implementation of SWAMP for DTW.

### Describe your proposed solution

As described in https://arxiv.org/ftp/arxiv/papers/2009/2009.07907.pdf

### Additional context

Wait for #1243 to be completed.

baraline:
We could also do the same for MSM distance based on https://arxiv.org/pdf/2301.01977.pdf

baraline:
Closing this as similarity search will not consider motif search in future updates

PR #1508, opened 2024-05-09
[MNT] Update similarity search with new base classes : Query Search

baraline:
#### Reference Issues/PRs

Part of #1243

As described in #1243, this is a first PR that implement base classes for query search task, and transfers the existing code under this new submodule.

#### Remaining TODOs:
- [x] Adapt distance profile function for unequal length
- [x] Add tests for unequal length for both dummy and top-k
- [x] Fix missing docstrings

Reported to another PR as this one is already big enough
- [x] Add typing to function/class parameters


aeon/similarity_search/tests/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+from numpy.testing import assert_almost_equal, assert_array_equal, assert_equal
+
+from aeon.similarity_search import BaseSimiliaritySearch
+from aeon.testing.utils.data_gen import (
+    make_example_3d_numpy,
+    make_example_unequal_length,
+)
+
+
+class BaseInstance(BaseSimiliaritySearch):
```
MatthewMiddlehurst, line 14: 1. Does it make sense to create a test class for this? I assumes it was abstract and each submodule would have its own base class
2. I would continue with the "MockEstimator" format we current have in `testing` if possible.

aeon/similarity_search/series_search/__init__.py
```diff
@@ -0,0 +1 @@
+"""Series search module."""
```
MatthewMiddlehurst, line 1: Empty

aeon/similarity_search/query_search/dummy.py
```diff
@@ -0,0 +1,85 @@
+"""Dummy similarity seach estimator."""
+
+__maintainer__ = ["baraline"]
+
+import numpy as np
+
+from aeon.similarity_search.query_search import BaseQuerySearch
+
+
+class DummyQuerySearch(BaseQuerySearch):
```
MatthewMiddlehurst, line 10: Similar to a different comment, but I would say the dummy estimators provide a "reasonable" baseline with a strategy (i.e. intentional random guessing or mean of the label), while mock estimators try to go through the interface as quick as possible even if the output is nonsense. May not apply here, but the top of the docstring gives that impression.

MatthewMiddlehurst (changes requested):
Some comments. All experimental so there some leeway, not going to say I went through every function with a fine tooth comb. Some of the comments may be a bit more general and apply to more than the highlighted code.

Would be good to update https://github.com/aeon-toolkit/aeon-admin/pull/1 with the contents of your comments just so there a centralised document for interested people until we decide to make this non-experimental.

I would be careful with the base class parameters, if it does not apply to all (or the vast majority) of inheriting classes then it probably should not be here. Not saying that is the case, but there are some bases i.e. clustering with `n_clusters` where we are going to run into issues later.

aeon/similarity_search/distance_profiles/naive_distance_profile.py
```diff
@@ -106,55 +117,62 @@ def normalized_naive_distance_profile(
...
+    distance_profiles = _normalized_naive_distance_profile(
         X, q, mask, X_means, X_stds, q_means, q_stds, dist_func
     )
+    # If input was not unequal length, convert to 3D np array
+    if isinstance(X, np.ndarray):
+        distance_profiles = np.asarray(distance_profiles)
+    return distance_profiles
 
 
-@njit(cache=True, fastmath=True)
```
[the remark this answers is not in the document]
MatthewMiddlehurst: If this is kept, at the very least it should be noted in the docstring and we should change the test config to disable numba threading.

aeon/similarity_search/query_search/dummy.py
baraline: The goal of dummy search is to return only the best match to the query, while other estimator may return something else (e.g. top k matches), changed doc string accordingly.

aeon/similarity_search/tests/test_base.py
baraline: The only thing of interest to test, that will be common to all future similarity search classes, here is the mean/std computation function, would this be a case for a Mock or a Dummy estimator ? Or just test these functions outside of this context ?

MatthewMiddlehurst: If the main purpose is the test the function, I would just test it directly. If you are also testing other base class functionality i.e. the contents of `fit` and `predict` I would create a "mock" estimator to do so in the style of other modules.

aeon/similarity_search/distance_profiles/naive_distance_profile.py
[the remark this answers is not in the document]
baraline: Added set_num_threads back to prev_num_threads as in rocket, and #1604

aeon/similarity_search/tests/test_base.py
baraline: After some reflection, I removed the base class for similarity search and moved its content to the base class of query search. A mock estimator has been created for query search base class.

MatthewMiddlehurst:
Any reason for the new base class direction? Do they other base classes not have much shared functionality?

baraline:
> Any reason for the new base class direction? Do they other base classes not have much shared functionality?

Yeah, after doing some pen and paper for the other two classes, the base class for similarity search would be pretty much empty. For example, index search will actually do things differently during fit as it need to build a similarity model, while series-search in the most naive way (without computational optimisations like MP) is simply looping over a query search for all possible candidates. The computational optimisations do require some rethinking of the fit also compared to query search.
So if we consider the three submodule, at least for now, I don't see a use for BaseSimilaritySearch class. Still possible to refactor afterward if we find a good reason to.

baraline:
On second thought, we could do the preprocess of the collection in it ... Would that be better for structure’s sake's ?

MatthewMiddlehurst:
Couple of things to consider but ofc can change later as you say.

If there are a significant number of shared parameters/attributes/functions used then it may be a good idea to keep even if its mostly abstract methods.

Also there may be some situations where you want to use `isinstance` to cover all of them? Maybe not also, not thought about it that much 🙂.

baraline:
> If there are a significant number of shared parameters/attributes/functions used then it may be a good idea to keep even if its mostly abstract methods.

For some reason, I forgot to consider this ... a bit out of touch today ! Swapping structure to add it back with the adjustment.

baraline:
Noticed some issue with the base class structure for the case of #1311, where the optimisation relies on lower bounding and not returning the distance profile fully computed, so I'll revamp it to be useable for all type of optimisations.

baraline:
Sorry for the mess in this PR ... To summarize :

The base query search class was made in the spirit of allowing any matching condition at the ouput of the search (e.g. top k matches, all matches below a distance threshold, ...). In practice, as there are only few plausible conditions: top k and/or threshold and worse-k and/or threshold, I just made a `QuerySearch` estimator with `k`, `threshold` and `inverse_distance` parameter that cover all these cases.

Additionally, there was the problem of computational optimisations that only compute part of the distance profiles (e.g. dtw lower bounding), but these optimisations need to know the matching condition (top-k, ...) to work. The previous structure would not have allowed that has distance profile computations were happening in the `BaseQuerySearch` class, which didn't have access to the matching condition.

aeon/similarity_search/base.py
```diff
@@ -104,364 +93,24 @@ def fit(self, X, y=None):
...
-        self._X = X.astype(float)
+        prev_threads = get_num_threads()
+        X = self._preprocess_collection(X)
+        # Store minimum number of n_timepoints for unequal length collections
+        self.min_timepoints_ = min([X[i].shape[-1] for i in range(len(X))])
+        self.n_channels_ = X[0].shape[0]
+        self.n_cases_ = len(X)
+        if self.metadata_["unequal_length"]:
+            X = List(X)
+        set_num_threads(self._n_jobs)
```
MatthewMiddlehurst, line 105: where is `self._n_jobs` defined? may be missing it.

aeon/similarity_search/base.py
```diff
@@ -104,364 +93,24 @@ def fit(self, X, y=None):
...
-
-        self._X = X.astype(float)
+        prev_threads = get_num_threads()
+        X = self._preprocess_collection(X)
+        # Store minimum number of n_timepoints for unequal length collections
+        self.min_timepoints_ = min([X[i].shape[-1] for i in range(len(X))])
+        self.n_channels_ = X[0].shape[0]
+        self.n_cases_ = len(X)
+        if self.metadata_["unequal_length"]:
+            X = List(X)
```
MatthewMiddlehurst, line 104: May be worth just doing this always? not sure how much the code differs later because of the different from lists and numpy arrays. If it has to be separate anyway no issue.

Doesnt have to be now, feel free to ignore.

baraline: I'm asking myself the same question, as you have to use the List indexing anyway to be compatible with unequal length... I think there was something performance related, but I'll have to take a closer look and do some benchmarking to be sure.

aeon/similarity_search/base.py
baraline: It is defined in the preprocess collection method of the BaseCollectionEstimator [here](https://github.com/aeon-toolkit/aeon/blob/d37a77fcd506757112e1f4dfccfca64c2b64acee/aeon/base/_base_collection.py#L46C9-L46C31)

MatthewMiddlehurst (approved):
Fine with this going in if you are happy with it @baraline. If any issues arise can be dealt with in a separate PR.

PR #1512, opened 2024-05-10
[MNT] Fix `BaseSeriesEstimator` test

MatthewMiddlehurst:
Series is no longer a valid inner type.


TonyBagnall:
I think I have figured out the cause, I'll attempt the fix then make more significant changes in a second PR

TonyBagnall:
see #1513

TonyBagnall (approved):

Issue #1498, opened 2024-05-07
[BUG]  matrix profile distance index out of bounds test failure


MatthewMiddlehurst:
### Describe the bug

The matrix profile distance is currently failing on the code coverage tests which disable numba. Most likely from changes in #1415.

Why does the `test_mpdist` test fail and the other distances ones don't, is the distance function only being used here?

### Steps/Code to reproduce the bug

Run the `test_mpdist` test without numba

https://github.com/aeon-toolkit/aeon/actions/runs/8978393264/job/24658747510

### Expected results

Test passes with and without numba

### Actual results

Test fails with no numba

### Versions

See CI failure

TonyBagnall:
I can reproduce this, its weird!

TonyBagnall:
I dont think a zero length subsequence makes much sense, and m should be constrained to be >0

TonyBagnall:
but then in _stomp_ab we have
```
    len_x = len(x)
    len_y = len(y)
    if m == 0:
        if len_x > len_y:
            m = int(len_x / 4)
        else:
            m = int(len_y / 4)
```
so zero is indicating set to 1/4 length really. I should go back and remind myself how it works lol

PR #1513, opened 2024-05-11
[BUG] mpdist window length

TonyBagnall:
fixes https://github.com/aeon-toolkit/aeon/issues/1498

the window length (called m for some reason) has as default value of 0, which means its set to a quarter of the length of the series. However, this check was happening after forming the windows, so it was trying to form windows with 0 length. 

I think it might be better to just use stumpy for this, m = None should be the default and k/threshold should definitely be a parameter. I will see if it makes a difference in a separate PR


MatthewMiddlehurst:
could you make a change to run the CI again? If there are other problems would be good to create an issue.

aeon/distances/_mpdist.py
```diff
@@ -69,6 +85,11 @@ def _mpdist(x: np.ndarray, y: np.ndarray, m: int) -> float:
     threshold = 0.05
     len_x = len(x)
     len_y = len(y)
+    if m == 0:
+        if len_x > len_y:
+            m = int(len_x / 4)
+        else:
+            m = int(len_y / 4)
```
baraline, line 92: If you do this operation here but `len_x > 4 * len_y` you are going to throw errors down the line.

aeon/distances/_mpdist.py
```diff
@@ -56,9 +63,18 @@ def mpdist(x: np.ndarray, y: np.ndarray, m: int = 0) -> float:
...
+            "to 1/4 of the length of the shortest time series"
+        )
     x = np.squeeze(x)
     y = np.squeeze(y)
-
+    if m > len(x) or m > len(y):
+        raise ValueError(
+            "subseries length must be less than or equal to the length "
+            "of both time series"
+        )
```
baraline, line 77: This should be after the handling of the `m=0` case

baraline (changes requested):

baraline:
Overall I agree that having m=0 just feels weird as default value, None would be better !

baraline (approved):
LGTM to me now !

PR #1524, opened 2024-05-13
[DEP] Rename `BaseDeepNetwork` and deprecate inheritance of `BaseObject`

MatthewMiddlehurst:
Decouples deep learning networks from the `BaseObject` structure, There is so much unnecessary baggage brought over from the base class that will never see usage in my opinion. The only tag used it for dependencies, which can easily be replaced with two parameters and function calls in the base class.


hadifawaz1999 (approved):
LGTM !

TonyBagnall (approved):
lets get this in since it touches a lot of files

PR #1526, opened 2024-05-14
[DOCS] Add any papers

TonyBagnall:
will just leave this here for a bit in case anyone has papers to add, will prob wait until after regression and jmlr are confirmed before merging


dguijo (approved):
LGTM, Tony! Thank you!!

Issue #1502, opened 2024-05-08
[BUG] load_classification function not working and time-series classification's website down


fmorenopino:
### Describe the bug

Hi, I think that the website https://www.timeseriesclassification.com/index.php has been down for a couple of days now. Further, on python, I cannot use the function load_classification (from aeon.datasets import load_classification) to download non-previously used datasets (for example, X, Y = load_classification("PhonemeSpectra") is not working).

### Steps/Code to reproduce the bug

```
from aeon.datasets import load_classification
X, Y = load_classification("PhonemeSpectra")
```

### Expected results

No error is thrown

### Actual results

```
Traceback (most recent call last):
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 1283, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 1329, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 1278, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 1038, in _send_output
    self.send(msg)
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 976, in send
    self.connect()
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 1448, in connect
    super().connect()
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/http/client.py", line 942, in connect
    self.sock = self._create_connection(
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/socket.py", line 824, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/nfs/home/fernandom/miniconda3/envs/contrastive/lib/python3.10/socket.py", line 955, in getaddrinfo
[...]

PR #1531, opened 2024-05-15
[ENH] Load tsc problems from zenodo as a back up to tsc.com

TonyBagnall:
due to the ongoing issues I am having with timeseriesclassification.com, we are backing up the UCR data to zenodo. 
https://github.com/aeon-toolkit/aeon/issues/1502

briefly, we noticed tsc.com was becoming very unreliable, leading to overnight test fails. For historical reasons, tsc.com was an addon to a ISP account I have held for 15 years. I thought switching to making tsc.com the main account might improve things. This involved backing up everything then deleting the addon. The ISP (uk2.net) are being very slow in responding to tickets, and whilst they have reassigned I dont know how to make the site live again.

This PR makes the load_classification function look on zenodo if tsc.com is down. I'll update the keys as the go up and then make this non draft


aeon/datasets/tsc_datasets.py
```diff
@@ -465,6 +465,170 @@
     "JapaneseVowels",
     "SpokenArabicDigits",
 }
+
+# 158 tsml time series classification problems [1]
```
MatthewMiddlehurst, line 469: cite may be outdated since its both mv and uv. Some comments to segment the dict by archive could be helpful for later.

TonyBagnall: I would rather move towards having one archive containing both uni and multi

MatthewMiddlehurst (approved):

PR #1543, opened 2024-05-20
[MNT] Remove stale branches

MatthewMiddlehurst:
Updates the branch workflow to delete stale branches after a while. 

Removes the self assign condition that the bot @ must be at the start.


TonyBagnall (dismissed):

TonyBagnall:
I've approved, but on reflection could you set to more than 175 for a while? There are a few that will get taken down automatically immediately, and I'd like to ask the authors first

MatthewMiddlehurst:
Its been 9999 for over a month at this point, you can see on the current issues created by the bot that they have been notified multiple times. I'll wait so it is not run this week though at least.

TonyBagnall (approved):

MatthewMiddlehurst:
@chrisholder this will kill your old branches on Monday if not updated

PR #1549, opened 2024-05-21
[ENH] Results lists and loading

TonyBagnall:
this changes the results loaders to use http://timeseriesclassification.com because there is a current issue with the security certificate after website transfer. Also corrects a typo in the univariate2015 list.

### DIFF
--- a/aeon/benchmarking/results_loaders.py
+++ b/aeon/benchmarking/results_loaders.py
@@ -275,7 +275,7 @@ def get_estimator_results(
     default_only=True,
     task="classification",
     measure="accuracy",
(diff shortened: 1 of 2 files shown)


hadifawaz1999 (approved):
LGTM

MatthewMiddlehurst:
Should maybe make an issue so we remember to change it back

TonyBagnall:
I actually need to check the data loaders too, may as well do it on this PR

TonyBagnall:
> I actually need to check the data loaders too, may as well do it on this PR

actually dont need to do the data loaders

PR #1146, opened 2024-02-07
[ENH] Function to write to Monash tsf style

itsdivya1309:
#### Reference Issues/PRs
Fixes #469 

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).

##### For developers with write access
- [ ] Optionally, I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #1552, opened 2024-05-21
[DEP] convert Scaled Logit

TonyBagnall:
part of #1146


ltsaprounis (approved):
Ah the scaled logit, very underrated transformer imo, but I digress. LGTM

PR #1564, opened 2024-05-24
[ENH] Improve and refactor collection test data generation utilities

MatthewMiddlehurst:
Changes to unify collection test data generation. Removes unused functions, updates present ones and adds new generators for missing data types. Adds documentation to all "make_example" functions.

- Reworks some of the data generation functions to follow the format of the 3d numpy and numpy list generators, i.e. allowing the generation of unequal length if possible and using the same parameters
- Refactors the name of some functions i.e. `make_example_unequal_length` -> `make_example_3d_numpy_list`
- Refactors some old but still used functions into a `_legacy` subdirectory until they can be removed.
- Moves the pre-generated testing data to `testing_data.py` and expands them to include all current collection data types
- Edits to allow 2d data in `is_collection` and a fix for multi-index data in `is_equal_length` 

Minor refactors from the above covers the majority of file changes, altering the parameters used, names and imports in test functions.


TonyBagnall (commented):
- allow 2d data in is_collection

Not sure I agree with this, although with a parameter I guess its ok. IMO the 2D should be handled as an exception in specific classes (as with classifiers) that inflate to 3D, otherwise we cannot ever tell 2d collection of univariate from 2d multivariate single series.

MatthewMiddlehurst:
Yeah I agree, I would only have it with a parameter to explicitly allow it really.

TonyBagnall (approved):

Issue #159, opened 2023-03-06
[ENH] Proximity Forest implementation using numpy 


TonyBagnall:
**Is your feature request related to a problem? Please describe.**

Proximity Forest is a distance based ensemble of decision trees. We do not currently have a working version of PF in aeon, but would very much like to have one. This is moderately complicated 

https://link.springer.com/article/10.1007/s10618-019-00617-3
https://arxiv.org/abs/1808.10594

The original java version is here
https://github.com/benjaminmlucas/proximity-forest

We have a java version here 
https://github.com/time-series-machine-learning/tsml-java/blob/master/src/main/java/tsml/classifiers/distance_based/proximity/ProximityForest.java

There is a version in the old repo, but it is broken and uses dataframes
https://github.com/sktime/sktime/blob/main/sktime/classification/distance_based/_proximity_forest.py

**Describe the solution you'd like**
I would like an implementation optimised for speed, that uses our distance functions and that can achieve accuracy results that are not significantly different to those published 

**Additional context**
We would be happy to help the development and test performance against other implementations.

PR #1566, opened 2024-05-24
[ENH] Implement Proximity Tree classifier using aeon distances

itsdivya1309:
Related to: #159 

We implemented the `ProximityTree` class using aeon distances.
This will be used to create the Proximity Forest class.

- Coded the `ProximityTree` class
- Wrapped Gini functions with Numba
- Wrote unit tests
- Added the class to docs/api_reference


TonyBagnall:
this is good. From our end @MatthewMiddlehurst we should set up a pipeline for running experiments to check equivalence to the Java results, as that was the heart of the problem last time. No rush of course, project has not started yet :)

itsdivya1309:
I have completed the framework for Proximity Tree along with it's API.

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,479 @@
...
+import numpy as np
+from numba import njit
+from sklearn.utils import check_random_state
+
+from aeon.classification.base import BaseClassifier
+from aeon.distances import distance
+
+
+class Node:
+
```
baraline, line 12: The docstring for the Node class parameter is missing.

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,479 @@
...
+        Takes a time series dataset and a set of parameterized
+        distance measures to create a candidate splitter, which
+        contains a parameterized distance measure and a set of exemplars.
+
+        Parameters
+        ----------
+        X : np.ndarray shape (n_cases, n_timepoints)
+            The training input samples.
+        y : np.array shape (n_cases,) or (n_cases,1)
```
baraline, line 172: Missing the description of `y`.

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,479 @@
...
+            A distance and its parameter values and a set of exemplars.
+        """
+        rng = check_random_state(self.random_state)
+
+        exemplars = {}
+        for label in np.unique(y):
+            y_new = y[y == label]
+            X_new = X[y == label]
+            id = rng.randint(0, X_new.shape[0])
+            exemplars[y_new[id]] = X_new[id, :]
```
baraline, line 188: You could simplify this bit to avoid creating the `X_new` and `y_new` arrays. For example, you could create at the beginning of fit a private class attribute that store the class locations :
``` python
self.class_locs = []
for label in np.unique(y):
    self.class_locs.append(np.where(y==label)[0]) # use [0] as y is 1D
```
Then you could use this attribute to pick a sample by `id` from it directly
```python
for label in np.unique(y):
      id = rng.randint(0, self.class_locs[label].shape[0])
      exemplars[label] = X[id, :]
```

baraline (commented):
Some first comment on “surface” content, I'll have a proper look at the algorithmic correctness a bit later

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,498 @@
...
+    A Proximity Tree is similar to a decision tree classifier where
+    the splits are based on similarity to chosen time series exemplars.
+    A Proximity Tree splits the data based on the proximity of each instance to
+    each of a set of class exemplars based on a parameterized similarity measure. The
+    tree is constructed recursively from the root node down to the leaf nodes.
+    At each node (other than leaf nodes) a pool of candidate splitters are evaluated.
+    Each splitter has a set of exemplar time series for each class chosen randomly,
+    and a parameterized distance measure, also chosen randomly. The splitter that
+    maximizes the difference between the Gini impurity of the parent node and the
+    weighted sum of Gini impurity of the child nodes.
```
MatthewMiddlehurst, line 59: Think this needs another read-through, we can help if required.

i.e. "A Proximity Tree is similar to a decision tree classifier" It is a decision tree classifier rather than similar to and the final sentence ends a bit abruptly.

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,498 @@
...
+        The maximum depth of the tree. If None, then nodes are expanded until all
+        leaves are pure or until all leaves contain less than min_samples_split samples.
+    min_samples_split: int, default = 2
+        The minimum number of samples required to split an internal node.
+    random_state : int, RandomState instance or None, default=None
+        If `int`, random_state is the seed used by the random number generator;
+        If `RandomState` instance, random_state is the random number generator;
+        If `None`, the random number generator is the `RandomState` instance used
+        by `np.random`.
+    n_jobs : int, default = None
```
MatthewMiddlehurst, line 75: defaults to 1

aeon/classification/distance_based/_proximity_tree.py
```diff
@@ -0,0 +1,498 @@
...
+                y_subs[sub].append(y[j])
+            y_subs = [np.array(ele, dtype=y.dtype) for ele in y_subs]
+            gini_index = gini_gain(y, y_subs)
+            if gini_index > max_gain:
+                max_gain = gini_index
+                best_splitter = splitter
+        return best_splitter
+
+    def _build_tree(self, X, y, depth, node_id, parent_target_value=None):
+
```
MatthewMiddlehurst, line 249: It would be nice to have a bit of documentation here summarising the function.

aeon/classification/distance_based/tests/test_proximity_tree.py
MatthewMiddlehurst, line 1: I think it would be good to add a test giving the tree an extremely easy problem (i.e. random data where one half has larger values for class 0 and the other half has larger values for class 1). You can then test that the `accuracy_score` after `fit` and `predict` is `>=0.9` or similar. It should be able to do this IMO.

MatthewMiddlehurst (changes requested):
Looks good for the most part, left a few comments. We may revisit this later down the line if there are issues, but appears to be fine for now.

I would add this to the API documentation in `docs/` in this PR.

There are some `if` statements here which are not covered by testing. Not going to block this PR because of them, but if you find the time and can figure out how to get to them reasonably it would be nice. https://app.codecov.io/gh/aeon-toolkit/aeon/pull/1566/blob/aeon/classification/distance_based/_proximity_tree.py

MatthewMiddlehurst:
Try merging the main branch for the `pre-commit` test, I will take a further look if it does not work after.

aeon/classification/distance_based/_proximity_tree.py
itsdivya1309: Hi, thanks for the suggestion, but when I define the `class_locs` in the fit method, it gives an error, probably because the splitter method is called at every node, and only parts of the data are present there.

So, would it be OK to define the `class_locs` attribute in the splitter method? I tried it, and it doesn't raise an error.

baraline: Oh, what type of errors exactly ? The point is to avoid recomputations when possible, especially as the method will be called many time

itsdivya1309: It gives Index Error, index out of bound.

itsdivya1309:
> Try merging the main branch for the `pre-commit` test, I will take a further look if it does not work after.

I have merged the main branch, but it is still failing.

aeon/classification/distance_based/tests/test_proximity_tree.py
itsdivya1309: I have added a test for this.

aeon/classification/distance_based/_proximity_tree.py
itsdivya1309: Please have a look, I have updated the text.

aeon/classification/distance_based/_proximity_tree.py
baraline: Ok, then don't worry about it for now, we'll look at such optimisations once the base structure will be validated

TonyBagnall (dismissed):
non blocking changes from me, looks good

TonyBagnall (approved):
lgtm, thanks

PR #1571, opened 2024-05-25
[ENH] Convert stray and add max iterations for MERLIN

MatthewMiddlehurst:

Converts `STRAY` to the new anomaly detection format. Currently only outs booleans, previous some sort of score was stored in a class attribute.

Adds max iterations to `MERLIN` and stops it from processing NaNs


aeon/anomaly_detection/_merlin.py
```diff
@@ -86,24 +86,47 @@ def _predict(self, X):
         r = 2 * np.sqrt(self.min_length)
         distances = np.full(len(lengths), -1.0)
         indicies = np.full(len(lengths), -1)
+
+        it = 0
         while distances[0] < 0:
+            if it > 500:
+                indicies[0] = np.nan
+                distances[0] = 2 * np.sqrt(self.min_length)
```
CodeLionX, line 94: Where is the upper-bound distance `2 * np.sqrt(self.min_length)` coming from? `r` is also initialized to it. Should we create a constant for it?

aeon/anomaly_detection/_merlin.py
```diff
@@ -86,24 +86,47 @@ def _predict(self, X):
         r = 2 * np.sqrt(self.min_length)
         distances = np.full(len(lengths), -1.0)
         indicies = np.full(len(lengths), -1)
+
+        it = 0
         while distances[0] < 0:
+            if it > 500:
```
CodeLionX, line 92: Why 500 iterations? Should this be a parameter?

CodeLionX (changes requested):
Having the distance-calculation loop repeated three times really hurts my eyes. There are some minor differences from loop to loop, but maybe we can improve this. In addition, I don't know how fast this MERLIN implementation is, but the loops in the `_predict`-function also look like good candidates for jitting.

I have also found this useless comparison of numbers with `np.nan`, which always holds:

https://github.com/aeon-toolkit/aeon/blob/4135913df01fd1d84658858497d38d15649691d7/aeon/anomaly_detection/_merlin.py#L135-L137

`np.nan != np.nan` is also true:
image

aeon/anomaly_detection/_merlin.py
MatthewMiddlehurst: done

aeon/anomaly_detection/_merlin.py
MatthewMiddlehurst: That just happens to be the length for this loop, it's reset to the current length in later ones. 

This is not part of the algorithm, but it seems sensible to reset the upper bound if we have to cancel due to no anomalies being found for the length, given that the later `r` values are derived from the previous distances.

MatthewMiddlehurst:
Removed the `np.nan` and moved the loop to a function. Could probably work it around to jit it, but would have to remove the function input

aeon/anomaly_detection/_merlin.py
CodeLionX: Ah, I see. This was not obvious to me.

CodeLionX (approved):
LGTM now.

PR #1577, opened 2024-05-27
[ENH] Add AEDRNNNetwork

aadya940:
Implements Network based on Dilated Recurrent Neural Networks.


aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,157 @@
...
+    n_units : List[int], default="None"
+        Number of units in each GRU layer, by default None.
+        If None, default to [100, 50, 50].
+    """
+
+    def __init__(
+        self,
+        latent_space_dim=128,
+        n_layers=3,
+        dilation_rate=None,
```
hadifawaz1999, line 33: even if decoder doesnt have dilation, define dilation_rate_decoder and default to1, make sure to put this in the docstrings

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,157 @@
...
+    Parameters
+    ----------
+    latent_space_dim : int, default = 128
+        Dimensionality of the latent space.
+    n_layers : int, default = 3
+        Number of GRU layers in the encoder.
+    dilation_rate : List[int], default = None
+        List of dilation rates for each layer, by default None.
+        If None, default = powers of 2 up to `n_stacked`.
+    activation : str, default="relu"
```
hadifawaz1999, line 20: if the input is a list keep a list, by default none transformed to a list of default values, and if one element wrap it to list

hadifawaz1999 (changes requested):

aadya940:
@MatthewMiddlehurst Do the tests look Okay?

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,234 @@
...
+    temporal_latent_space : bool, default = False
+        Flag to choose whether the latent space is an MTS or Euclidean space.
+    """
+
+    def __init__(
+        self,
+        latent_space_dim=128,
+        n_layers_encoder=3,
+        n_layers_decoder=1,
+        dilation_rate=None,
```
hadifawaz1999, line 42: dilation_rate_encoder

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,234 @@
...
+        activation_encoder="relu",
+        activation_decoder=None,
+        n_units=None,
```
hadifawaz1999, line 45: also 2 parameters for n_units, encoder and decoder

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,234 @@
...
+    activation_decoder : Union[str, List[str]], default=None
+        Activation function of the single GRU layer in the decoder.
+        If None, defaults to relu.
+    n_units : List[int], default="None"
+        Number of units in each GRU layer, by default None.
+        If None, default to [100, 50, 50].
+    dilation_rate : Union[int, List[int]], default = None
+        List of dilation rates for each layer, by default None.
+        If None, default to a list of ones.
+    temporal_latent_space : bool, default = False
```
hadifawaz1999, line 33: would be good to have it as the first parameter next to latent space dim

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,234 @@
...
+            final, output = self._bidir_gru(
+                x,
+                self._n_units[-1],
+                activation=self._activation_encoder[-1],
+                return_sequences=True,
+            )
+
+            encoder_output_layer = tf.keras.layers.Conv1D(
+                self.latent_space_dim,
+                activation="relu",
```
hadifawaz1999, line 168: activation here should be linear as for the dense in the case its not a temporal latent space

hadifawaz1999 (changes requested):

aeon/networks/tests/test_ae_drnn.py
```diff
@@ -0,0 +1,54 @@
...
+
+import random
+
+import pytest
+
+from aeon.networks import AEDRNNNetwork
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
+def pytest_generate_tests():
```
MatthewMiddlehurst, line 11: This seems very overkill, if you want to parameterise I would just do it in the decorator rather than using a function to generate it.

This is probably generating way too many tests. Unless there is a reason to have them together, try split some off into different functions so it isn't growing the parameter set multiplicatively.

aeon/networks/tests/test_ae_drnn.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["tensorflow"], severity="none"),
+    reason="Tensorflow soft dependency unavailable.",
+)
+@pytest.mark.parametrize("params", params)
+def test_aedrnnnetwork_init(params):
+    """Test whether AEDRNNNetwork initializes correctly for various parameters."""
+    aedrnn = AEDRNNNetwork(**params)
+    encoder, decoder = aedrnn.build_network((1000, 5))
```
MatthewMiddlehurst, line 52: What is the input here? Does this need to be 1000, and does the input size impact how long this takes to run? 

We want these to be as quick as possible while remaining useful.

aadya940: Nope, when we call `build_network` , tensorflow will only compile the code to Tensorflow Graph (tf.Graph) which is invariant of the input shape. It's just a dummy input shape.

aeon/networks/_ae_drnn.py
aadya940: From what I understand, number of units in the decoder are computed using number of units in the encoder as per this line "The decoder is composed of a single RNN layer with GRU units of size (100+50+50) × 2 = 400". Hence, I've done this deliberately. What do you think?

aeon/networks/tests/test_ae_drnn.py
MatthewMiddlehurst: What exactly is the input doing if it has no impact on performance? I assume theres no actual building going on in the method its all just initialisation. I would still reduce it if possible just in case its using extra memory under the hood.

MatthewMiddlehurst:
Coverage reports are being uploaded now, see the checks. 

Report for the network is here: https://app.codecov.io/gh/aeon-toolkit/aeon/pull/1577/blob/aeon/networks/_ae_drnn.py

aeon/networks/tests/test_ae_drnn.py
aadya940: So what happens is when we call the `build_network` method, it will Compute the shape of the tensor before/after passing through each layer (use `model.summary()` to take a look). However, it doesn't allocate any memory under the hood until we call the `fit` method. Hence, the magnitude of the input shape doesn't amount to any additional overhead until we call the `fit` method on the model.

aadya940: I'll reduce it though.

aadya940:
Can we recheck the coverage? @MatthewMiddlehurst

MatthewMiddlehurst:
You can navigate to it in the actions tab.

aadya940:
Now its 95% coverage, which is Okay I guess? @MatthewMiddlehurst

MatthewMiddlehurst:
It would be good to cover the partials if possible, but I'm not going to demand 100%. Keep in mind testing is more than coverage, there probably not a lot to do here since its a single method with general testing also, but we do want to make sure that the output is correct and errors are properly raised as well where possible.

aeon/networks/_ae_drnn.py
hadifawaz1999: the default behavior can be like that, but i think its best to give ability to the user to control this parameter
if none is chosen for the number of units decoder, than it can be this as default

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,253 @@
...
+        If None, defaults to relu.
+    n_units_encoder : List[int], default="None"
+        Number of units in each GRU layer of the encoder, by default None.
+        If None, default to [100, 50, 50].
+    n_units_decoder : List[int], default="None"
+        Number of units in each GRU layer of the decoder, by default None.
+        If None, default to two times sum of units of the encoder.
+    """
+
+    _config["auto-encoder"] = True
```
MatthewMiddlehurst, line 40: Would this work? I was assuming you would have to redefine the dictionary here.

It's fine I guess as long as it is isolated to this class and does not impact others.

aadya940: Yes , it will work :)

image

aadya940: However, this will also tweak the "auto-encoder" key of the base class as well.

aadya940: So, we should also explicitly define _config["auto-encoder"]=False for non auto encoders as it could cause issues while spawning multiple threads?

aadya940: Then we could remove `_config` from the Base Class, else wouldn't it be the same thing?

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,257 @@
...
+
+    Parameters
+    ----------
+    latent_space_dim : int, default = 128
+        Dimensionality of the latent space.
+    temporal_latent_space : bool, default = False
+        Flag to choose whether the latent space is an MTS or Euclidean space.
+    n_layers_encoder : int, default = 3
+        Number of GRU layers in the encoder.
+    n_layers_decoder : int, default = 3
```
hadifawaz1999, line 19: default in doc should be 1 not 3

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,257 @@
...
+                    50 for _ in range(self.n_layers_encoder - 1)
+                ]
+        else:
+            self._n_units_encoder = self.n_units_encoder
+            assert isinstance(self.n_units_encoder, list)
+            assert len(self.n_units_encoder) == self.n_layers_encoder
+
+        if self.n_units_decoder is None:
+            self._n_units_decoder = sum(self._n_units_encoder) * 2
+            self._n_units_decoder = [
```
hadifawaz1999, line 126: i would name the two variables 2 different things
so i would do this:
```
self._n_units_decoder_ = sum(self._n_units_encoder) * 2
self._n_units_decoder = [
                self._n_units_decoder_ for _ in range(self.n_layers_decoder)
            ]
```

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+            expanded_latent_space = decoder_input_layer
+
+        for i in range(self.n_layers_decoder):
+            decoder_gru = tf.keras.layers.GRU(
+                self._n_units_decoder[i],
+                return_sequences=True,
+                activation=self._decoder_activation[i],
+            )(expanded_latent_space)
+            if i < self.n_layers_decoder - 1:
```
hadifawaz1999, line 217: why dilate only on last layer ?

hadifawaz1999 (changes requested):
just left some final comments

aeon/networks/_ae_drnn.py
aadya940: Isn't it all layers except the last layer?

hadifawaz1999: gonna reformulate my question then, why not dilate the last layer:p

aadya940: From what I remember it was to make it symmetric with the encoder. In the encoder we dilate the input for all layers except the last one and so we follow the same in the decoder.

hadifawaz1999: but in this case the decoder should dilate every layer except the first one
to be symmetrical
if the encoder has 4 layers, the first three dilated the last one no, then the decoder should be made of 4 layers, the first not dilated and the last three yes

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,257 @@
...
+                decoder_gru = tf.keras.layers.Lambda(
+                    self._dilate_input,
+                    arguments={"dilation_rate": self._dilation_rate_decoder[i]},
+                )
```
hadifawaz1999, line 221: two issues here:

- decoder_gru is calling Lambda layer wihtout taking any input
- the loop seems to take the same input again and again "expanded_latent_space"

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,269 @@
...
+
+__maintainer__ = []
+
+from aeon.networks.base import BaseDeepLearningNetwork
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+if _check_soft_dependencies(["tensorflow"]):
+    import tensorflow as tf
+
+    class _TensorDilation(tf.keras.layers.Layer):
```
hadifawaz1999, line 11: the fail on drnn i think is coming from here, because u didnt add a docstring to the new class
simply put a sentence that ends will full stop between """ """ under the class definition
image.png

hadifawaz1999: @aadya940 probably should check python version as well

hadifawaz1999 (changes requested):

aeon/networks/_ae_drnn.py
```diff
@@ -0,0 +1,271 @@
+"""Auto-Encoder based Dilated Recurrent Neural Networks (DRNN)."""
+
+__maintainer__ = []
+
+from aeon.networks.base import BaseDeepLearningNetwork
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+if _check_soft_dependencies(["tensorflow"], severity=None):
```
hadifawaz1999, line 8: its "none" to my knowledge not None

hadifawaz1999 (changes requested):
For this network, some changes should be done as follows:

- Given the encoder dilates all layers except the last one, then the symmetrical view is to dilate all layers in the decoder except the first one, the code is not currently doing this
- Following the above comment, in the case of default setup where the decoder has only one layer, simply follow what the original paper decided to do, if they dilate it then yes if not then no
- When temporal latent space is chosen, the logic of the whole network changes as the final states are no longer being used, given this is not the in the original proposal and more an enhancement we are adding, am happy to leave the logic as is for now like you did (i dont see other options now) but we should brainstorm a better way to do it

aeon/networks/_ae_drnn.py
hadifawaz1999: python version here probably is breaking CI as for some reason CI is failing on 3.12 (talking about the clusterer PR depends on this one)

aadya940:
Looks good now? @hadifawaz1999

hadifawaz1999:
> Looks good now? @hadifawaz1999

you seem to be dilating all encoder's layers now, why the change ? @aadya940

aadya940:
Oh by the way @hadifawaz1999, I forgot to tell you that dilation layers are placed as they currently are because this is the configuration that works with `temporal_latent_space = True`, this is a little different from the paper implementation because they don't use dilation in the decoder but we decided to use it. I hope it makes sense :)) 

If we aim for symmetrical `_TensorDilation` layer configurations for encoder and decoder, the output shape of the decoder deviates from what it should be eg: Instead of (100, 2), the output shape would turn out (50, 2). Thats why I've built the network in such a way that we don't have this issue.

hadifawaz1999:
> Oh by the way @hadifawaz1999, I forgot to tell you that dilation layers are placed as they currently are because this is the configuration that works with `temporal_latent_space = True`, this is a little different from the paper implementation because they don't use dilation in the decoder but we decided to use it. I hope it makes sense :))
> 
> If we aim for symmetrical `_TensorDilation` layer configurations for encoder and decoder, the output shape of the decoder deviates from what it should be eg: Instead of (100, 2), the output shape would turn out (50, 2). Thats why I've built the network in such a way that we don't have this issue.

aaahh okay okay i see what you did there :) i understand why you did the condition now, okay keep it like that then thanks for the explanation

hadifawaz1999:
should merge main to remove pre-commit fail @aadya940

hadifawaz1999 (approved):
all looks good should just add it to api

aadya940:
#1962

PR #1583, opened 2024-05-31
[ENH] Add AEBiGRUNetwork

aadya940:
Implements Auto-Encoder based on BiDirectional GRU Network.


aeon/networks/_ae_bgru.py
```diff
@@ -0,0 +1,110 @@
...
+from aeon.networks.base import BaseDeepNetwork
+
+
+class AEBiGRUNetwork(BaseDeepNetwork):
+    """
+    A class to implement an Auto-Encoder based on Bidirectional GRUs.
+
+    Parameters
+    ----------
+        latent_space_dim : int, default=128
```
hadifawaz1999, line 12: temporal_latent_space, default False, and if set to True, make sure that the output of the encoder is temporal series

hadifawaz1999 (changes requested):

aeon/networks/tests/test_ae_bgru.py
```diff
@@ -0,0 +1,54 @@
+"""Tests for the AEBiGRU Model."""
+
+import random
+
+import pytest
+
+from aeon.networks import AEBiGRUNetwork
```
hadifawaz1999, line 7: would be good to test if when temporal_latent_space is True does the space really become temporal or not, just checking the shapes i mean

hadifawaz1999: can leave that for later though :)
wanna get the networks in

hadifawaz1999 (changes requested):
this is good to go, just need the test to assert the shapes of input and output of the network are the same

hadifawaz1999 (approved):
lgtm

aadya940:
> this is good to go, just need the test to assert the shapes of input and output of the network are the same

Already done in `test_all_networks.py`.

aadya940:
There's a mistake here.
`_config.structure` should be "auto-encoder" not "encoder".

aadya940:
#1962

PR #1589, opened 2024-06-02
[REF] Refactor `testing.utils.data_gen` into `testing.data_generation`

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #1564.

#### What does this implement/fix? Explain your changes.

This is a clearer name to me. Would rather it is top level than in a utilities director since it's rather core to all our testing.

PR #1592, opened 2024-06-03
[ENH] Rework series test data generators

MatthewMiddlehurst:
#### Reference Issues/PRs

Relies on #1589 

- Renames 2d collection data generators to include collection in the name to reduce ambiguity
- Moves the old contents of `_series.py` and `_forecasting.py` to the legacy directory so they can be slowly phased out or reintroduced following the new format
- Adds new example data generators for our accepted series formats. These are simpler than the collection ones, only using a few lines of code currently.
- Testing for the above


TonyBagnall (approved):
I though this was already in, soz

PR #1597, opened 2024-06-03
[BUG] Fixing wrong output activation in MLPClassifier

### LINKED ISSUE #1596: [BUG] Wrong output activation in MLPClassifier
### Describe the bug

The activation used in MLPClassifier at the last layer is using the self.activation, however this parameter is for internal activation, the output activation should be softmax

_No response_

### Expected results

output probas sum to 1

### Actual results

output 0/1

### Versions

_No response_

hadifawaz1999:
Fix #1596

### DIFF
--- a/aeon/classification/deep_learning/_mlp.py
+++ b/aeon/classification/deep_learning/_mlp.py
@@ -166,7 +166,7 @@ class MLPClassifier(BaseDeepClassifier):
         input_layer, output_layer = self._network.build_network(input_shape, **kwargs)
 
         output_layer = keras.layers.Dense(
-            units=n_classes, activation=self.activation, use_bias=self.use_bias
+            units=n_classes, activation="softmax", use_bias=self.use_bias
         )(output_layer)
 
         self.optimizer_ = (


TonyBagnall (approved):
LGTM

aeon/classification/deep_learning/_mlp.py
```diff
@@ -166,7 +166,7 @@ def build_model(self, input_shape, n_classes, **kwargs):
         input_layer, output_layer = self._network.build_network(input_shape, **kwargs)
 
         output_layer = keras.layers.Dense(
-            units=n_classes, activation=self.activation, use_bias=self.use_bias
```
lngdet, line 169: This same issue is present in `CNNClassifier`, see [here](https://github.com/aeon-toolkit/aeon/blob/708db2e76914d78c1d36044d73b47ad23941996a/aeon/classification/deep_learning/_cnn.py#L205-L207).

hadifawaz1999: This was actually intentional in the case of CNNs because the original authors proposed to use sigmoid as activation at the last layer, and simply do a mean squared error loss, instead of cross entropy (when the activation is softmax)

hadifawaz1999: as you see the default loss in case of CNNclassifier its not cross entropy its mse

lngdet: Thanks for the feedback, @hadifawaz1999

PR #1601, opened 2024-06-04
[MNT] Fix `pre-commit` pushes for forks (hopefully) and easier CI label interactions

MatthewMiddlehurst:

The current pre-commit fix action does not work so well for forks. Given that we have a lot of new contributors without write access, hopefully this fix the issue. Also includes some checkboxes for CI labels i.e. full testing and codecov. I'm not sure you can check these without write access, but at least it shows that they exist and allows devs to use them easier.

- Changes name for checkout action, as a lot of these arent for `main`
- Adds checkboxes to the initial PR comment which can be checked to add or remove CI labels
- Uses a different action which does not rely on an app for pushing CI fixes which should work for forks


TonyBagnall (approved):
In the absence of any other comments, I'll approve

PR #1604, opened 2024-06-05
[MNT] Add threads limit for numba and TF

### LINKED ISSUE #1603: [MNT] Restrict the number of threads for numba and tensorflow in CI
### Describe the issue

Force numba and tensorflow to use only one thread during CI.

_No response_

### Additional context

_No response_

baraline:
#### Reference Issues/PRs

Fixes #1603 

Restrict thread base parallelism to 1 thread during CI.

#### Any other comments?

Do we need to check soft dep here ?


MatthewMiddlehurst (commented):
Some other ones possibly.

```
if _check_soft_dependencies("torch", severity="none"):
    import torch
    torch.set_num_threads(1)
```

```
os.environ["MKL_NUM_THREADS"] = "1"  # must be done before numpy import!!
os.environ["NUMEXPR_NUM_THREADS"] = "1"  # must be done before numpy import!!
os.environ["OMP_NUM_THREADS"] = "1"  # must be done before numpy import!!
```

Should probably move the imports to the function itself

baraline:
Also found two other env variable mentioned in some issues :
```
os.environ["OPENBLAS_NUM_THREADS"] = "1"
os.environ["VECLIB_MAXIMUM_THREADS"] = "1"
```
Moved all imports in the function

MatthewMiddlehurst (approved):

MatthewMiddlehurst:
Approved, but I will note the tests aren't actually being run. May be best to make a small change elsewhere to trigger them and add the full test label.

conftest.py
```diff
@@ -28,5 +26,35 @@ def pytest_addoption(parser):
...
+    os.environ["OMP_NUM_THREADS"] = "1"
+    os.environ["OPENBLAS_NUM_THREADS"] = "1"
+    os.environ["VECLIB_MAXIMUM_THREADS"] = "1"
+
+    from numba import set_num_threads
+
+    from aeon.testing import test_config
+    from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+    set_num_threads(1)
```
MatthewMiddlehurst, line 43: not a blocker but `numba.set_num_threads` would be better since its such a generic function name (`torch` uses it below as well).

MatthewMiddlehurst (approved):
LGTM. Better this way IMO, as we use all available threads in CI for each test.

PR #1608, opened 2024-06-05
[ENH] Implement AEAttentionBiGRUNetwork

aadya940:
Implements Attention Bidirectional Gated Recurrent Unit based Auto Encoder .


aadya940:
@hadifawaz1999 How can we use the Keras Subclassing API in this case, since the tests that don't have tensorflow will fail?

MatthewMiddlehurst:
See the Hydra transformation for ideas, there is ways to get around it. Not sure you can do local classes in Python? Probably dont want to either way 🙂

aadya940:
@MatthewMiddlehurst Thanks :-)

aadya940:
@hadifawaz1999 What will change if we add support for `temporal_latent_space` in this? Do we apply the `Attention` and `self._gate` if `temporal_latent_space` is True since `self._gate` will Flatten the input tensor and convert it from 3 Dimensions to 2 Dimension and the output will no longer be temporal?

hadifawaz1999:
> See the Hydra transformation for ideas, there is ways to get around it. Not sure you can do local classes in Python? Probably dont want to either way 🙂

yup should work like we have in hydra

hadifawaz1999:
> @hadifawaz1999 What will change if we add support for `temporal_latent_space` in this? Do we apply the `Attention` and `self._gate` if `temporal_latent_space` is True since `self._gate` will Flatten the input tensor and convert it from 3 Dimensions to 2 Dimension and the output will no longer be temporal?

let me check the architecture and get back to you

hadifawaz1999:
> > @hadifawaz1999 What will change if we add support for `temporal_latent_space` in this? Do we apply the `Attention` and `self._gate` if `temporal_latent_space` is True since `self._gate` will Flatten the input tensor and convert it from 3 Dimensions to 2 Dimension and the output will no longer be temporal?
> 
> let me check the architecture and get back to you

hadifawaz1999:
if i understand correctly, the number of params for RNN layers depend on the batch size
right ? as it depends on x.shape[0]

aadya940:
Yep, that's what I understood.

aadya940:
@hadifawaz1999 Can we really parametrize the number of layers here? Since,  after the input passes through a Dense layers , its dimensions decrease from 3D to 2D and hence, we cant pass it to the next GRU layer. Also, the implementation here uses just 1 layer in the encoder, as well as the decoder:
https://github.com/blafabregue/TimeSeriesDeepClustering/blob/main/networks/attention_rnn.py

aadya940:
Could we just expand the dims of the tensor after each dense layer as follows?
```
if not (i == self.n_layers_encoder - 1):
                x = tf.expand_dims(x, -1)
```

hadifawaz1999:
should merge main here too @aadya940

aadya940:
#1962

aeon/networks/_ae_abgru.py
```diff
@@ -0,0 +1,172 @@
+"""Implements an Auto-Encoder based on Attention Bidirectional GRUs."""
+
+from aeon.networks.base import BaseDeepLearningNetwork
+
+
```
hadifawaz1999, line 5: its missing a reference

hadifawaz1999 (approved):

PR #1610, opened 2024-06-05
[BUG] lcss bounded below to 0

### LINKED ISSUE #761: [BUG] lcss distance producing negative values in pairwise dist matrix
### Describe the bug

When running tests on distance module locally, noticed a random fail (not always) on the lcss distance in the test_sklearn_compatibility:test_clusterer. Apparently, after i printed the pairwise dist matrix, its producing a very small negative number on some random data, the value is in order of 10 to the power -17 but its still negative.

Run the mentioned test many times to see it.

### Expected results

it should be all positive

### Actual results

positive values only

### Versions

_No response_

TonyBagnall:
fixes #761 by just lower bounding LCSS to zero to avoid numba numerics, adds a test for all distances >=0


CodeLionX (approved):

hadifawaz1999 (approved):
nice fix

Issue #1606, opened 2024-06-05
[BUG] Problems with the size of data


rruizdeaustri:
### Describe the bug

Hi,

I want to use rocket algorithms to classify Gravitational waves. The size of my data is (400000, 2, 2048) where 2 is the number of channels and 2048 is the length of each time series. It does not work. 

Thank you !

Roberto

### Steps/Code to reproduce the bug
``` python
import sys
import numpy as np
import h5py
import time
from pathlib import Path

import tensorflow as tf
import matplotlib.pyplot as plt

from utils.configfiles import get_config
from utils.datasets import load_data_into_numpy, InjectionSNR
from utils.metrics import auc_snr_eval
import json

from aeon.classification.deep_learning import LITETimeClassifier
from aeon.classification.hybrid import HIVECOTEV1, HIVECOTEV2
from aeon.classification.convolution_based import Arsenal

from sklearn.metrics import roc_auc_score

# -----------------------------------------------------------------------------
# MAIN CODE
# -----------------------------------------------------------------------------

model = 'multirocket'

if __name__ == '__main__':

    # -------------------------------------------------------------------------
    # Preliminaries
    # -------------------------------------------------------------------------
    print(tf.config.list_physical_devices('GPU'))

    # Example usage with your configuration settings
    config = get_config()
    xtrain, ytrain = load_data_into_numpy(config['data']['training'])
    xtest, ytest = load_data_into_numpy(config['data']['testing'])
 
    injections_snr = InjectionSNR()

    if model == 'LITETime': 
     clf = LITETimeClassifier(batch_size=32, n_classifiers=5, n_epochs=50, file_path='checkpoints/', s
ave_best_model=True, best_file_name="best_model", verbose=True)
    elif model == 'hivecote':
     clf = HIVECOTEV2(time_limit_in_minutes=0.2, verbose=1)   
    elif model == 'multirocket':
     clf = Arsenal(rocket_transform="multirocket")
    else:
     print('wrong model')   
     sys.exit()   

    # Check unique values in data
    #unique_values, counts = np.unique(ytrain, return_counts=True)
    #print(f"Unique values in predictions: {unique_values}")
    #print(f"Counts of unique values: {counts}")
    #sys.exit()
    print(f"xtrain shape: {xtrain.shape}, type: {xtrain.dtype}")
    clf.fit(xtrain, ytrain)

    #Compute AUC versus SNR and plot
    ypred = clf.predict(xtest)

    # Check unique values in predictions
[...]

PR #1612, opened 2024-06-06
[BUG] Convert rockets to only use 32 bit floats internally

TonyBagnall:
See #1606 

Currently some rocket transforms convert to 32 bit, some convert to 64 bit, some do not conversion at all. This PR makes them all 32 bit

1. MultiRocket
currently converts 64 bit in fit, does nothing in transform but expects 64 bit
fit:        X = X.astype(np.float64)
transform: nothing
**now**
fit and transform to 32, number decorator in _transform now
@njit(
    "float32[:,:](float32[:,:],float32[:,:],Tuple((int32[:],int32[:],float32[:])),"

2.  MultiRocketMultivariate
fit:   X = X.astype(np.float64)
transform: nothing
**now**
fit and transform to 32, number decorator in _transform now
@njit(
    "float32[:,:](float32[:,:],float32[:,:],Tuple((int32[:],int32[:],float32[:])),"

3. MiniRocket
fit and transform:         X = X[:, 0, :].astype(np.float32)
now squeeze for same as others

4. MiniRocketMultivariate already converts to 32

5. MiniRocketMultivariateVariable, does some weird shit, seems 32 bit
6. Hydra: not numba afaik

### DIFF
--- a/aeon/transformations/collection/convolution_based/_minirocket.py
+++ b/aeon/transformations/collection/convolution_based/_minirocket.py
@@ -92,8 +92,8 @@ class MiniRocket(BaseCollectionTransformer):
         random_state = (
             np.int32(self.random_state) if isinstance(self.random_state, int) else None
         )
-
-        X = X[:, 0, :].astype(np.float32)
+        X = X.squeeze()
+        X = X.astype(np.float32)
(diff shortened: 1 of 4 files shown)


baraline (approved):
LGTM !

PR #1622, opened 2024-06-07
[ENH] Lower Bounding Distances for SAX and SFA

patrickzib:

This implements four Euclidean lower bounding distances - two for SFA and two for SAX:

1. dft_sfa_mindist: compare dft to sfa words
2. paa_sax_mindist: compare paa to sax words
3. sax_mindst: compare two sax words
4. sfa_mindst: compare two sfa words

These can be used for indexing. 

Further changes:

1. The PR adds a test case, too.
2. It fixes two bugs in PAA/SAX for z-normalization with 0 std (dividing my 0) and for segments of size 0.


aeon/distances/_dft_sfa_mindist.py
```diff
@@ -0,0 +1,151 @@
+__maintainer__ = []
+
+import numpy as np
+from numba import njit, prange
+
+from aeon.distances._utils import reshape_pairwise_to_multiple
+
+
+# @njit(cache=True, fastmath=True)
```
baraline, line 9: Is this supposed to stay commented ?

aeon/distances/tests/test_symbolic_mindist.py
```diff
@@ -0,0 +1,142 @@
...
+        #    print("ED is:         \t", ed)
+
+        assert mindist_paa_sax <= ed
+        #    print("mindist_paa_sax is:\t", mindist_paa_sax, mindist_sax)
+        #    print("ED is:             \t", ed)
+
+    # print("\n")
+    # print("All ok:")
+    # print("PAA-SAX-Min-Dist Tightness:\t", tightness_paa_sax / X_train.shape[0])
+    # print("SAX-Min-Dist Tightness:    \t", tightness_sax / X_train.shape[0])
```
baraline, line 64: Some print comments to clean out

aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -1133,3 +1230,59 @@ def shorten_words(words, amount, letter_bits):
...
+#         return n
+#
+#     # Find the position of the most significant bit
+#     msb = 0
+#     while n > 0:
+#         n >>= 1
+#         msb += 1
+#
+#     # Return the next power of two
+#     return 1 << msb
```
baraline, line 1288: Comments to clean up

baraline (commented):
Hey Patrick, just a quick review of the surface for now, I'll have a look at the paper during the week to refresh myself on the core of it and come back to it, but overall looks good !
The comment on the @njit comment is true for most new distances you introduced, was there something wrong with numba ?

aeon/distances/_dft_sfa_mindist.py
patrickzib: `squeeze()` is not supported by numba. However, `squeeze()` was just a convenience method, and can be removed. Thus, I did add njit again.

aeon/distances/tests/test_symbolic_mindist.py
patrickzib: Thank you. I cleaned it up

aeon/transformations/collection/dictionary_based/_sfa_fast.py
patrickzib: Done :)

baraline:
Great ! An alternative to `squeeze`, if you are sure that there is only 1D might be to use `ravel`, which should be supported.

patrickzib:
> Great ! An alternative to `squeeze`, if you are sure that there is only 1D with "empty" subdimensions might be to use `flatten` or `ravel`, which should both be supported.

`ravel()` and `flatten()` would turn any 2d array to a 1d array, right? For this to work, the input must be 1d with one empty subdimension. Otherwise, I kind of break the logic. Might be better to just leave it and throw the ValueError.

aeon/distances/tests/test_symbolic_mindist.py
```diff
@@ -0,0 +1,124 @@
...
+        mindist_paa_sax = paa_sax_mindist(
+            PAA_train[i], SAX_test[i], SAX_transform.breakpoints, X_train.shape[-1]
+        )
+
+        # Euclidean Distance
+        ed = np.linalg.norm(X[0] - Y[0])
+
+        if ed > 0:
+            tightness_sax += mindist_sax / ed
+            tightness_paa_sax += mindist_paa_sax / ed
```
baraline, line 51: The tightness values are not used for in the following, did you want to assert that the tightness of paa_sax will be higher than sax ?

aeon/distances/tests/test_symbolic_mindist.py
```diff
@@ -0,0 +1,124 @@
...
+            mindist_dft_sfa = dft_sfa_mindist(
+                SFA_train_dfts[i], Y_train_words[i], sfa.breakpoints
+            )
+
+            # Euclidean Distance
+            ed = np.linalg.norm(X[0] - Y[0])
+
+            if ed > 0:
+                tightness_sfa += mindist_sfa / ed
+                tightness_dft_sfa += mindist_dft_sfa / ed
```
baraline, line 120: Same comment on tightness values not used

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -114,7 +114,13 @@ def _get_paa(self, X):
             The output of the PAA transformation
         """
         if not self.znormalized:
-            X = scipy.stats.zscore(X, axis=-1)
+            # Safe, if std is 0
+            X = (X - np.mean(X, axis=-1, keepdims=True)) / (
+                np.std(X, axis=-1, keepdims=True) + 1e-8
+            )
```
baraline, line 120: You could use the functions defined in `aeon.utils.numba.general` (which ever of "z_normalise_series", "z_normalise_series_2d" and "z_normalise_series_3d" you need) for this, but this works fine as well

aeon/transformations/collection/dictionary_based/_sfa.py
```diff
@@ -147,9 +151,16 @@ def __init__(
...
         self.norm = norm
         self.lower_bounding = lower_bounding
-        self.inverse_sqrt_win_size = (
-            1.0 / math.sqrt(window_size) if not lower_bounding else 1.0
-        )
+        self.lower_bounding_distances = lower_bounding_distances
+
+        if lower_bounding_distances:
+            self.inverse_sqrt_win_size = (
+                1.0 / math.sqrt(window_size) if lower_bounding_distances else 1.0
```
baraline, line 158: Don't you mean `1.0 / math.sqrt(window_size) if lower_bounding else 1.0` here instead of `if lower_bounding_distances else 1.0` ?

aeon/transformations/collection/dictionary_based/_sfa.py
```diff
@@ -400,6 +433,56 @@ def _transform_case(self, X, supplied_dft=None):
...
+
+        Returns
+        -------
+        Array of words
+        """
+        if X.ndim == 3:
+            X = X.squeeze(1)
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=NumbaTypeSafetyWarning)
```
baraline, line 466: What were the warning raised here, if you remember ?

Additionally, I'm not sure filtering the warning by default is a good solution, do we have some kind of policy on this @MatthewMiddlehurst ?

aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -171,9 +174,16 @@ def __init__(
...
         self.norm = norm
         self.lower_bounding = lower_bounding
-        self.inverse_sqrt_win_size = (
-            1.0 / math.sqrt(window_size) if not lower_bounding else 1.0
-        )
+        self.lower_bounding_distances = lower_bounding_distances
+
+        if lower_bounding_distances:
+            self.inverse_sqrt_win_size = (
+                1.0 / math.sqrt(window_size) if lower_bounding_distances else 1.0
```
baraline, line 181: Same comment on `lower_bounding_distances` than above

aeon/transformations/collection/dictionary_based/_sfa_fast.py
baraline: You might have forgotten it ? Or maybe it will be in the next commit, then nevermind.

baraline (changes requested):
Didn't find much to say on the distance themselves, looks good to me. Some comment on possible typos.

Additionally, more out of curiosity than for this PR, would a multivariate version of the lower bounds where you consider each channel independently and sum the result work, as we are working with (squared) Euclidean distance ?

aeon/distances/tests/test_symbolic_mindist.py
patrickzib: Good catch :)

aeon/transformations/collection/dictionary_based/_sax.py
patrickzib: I was not aware of this method 👍 

The issue with z_normalise_series is that I have to know if the data is 1d, 2d or 3d and call the corresponding one, right? The code above will work regardless of the input dimension. Thus, I would prefer to keep it like that, ok?

aeon/transformations/collection/dictionary_based/_sfa.py
patrickzib: Good question. I have not touched this part of the code.

When debugging, I do not see any warnings. Not sure, if pytest catches these. I have removed the filter.

aeon/transformations/collection/dictionary_based/_sfa.py
patrickzib: This one is tricky. When I implemented the mindistances, I have noticed that the semantics of `lower_bounding` are wrong. To allow for lower-bounding distances, it must be set to `lower_bounding=False`, which is rather unintuitive. 

Thus, I have added a new `lower_bounding_distances` to the class. `not lower_bounding = lower_bounding_distances`. I have simplified the code.

I cannot simply change `lower_bounding` as it is heavily used in dictionary-based classification.

patrickzib:
> Didn't find much to say on the distance themselves, looks good to me. Some comment on possible typos.
> 
> Additionally, more out of curiosity than for this PR, would a multivariate version of the lower bounds where you consider each channel independently and sum the result work, as we are working with (squared) Euclidean distance ?

Very Interesting research question! I am not aware of a multivariate lower bounding distance measure. Yet, as you suggest, it should be possible to transform each channel ro SFA/SAX and sum up the lower bounds on each channel.

baraline (approved):
Changes have been adressed and things looks good to me now. 
Althougt I would consider for a future PR is to change the naming on `lower bounding` and `lower bounding distance` to avoid confusions in the future.

PR #1625, opened 2024-06-07
[ENH] `axis=1` function default and removal of axis class default for series transformation and segmentation

MatthewMiddlehurst:
Updates the other series base classes to default to `axis=1` like the anomaly detection base class.


TonyBagnall (approved):
well if we do it I agree we should be consistent

PR #1631, opened 2024-06-09
[MNT] Enhance networks module testing suite

aadya940:
- Re-implemented `test_all_networks.py` to create a more robust testing Mechanism.
- Added `_config` dict to keep track of library dependencies and compatible python versions and the network structure. 
- Checked if `build_network` returns correct return type, that is, KerasTensor if structure is encoder, Model object otherwise.


aadya940:
@MatthewMiddlehurst  let's check if this improves coverage or not.

MatthewMiddlehurst:
You will have to push another commit for the test to run unfortunately.

TonyBagnall (commented):
I think we can put this in without testing coverage, seems a good thing to do!

TonyBagnall (dismissed):

MatthewMiddlehurst (commented):
Taking a look, I don't think this is needed anymore actually, see line 33 of the new file where it skips individual estimators if there is a missing module.

Honestly, this test could be improved a lot though, the skip list and try/catch are a bit hacky. Should try to integrate `_check_soft_dependencies` without skipping the whole test.

aadya940:
I think we could skip the complete test if the `tensorflow` is not present. However, we could skip the networks which use tensorflow-addons?

MatthewMiddlehurst:
Ideally we shouldnt skip anything other than the base class, the only good reason to skip currently is it has a dependency which is not installed

MatthewMiddlehurst (commented):
It could be good to remove the current skipped classes and replace the except with one using the `_check_soft_dependencies(soft_dependencies)` and `_check_python_version(python_version)` methods. You may have to edit the new base class a bit to get the correct inputs.

If you can think of anything else we would want to add to this test, that would be good. Currently it does not check anything from the output.

aadya940:
> It could be good to remove the current skipped classes and replace the except with one using the `_check_soft_dependencies(soft_dependencies)` and `_check_python_version(python_version)` methods. You may have to edit the new base class a bit to get the correct inputs.
> 
> If you can think of anything else we would want to add to this test, that would be good. Currently it does not check anything from the output.

Could you please elaborate on this? I didn't quite understand. Thanks :)

aadya940:
How does this look? @MatthewMiddlehurst

aadya940:
Do we skip this using `pytest.mark.skipif` if the version is `3.10`? Looks like `tensorflow.keras` is incompatible with `3.10`.

aeon/networks/tests/test_all_networks.py
```diff
@@ -15,20 +20,47 @@ def test_all_networks_functionality():
...
+                    my_network = network_classes[i]()
+                else:
+                    continue
+        except Exception as e:
+            logging.error(f"Error instantiating {network_classes[i]}: {e}")
             continue
 
-        input_layer, output_layer = my_network.build_network(input_shape=input_shape)
+        if my_network is None:
+            continue
```
MatthewMiddlehurst, line 46: I would remove all this. Best way would be to check via `_check_soft_dependencies()` and `_check_python_version()` if the dependency for a class exists. There probably not a good way to do this currently for  `BaseDeepLearningNetwork` though.

i.e. rough psuedocode:
```
if _check_soft_dependencies() and _check_python_version():
    my_network = network_classes[i]()
else:
    continue
```

aeon/networks/tests/test_all_networks.py
```diff
@@ -15,20 +20,47 @@ def test_all_networks_functionality():
...
+            else:
+                input_layer, output_layer = my_network.build_network(
+                    input_shape=input_shape
+                )
+                assert input_layer is not None
+                assert output_layer is not None
+        except AttributeError as e:
+            logging.error(f"Error in network {network_classes[i]}: {e}")
+        except AssertionError as e:
+            logging.error(f"Assertion failed for network {network_classes[i]}: {e}")
```
MatthewMiddlehurst, line 66: just have the test fail, no need to catch

aeon/networks/tests/test_all_networks.py
```diff
@@ -15,20 +20,47 @@ def test_all_networks_functionality():
...
+                )
+                assert (
+                    encoder.layers[0].input_shape[0] == decoder.layers[-1].output_shape
+                )
+            else:
+                input_layer, output_layer = my_network.build_network(
+                    input_shape=input_shape
+                )
+                assert input_layer is not None
+                assert output_layer is not None
```
MatthewMiddlehurst, line 62: This is very specific, and it would be more appropriate to have sub-base class or tag rather then look at the name if this is something we want to generally test. I was thinking just check if input/output is a valid type rather then the current `is not None`

aeon/networks/tests/test_all_networks.py
```diff
@@ -15,20 +20,47 @@ def test_all_networks_functionality():
...
     for i in range(len(network_classes)):
-        if (
-            "BaseDeepNetwork" in str(network_classes[i])
-            or "BaseDeepLearningNetwork" in str(network_classes[i])
-            or "AEFCNNetwork" in str(network_classes[i])
-            or "EncoderNetwork" in str(network_classes[i])
-        ):
+
+        if "BaseDeepNetwork" in str(
+            network_classes[i]
```
MatthewMiddlehurst, line 25: probably dont need to include this as its deprecated.

aeon/networks/tests/test_all_networks.py
MatthewMiddlehurst: This would require changing the current networks to have a class attribute you can draw from for dependencies. Alternatively, we can just keep the old try/catch for now.

aadya940: Could we just port over the `_tags` from `BaseDeepNetwork` into `BaseDeepLearningNetwork`?

aadya940:
@MatthewMiddlehurst Can you please check why these are failing. Not very familiar with how to implement `_tags` tbh. :))

MatthewMiddlehurst:
There does not have to be a system not does it have to be added to the registry. `tags` can just be a regular class variable.

`network_classes[i]._tags["auto_encoder"]` does not work because there is no `"auto_encoder"` key in the class dictionary.

MatthewMiddlehurst (commented):
Rough code example:

```python
if _check_soft_dependencies(network_classes[i]._config["python_dependencies"], severity="none") 
      and _check_python_version(network_classes[i]._config["python_version"], severity="none")

     layer1, layer2 = my_network.build_network(input_shape=input_shape)

     if network_classes[i]._config["auto_encoder"]:
           encoder, decoder = my_network.build_network(input_shape=input_shape)
           assert layer1.layers[-1].output_shape[1:] == (my_network.latent_space_dim,)
           assert layer1.layers[0].input_shape[0] == layer2.layers[-1].output_shape

     if "tensorflow" in (network_classes[i]._config["python_dependencies"]:
          from tensorflow.keras import layers

          assert isinstance(laye1r, layers.Layer)
          assert isinstance(layer2, layers.Layer)
     else:
           raise error "not supported yet"
```

MatthewMiddlehurst:
Whether you need a tag for autoencoders or this is the structure you want all those networks to follow is up to you and @hadifawaz1999

aadya940:
Got it :+1: , shall update it once we take a decision. Thanks :))

aeon/networks/tests/test_all_networks.py
```diff
@@ -2,33 +2,73 @@
...
+def test_network_config(network):
+    """Tests if the config dictionary of classes is correctly configured."""
+    assert "python_dependencies" in network._config.keys()
+    assert "python_version" in network._config.keys()
+    assert "structure" in network._config.keys()
+    assert isinstance(network._config["python_dependencies"], list) and (
+        "tensorflow" in network._config["python_dependencies"]
+    )
+    assert isinstance(network._config["python_version"], str)
+    assert isinstance(network._config["structure"], str)
```
MatthewMiddlehurst, line 30: What are the possible options for this? Is it just "encoder" and "auto-encoder"? May be worth ensuring it is one of these if so, given that the testing below is a strict if/elif for them.

aeon/networks/tests/test_all_networks.py
```diff
@@ -2,33 +2,73 @@
...
+        if _check_soft_dependencies(
+            network._config["python_dependencies"], severity="none"
+        ) and _check_python_version(network._config["python_version"], severity="none"):
+            if "EncoderNetwork" == str(network):
+                if _check_soft_dependencies(["tensorflow-addons"], severity="none"):
+                    my_network = network()
+                else:
+                    with pytest.raises(ModuleNotFoundError):
+                        my_network = network()
+            else:
```
MatthewMiddlehurst, line 51: Not sure we need this? Would be caught in the previous if statement and it is already skipped above.

aeon/networks/tests/test_all_networks.py
aadya940: Right Now, its only "encoder" and "auto-encoder". However, as we've discussed in Slack, this list will likely extend in the future as we add networks for other purposes.

MatthewMiddlehurst (approved):
LGTM. Still some minor issues but does not need to be done here.

hadifawaz1999 (approved):
LGTM! would be good to check coverage now too

PR #1632, opened 2024-06-09
[ENH] Adds a SeriesToCollection broadcaster

TonyBagnall:
superseeds #1244 which has got a bit entangled with other things. 

Adds a BaseCollectionTransformer that takes a BaseSeriesTransformer and broadcasts it across all instances in the collection, also returning a collection. Simple first implementation for clarity.

If the series trans fit method is empty, this uses the same transformer on all instances. This complicates threading, so will defer that to a later date (it threading will always need a separate transformer for each instance to avoid interference).

its simple atm and can no doubt be improved, but would like one in there to use


hadifawaz1999 (approved):
LGTM

PR #1656, opened 2024-06-13
[MNT] Disable `pytest` threading by default for local runs

MatthewMiddlehurst:

From discussion in Slack. Threading removes defaults and does other things which can make debugging more complicated, which is usually the purpose of local `pytest` runs.

Removes the default configuration for threading and adds it manually to `pytest` calls in the CI.

The CI can still be threaded locally but now need an argument to do so.


TonyBagnall (approved):
lgtm

CodeLionX (approved):

PR #1658, opened 2024-06-13
[MNT] Exclude `docs` and `examples` from release source

MatthewMiddlehurst:

This PR would remove the `docs` and `examples` directories from being included in PyPi release downloads.

Upside: Our uploaded source files would be smaller in side, no longer having to include the documentation and images

Downside: Users will no longer be able to access our webpage documentation and notebook examples through the uploaded source files.


TonyBagnall (approved):
docs and examples are about 30MB together so I think I approve

chrisholder (dismissed):

CodeLionX (dismissed):
I think the smaller artifact size outweighs offline documentation support. Especially because the API documentation is still available using IntelliSense et al.

CodeLionX (approved):

Issue #1393, opened 2024-04-09
[ENH] Checklist for deprecation in V0.10.0


TonyBagnall:
#### Open PR
- [x] #1539 dobin.py: remove DOBIN #1671
- [x] #1473 remove ACF and PACF BaseTransformer wrappers #1678
- [x] #1550 theta.py: ThetaLinesTransformer #1676 
- [x] #1494 remove IndexSubset, ColumnSelect #1675
- [x] #1558 pca.py: PCATransformer  #1674 
- [x] #1565 clear_sky.py: ClearSky #1673 
- [x] #1493 Move BoxCox and Log BaseTransformers to legacy #1672
- [x] #1561 kalman_filter.py: BaseKalmanFilter, KalmanFilterTransformer #1669 
- [x] #1560 bkfilter.py: BKFilter statsmodels wrapper #1669 
- [x] #1559 timesince.py: TimeSince #1669
- [x] #1494 remove TimeBinning, CosineTransformer, SqrtTransformer, Augmenters #1668
- [x] #1378 Deprecate _serialize and load #1667 
- [x] #1369 Make VectorizedDF private #1666
- [x] #1534 remove difference.py: Differencer #1645 

#### Done

- [x] #1442 Remove check_X from utils.validation.forecasting.py
- [x] Remove save transform from RotationForestRegressor, RotationForestClassifier, FreshPRINCERegressor #1701
- [x] #1042 missed deprecation of utils.estimator_checks check_estimator function 
- [x] #1439 Remove check_scorer function from utils.validation.forecasting.py #1644
- [x] #1517 remove exceptions.py  #1644 
- [x] #1436 Remove probabilistic performance metric classes  #1630

TonyBagnall:
all done

PR #1669, opened 2024-06-14
[DEP] Remove kalman filter and time since

TonyBagnall:
scheduled deletion, see #1561 and #1559 part of #1393


MatthewMiddlehurst (approved):

MatthewMiddlehurst (changes requested):
Should probably remove the dependency as well actually.

Can close any Kalman filter issues and/or raise one for a proper implementation.

MatthewMiddlehurst (approved):

PR #1671, opened 2024-06-14
[DEP] v0.10 remove dobin

TonyBagnall:
see #1539 part of #1393


MatthewMiddlehurst (commented):
This one might be nice to keep IMO

TonyBagnall:
sure, go for it, but need to convert it to numpy and sort the axis out though, I dont think we should have axis 0 series transformers.

PR #1683, opened 2024-06-14
[DEP] Deprecate BaseTransformer

TonyBagnall:
last one out, turn off the lights, The whole lot will be removed, refactored or brushed under the carpet by release 0.11


MatthewMiddlehurst:
Could you run one of the currently deprecated estimators and make sure there are not 2 warning printed because of this?

MatthewMiddlehurst:
I'm going to propose we just roll this into v1.0.0 with one single package-wide deprecation message

PR #1687, opened 2024-06-16
[ENH,BUG] LearnedShapelets interfaced functions

### LINKED ISSUE #1387: [BUG] Fix `LearningShapeletClassifier` non-standard methods
### Describe the bug

The non-standard methods in `LearningShapeletClassifier` were breaking the CI, likely due to the naming of `transform`. These have been commented out for now.

They should be included again and some testing should be done to figure out if this is why it breaks.

N/A

### Expected results

These methods can be included without CI breakage.

### Actual results

CI breaks.
i.e. https://github.com/aeon-toolkit/aeon/actions/runs/8560508379/job/23459706210

### Versions

_No response_

TonyBagnall:
fixes #1387 by introducing a save_transformed_data parameter, same as RDST, then changing the contentious methods to get_transform and get_locations.  

Also have some more testing, we should have 100% coverage of all shapelet classifiers except for MrSQM.


aeon/classification/shapelet_based/_ls.py
```diff
@@ -109,6 +109,7 @@ def __init__(
         max_size: Union[int, None] = None,
         scale: bool = False,
         random_state: Union[int, None] = None,
+        save_transformed_data: bool = False,
```
MatthewMiddlehurst, line 112: This needs to be added to the docstring as well as the new public attribute. Make sure to include the attribute which it saves data to in the descriptions.

aeon/classification/shapelet_based/tests/test_rdst.py
```diff
@@ -0,0 +1,20 @@
...
+    """RDST tests."""
+    X = make_example_3d_numpy(return_y=False, n_cases=10)
+    y = np.array([0, 1, 0, 1, 0, 1, 0, 1, 0, 1])
+
+    clf = RDSTClassifier(estimator=RandomForestClassifier(n_jobs=1))
+    clf.fit(X, y)
+    p = clf._predict_proba(X)
+    assert p.shape == (10, 2)
+    p = clf._predict(X)
+    assert p.shape == (10,)
```
MatthewMiddlehurst, line 20: why do we need to test these directly? should all be covered by the general tests right?

aeon/classification/shapelet_based/tests/test_ls.py
```diff
@@ -0,0 +1,33 @@
...
+    with pytest.raises(ValueError):
+        clf.get_locations(X)
+    clf = LearningShapeletClassifier(max_iter=100, save_transformed_data=True)
+    with pytest.raises(ValueError):
+        clf.get_transform(X)
+    with pytest.raises(ValueError):
+        clf.get_locations(X)
+    clf.fit(X, y)
+    t = clf.get_transform(X)
+    assert isinstance(t, np.ndarray)
```
MatthewMiddlehurst, line 33: If testing multiple things in a single function could you comment above each section what we are testing

MatthewMiddlehurst (dismissed):
Test comments i.e. "Learned Shapelets tests." are a bit terse 🙂.

It is going to be confusing down the line if undocumented when we are testing for multiple things in a single test. Left one comment on the LS one, but it applied to the other tests also really.

aeon/classification/shapelet_based/tests/test_rdst.py
TonyBagnall: this is for coverage with a non standard classifier

aeon/classification/shapelet_based/tests/test_ls.py
TonyBagnall: I will but it seems self evident if you read it, one is with save transformed data, one without. This is testing, not a user function

MatthewMiddlehurst (approved):

Issue #669, opened 2023-08-17
[BUG] Implementation of DTW missing the square root


sylvaincom:
### Describe the bug

Hi,

Thanks for the great work at `aeon`.

Your implementation of DTW (`aeon.distances.dtw_distance`) differs from the one by `tslearn` (`tslearn.metrics.dtw`) as it does not seem the apply the square root.

### Steps/Code to reproduce the bug

The following returns an `AssertionError`:
```python
# Imports
import numpy as np
from aeon.datasets import load_arrow_head
from aeon.distances import dtw_distance as dtw_aeon
from tslearn.metrics import dtw as dtw_tslearn
from scipy.spatial.distance import pdist, squareform
from sklearn.cluster import AgglomerativeClustering
from sklearn.metrics import adjusted_rand_score

# Load some data
X, _ = load_arrow_head(return_type="numpy2d")

# Compute the distance matrix for different implementations
distance_matrix_dtw_aeon = squareform(pdist(X, dtw_aeon))
distance_matrix_dtw_tslearn = squareform(pdist(X, dtw_tslearn))

# Check if the distance matrix are the same
err_msg = "The distance matrix from DTW differs between aeon and tslearn"
assert np.allclose(distance_matrix_dtw_aeon, distance_matrix_dtw_tslearn), err_msg
```
The following returns `True`:
```python
distance_matrix_dtw_aeon_sqrt = np.sqrt(distance_matrix_dtw_aeon)
np.allclose(distance_matrix_dtw_aeon_sqrt, distance_matrix_dtw_tslearn)
```

### Expected results

No error is thrown

### Actual results

```python-traceback
AssertionError: The distance matrix from DTW differs between aeon and tslearn
```

### Versions

<details>
System:
    python: 3.10.6 (main, Oct 24 2022, 11:04:34) [Clang 12.0.0 ]
executable: [/Users/my_username/opt/anaconda3/envs/my_env/bin/python](https://file+.vscode-resource.vscode-cdn.net/Users/my_username/opt/anaconda3/envs/<my_env>/bin/python)
   machine: macOS-10.16-x86_64-i386-64bit

Python dependencies:
          pip: 22.1.2
   setuptools: 59.8.0
      sklearn: 0.0
         aeon: 0.4.0
  statsmodels: 0.14.0
        numpy: 1.23.5
        scipy: 1.9.3
       pandas: 2.0.3
   matplotlib: 3.5.3
       joblib: 1.2.0
[...]

PR #1697, opened 2024-06-18
[BUG] Remove squaring distances in KNN regression

### LINKED ISSUE #1355: [BUG] Squaring weights in `KNeighborsTimeSeriesRegressor`
### Describe the bug

refer: #1234 and #1240 
see [this](https://github.com/aeon-toolkit/aeon/blob/2240e5c59d1e8e6db9442343c5fc6a4b95258e98/aeon/regression/distance_based/_time_series_neighbors.py#L175) we dont need to square the weights here !

in `class KNeighborsTimeSeriesRegressor(BaseRegressor):`

```python
        if self.weights == "distance":
            ws = distances[closest_idx]
            ws = ws**2 # after removing this

            # Using epsilon ~= 0 to avoid division by zero
            ws = 1 / (ws + np.finfo(float).eps)
```

### Expected results

the [`test_time_series_neighbors.py`](https://github.com/aeon-toolkit/aeon/blob/2240e5c59d1e8e6db9442343c5fc6a4b95258e98/aeon/regression/distance_based/tests/test_time_series_neighbors.py) should return :
Expected predictions: [-216.06541863 -4.54133078 -324.7624233 ]

### Actual results

But it returns : 
```python
AssertionError: assert 94.78595550443961 < 1e-06
```
and Actual predictions : [-144.41037008 -25.55876587 -229.9764678 ]

### Versions

_No response_

TonyBagnall:
AS pointed out here
https://github.com/aeon-toolkit/aeon/issues/1355

we square the weights of the distance functions in regression, but the are not sqrted see #669 

squaring simply gives more influence to closer instances, much in the way HC2 weights. Its not necessarily wrong, but I dont think we should do it automatically. It will be a new issue to assess whether it makes a difference in performance.

fixes #1355


dguijo (approved):
LGTM! Thank you Tony!

PR #1698, opened 2024-06-18
[ENH] Remove two  hard coded arrays from minirocket

TonyBagnall:
part of #1699 currently in minirocket we have two huge hard coded arrays in both fit and transform, this replaces both with a numba function to work out permutations.

Could instead replace by a function that returns the hard coded 84x3 array, but this will not be costly and its neater.


aeon/transformations/collection/convolution_based/_minirocket.py
```diff
@@ -148,264 +149,7 @@ def _fit_biases(X, dilations, num_features_per_dilation, quantiles, seed):
     # equivalent to:
     # >>> from itertools import combinations
     # >>> indices = np.array([_ for _ in combinations(np.arange(9), 3)])
```
patrickzib, line 151: Why not use this code (that is commented out?

TonyBagnall: I assume because it wont work with numba, but I confess I didnt try! It was commented out on main

TonyBagnall: can confirm it doesnt work with numba

  File "C:\Code\aeon\venv\lib\site-packages\numba\core\compiler_machinery.py", line 311, in _runPass
    mutated |= check(pss.run_pass, internal_state)
  File "C:\Code\aeon\venv\lib\site-packages\numba\core\compiler_machinery.py", line 273, in check
    mangled = func(compiler_state)
  File "C:\Code\aeon\venv\lib\site-packages\numba\core\untyped_passes.py", line 86, in run_pass
    func_ir = interp.interpret(bc)
  File "C:\Code\aeon\venv\lib\site-packages\numba\core\interpreter.py", line 1372, in interpret
    flow.run()
  File "C:\Code\aeon\venv\lib\site-packages\numba\core\byteflow.py", line 124, in run
    runner.dispatch(state)
  File "C:\Code\aeon\venv\lib\site-packages\numba\core\byteflow.py", line 357, in dispatch
    raise UnsupportedError(msg, loc=self.get_debug_loc(inst.lineno))
numba.core.errors.UnsupportedError: Failed in nopython mode pipeline (step: analyzing bytecode)
Use of unsupported opcode (IMPORT_NAME) found
File "..\transformations\collection\convolution_based\_multirocket_multivariate.py", line 656:
def _transform(X, X1, parameters, parameters1, n_features_per_kernel=4):
    <source elided>
    # equivalent to:
    from itertools import combinations
    ^

patrickzib: Hmm, why not:

```python
    def __init__(...):
        [...]
        self.indices = np.array(
            [_ for _ in combinations(np.arange(9), 3)], dtype = np.int32
        )        
       [...]
```

```python

 X = _transform(
      X,
      X1,
      self.parameter,
      self.parameter1,
      self.n_features_per_kernel,
      self.indices
)

@njit(
      fastmath=True,
      parallel=True,
      cache=True,
)
def _transform(X, X1, parameters, parameters1, n_features_per_kernel, indices):
    [...] remove the indices construction
```

```python

biases = _fit_biases(
    X, dilations, num_features_per_dilation, quantiles, self.random_state, self.indices
)

@njit(
    fastmath=True,
    parallel=False,
    cache=True,
)
def _fit_biases(X, dilations, num_features_per_dilation, quantiles, seed, indices):
    [...] remove the indices construction
```


Just tested it, and it does work. This way, you do not need any numba implementations?

patrickzib: There is no real reason to create the indices within the numbarized-funcations. Also not to do it for every call to these methods. Would be much simpler to construct them once in init.

baraline: Agree with Patrick on this one, the generation of the combination is not likely to be a computing intensive part, neither it is a big array to store.  Just storing it as a private class attribute and using it in the call to transform function seems a good way to go about it.

TonyBagnall: sure, although I thought we were meant to not do things in the constructor? Its the most efficient way in any case. I just want to remove the 1000 lines of array.

TonyBagnall: the alternative is to hard code it in one place, but its horrible to do so

TonyBagnall: could I not do it once as a class variable? That would be the most efficient

TonyBagnall: " numbarized-funcations." sums up numba to me :)

TonyBagnall: so indices also used in a numba function also called from a static method), (strangely not numba tho), so object attribute in init wont work

TonyBagnall: class variable seems fine tho and is the most appropriate imo

TonyBagnall:
@patrickzib @baraline how about this?

baraline:
> @patrickzib @baraline how about this?

Don't know much about the implications of initializing it like this, I thought you could do it at the start of _fit, but both should be somewhat similar I guess ?

TonyBagnall:
> > @patrickzib @baraline how about this?
> 
> Don't know much about the implications of initializing it like this, I thought you could do it at the start of _fit, but both should be somewhat similar I guess ?

In Java this would be called once as a singleton first time an instance was created. If you do it in fit then you may as well use the numba version I implemented before.

baraline (approved):
In light of Tony's answer, this LGTM

TonyBagnall:
just confirmed via the font of all knowledge:)

image

PR #1702, opened 2024-06-19
[ENH] Add AEDCNNNetwork

aadya940:
Implements the AEDCNNNetwork.


aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,269 @@
...
+    def __init__(
+        self,
+        latent_space_dim=128,
+        temporal_latent_space=False,
+        num_layers_encoder=4,
+        num_layers_decoder=4,
+        kernel_size_encoder=3,
+        kernel_size_decoder=None,
+        activation_encoder="relu",
+        activation_decoder="relu",
```
hadifawaz1999, line 71: the decoder parameters i wouldnt make them input arguments, as i dont want to give the user ability to have a non symetrical view of the auto encoder, you can keep the parameters but internally just not as input args

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,269 @@
...
+            x = self._dcnn_layer(
+                x,
+                self._num_filters_encoder[i],
+                self._dilation_rate_encoder[i],
+                _activation=self._activation_encoder[i],
+                _kernel_size=self._kernel_size_encoder[i],
+            )
+
+        if not self.temporal_latent_space:
+            shape_before_flattent = x.shape[1:]
```
hadifawaz1999, line 206: typo "shape_before_flattent" should flatten not flattent
and use tf.keras.layers.Flatten() instead of GAP, in case of auto encoder, theoretically, GAP doesnt make sense

hadifawaz1999 (changes requested):
small changes requested

aeon/networks/_ae_dcnn.py
hadifawaz1999: for example you can see the original FCNNetwork does a GAP at the end, but in AEFCNNetwork, we use a flatten, because the GAP removes any information about the time, so wouldnt be possible to reconstruct it, so using  a flatten is more suitable

aadya940: Oh, didn't know. Will do it. :))

aadya940:
@MatthewMiddlehurst @TonyBagnall  Hello, the fail is not related to this PR.

hadifawaz1999:
@aadya940 update to main to resolve the fail

hadifawaz1999 (changes requested):
The decoder should contain transpose convolutions not standard ones, check how AEFCN and AEResNet are made

2nd thing, why is there AEBiGRU on this PR as well ?

aadya940:
> The decoder should contain transpose convolutions not standard ones, check how AEFCN and AEResNet are made
> 
> 2nd thing, why is there AEBiGRU on this PR as well ?

They use normal convolutions in the decoder as well, check here:

[https://github.com/blafabregue/TimeSeriesDeepClustering/blob/main/networks/dilated_causal_cnn.py#L250](https://github.com/blafabregue/TimeSeriesDeepClustering/blob/main/networks/dilated_causal_cnn.py#L250)

hadifawaz1999:
> > The decoder should contain transpose convolutions not standard ones, check how AEFCN and AEResNet are made
> > 2nd thing, why is there AEBiGRU on this PR as well ?
> 
> They use normal convolutions in the decoder as well, check here:
> 
> https://github.com/blafabregue/TimeSeriesDeepClustering/blob/main/networks/dilated_causal_cnn.py#L250

@aadya940 
I had this discussion with the first author of the deep clustering bake off, they used normal convolution because back then there was no convolution transpose defined, so we decided if we want to get them into aeon we will get them with the technique done in AEFCNNetwork, so the same should be applied on AEDCNNNetwork

aadya940:
Oh Okay, however, Conv1DTranspose doesn't have padding="causal" kwarg. What to do?

hadifawaz1999:
> @hadifawaz1999 Oh Okay, however, Conv1DTranspose doesn't have padding="causal" kwarg. What to do?

i see, yes because it doesn't make sense mathematically to do causal conv transpose, put it as convT with padding same for now, we'll see later if we add another option to use conv normal with causal

hadifawaz1999:
@aadya940 you should be able to merge main now to fix the fail

aeon/networks/_ae_dcnn.py
hadifawaz1999: could you take care of this @aadya940 when you have time ? to be like AEFCN and AEResNet, as long as original authors didnt propose a non symmetrical version, i prefer obligating it

hadifawaz1999 (changes requested):
small request, and should n_ not num

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,219 @@
...
+    activation: str, default="relu"
+        The activation function used by convolution layers of the encoder.
+    n_filters: int, default=None
+        Number of filters used in convolution layers of the encoder.
+    dilation_rate: list, default=None
+        The dilation rate for convolution of the encoder.
+
+    References
+    ----------
+    .. [1] Network originally defined in:
```
hadifawaz1999, line 35: reference in non bib format, and no need to say "Network originally.." just the citation i think

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,219 @@
...
+
+__maintainer__ = []
+
+import numpy as np
+
+from aeon.networks.base import BaseDeepLearningNetwork
+
+
+class AEDCNNNetwork(BaseDeepLearningNetwork):
+    """Establish the network structure for a DCNN-Model.
```
hadifawaz1999, line 11: can u mention its the auto encoder version ?

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Dilated Convolutional Neural Network based Model
+    for low-rank embeddings.
+
+    Parameters
```
hadifawaz1999, line 16: if a parameter can be a list, it should be mentioned in the docstring

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,219 @@
...
+        elif not self.temporal_latent_space:
+            input_layer_decoder = tf.keras.layers.Input((self.latent_space_dim,))
+            dense_layer = tf.keras.layers.Dense(units=np.prod(shape_before_flatten))(
+                input_layer_decoder
+            )
+
+            reshape_layer = tf.keras.layers.Reshape(target_shape=shape_before_flatten)(
+                dense_layer
+            )
+            input_layer_decoder = reshape_layer
```
hadifawaz1999, line 153: why re-assign "input_layer_decoder" here to reshape_layer ?

hadifawaz1999 (changes requested):

aadya940:
#1962

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,218 @@
...
+        Dimension of the models's latent space.
+    temporal_latent_space : bool, default = False
+        Flag to choose whether the latent space is an MTS or Euclidean space.
+    n_layers: int, default=4
+        Number of convolution layers in the autoencoder.
+    kernel_size: Union[int, List[int]], default=3
+        Size of the 1D Convolutional Kernel of the encoder.
+    activation: Union[str, List[str]], default="relu"
+        The activation function used by convolution layers of the encoder.
+    n_filters: Union[int, List[int]], default=None
```
hadifawaz1999, line 28: can you add a sentence to n_filters and dilation_rate to how the default values are calculated ?

aeon/networks/_ae_dcnn.py
```diff
@@ -0,0 +1,218 @@
...
+
+__maintainer__ = []
+
+import numpy as np
+
+from aeon.networks.base import BaseDeepLearningNetwork
+
+
+class AEDCNNNetwork(BaseDeepLearningNetwork):
+    """Establish the Auto-Encoder based structure for a DCN Network.
```
hadifawaz1999, line 11: padding should be a parameter, and default to "same", having causal in encoder and same in decoder won't work i think, won't be an issue if we no longer use causal

aeon/networks/_ae_dcnn.py
hadifawaz1999: yes but the input of the decoder is changing, it should not be overridden it won't work like that

hadifawaz1999: you should define a temp layer that in the case of temporal latent space is set to input layer, and the other case is set to reshape layer, then y would be set to the temp layer

hadifawaz1999 (changes requested):
some changes and bug fix for the input decoder

aeon/networks/_ae_dcnn.py
aadya940: From what I remember "causal" in encoder & "same" in decoder does work.

aadya940: The tests do run successfully.

hadifawaz1999 (approved):

Issue #1699, opened 2024-06-18
[ENH] Tidy up the rocket transformers


TonyBagnall:
### Describe the feature or idea you want to propose

Its time to tidy up the convolutional transformers, will collate all issues here and make tasks for smaller PRs. Replaces #208 

### To Do

- [ ] Combine minirocket and minirocket variable length
- [ ] GPU implementations, benchmark vs CPU #313  #1248 
- [ ] Profile memory usage #1126
- [ ] Tidy up multirocket implementation, remove duplicate methods and repeated code from minirocket
- [ ] Decorate numba functions in rocket, test if any speed up
- [x] Test for numpy 2 compatibility, and optimise (e.g.     G = _X + _X + _X)
 
### Done
- [x] Minirocket has hard coded arrays that reduce readability. Replace with a numba function Done: #1698 
- [x] Combine minirocket and minirocket multivariate #1708 
- [x] Benchmark variants against alternative implementations #214 
- [x] Combine two multirocket variants into one #1711

TonyBagnall:
Time wise results as expected, also comparable to other implementations

![image](https://github.com/aeon-toolkit/aeon/assets/9594042/1fd246ec-51ca-4f8e-9a5b-4912ea452615)

TonyBagnall:
after #1781, next issue is to allow variable length. #1746 PR does it by taking numba inside the loop, but it slows it down unacceptably. An alternative, which is already in Rocket, is not to give types in njit arguments. Surprisingly, Rocket does not do this

```python


@njit(fastmath=True, cache=True)
def _apply_kernel_univariate(X, weights, length, bias, dilation, padding):
    n_timepoints = len(X)

    output_length = (n_timepoints + (2 * padding)) - ((length - 1) * dilation)
```

whereas minirocket does 
```python

@njit(
    "float32[:,:](float32[:,:],Tuple((int32[:],int32[:],int32[:],int32[:],float32["
    ":])), int32[:,:])",
    fastmath=True,
    parallel=True,
    cache=True,
)
def _static_transform_uni(X, parameters, indices):
```
without them, its much easier to use multivariate, but crucially need to assess any performance hit. So some timing experiments

1. Run minirocket both with and without the annotation and time it, passing 3D numpy
2. Run rocket both with and without the annotation and time it (possible speed up for main),

based on these results, either remove type annotation and adapt to lists of numpy or add them and use split functions

TonyBagnall:
Code to test 1 and 2
[...]

PR #1708, opened 2024-06-19
[ENH] Merge MiniRocket and MiniRocketMultivariate

TonyBagnall:
merge these two transformers, part of #1699, deprecate multivariate version

based on advice from @baraline, tried merging just using the multivariate version and it worked. Will explore any performance hit later.  Prior to #1698, Minirocket and Minirocket multvariate were about 2000 lines long. After this PR combined both will be under 400 lines.

It could be done better, but at this point, with the issue #208 over a year old, I just want to move things forward.


baraline:
Is there a specific reason why you introduce two versions of fit/transform (multi, uni) ? Couldn't we use the multivariate code with `n_channels=1` ? 
Otherwise, the rest looks good to me at first glance.

TonyBagnall:
> Is there a specific reason why you introduce two versions of fit/transform (multi, uni) ? Couldn't we use the multivariate code with `n_channels=1` ? Otherwise, the rest looks good to me at first glance.

because its very complicated and I was looking for the simplest solution. Anything more can be done after imo

TonyBagnall:
> Is there a specific reason why you introduce two versions of fit/transform (multi, uni) ? Couldn't we use the multivariate code with `n_channels=1` ? Otherwise, the rest looks good to me at first glance.

I'll try it though

aeon/transformations/collection/convolution_based/__init__.py
```diff
@@ -3,6 +3,7 @@
 __all__ = [
     "Rocket",
     "MiniRocket",
+    "MultiRocketMultivariate",
```
baraline, line 6: I guess this was aider back for deprecation policy reasons ?

aeon/transformations/collection/convolution_based/_minirocket.py
```diff
@@ -220,86 +164,207 @@ def _quantiles(n):
...
     if a > b:
         return 1
-    else:
-        return 0
+    return 0
 
 
 @njit(
-    "float32[:,:](float32[:,:],Tuple((int32[:],int32[:],float32[:])), int32[:,:])",
+    "float32[:,:](float32[:,:,:],Tuple((int32[:],int32[:],int32[:],int32[:],float32["
```
baraline, line 220: Is there any known reason why we use 32bit precision ? Otherwise it might be a good thing to benchmark it one day, as when we use fastmath option I'm not sure it matters that much

baraline (approved):
Non blocking comments, but computing logic looks fine, test are passing even without numba, so LGTM

aeon/transformations/collection/convolution_based/__init__.py
TonyBagnall: yep

aeon/transformations/collection/convolution_based/_minirocket.py
TonyBagnall: I'm not sure, I think for memory issues? Can investigate as part of #1699

Issue #307, opened 2023-04-19
[MNT] `binder` install option does not install optional dependencies used in notebooks


MatthewMiddlehurst:
The `binder` install option should install these so that all notebooks can be run. I'm not even sure if all notebooks run in their current state, so this would require a bit of a review.

RohitP2005:
@MatthewMiddlehurst  hey i am interested in solving this issue, if its not yet assigned or worked upon

baraline:
Hey @RohitP2005, I don't think anyone is working on this currently, as this is an old issue, we should first check its status.
A first step would be to try to run the notebooks on binders to see if you get any failures due to missing optional dependencies.

baraline:
like this :
@aeon-actions-bot assign @RohitP2005

RohitP2005:
Thanks @baraline

MatthewMiddlehurst:
Is this still being worked on?

RohitP2005:
Yes , i will probably finish it within two days, sorry for the delay

RohitP2005:
The last time I tried, the build was successfully , but didn't run the notebooks.
It seems that the binder site is not working currently.
Can u guide me forwards

MatthewMiddlehurst:
Seems the site is currently down.

RohitP2005:
I guess, I will wait for the site to be back up running

RohitP2005:
Hey @MatthewMiddlehurst, if you don't mind, could you help by building the Binder and sharing the URL? My builds are currently failing. Once it's up, I'll check the notebooks and update the Binder dependencies myself.

MatthewMiddlehurst:
I don't think it works like that unfortunately. but I'm also not so familiar 🙂. I just tried the link and It works for me, just takes a bit if it the first time an image is created for the commit.

RohitP2005:
hey @MatthewMiddlehurst  i tried runnig those notebooks and most of them ran with no issue, but several notebooks ran into this erro
```OMP: Info #276: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.```

am i missing smth here?

MatthewMiddlehurst:
That should not be the case. I just checked `tsfresh.ipynb` and got the following as expected.

![Image](https://github.com/user-attachments/assets/3cfca779-53cb-4b57-bc7a-b20cce83f345)

RohitP2005:
> That should not be the case. I just checked `tsfresh.ipynb` and got the following as expected.
[...]

PR #1719, opened 2024-06-25
[MNT] Only install binder dependencies for examples tests

MatthewMiddlehurst:
Completes #307. The notebook CI jobs installed `all_extras` alongside `binder`, which hid gaps in the binder extra. Those gaps only showed up on binder.

- Notebook CI now installs `.[binder,dev]` only, so a missing binder dependency fails CI.
- Added `keras`, `pydot`, `roughpy`, `stumpy` and `tsfresh` to the `binder` extra, all used by example notebooks. Pins match `all_extras`.
- Dockerfile: `quay.io` base image (Docker Hub `jupyter/*` is no longer updated), Python 3.12, graphviz for `plot_network`, and the llvmlite workaround removed.
- `signature_method.ipynb` is no longer excluded from CI, since RoughPy (#3754) has macOS wheels.
- Fixed the `shapelet_based.ipynb` failure in periodic tests: `_classifier` was renamed to `classifier_` in #3730.

`shapelet_based.ipynb` is skipped on PR runs, so CI only covers the fix with the `full examples run` label.


MatthewMiddlehurst:
@copilot resolve the merge conflicts in this pull request

TonyBagnall (changes requested):
my oracle says this:

dropping all_extras from the two example-notebook workflows silently removes torch, and the new binder extra doesn't add it back. In full-run mode this breaks two classification notebooks.

The blocking finding, reproduced. Both workflows install .[binder,dev]. The exclusion in run_examples.sh is gated on $1: normal PR runs pass $1=true and skip the heavy notebooks, but periodic_tests.yml runs run_examples.sh false true ($1=false, full set) and pr_examples.yml passes $1=false on push-to-main and on any full examples run-labelled PR. In the full set, classification/convolution_based.ipynb (HydraClassifier, MultiRocketHydraClassifier) and classification/interval_based.ipynb (QUANTClassifier) execute torch-backed estimators. With aeon installed minus torch:

HydraClassifier: ModuleNotFoundError: HydraClassifier requires package 'torch' ... 'torch' was not found
MultiRocketHydraClassifier: ModuleNotFoundError: ... requires package 'torch' ...
QUANTClassifier: ModuleNotFoundError: ... requires package 'torch' ...

torch is in all_extras but not in the new binder, so this is a straight regression from dropping all_extras. The trap is that this PR's own PR-examples check goes green (it runs with $1=true, which excludes both notebooks); the failure only surfaces on the periodic run and on the push-to-main examples run after merge. Fix: add torch>=1.13.1 (matching all_extras) to the binder extra. I confirmed torch 2.14.0 ships a cp312 macosx_14_0_arm64 wheel, so it installs on the macos-14 runners both jobs use.

MatthewMiddlehurst:
Good catch, added it to the list. Installing the CPU version for binder to reduce bloat a bit.

TonyBagnall (approved):

PR #1729, opened 2024-06-27
[ENH] Implement Proximity Forest classifier

### LINKED ISSUE #159: [ENH] Proximity Forest implementation using numpy

itsdivya1309:
#### Reference Issues/PRs
Closes #159 

Implementation of Proximity Forest Algorithm using the Proximity Trees.


aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,140 @@
...
+        self.n_jobs = n_jobs
+        super().__init__()
+
+    def _fit(self, X, y):
+        # Check dimension of X
+        if X.ndim == 3:
+            if X.shape[1] == 1:
+                X = np.squeeze(X, axis=1)
+            else:
+                raise ValueError("X should be univariate.")
```
MatthewMiddlehurst, line 102: Not sure about this here and in the tree as well actually. Probably better to just have "numpy2D" as the input type and let the base class handle conversion.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,140 @@
...
+                n_splitters=self.n_splitters,
+                max_depth=self.max_depth,
+                min_samples_split=self.min_samples_split,
+                random_state=self.random_state,
+                n_jobs=self.n_jobs,
+            )
+            clf.fit(X, y)
+            self.trees_.append(clf)
+
+        self._is_fitted = True
```
MatthewMiddlehurst, line 117: Shouldn't need this, handled in the base class.

MatthewMiddlehurst (commented):
Great work, we'll have to give it a run through on the UCR archive datasets as we discussed. Next steps are up to you really, we can discuss on Slack.

MatthewMiddlehurst:
This needs to be included in the API documentation also.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,129 @@
...
+    def _fit_tree(self, X, y):
+        clf = ProximityTree(
+            n_splitters=self.n_splitters,
+            max_depth=self.max_depth,
+            min_samples_split=self.min_samples_split,
+            random_state=self.random_state,
+            n_jobs=self.n_jobs,
+        )
+        clf.fit(X, y)
+        return clf
```
baraline, line 112: Similar comment for predict, but I think it might be better to define the function you parallelize with joblib outisde of the object you call them from. Something to do with pickling the objects, if I remember right ? 

For example, here, you would define `_fit_tree` as a function outside `ProximityForest`, as it don't even use the `self` reference

itsdivya1309: Thanks for pointing this out.

aeon/classification/distance_based/tests/test_proximity_forest.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_proximity_forest(time_series_dataset):
+    """Test the fit method of ProximityTree."""
+    X, y = time_series_dataset
+    clf = ProximityForest(n_trees=5, n_splitters=3, max_depth=4)
+    clf.fit(X, y)
+    X_test, y_test = time_series_dataset
+    y_pred = clf.predict(X_test)
+    score = accuracy_score(y_test, y_pred)
+    assert score >= 0.9
```
baraline, line 37: Testing the classifier for accuracy is a good practice, but doing it on randomly generated data might cause the test to fail at some point. If you want to do this, you can use the [unit test dataset](https://github.com/itsdivya1309/aeon/blob/2d74d4d3f8ed8041ca6b17eeb0443fb0ded81f83/aeon/datasets/_single_problem_loaders.py#L183).

baraline (changes requested):
Oher than this testing issue, the rest LGTM !

aeon/classification/distance_based/_proximity_forest.py
MatthewMiddlehurst: Is this true, I think we have functions elsewhere that do this. Interesting to see if that needs to be changed.

MatthewMiddlehurst (commented):
You should investigate how we use joblib in other estimators in the classification module. Typically we use "threads" as a default backend, with a parameter to change that,

The docstring `n_jobs` needs updating.

aeon/classification/distance_based/tests/test_proximity_forest.py
itsdivya1309: I have updated the code to use unit test. Thank you for suggesting this.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,140 @@
...
+                self.max_depth,
+                self.min_samples_split,
+                self.random_state,
+                self.n_jobs,
+            )
+            for _ in range(self.n_trees)
+        )
+
+    def _predict_proba(self, X):
+        output_probas = Parallel(n_jobs=self._n_jobs, prefer="threads")(
```
baraline, line 113: We discussed the need of a parameter for the joblib backend (i.e. threads vs processes), you should add a class parameter that default to threads

baraline (changes requested):
Only this small parameter missing and it should be good to go !

aeon/classification/distance_based/_proximity_forest.py
itsdivya1309: I thought I had added the parameter; thank you, I have added the parameter

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,149 @@
...
+        self.n_splitters = n_splitters
+        self.max_depth = max_depth
+        self.min_samples_split = min_samples_split
+        self.random_state = random_state
+        self.n_jobs = n_jobs
+        self.parallel_backend = parallel_backend
+        super().__init__()
+
+    def _fit(self, X, y):
+        self.classes_ = list(np.unique(y))
```
MatthewMiddlehurst, line 105: This is done in the base class, don't want to do so here. Caused an issue when trying to run it 🙂

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,149 @@
...
+            n_jobs=self._n_jobs, backend=self.parallel_backend, prefer="threads"
+        )(
+            delayed(_fit_tree)(
+                X,
+                y,
+                self.n_splitters,
+                self.max_depth,
+                self.min_samples_split,
+                self.random_state,
+                self.n_jobs,
```
MatthewMiddlehurst, line 116: The random state needs to be different. See the Arsenal classifier for an example. If we are multithreading the forest, we probably dont want to thread the trees also.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -0,0 +1,149 @@
...
+            for _ in range(self.n_trees)
+        )
+
+    def _predict_proba(self, X):
+        output_probas = Parallel(
+            n_jobs=self._n_jobs, backend=self.parallel_backend, prefer="threads"
+        )(delayed(_predict_proba_tree)(tree, X) for tree in self.trees_)
+        output_probas = np.sum(output_probas, axis=0)
+        output_probas = np.divide(output_probas, self.n_trees)
+        return output_probas
```
MatthewMiddlehurst, line 127: Rather than averaging probabilities, it may be better to just sum votes from each tree i.e. https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/estimator/interval_based/base_interval_forest.py#L252

Unless this is how the original does it, then we should keep it as is.

aeon/classification/distance_based/_proximity_forest.py
itsdivya1309: Sorry, I'll keep this in mind.

aeon/classification/distance_based/_proximity_forest.py
itsdivya1309: The paper indeed uses majority voting.

itsdivya1309: I think it is also averaging the probabilities: https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/estimator/interval_based/base_interval_forest.py#L267
Correct me if I am wrong, but I can't see the difference.

MatthewMiddlehurst: ah sorry bad link. this should be better, but with 1 rather than a weight https://github.com/aeon-toolkit/aeon/blob/main/aeon/classification/dictionary_based/_tde.py#L422.

Just want to count the predictions and then normalise the array basically. Have to be careful for string labels 🙂

aeon/classification/distance_based/_proximity_forest.py
MatthewMiddlehurst: No problem, was good to catch it.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -85,36 +91,47 @@ def __init__(
...
+        )(
             delayed(_fit_tree)(
                 X,
                 y,
                 self.n_splitters,
                 self.max_depth,
                 self.min_samples_split,
-                self.random_state,
-                self.n_jobs,
+                check_random_state(rng.randint(np.iinfo(np.int32).max)),
```
baraline, line 116: Very unlikely to happen, but you could still get trees with the same random state using this. Initializing the random states values with something like ```rng_list = rng.choice(np.iinfo(np.int32).max, size=self.n_trees, replace=False)``` before the Parallel loop would fix this (again, very unlikely to happen) issue.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -85,36 +91,47 @@ def __init__(
...
         self.n_jobs = n_jobs
+        self.parallel_backend = parallel_backend
         super().__init__()
 
     def _fit(self, X, y):
-        self.classes_ = list(np.unique(y))
-        self.trees_ = Parallel(n_jobs=self._n_jobs, prefer="threads")(
+        rng = check_random_state(self.random_state)
+        self.trees_ = Parallel(
+            n_jobs=self._n_jobs, backend=self.parallel_backend, prefer="threads"
```
baraline, line 108: If `backend` is specified, you can remove the `prefer` parameter here and default `backend` to `threading` during initialization.

aeon/classification/distance_based/_proximity_forest.py
```diff
@@ -85,36 +91,47 @@ def __init__(
...
 
     def _predict_proba(self, X):
-        output_probas = Parallel(n_jobs=self._n_jobs, prefer="threads")(
-            delayed(_predict_proba_tree)(tree, X) for tree in self.trees_
-        )
-        output_probas = np.sum(output_probas, axis=0)
-        output_probas = np.divide(output_probas, self.n_trees)
+        classes = list(self.classes_)
+        preds = Parallel(
+            n_jobs=self._n_jobs, backend=self.parallel_backend, prefer="threads"
```
baraline, line 124: Same comment for this `prefer` parameter.

aeon/classification/distance_based/_proximity_forest.py
itsdivya1309: When I add this code, the algo becomes very slow, probably because of the large range.

baraline: Maybe my bad here, does storing `np.iinfo(np.int32).max` in a variable and using that in the choice function makes it better ? Can't test it right now sorry :/ 

Its weird anyway ... could you provide some timing with and without the change ? (you can use the %timeit magic in a IPython environment to do that)

itsdivya1309: It's taking more than 45 mins, just to run this single line of code `rng_list = rng.choice(np.iinfo(np.int32).max, size=self.n_trees, replace=False)`, even after using a variable.
Maybe we can use `self.n_trees * 100` and do it like this:
`rng_list = rng.choice(self.n_trees*100, size=self.n_trees, replace=False)`

MatthewMiddlehurst: I think the risk here is minimal. We seed ensemble members in a similar way elsewhere. this is essentially a 1/2,147,483,647 situation, no?

aeon/classification/distance_based/_proximity_forest.py
MatthewMiddlehurst: This is now a bit inconsistent with how to do it in other estimators, we mention in the docs that `threading` is the default.

aeon/classification/distance_based/_proximity_forest.py
MatthewMiddlehurst: Same as above.

aeon/classification/distance_based/_proximity_forest.py
baraline: Just seen your comment Matthew, will try to provide some benchmark script.

aeon/classification/distance_based/_proximity_forest.py
itsdivya1309: I've revert the changes.

MatthewMiddlehurst (approved):
LGTM

MatthewMiddlehurst:
@baraline any more improvements you think this needs? No rush to merge if so.

baraline (approved):
Lgtm for now, i'll try to do the joblib testing this evening, we can iterarate on this afterward.

PR #1735, opened 2024-06-27
[ENH] change input type for MrSQMClassifier and deprecate prior to moving package

### LINKED ISSUE #1685: [ENH] MrSQM uses nested data frames

MrSQM still uses
```python
        "X_inner_type": "nested_univ",  # we don't like this, but it's the only input!
```

it would be really good if the original package switched to using numpy3D/np_list, 

any thoughts @heerme, @Inthach @mbalatsko?

we would be happy to help. Failing that, I think we should internalise the conversion with a private local function, it must be the last usage

### LINKED ISSUE #1738: [ENH] Categorisation of MrSQM and MrSEQL algorithms
### Describe the issue

MrSQM and MrSEQL are  currently categorised under TSC shapelet methods. We think the categorisation  should change to dictionary_methods. We don't think of them as shapelet methods, because they are more similar to dictionary-methods of the WEASEL type, than shapelet methods in how they work, e.g. there is no distance involved in the MrSQM/MrSEQL algorithms as in shapelet methods.

Both algos can be easily installed with pip:
pip install mrseql
pip install mrsqm

MrSQM can also be used as a transformer to extract features, see here:
https://github.com/mlgig/mrsqm/blob/main/example/Time_Series_Classification_and_Explanation_with_MrSQM.ipynb

_No response_

### Additional context

_No response_

TonyBagnall:
fixes #1685
fixes #1738 

currently MrSQM used the deprecated "nested_univ" internal type. In order to allow that to be removed, this PR moves the conversion from numpy 3D to pd.DataFrame with each cell a series to the only place its used, this function


baraline (approved):
Should fix linked issues. LGTM !

lnthach:
Hi @TonyBagnall 
Regarding the input type, the latest mrsqm ([0.0.7](https://github.com/mlgig/mrsqm/releases/tag/v.0.0.7)) supports numpy 3D so the conversion shouldn't be needed. Can I create another pull request to remove the conversion ?

MatthewMiddlehurst:
@lnthach Hi Thach, I have done this as part of #1756 actually 🙂. Feel free to make your own PR if you want however as that has other stuff.

I saw that you had uploaded `MrSEQL` to pypi so was interested in adding a wrapper for it, but was having a bit of trouble figuring out what parameters it has and what defaults I should set for them. Would you be able to help me with that?

PR #1737, opened 2024-06-27
[ENH] Tidy up unequal length collection transformers

### LINKED ISSUE #1156: [BUG] TSInterpolator does not interpolate, it resizes
### TSInterpolator

I had assumed that the transformer TSInterpolator interpolated missing values. Instead, it resamples data to make it different length. To me this is resampling not interpolation. It does not have the capability to remove missing values, and even if you give it that capability, it does not remove them. So

```python
from aeon.utils.validation import has_missing
X = np.random.random(size=(10,1,100))
X[0][0][55] = np.nan
print(has_missing(X))
from aeon.transformations.collection import TSInterpolator
interpolator = TSInterpolator(X.shape[2])
X = interpolator.fit_transform(X)
print(has_missing(X))
```
raises 
```python
ValueError: Data seen by instance of TSInterpolator has missing values, but TSInterpolator cannot handle missing values.
```
If I change ``        "capability:missing_values": True,``
it does not remove the missing, even if I resize

```python
from aeon.transformations.collection import TSInterpolator
interpolator = TSInterpolator(X.shape[2]-10)
X = interpolator.fit_transform(X)
print(X.shape," has missing = ",has_missing(X))
```
even the comments refer to resizing, so its just badly named imo. I think we should rename this to Resizer then having a simple transform Imputer that simply fills in missing values through linear interpolation.

TonyBagnall:
fixes #1156

We have three transformers to make unequal length series equal length. One pads, one truncates and one stretches. This PR renames them Truncator, Padder and Resizer, makes the files private and reworks the  associated notebook, which could ultimately be merged with the one on unequal length series.


MatthewMiddlehurst (commented):
Looks good, but needs deprecation. Pretty sure everything can stay as it is now just add the old classes and `__init__` with a warning

TonyBagnall:
> Looks good, but needs deprecation. Pretty sure everything can stay as it is now just add the old classes and `__init__` with a warning

I thought that collections was experimental? Conflicted with deprecation in any case, might just leave this until after the next release

MatthewMiddlehurst:
Series is but collection is not. Have it listed in the README

MatthewMiddlehurst (approved):

Issue #882, opened 2023-10-31
[BUG] LCSS and EDR PAM doesnt assign correct amount of clusters


chrisholder:
### Describe the bug

For PAM-LCSS and PAM-EDR sometimes the correct amount of clusters are not assigned. Specifically LCSS on Adiac and EDR on OliveOil assign incorrect number of clusters.

### Expected results

Should have clusters = n_clusters

### Actual results

cluster != n_clusters

TonyBagnall:
I can reproduce this like so

```python
from aeon.clustering import TimeSeriesKMedoids
from aeon.datasets import load_classification
from aeon.distances import lcss_distance, edr_distance

trainX, trainy = load_classification(name="OliveOil",split="train")
testX, testy = load_classification(name="OliveOil",split="test")

cls = np.unique(trainy)
cls2 = np.unique(testy)

print(" Train classes = ",cls, " length ",len(cls))
print(" Train classes = ",cls2, " length ",len(cls2))

knn = TimeSeriesKMedoids(distance=lcss_distance, n_clusters=len(cls), method="pam")
knn.fit(trainX)
preds = knn.predict(trainX)
preds2 = knn.predict(testX)
print(" Predicted classes = ",np.unique(preds), " Test classes = ",np.unique(preds2))


trainX, trainy = load_classification(name="Adiac",split="train")
testX, testy = load_classification(name="Adiac",split="test")

cls = np.unique(trainy)
cls2 = np.unique(testy)

print(" Train classes = ",cls, " length ",len(cls))
print(" Train classes = ",cls2, " length ",len(cls2))

knn = TimeSeriesKMedoids(distance=lcss_distance, n_clusters=len(cls), method="pam")
knn.fit(trainX)
preds = knn.predict(trainX)
preds2 = knn.predict(testX)
print(" Predicted classes = ",np.unique(preds), " Test classes = ",np.unique(preds2))
```
output for olive oil 
```python-traceback
Train classes =  ['1' '2' '3' '4']  length  4
 Train classes =  ['1' '2' '3' '4']  length  4
 Predicted classes =  [0]  Test classes =  [0]
```

chrisholder:
I think this is now fixed I get the output:
 Predicted classes =  [ 0  1  7  8 10]  Test classes =  [ 0  1  3  9 10 17]

dguijo:
I'm facing some issues with these two distances too. 
In this case, I'm playing around kNN versions of aeon and sklearn.

[...]

PR #1739, opened 2024-06-29
[BUG] Cluster predict proba edge cases

chrisholder:

There were two edge cases with the clustering predict_proba function that would cause crashes.

1) If there wasn't at least 1 prediction in every cluster
2) If the prediction didn't use cluster values of 0-n_clusters. For example I used a k-medoids package that when predict was called returned [22, 22, 32, 3, 32, 22] where instead of the cluster values being a value from 0-n_clusters it was referring to the index of the cluster centre that each value was assigned to. 

This PR makes two main changes to address the above.
1) Instead of using .shape[0] to create the dist array n_clusters is used (when available isn't required):

```
            n_clusters = self.n_clusters
            if n_clusters is None:
                n_clusters = int(max(preds)) + 1
            dists = np.zeros((X.shape[0], n_clusters))
```

2) A additional sanitisation of the predict result is done to convert predictions value to 0-n_clusters. This allows the other logic to work if the predictions doesn't return 0-n_clusters:

```
  unique = np.unique(preds)
  for i, u in enumerate(unique):
      preds[preds == u] = i
```

##### For all contributions

### DIFF
--- a/aeon/clustering/base.py
+++ b/aeon/clustering/base.py
@@ -186,6 +186,9 @@ class BaseClusterer(BaseCollectionEstimator, ABC):
             (i, j)-th entry is predictive probability that i-th instance is of class j
         """
         preds = self._predict(X)
+        unique = np.unique(preds)
+        for i, u in enumerate(unique):
+            preds[preds == u] = i
         n_cases = len(preds)
         n_clusters = self.n_clusters
         if n_clusters is None:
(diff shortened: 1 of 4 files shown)


TonyBagnall:
is this related to https://github.com/aeon-toolkit/aeon/issues/882?

chrisholder:
No this is just a general bug fix

TonyBagnall (approved):

PR #1745, opened 2024-07-02
[DOC] remove ProximityTree from distance classifier notebook results display

TonyBagnall:
addition of the proximity tree classifier broke a notebook only tested on overnights, because we do not have the results for it on TSC.com. Changed the notebook to hard code the ones we do have


MatthewMiddlehurst:
Could you double check outputs, seems like an error was introduced for one.

TonyBagnall:
seems the https certificate is still not working properly.

hadifawaz1999 (approved):
LGTM

PR #1756, opened 2024-07-03
[ENH] MrSEQLClassifier wrapper

### LINKED ISSUE #1738: [ENH] Categorisation of MrSQM and MrSEQL algorithms

MatthewMiddlehurst:
Fixes #1738

Start of a `mrseql` wrapper. Unfortunately, I can't quite find how to parameterise it currently after installing on Windows 🙂.

If anyone can find it in the source that that would be helpful: https://github.com/lnthach/Mr-SEQL

`mrseql`


MatthewMiddlehurst:
@lnthach Hi Thach, I am having a bit of trouble figuring out what parameters your MrSEQL pypi package has and what defaults I should set for them. Would you be able to comment and help me with this?

lnthach:
@MatthewMiddlehurst Hi Matthew, I just updated mrseql (v0.0.4) on pypi so I hope it would make things easier for you. I also remove sktime dependency but use fftw3 library instead (same dependencies as mrsqm).

Currently there are three parameters for MrSEQLClassifier and the default values are as follows:
`
MrSEQLClassifier(seql_mode='fs', symrep=('sax'), custom_config=None)`

Please let me know if it works for you.

MatthewMiddlehurst:
@lnthach Thanks for this. Raised the version found but it seems to be failing to install on our Linux runner (https://github.com/aeon-toolkit/aeon/actions/runs/10247769782/job/28347641403?pr=1756). mrsqm installs fine by the look of things

lnthach:
@MatthewMiddlehurst This one is strange. Both uses the same sfa implementation so both depends on fftw3. One possible reason is that mrsqm installed from built distribution hence the installation somehow bypassed this dependency (not sure if this is possible to be honest). I have some issues with generating the built distribution for mrseql so I uploaded only the source distribution. Do you think this could be the reason ?

MatthewMiddlehurst:
@lnthach I think thats likely, not really sure how to resolve it through as I dont have much experience wrapping C, let alone distributing it 🙂. If the parameters used are the same I can restrict it to 0.0.2 for now?

lnthach:
@MatthewMiddlehurst I just uploaded the built distribution for v0.0.4. Could please you check if it still fails ?

MatthewMiddlehurst:
Seems to all be running fine @lnthach

TonyBagnall (approved):

PR #1758, opened 2024-07-03
[BUG] Store n_timepoints and n_channels in fit to check in transform

### LINKED ISSUE #1712: [BUG] classification base class does not check for different length series in `predict`
### Describe the bug

From #1696

The classification base class does not check for a different series length in `predict` than the one used in `fit`. This should raise an exception unless the classifier can handle unequal length.

Possibly impacts other base classes, but have not checked.

The below example has two estimators raise an exception like excepted, but this in internal to the specific estimator from the following lines:

```python
        n_cases, n_channels, n_timepoints = X.shape

        if n_channels != self.n_channels_:
            raise ValueError(
                "The number of channels in the train data does not match the number "
                "of channels in the test data"
            )
        if n_timepoints != self.n_timepoints_:
            raise ValueError(
                "The series length of the train data does not match the series length "
                "of the test data"
            )
```

```python
from aeon.classification.convolution_based import Arsenal
from aeon.classification.hybrid import HIVECOTEV2
from aeon.classification.interval_based import DrCIFClassifier
import numpy as np

cls = [Arsenal, HIVECOTEV2, KNeighborsTimeSeriesClassifier, DrCIFClassifier]

for c in cls:
    print(c)
    X = np.random.random((10,20))
    X2 = np.random.random((10,20))
    X3 = np.random.random((1,20))
    X4 = np.random.random((1,200))
    X5 = np.random.random((10,200))
    y = np.array([0,1,0,1,0,1,0,1,0,1])

    afc = c()
    afc.fit(X,y)

    try:
        print(afc.predict(X2))
    except Exception as e:
        print("X2 failed")
        print(e)
    try:
        print(afc.predict(X3))
    except Exception as e:
        print("X3 failed")
        print(e)
    try:
        print(afc.predict(X4))
    except Exception as e:
        print("X4 failed")
        print(e)
    try:
        print(afc.predict(X5))
    except Exception as e:
        print("X5 failed")
        print(e)
```

### Expected results

All classifiers except for KNN which can handle unequal length throw an exception in the base class.

### Actual results

Only DrCIF and HC2 (which contains DrCIF) throw exceptions, neither in the base class.

```python-traceback
<class 'aeon.classification.convolution_based._arsenal.Arsenal'>
[1 1 1 0 1 1 1 0 0 1]
[1]
[1]
[1 1 1 1 1 1 1 1 1 1]
<class 'aeon.classification.hybrid._hivecote_v2.HIVECOTEV2'>
[0 0 1 1 0 1 0 0 1 0]
[0]
X4 failed
The series length of the train data does not match the series length of the test data
X5 failed
The series length of the train data does not match the series length of the test data
<class 'aeon.classification.distance_based._time_series_neighbors.KNeighborsTimeSeriesClassifier'>
[1 1 1 1 0 1 1 1 1 0]
[1]
[0]
[1 0 0 0 0 1 1 0 1 0]
<class 'aeon.classification.interval_based._drcif.DrCIFClassifier'>
[1 0 1 1 1 1 0 1 1 1]
[1]
X4 failed
The series length of the train data does not match the series length of the test data
X5 failed
The series length of the train data does not match the series length of the test data
```

### Versions

N/A

TonyBagnall:
fixes #1712 

Stores the length of the first series and number of channe;ls in the metadata in fit, which is then checked against the series length/number of channels  in predict and predict proba. Note

1. Because of the number of supported data types, best way is to call a method get_n_timepoints/n_timepoints in validation
2. For variable length data structures it just looks at the length of the first element, and does not do the test if the classifier has unequal_length capability. If the collection is in fact unequal length, then it fails before here. Only case is where it is in an np-list but they are all in fact equal length. This tests if that common length is different to that seen in train 
3. n_channels only checked for multivariate capable classifiers
4. will move the checking function up to base collection estimator because it is common to all collection estimators

### DIFF
--- a/aeon/base/_base_collection.py
+++ b/aeon/base/_base_collection.py
@@ -10,6 +10,7 @@ from aeon.utils.validation import check_n_jobs
 from aeon.utils.validation._dependencies import _check_estimator_deps
 from aeon.utils.validation.collection import (
     get_n_cases,
+    get_n_timepoints,
     has_missing,
     is_equal_length,
     is_univariate,
@@ -216,5 +217,7 @@ class BaseCollectionEstimator(BaseEstimator):
         metadata["missing_values"] = has_missing(X)
         metadata["unequal_length"] = not is_equal_length(X)
         metadata["n_cases"] = get_n_cases(X)
+        # Length of first series, used if equal length is True
+        metadata["n_timepoints"] = get_n_timepoints(X)
 
         return metadata
--- a/aeon/base/_base_collection.py
+++ b/aeon/base/_base_collection.py
@@ -10,6 +10,7 @@ from aeon.utils.validation import check_n_jobs
 from aeon.utils.validation._dependencies import _check_estimator_deps
 from aeon.utils.validation.collection import (
     get_n_cases,
+    get_n_channels,
     get_n_timepoints,
     has_missing,
     is_equal_length,
@@ -219,5 +220,6 @@ class BaseCollectionEstimator(BaseEstimator):
         metadata["n_cases"] = get_n_cases(X)
         # Length of first series, used if equal length is True
         metadata["n_timepoints"] = get_n_timepoints(X)
+        metadata["n_channels"] = get_n_channels(X)
 
         return metadata
--- a/aeon/base/tests/test_base_collection.py
+++ b/aeon/base/tests/test_base_collection.py
@@ -17,6 +17,8 @@ def test__get_metadata(data):
     assert not meta["multivariate"]
     assert not meta["missing_values"]
     assert not meta["unequal_length"]
+    assert meta["n_channels"] == 1
+    assert meta["n_timepoints"] == 20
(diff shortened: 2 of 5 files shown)


baraline (commented):
Code LGTM, but wouldn't we need to apply this to other base classes like fitted transformations ? And in such cases shouldn't we consider the axis parameter in the `get_n_channels` and `get_n_timepoints` methods ?

TonyBagnall:
> Code LGTM, but wouldn't we need to apply this to other base classes like fitted transformations ? And in such cases shouldn't we consider the axis parameter in the `get_n_channels` and `get_n_timepoints` methods ?

yes it should be in other estimators, but I thought I'd get approval for doing it like this before doing the other collection estimators. I will do them when this is in. Axis is not an issue with collection estimators, we always assume (n_cases, n_channels, n_timepoints),

baraline (approved):
In that case, LGTM !

Issue #1477, opened 2024-04-26
[BUG] AEON does not work with tensorflow 2.13


rruizdeaustri:
### Describe the bug

Dear authors,
I can't run the LITE model in aeon with tensorflow 2.13

Thanks !

### Steps/Code to reproduce the bug

```python
from aeon.datasets import load_classification
from aeon.classification.deep_learning import LITETimeClassifier

xtrain, ytrain = load_classification(name="Coffee", split="train")
xtest, ytest = load_classification(name="Coffee", split="test")


clf = LITETimeClassifier(n_classifiers=5)

clf.fit(xtrain, ytrain)

print(clf.score(xtest, ytest))
```

### Expected results

I don't know since it does not work. 

### Actual results
```python-traceback
File "main_aeon.py", line 10, in <module>
    clf.fit(xtrain, ytrain)
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/aeon/classification/base.py", line 129, in fit
    self._fit(X, y)
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/aeon/classification/deep_learning/_lite_time.py", line 200, in _fit
    cls.fit(X, y)
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/aeon/classification/base.py", line 129, in fit
    self._fit(X, y)
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/aeon/classification/deep_learning/_lite_time.py", line 520, in _fit
    self.history = self.training_model_.fit(
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/keras/src/utils/traceback_utils.py", line 70, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/lhome/ext/ifca124/ifca1241/software/anaconda3/envs/tensorflow/lib/python3.8/site-packages/keras/src/saving/saving_api.py", line 142, in save_model
    raise ValueError(
ValueError: The following argument(s) are not supported with the native Keras format: ['options']
```

baraline:
Hi, thanks for raising the issue. Could you please provide the versions of your environment by pasting the output of the followin function here or in your answer ?

```python
from aeon import show_versions
[...]

PR #1765, opened 2024-07-04
[MNT] Raise tensorflow bound to >2.13

TonyBagnall:
aeon does not work properly with tensorflow 2.13, see https://github.com/aeon-toolkit/aeon/issues/1477. The current version is 2.16. Conversations with @hadifawaz1999 suggest it is not worth the effort to make things work as the signatures keeps changing. In my experience tensorflow is notoriously bad at backward compatibility, so it makes sense just to enforce a bound I think. Interaction with python 3.12 is still an issue!

Closes https://github.com/aeon-toolkit/aeon/issues/1477

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ all_extras = [
     "stumpy>=1.5.1",
     "ruptures>=1.1.9",
     "tbats>=1.1.0",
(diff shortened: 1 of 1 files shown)


hadifawaz1999 (approved):
thanks for this

PR #1771, opened 2024-07-08
[BUG] Hydra bugfix

MatthewMiddlehurst:

Hydra does not have a `_fit` currently unlike other convolution transforms. It should have, as it does not really make sense to generate new kernels each transform.

#### Any other comments?

Found by @angus924


TonyBagnall (approved):
should it not be hydra_ instead of hydra_, given it is set in fit?

MatthewMiddlehurst:
Can possibly change in the future, but I don't think we want people accessing this. I usually go for `_x` for that kind of stuff, `x_` if people may want to look at it.

PR #1776, opened 2024-07-09
[ENH] Adds class_weight to those classifier that support it. Required for imbalanced datasets.

patrickzib:
- This PR adds `class_weight` to those classifiers that support it by sklearn design. `class_weight` is intended for training with imbalanced datasets.

From ExtraTreesClassifier:

```python
"""
class_weight{“balanced”, “balanced_subsample”}, dict or list of dicts, default=None
Weights associated with classes in the form {class_label: weight}. If not given, all classes are supposed to have weight one. For multi-output problems, a list of dicts can be provided in the same order as the columns of y.
Note that for multioutput (including multilabel) weights should be defined for each class of every column in its own dict. For example, for four-class multilabel classification weights should be [{0: 1, 1: 1}, {0: 1, 1: 5}, {0: 1, 1: 1}, {0: 1, 1: 1}] instead of [{1:1}, {2:5}, {3:1}, {4:1}].
The “balanced” mode uses the values of y to automatically adjust weights inversely proportional to class frequencies in the input data as n_samples / (n_classes * np.bincount(y))
The “balanced_subsample” mode is the same as “balanced” except that weights are computed based on the bootstrap sample for every tree grown.
For multi-output, the weights of each column of y will be multiplied.
Note that these weights will be multiplied with sample_weight (passed through the fit method) if sample_weight is specified.
"""
```

- Fixes a bug in Quant, as the `random_state parameter` was not passed to the internal estimator.


patrickzib:
There was also a bug in Hydra, as the random_state parameter was not passed to the classifier. Now fixed

hadifawaz1999:
i think you mean you fixed the random_state thing in quant, not hydra @patrickzib no ?

patrickzib:
> i think you mean you fixed the random_state thing in quant, not hydra @patrickzib no ?

Uh, yes :)

hadifawaz1999:
will raise an issue to keep in mind doing this for the rest of classifiers as lots others use sklearn based estimators

hadifawaz1999 (approved):
LGTM! check #1777 for the future

TonyBagnall (approved):
lgtm, be interested to see if it makes any difference. Just saw the conflict

patrickzib:
Thank you @TonyBagnall @hadifawaz1999 . Would you mind approving again? I had to merge main.

TonyBagnall (approved):

hadifawaz1999 (approved):

PR #1784, opened 2024-07-10
[ENH] Add AEDRNNClusterer

aadya940:
Implements Clusterer based on Dilated Recurrent Neural Network (DRNN).


hadifawaz1999:
@aadya940 i think an issue is happening when importing the custom layer from the networks package, one possibility is because its wrapped under a condition, to debug this (i dont think its the best solution) can you implement as well the custom layer inside ae_drnn.py but the clusterer ? it is just to see if CI is failing because of this

hadifawaz1999 (changes requested):
.

aadya940:
@hadifawaz1999 Now they are the same :))

aadya940:
#1962

hadifawaz1999 (changes requested):
- the default for clustering_algorithm should be "deprecated" in the docstring
- the example should not use n_clusters
- the _score function should use self._estimator not self.clusterer
- the test parameters function should set the estimator to DummyClusterer, to make testing fast, and not set the old parameters to be deprecated

aeon/clustering/deep_learning/_ae_drnn.py
```diff
@@ -0,0 +1,348 @@
...
+
+    Examples
+    --------
+    >>> from aeon.clustering.deep_learning import AEDRNNClusterer
+    >>> from aeon.datasets import load_unit_test
+    >>> X_train, y_train = load_unit_test(split="train", return_X_y=True)
+    >>> X_test, y_test = load_unit_test(split="test", return_X_y=True)
+    >>> from aeon.clustering import DummyClusterer
+    >>> _clst = DummyClusterer(n_clusters=2)
+    >>> aefcn = AEDRNNClusterer(n_clusters=2,
```
hadifawaz1999, line 106: n_clusters should not be set in AEDRNNCluster as it is set in the estimator used

aeon/clustering/deep_learning/_ae_drnn.py
```diff
@@ -0,0 +1,348 @@
...
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`.
+        """
+        param1 = {
+            "n_clusters": 2,
+            "n_epochs": 1,
+            "batch_size": 4,
+            "n_layers_encoder": 1,
+            "n_layers_decoder": 1,
+            "clustering_params": {
```
hadifawaz1999, line 340: test params should be changes, n_clusters and clustering_algorithm should be removed here and replaced by estimator set to dummy clusterer

hadifawaz1999 (dismissed):

hadifawaz1999 (approved):

Issue #1611, opened 2024-06-05
[ENH] Checklist for deprecation in V0.11.0


TonyBagnall:
- [x] Remove SplitTimeSeries #1613 
- [x] Remove PandasAdaptor #1614
- [x] Move compose to _legacy #1679 
- [x] Remove `BaseDeepNetwork`
- [x] Remove MiniRocket multivariate #1708 
- [x] Remove MultiRocketMultivariate
- [x] Move MrSQM to dictionary_based
- [x] Delete or movbe to _legacy all BaseTransformers 
- [x] Remove BKFilter #1788

PR #1788, opened 2024-07-13
[DEP] Remove BKFilter BaseTransformer, redesign equivalent series transformer

TonyBagnall:
removes deprecated BaseTransformer BKFilter and adapts the BaseSeriesTransformer so it does not require statsmodels to do pointless conversions to and from data frames

part of #1611


MatthewMiddlehurst (approved):
Cant see anything off, but what strange code. Are the imports really just for datatype stuff rather than any actual processing?

TonyBagnall:
> Cant see anything off, but what strange code. Are the imports really just for datatype stuff rather than any actual processing?

yes its baffling, statsmodels has some strange things. Output tested identical

PR #1793, opened 2024-07-13
[DEP] remove several BaseTransformers 

TonyBagnall:
all part of #1611 either remove or move to legacy if used in forecasting


TonyBagnall:
going to close this and do them individually, too much hassle

PR #1800, opened 2024-07-14
[EN] Remove internal setting of n_jobs in minirocket

TonyBagnall:
this removes the setting of the number of jobs in the private transform function. I cant remember why we put this here, I think it was something to do with the use of prange in the numba function. However, I think remove it because

1, There is a measurable effect on performance, 
image
2. I dont think we should be setting the number of threads in this function, if required it should be done in base class. 

so even if there was a reason to do this, I think we should look to fix it in a different way./


TonyBagnall:
actually this is my misunderstanding, minirocket uses prange to parallelise and this code is to stop it using all cores by default. I think we need to sort out n_jobs though!

Issue #948, opened 2023-11-30
[ENH] Adaptive linear segmentation algorithm


TonyBagnall:
### Describe the feature or idea you want to propose

It would be great to have an implementation of this piecewise linear approximation method for segmenting time series

https://ieeexplore.ieee.org/abstract/document/989531

### Describe your proposed solution

Implement as a base segmenter, internally store the fitted slopes

Moonzyyy:
Will be taking doing this now!

patrickzib:
This approach, in my assessment, does not fall under the category of segmentation algorithms such as "change point detection" or "semantic segmentation." Rather than being classified among segmentation techniques, it aligns more closely with data representation methods like Piecewise Aggregate Approximation (PAA), Symbolic Aggregate approXimation (SAX), or Symbolic Fourier Approximation (SFA). These methods focus on summarizing or approximating time series data rather than identifying distinct segments or change points within the data.

PR #1807, opened 2024-07-15
[ENH] Piecewise Linear Approximation (PLA)

### LINKED ISSUE #1384: [ENH] Update Catch22 transformer

Since its implementation, the numba version of `Catch22` ([code here](https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/feature_based/_catch22.py)) has become slightly out of date compared to the [original implementation](https://github.com/DynamicsAndNeuralSystems/pycatch22).

The features which now differ should be updated, and parity with the original output should be verified

Look at the [authors implementation](https://github.com/DynamicsAndNeuralSystems/pycatch22) (C code), and update our version of the features in numba to match.

The transformer has a parameter to use this package, so the results can be compared to verify they are the same (or similar).

_No response_

### Additional context

_No response_

Moonzyyy:
Implements #948. Added tests for new series transformer PLA.

No new dependency


aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,413 @@
...
+    Attributes
+    ----------
+    segment_dense : np.array
+        The endpoints of each found segment of the series for transformation
+
+    References
+    ----------
+    .. [1] Keogh, E., Chu, S., Hart, D. and Pazzani, M., 2001, November.
+    An online algorithm for segmenting time series. (pp. 289-296).
+    """
```
TonyBagnall, line 38: could you add an example?

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,413 @@
...
+
+    class Transformer(Enum):
```
TonyBagnall, line 40: I think I would rather it were called something else, overloading the term Transformer can be confusing. Alternatively could make it private?

TonyBagnall (dismissed):
couple very minor things, but looks good to me. Could you change the title of the PR, this is just PLA not catch22

TonyBagnall:
might be worth looking at numba at some point?

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,430 @@
...
+        """An enum class specifically for PLA."""
+
+        SlidingWindow = "SlidingWindow"
+        TopDown = "TopDown"
+        BottomUp = "BottomUp"
+        SWAB = "Swab"
+
+    _tags = {
+        "fit_is_empty": True,
+        "python_dependencies": "sklearn",
```
MatthewMiddlehurst, line 67: this is not required, sklearn is always installed.

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    def __init__(self, transformer, max_error, buffer_size=None):
+        if not isinstance(transformer, self.Transformer):
+            raise ValueError("Invalid transformer: please use Transformer class.")
+        if not isinstance(max_error, (int, float, complex)):
+            raise ValueError("Invalid max_error: it has to be a number.")
+        if not (buffer_size is None or isinstance(buffer_size, (int, float, complex))):
+            raise ValueError("Invalid buffer_size: use a number only or keep empty.")
```
MatthewMiddlehurst, line 76: better to do error checking like this in fit

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,430 @@
...
+        self.transformer = transformer
+        self.max_error = max_error
+        self.buffer_size = buffer_size
+        self.segment_dense = np.array([])
```
MatthewMiddlehurst, line 80: same with this, better to initialise in fit so it resets on multiple runs. Though in this case there is no fit, so we dont want to store it at all really.

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,430 @@
+__maintainer__ = []
```
MatthewMiddlehurst, line 1: There should be  a docstring at the top of the file

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,430 @@
...
+    >>> print(ts)
+    [ 573.  375.  301.  212.   55.   34.   25.   33.  113.  143.  303.  615.
+    1226. 1281. 1221. 1081.  866. 1096. 1039.  975.  746.  581.  409.  182.]
+    >>> print(results)
+    [ 538.8  423.1  307.4  191.7   48.    40.5   33.    25.5   43.6  210.2
+    376.8  543.4 1276.5 1227.  1177.5 1128.   953.5  980.5 1007.5 1034.5
+    759.1  572.7  386.3  199.9]
+    """
+
+    class Algorithm(Enum):
```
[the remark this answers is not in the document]
Moonzyyy: Was thinking of doing that, will make SWAB the default one and then allow users to just swap algorithm

aeon/transformations/series/_pla.py
Moonzyyy: So are the changes now fine? I can't change the parameters of _fit as its an inherited method

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,436 @@
...
+            results = self._SWAB(X)
+        else:
+            raise RuntimeError("No transformer was called.")
+
+        if len(results) > 1:
+            segment_dense = np.zeros([len(results) - 1])
+            segment_dense[0] = len(results[0])
+            for i in range(1, len(results) - 1):
+                segment_dense[i] = segment_dense[i - 1] + len(results[i])
+            self.segment_dense = segment_dense
```
MatthewMiddlehurst, line 145: This is eventually going to break testing as transform what finished for series transformers as it shouldn't change any attributes (`self.segment_dense = segment_dense`).

aeon/transformations/series/_pla.py
MatthewMiddlehurst: noticed other issues, needs to be structured differently but resolve this conversation.

aeon/transformations/series/_pla.py
MatthewMiddlehurst: Resoling this in favour of new comments, but this is only the case if fit should exist.

aeon/transformations/series/tests/test_pla.py
MatthewMiddlehurst, line 1: There is a bit of a testing gap here, as the string option for `transformer` is never used. I'm pretty sure in the current state this would result in a crash.

MatthewMiddlehurst (changes requested):
A few comments. I did not notice initially that `_fit` should not exist due to the tag setting, unless you think it is the tag which is wrong and the method should exist.

I am going to turn on the code coverage tests to its easier to see what lines are covered by unit tests.

aeon/transformations/series/_pla.py
Moonzyyy: Well honestly I don't think it has much use since we aren't segmenting the data

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,436 @@
...
+    >>> from aeon.transformations.series import PiecewiseLinearApproximation
+    >>> from aeon.datasets import load_electric_devices_segmentation
+    >>> ts, period_size, true_cps = load_electric_devices_segmentation()
+    >>> ts = ts.values
+    >>> pla = PiecewiseLinearApproximation(0.001, transformer=3)
+    >>> transformed_x = pla.fit_transform(ts)
+    """
+
+    _tags = {
+        "fit_is_empty": True,
```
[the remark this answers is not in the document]
Moonzyyy: Right, will remove it! I thought it was meant to see if the model has been fitted or not, my bad.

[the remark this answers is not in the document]
MatthewMiddlehurst: If there is no fitting to be done it can remain and `_fit` can go. I dont really see much currently which would require fit

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
+"""Piecewise Linear Approximation.
+
+A transformer which uses Piecewise Linear Approximation algorithms.
+"""
+
+__maintainer__ = []
```
TonyBagnall, line 6: Could you add TonyBagnall as the maintainer?

Moonzyyy: Should be done :)

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+    Parameters
+    ----------
+    transformer: int or str
+        The transformer to be used.
+        Default transformer is swab.
+        Valid transformers with their string and int:
+            Sliding Window: "sliding window", 1
+            Top Down: "top down" , 2
+            Bottom Up: "bottom up", 3
+            SWAB: "swab", 4
```
MatthewMiddlehurst, line 32: no ints, just lower case strings

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+    max_error: float
+        The maximum error value for the algorithm to find before segmenting the dataset.
+    buffer_size: float
+        The buffer size, used only for SWAB.
+
+    Attributes
+    ----------
+    segment_dense : np.array
+        The endpoints of each found segment of the series for transformation.
```
MatthewMiddlehurst, line 41: no longer exists

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+    .. [1] Keogh, E., Chu, S., Hart, D. and Pazzani, M., 2001, November.
+    An online algorithm for segmenting time series. (pp. 289-296).
+
+    Examples
+    --------
+    >>> from aeon.transformations.series import PiecewiseLinearApproximation
+    >>> from aeon.datasets import load_electric_devices_segmentation
+    >>> ts, period_size, true_cps = load_electric_devices_segmentation()
+    >>> ts = ts.values
+    >>> pla = PiecewiseLinearApproximation(0.001, transformer=3)
```
MatthewMiddlehurst, line 54: will need to be updated

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+            elif transformer.lower() == "bottom up":
+                transformer = 3
+            elif transformer.lower() == "swab":
+                transformer = 4
+            else:
+                raise ValueError(
+                    "Invalid transformer: wrong transformer: ", transformer
+                )
+        elif not (1 <= transformer <= 4):
+            raise ValueError("Invalid transformer: choose between 1-4")
```
MatthewMiddlehurst, line 83: looking now the first 3 of these would be better at the start of transform, and the third does not need to exist (but the error message can replace the current one in transform).

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+        ----------
+        X : np.array
+            1D time series.
+
+        Returns
+        -------
+        np.array
+            the linear regression of the time series.
+        """
+        return self._linear_regression(X)
```
MatthewMiddlehurst, line 407: Some of these functions seem a bit unnecessary, especially  `_create_segment` which just calls another function.

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,407 @@
...
+from aeon.transformations.series.base import BaseSeriesTransformer
+
+
+class PiecewiseLinearApproximation(BaseSeriesTransformer):
+    """Piecewise Linear Approximation (PLA) for time series transformation.
+
+    Takes a univariate time series as input. Approximates a time series using
+    linear regression and the sum of squares error (SSE) through an algorithm.
+    The algorithms available are two offline algorithms: TopDown and BottomUp
+    and two online algorithms: SlidingWindow and SWAB (Sliding Window and Bottom Up).
```
MatthewMiddlehurst, line 21: Could you expand on this a bit, what is the output of the transformation (and how would the different methods effect this)

aeon/transformations/series/tests/test_pla.py
```diff
@@ -0,0 +1,222 @@
...
+            1221.0,
+            1081.0,
+            866.0,
+            1097.16666667,
+            1036.66666667,
+            976.16666667,
+            747.16666667,
+            578.66666667,
+            410.16666667,
+            182.0,
```
MatthewMiddlehurst, line 73: this seems very similar to the input series

aeon/transformations/series/tests/test_pla.py
```diff
@@ -0,0 +1,222 @@
...
+            1221.0,
+            1081.0,
+            866.0,
+            1097.16666667,
+            1036.66666667,
+            976.16666667,
+            746.0,
+            581.0,
+            409.0,
+            182.0,
```
MatthewMiddlehurst, line 111: this also 🙂

aeon/transformations/series/tests/test_pla.py
```diff
@@ -0,0 +1,222 @@
...
+            1177.5,
+            1128.0,
+            953.5,
+            980.5,
+            1007.5,
+            1034.5,
+            759.1,
+            572.7,
+            386.3,
+            199.9,
```
MatthewMiddlehurst, line 187: This seems to be the same as bottom up

MatthewMiddlehurst (commented):
Still a bit to go I think. Not fully checked the transformer functions yet.

In the testing please use keywords for parameters

aeon/transformations/series/tests/test_pla.py
Moonzyyy: It is bound to be similar to bottom up as it uses bottom up for transforming a time series

aeon/transformations/series/tests/test_pla.py
Moonzyyy: image

From the research paper the output will be a more sharpened version of the input series, however I can decrease the error margin if that is what is wanted in the tests

aeon/transformations/series/tests/test_pla.py
MatthewMiddlehurst: unless I'm misreading it is exactly the same however, would be better to find an example which changes and verify the output is what you expect

aeon/transformations/series/tests/test_pla.py
MatthewMiddlehurst: There does not seem to be any large differences like the example shown here though. Only the last few values change and it is small.

aeon/transformations/series/_pla.py
Moonzyyy: I removed functions that only appeared once or called other functions. The only functions that remain are sliding window, top up, bottom up, Swab, SEE and linear regression.

aeon/transformations/series/_pla.py
Moonzyyy: I completely removed numbering, only leaving it to strings. Therefore there are less given error messages and the code should look more clean

aeon/transformations/series/tests/test_pla.py
Moonzyyy: Should be changed now, all tests should be different

MatthewMiddlehurst:
Could you plot the test series and each of the options similar to the image above? We have utilities in the visualisation module which can do this.

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,367 @@
...
+
+from aeon.transformations.series.base import BaseSeriesTransformer
+
+
+class PLASeriesTransformer(BaseSeriesTransformer):
+    """Piecewise Linear Approximation (PLA) for time series transformation.
+
+    Takes a univariate time series as input. Approximates a time series using
+    linear regression and the sum of squares error (SSE) through an algorithm.
+    The algorithms available are two offline algorithms: TopDown and BottomUp
```
MatthewMiddlehurst, line 20: no space between the first two options here

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,367 @@
...
+        Default transformer is swab.
+        Valid transformers with their string:
+            Sliding Window: "sliding window"
+            Top Down: "top down"
+            Bottom Up: "bottom up"
+            SWAB: "swab"
+    max_error: float
+        The maximum error value for the algorithm to find before segmenting the dataset.
+    buffer_size: float
+        The buffer size, used only for SWAB.
```
MatthewMiddlehurst, line 48: The default value should be included on the first line. See other classes.

aeon/transformations/series/_pla.py
```diff
@@ -0,0 +1,367 @@
...
+    with infinite data, the algorithm will infinitely read the data, never
+    processing it. On the other hand, online algorithms read a subsequence of data
+    and process the subsequence. Once processing is done, more data is read in.
+    This means that online algorithms can process even infinity data.
+
+    From the four algorithms, Sliding Window is known to give the worst performance in
+    transformation. While Bottom Up is known to give the best performance, yet has its
+    inefficiencies due to being an offline aglorithm. SWAB uses Bottom Up's performance
+    with the combination of Sliding Window to create an efficient online algorithm. It's
+    performance is only slightly reduced versus it's competitor Bottom Up.
```
MatthewMiddlehurst, line 33: I don't feel these new paragraphs explain the output for each algorithm. What do they do differently and when would you want to use one over the other? Is there a source for the second paragraph? 

If you can't do this we can just remove these bits and move on for now.

Moonzyyy: I re-wrote the paragraphs and added the reference paper

PR #1833, opened 2024-07-21
[DEP] Move summarizer to legacy

TonyBagnall:
this is the last one, also removed the transformers notebook and move a rocket test to the correct place


baraline (commented):
Is it normal that you deleted the whole transformation notebook?

baraline (approved):
NVM, it is mostly about deprecated stuff, I'll open a doc issue for us to remake it afterward.

Issue #1233, opened 2024-02-23
[ENH] Deep learning networks coverage


TonyBagnall:
### Describe the feature or idea you want to propose

following dev discussion about codecov, looking at lack of testing coverage for networks. I think its because it lacks tests of certain 
parameters.  So, for example, cnn lacks coverage because its not tested with list parameters 
![image](https://github.com/aeon-toolkit/aeon/assets/9594042/f199f966-b851-4838-bdeb-8b35f008a9d1)

its not critical, cant really but it would be good to cover these

### Describe your proposed solution

I'm not sure how much of this is common between networks, maybe we could have a function that tests list input and non list input?

hadifawaz1999:
good spotting @TonyBagnall , maybe its waiting to cover all possibilities, will keep that in mind

Cyril-Meyer:
Aeon-Assign bot assign @Cyril-Meyer : test coverage for parameters which can be list or not.

TonyBagnall:
seems @Cyril-Meyer fixed many of the old networks (thanks) but now the ae_ networks are poorly covered

hadifawaz1999:
> @Cyril-Meyer that would be fantastic, thanks. But I will just check there are no open PRs actually, might be best to wait for all the clustering PRs maybe @hadifawaz1999 can advise?

would be best to wait till i finish with clustering PRs, would be better not to touch networks test this way the PRs don't get blocked from a fail, we can revisit this later

Cyril-Meyer:
> > @Cyril-Meyer that would be fantastic, thanks. But I will just check there are no open PRs actually, might be best to wait for all the clustering PRs maybe @hadifawaz1999 can advise?
> 
> would be best to wait till i finish with clustering PRs, would be better not to touch networks test this way the PRs don't get blocked from a fail, we can revisit this later

@hadifawaz1999 good to me, ping me here when it's done :)

PR #1851, opened 2024-07-27
[ENH] Improve deep learning networks test coverage for parameters which can be list

Cyril-Meyer:
#### Reference Issues/PRs

Improves part of #1233

Improve the `test_all_networks_functionality` tests, adding cases when some parameters can be a list.

| Before | After |
|:-:|:-:|
|before|after2|

No.

#### Any other comments?

Some tests won't work and are disabled as 'continue exceptions'. I don't know if this is the right way to do it.
In any case, these problems will need to be fixed, and are related to the implementations themselves.
Anyway, we are adding new tests, and these cases are just like before, without the improvement.

```python
# Exceptions to fix
if (
    attrname in ["kernel_size", "padding"]
    and network.__name__ == "TapNetNetwork"
):
    continue
if attrname in ["activation"] and network.__name__ == "AEBiGRUNetwork":
    continue
# LITENetwork does not seem to work with list args
if network.__name__ == "LITENetwork":
    continue
```

##### For all contributions


aeon/networks/tests/test_all_networks.py
```diff
@@ -41,20 +41,65 @@ def test_all_networks_functionality(network):
...
+                "dilation_rate",
+                "use_bias",
+            ]:
+                # Exceptions to fix
+                if (
+                    attrname in ["kernel_size", "padding"]
+                    and network.__name__ == "TapNetNetwork"
+                ):
+                    continue
+                if attrname in ["activation"] and network.__name__ == "AEBiGRUNetwork":
```
hadifawaz1999, line 61: can skip the "AE" networks as they will be having their own tests for special reasons (them being auto encoders)

hadifawaz1999:
Thanks for taking care of this, coverage seems great now !

We should open an issue after this getting merged for cases like LITE (still not parametrized as it should with list inputs)

aeon/networks/tests/test_all_networks.py
Cyril-Meyer: Skip added in b99b90df39eff26877ee7cb808ecca5ceda8b073 with deletion of all other "AE" models checks.

MatthewMiddlehurst: This should use the tag instead of the network name IMO

MatthewMiddlehurst (commented):
This should be a separate test IMO i.e. `test_all_networks_params`.

Will these parameters be in every network (even future ones)? May be missing something, but would this fail if a network does not have one from the list?

aeon/networks/tests/test_all_networks.py
hadifawaz1999: i agree @MatthewMiddlehurst but issue is that we still havent added the config to all networks, only the ones added with clustering project for AE based

hadifawaz1999:
> This should be a separate test IMO i.e. `test_all_networks_functionality`.
> 
> Will these parameters be in every network (even future ones)? May be missing something, but would this fail if a network does not have one from the list?

hmm good point @MatthewMiddlehurst , maybe we can have a test per network ? test_inception_parameters, test_lite_parameters etc.

aeon/networks/tests/test_all_networks.py
MatthewMiddlehurst: seems like an easy change given they all start with AE anyway?

Cyril-Meyer: @MatthewMiddlehurst we can use tags, but if we want to, we have to add a check that the networks has tags before.
```python
if network.__name__.startswith("AE")
```
can be replaced by something like
```python
if hasattr(network, "_config") and network._config.structure == "auto-encoder"
```

Cyril-Meyer:
> This should be a separate test IMO i.e. `test_all_networks_params`.

Not sure about this as it's exactly the same test as before but with different types for some parameters.
But yeah, it's easy to change as you asked :)

> hmm good point @MatthewMiddlehurst , maybe we can have a test per network ? test_inception_parameters, test_lite_parameters etc.

This is also a valid choice, but which requires a new test to be created for each new network.
For now, I was mimicking the actual tests that do not require that.

> Will these parameters be in every network (even future ones)? May be missing something, but would this fail if a network does not have one from the list?

The networks do not share the parameters, that's why we check that the network have the parameters or not. So no, this would not fail if a network does not have the parameter.
The only reason to have a fail is if a new network use the same parameters name but do not allow the usage of both types (e.g. like LITE for now).

MatthewMiddlehurst:
I do not mind it being a singular test for now instead of going through each file and adding a bunch of tests. Ideally the parameters would be used in more than one network though.

> Not sure about this as it's exactly the same test as before but with different types for some parameters.
> But yeah, it's easy to change as you asked :)

I think it's acceptable to lose a bit of testing efficiency here for readability. You don't have to assert the output etc like the current test. I would like to avoid forming the "omni-test" for all networks which just covers everything in a single function, it was kinda like that before 🙂.

hadifawaz1999:
> I added the new test as a new function, `test_all_networks_params` as proposed. I also changed the AE skip using tag instead of the network name and added a skip info.
> 
> Also, is `tensorflow-addons` requirement still needed ? [ref](https://github.com/aeon-toolkit/aeon/blob/c980d7f7cb631672f16396f37691825e0743c4ad/aeon/networks/_encoder.py#L49)

thanks for the changes @Cyril-Meyer ! the check for config existence is good solution for now, we should keep in mind removing it once we add the config to all other networks

great catch on tensorflow-addons, we dont need it no, we removed its usage before and project  no longer depends on it, probably forgot to remove it here

Cyril-Meyer:
> thanks for the changes @Cyril-Meyer ! the check for config existence is good solution for now, we should keep in mind removing it once we add the config to all other networks

The check was not necessary, as the base class has it :
```python
    _config = {
        "python_dependencies": ["tensorflow"],
        "python_version": "<3.12",
        "structure": "encoder",
    }
```
I just didn't see this before, but it's cool.


> great catch on tensorflow-addons, we dont need it no, we removed its usage before and project no longer depends on it, probably forgot to remove it here

Yeah, I think it will be more clear if we create a new issue and solve it with a new PR (I can handle that).

hadifawaz1999:
Looks good to me @MatthewMiddlehurst how about you ? (btw for some reason readthedocs says "pending" for 12 hours now)

MatthewMiddlehurst (approved):
Looks fine for now. Feel free to merge when you think its good.

MatthewMiddlehurst:
Can ignore the docs, says it has built fine when you click on details.

hadifawaz1999 (approved):
LGTM! thanks for taking care of this

PR #1862, opened 2024-07-29
[DOC] Fix DL classification notebook

MatthewMiddlehurst:

Fixes the currently breaking example notebook in the periodic tests

https://github.com/aeon-toolkit/aeon/actions/runs/10136750201/job/28025991094


hadifawaz1999:
ah this might be my fault with TimeCNN introduction, but weird it didnt break my PR when doing this, we sure this is the casue @MatthewMiddlehurst  ?

MatthewMiddlehurst:
It was not run by default, given the tests pass now I would say yeah this is correct.

hadifawaz1999 (approved):
LGTM!

PR #1864, opened 2024-07-29
[ENH] Documentation fixes and function changes for `BaseCollectionEstimator`

MatthewMiddlehurst:

- changes conversion function to return immediately if the input datatype is already in `X_inner_type`
- updates unequal length metadata to store `None` for unequal length data
- updates documentation for metadata and to match changes


aeon/base/_base_collection.py
```diff
@@ -155,28 +160,28 @@ def _check_X(self, X):
...
-        know if the data is unequal length in order to choose between different
-        allowed input types. If multiple types are allowed by self, then the best
-        one for the data is selected. So, for example, if X_inner_tag is `["np-list",
-        "numpy3D"]` and an `np-list` is passed containing equal length series,
-        X will be converted to numpy3D.
+        If the input data is already an allowed type, it is returned unchanged.
+
+        If multiple types are allowed by self, then the best one for the type of input
+        data is selected. So, for example, if X_inner_tag is `["np-list", "numpy3D"]`
+        and an `df-list` is passed containing equal length series, will be converted
```
TonyBagnall, line 173: would rather not reference df-list, its not supported really

TonyBagnall (approved):
LGTM

aeon/base/_base_collection.py
MatthewMiddlehurst: Just picked any because the two we like are already used really 🙂.  All internal as its a private method.

PR #1865, opened 2024-07-29
[BUG] Check data shape in `predict`-like functions

### LINKED ISSUE #1774: [ENH] Check number of channels and timepoints for equal length matches fit/predict for other BaseCollectionEstimator subclasses

collection estimators without the capability to handle unequal length should check the series lengths in fit and predict match

- [ ] Regression
- [ ] Clustering
- [ ] BaseCollectionTransformer
 

do as with classifier  here https://github.com/aeon-toolkit/aeon/pull/1758

_No response_

### Additional context

_No response_

### LINKED ISSUE #1775: [ENH] Remove redundant checks of n_timepoints in classifiers

https://github.com/aeon-toolkit/aeon/issues/1712 highlights that some classifiers check for a match in series length between fit and predict. #1758 moved it to the base class making these checks redundant. Im not sure exactly which ones do this, thoughts @MatthewMiddlehurst ?

remove checks from classifiers that are now done in the base class

_No response_

### Additional context

_No response_

### LINKED ISSUE #1936: [BUG] Proximity Forest Accuracy tests fail CI
### Describe the bug

The proximity forest test `accuracy >= 0.9` fails randomly sometimes.

A possible fix would be to set a tolerance as follow:
`accuracy + 0.1 >= 0.9` ?

### Expected results

Shouldn't fail.

### Actual results

Tests don't pass.

### LINKED ISSUE #1966: [ENH] test_proximity_tree.py is causing issues for pull requests
### Describe the issue

There seems to be an issue with test_proximity_tree.py causing runs to end in failure. "def test_proximity_tree():" tries to assert an accuracy of above 0.9, even though its lowest accuracy score ( I have seen ) can be as low as 0.85.

Not sure if that is the intended feature, but its causing some people to think there is an error with their implementation/spend more time on the issue as the tests have to be re-run.

The issue can be found here: https://github.com/aeon-toolkit/aeon/actions/runs/10368551328/job/28702397463?pr=1959

Not sure, as I don't really understand why we are asserting accuracy above 0.9 if it can be as low as 0.85, or even lower which we haven't seen yet.

Currently there are two ways to fix this, creating a random push or pull from main

### Additional context

_No response_

MatthewMiddlehurst:
#### Reference Issues/PRs

Fixes #1774
Fixes #1775
Fixes #1936
Fixes #1966

Adds a check to collection base classes which raises an error if length and channels does not match fit. Removed redundant check from interval base class.


MatthewMiddlehurst:
@TonyBagnall @hadifawaz1999 do these changes seem reasonable? DLs are breaking, but i think that is more of a DL issues than the changes here.

hadifawaz1999:
@MatthewMiddlehurst not only DL seems dictionary based are failing too [https://github.com/aeon-toolkit/aeon/actions/runs/10151952868/job/28072315499?pr=1865#step:7:260](here)
is there something that these changes assume sub classes are doing ? because it seems it assumes n_timepoints is an attribute somewhere defined, which most sub classes don't do

MatthewMiddlehurst:
It likely has to do with no longer allowing storage of metadata via `predict`. Just want to make sure the current changes are fine before we fix any of the broken stuff.

aeon/base/_base_collection.py
```diff
@@ -212,6 +212,33 @@ def _convert_X(self, X):
...
+            if get_n_timepoints(X) != self.metadata_["n_timepoints"]:
+                raise ValueError(
+                    "X has different length to the data seen in fit but "
+                    "this classifier cannot handle unequal length series."
+                    "length of train set was",
+                    self.metadata_["n_timepoints"],
+                    " length in predict is ",
+                )
+        if self.get_tag("capability:multivariate"):
+            if get_n_channels(X) != self.metadata_["n_channels"]:
```
hadifawaz1999, line 233: maybe redundent, but should we also check if same length is on all channels when "capability:unequal_length" is false ?

aeon/base/_base_collection.py
```diff
@@ -212,6 +212,33 @@ def _convert_X(self, X):
...
+                raise ValueError(
+                    "X has different number of channels to the data seen in fit "
+                    "number of channels in train set was",
+                    self.metadata_["n_channels"],
+                    "but in predict it is ",
+                    get_n_timepoints(X),
```
hadifawaz1999, line 239: should be get_n_channels no ?

aeon/classification/base.py
```diff
@@ -175,7 +171,7 @@ def predict(self, X) -> np.ndarray:
             n_cases = get_n_cases(X)
             return np.repeat(list(self._class_dictionary.keys()), n_cases)
 
-        X = self._preprocess_collection(X)
+        X = self._preprocess_collection(X, store_metadata=False)
```
hadifawaz1999, line 174: just as a reminding note to fix the breaks later, this is the change that caused so should adapt to it

hadifawaz1999 (changes requested):
Changes look good just a couple of comments @MatthewMiddlehurst

aeon/base/_base_collection.py
MatthewMiddlehurst: If there are a different number of channels at any point it should throw an error, thats not supported. 

Unequal length data where the unequal tag is false will be picked up in data preprocessing.

Only things we want to check are consistent in both `fit` to `predict` is these two I think.

aeon/base/_base_collection.py
MatthewMiddlehurst: yep

TonyBagnall:
basic structure looks good to me

MatthewMiddlehurst:
@chrisholder any issue with the cache change? currently I think it runs when collecting tests, or at least it did for some reason locally.

hadifawaz1999 (dismissed):

MatthewMiddlehurst:
@hadifawaz1999 there was a conflict with the testing PRs, could you approve again?

hadifawaz1999 (approved):

Issue #1874, opened 2024-07-31
[ENH] Parameterise the `dtw` and `adtw` distances


itsdivya1309:
### Describe the feature or idea you want to propose

Recent advances in similarity-based classification, like the [Proximity Forest 2.0](https://github.com/aeon-toolkit/aeon/issues/428) include parameterising cost functions used in similarity measures. However, the current aeon implementation of `dtw` uses squared distance as the cost function. The Minkowski distance is a generalised form of both the Euclidean and Manhattan distance, and hence can be used as the cost function for `dtw` and `adtw`.

### Describe your proposed solution

Use Minkowski distance instead of `squared_univariate_distance` to compute the `dtw` cost matrix.
`distance = np.sum((np.abs(x - y) ** p)) ** (1.0 / p)`

PR #1876, opened 2024-07-31
[ENH] Parameterise the cost function of `dtw` and `adtw` distance measures

itsdivya1309:
#### Reference Issues/PRs
Closes: #1874

As we discussed on the slack, I've implemented the parameterised DTW and ADTW as private functions to be used in the upcoming Proximity Forest 2.0.

##### For all contributions


hadifawaz1999:
maybe these can go in distance module as parameter to the original dtw based functions ? not saying in this PR as i know its only used for PF2 for now, but future perspective, what do you think @chrisholder  ? however we should have a discussion on how it can affect BA stuff (some math related stuff if we go into this as a general parameter)

MatthewMiddlehurst:
FYI there were some large conversations regarding this in the core_devs and distances Slack channels. It is not really compatible with the way those functions are currently set up, best to take a look at there.

For now we will do it like this just to move things forward.

MatthewMiddlehurst:
I assume these are the same functions as your previous changes, but just in a new fine @itsdivya1309? If so I dont really see any issue for now.

itsdivya1309:
> I assume these are the same functions as your previous changes, but just in a new fine @itsdivya1309? If so I dont really see any issue for now.

Yes, that's right.

MatthewMiddlehurst (commented):
I think it looks fine for now, but I would put it in the same file as PF 2.0 rather than its own.

MatthewMiddlehurst:
replaced by #1978

PR #1880, opened 2024-08-01
Using pandocs syntax for double reference

IRKnyazev:
Checking to see if pandocs syntax resolves the issue1879


IRKnyazev:
This PR will be closed and replaced with a standardized one after I have seen the affects on readthedocs

PR #1904, opened 2024-08-04
[MNT, DOC] Accelerating deep testing

### LINKED ISSUE #1761: [MNT] testing random state of deep clustering module is slow
### Describe the issue

The `test_random_state_deep_learning_clr` may need some refinement on the test params to make the test faster, especially as it is giving timeout when removing numba in CI

To do:

1. reduce data size
2. reduce param size
3. split clusterers into different tests

### Additional context

_No response_

hadifawaz1999:
Fix #1761 

- Changes test params of deep classifiers/regressors/clusterers so that testing is faster
- In deep specific tests, use the test parameters, as they were not used before, to accelerate testing
-  Fix random state deep clustering test, it was looping infinitely before because there was a bug, it was considering the encoder/decoder as layers before, not its fixed and ensures encoders and decoders are testing separately
- Add possibility to use dummy clusterer in deep clustering, for the clustering in latent space, use it as test parameter to accelerate testing especially in no numba label. Another reason for this addition is that with test parameters of AE based clusterers in deep learning, if not trained long enough, kmeans may not converge and CI will fail with "ValueError: Unable to find a valid cluster configuration with parameters specified (empty clusters kept forming). Try lowering your n_clusters or raising n_init." so it will be a pain to find the best seed, so replacing with dummy clusterer is best option
- Fix Encoder Network issue with Lambda layer, using now internal attention layer, as the original authors implemented the network before keras introduced Attention layers. The math isnt 100% the same but the idea of attention is the same
- Doc issue with AEResNet


aeon/clustering/deep_learning/base.py
```diff
@@ -115,7 +116,11 @@ def _fit_clustering(self, X):
         else:
             clustering_params_ = self.clustering_params
             # clustering_params_["n_clusters"] = self.n_clusters
-        if self.clustering_algorithm == "kmeans":
+        if self.clustering_algorithm == "dummy":
+            self.clusterer = DummyClusterer(
+                n_clusters=self.n_clusters, **clustering_params_
+            )
+        elif self.clustering_algorithm == "kmeans":
```
MatthewMiddlehurst, line 123: Can this not just accept any `BaseClusterer`? Creating a useless option solely for testing is not a great way to resolve this IMO.

aeon/networks/_ae_resnet.py
```diff
@@ -138,36 +138,43 @@ def build_network(self, input_shape, **kwargs):
         self._kernel_size_ = [8, 5, 3] if self.kernel_size is None else self.kernel_size
 
         if isinstance(self._n_filters_, list):
+            assert len(self._n_filters_) == self.n_residual_blocks
```
MatthewMiddlehurst, line 141: raise an actual error with a message instead of asserting

aeon/clustering/deep_learning/base.py
hadifawaz1999: i wanted to change that for accepting an estimator input instead of string, but thought it might be a lot for the PR, but to keep the PR for testing purpose this can be done, if you think its ok to get all in one PR i dont mind can do the changes here

aeon/networks/_ae_resnet.py
hadifawaz1999: will do, we should raise an issue to do that all over the networks module, my code my bad ! never thought about raising a message

aeon/clustering/deep_learning/base.py
hadifawaz1999: fixed

aeon/networks/_ae_resnet.py
hadifawaz1999: fixed

MatthewMiddlehurst: It would be better as a `ValueError` IMO

aeon/clustering/deep_learning/base.py
MatthewMiddlehurst: Don't mind if you do it here. The dummy option is not a good addition IMO.

hadifawaz1999: will add the changes then

TonyBagnall:
hi @hadifawaz1999 this is very conflicted, it might be better to start again and do it in multiple PRs?

hadifawaz1999:
> hi @hadifawaz1999 this is very conflicted, it might be better to start again and do it in multiple PRs?

if am not mistaken, i dropped this and did every bullet point in a separate pr, will check what's left and close this one

Issue #1914, opened 2024-08-06
[DOC] Confusing placement for shell error solution


phershbe:
### Describe the issue linked to the documentation

https://github.com/aeon-toolkit/aeon/blob/main/docs/developer_guide/dev_installation.md

Step 2
```
To build and install aeon from source, navigate to the local clone's root directory and type:
pip install --editable .[dev]
```
...is at the top

And then an explanation for a problem with it...

```
If this results in a "no matches found" error, it may be due to how your shell handles special characters. Try surrounding the dependency portion with quotes:
pip install --editable ."[dev]"
```
...is at the bottom after the explanations of the `.`, `--editable`, and `all_extras`

I didn't make it to the bottom initially, and had the problem and searched for it and it seems to be a really common problem, and then noticed it here later, so it could be more inviting to new contributors to change the order.

### Suggest a potential alternative/fix

Change the order to...

```
To build and install aeon from source, navigate to the local clone's root directory and type:
pip install --editable .[dev]
```
...and then...

```
If this results in a "no matches found" error, it may be due to how your shell handles special characters. Try surrounding the dependency portion with quotes:
pip install --editable ."[dev]"
```
...and then mention the explanations of `.`, `--editable`, and `all_extras`

The start of the solution is in bold already = **If this results in a "no matches found" error** = so having it immediately after the problem that it is trying to correct should help to make things clearer

baraline:
One alternative is to consider it as a formal definition, where introducing the terms used after the definition can be a bit confusing !

Would having the definition/explanation of  `.` and `--editable` before 
```
To build and install aeon from source, navigate to the local clone's root directory and type:
pip install --editable .[dev]
```
And then immediately followed by the possible error cause as you suggested be a good solution ? I think it would balance both point of views on this issue (parameter first vs expected error first).

phershbe:
That would make sense to me for introducing a function for example, but here I think that the most important thing is to make sure that the user can get the project built from source and move on to contributing. 

I do have a counter proposal that includes the idea of what you mentioned...

```
To build and install aeon from source, navigate to the local clone's root directory and type:
pip install --editable .[dev]

Alternatively, the . may be replaced with a full or relative path to the root directory.

If this results in a "no matches found" error, it may be due to how your shell handles special characters. Try surrounding the dependency portion with quotes:
pip install --editable ."[dev]"
```
[...]

PR #1915, opened 2024-08-06
[DOC] Change order of build and install instructions for clarity

phershbe:
Issue #1914

There is a problem where `pip install --editable .[dev]` needs to have quotation marks `pip install --editable ."[dev]"` because of how your shell handles special characters.

It was explained in the documentation, but the solution was at the bottom of the section not so close to the issue that it was explaining, so the solution was moved in order to be immediately after the issue.


TonyBagnall:
@all-contributors add @phershbe for docs

TonyBagnall (approved):
LGTM, thanks very much

phershbe:
Thank you too and to @baraline, such a cool project

PR #1923, opened 2024-08-07
[ENH] Add Multi Reconstruction loss to AEFCNClusterer

aadya940:
Mean Squared Error between encoder and decoder layers summed up for a symmetric deep auto-encoder architecture.

### DIFF
--- a/aeon/clustering/deep_learning/_ae_fcn.py
+++ b/aeon/clustering/deep_learning/_ae_fcn.py
@@ -273,14 +273,25 @@ class AEFCNClusterer(BaseDeepClusterer):
                 file_name=self.file_name_,
             )
 
-        self.history = self.training_model_.fit(
-            X,
-            X,
-            batch_size=mini_batch_size,
-            epochs=self.n_epochs,
-            verbose=self.verbose,
-            callbacks=self.callbacks_,
-        )
+        if not self.loss == "multi_rec":
+            self.history = self.training_model_.fit(
+                X,
+                X,
+                batch_size=mini_batch_size,
+                epochs=self.n_epochs,
+                verbose=self.verbose,
+                callbacks=self.callbacks_,
+            )
+
+        elif self.loss == "multi_rec":
+            self._fit_multi_rec_model(
+                autoencoder=self.training_model_,
+                inputs=X,
+                outputs=X,
+                batch_size=mini_batch_size,
+                epochs=self.n_epochs,
+                callbacks=self.callbacks_,
+            )
 
         try:
             self.model_ = tf.keras.models.load_model(
@@ -302,6 +313,79 @@ class AEFCNClusterer(BaseDeepClusterer):
         X = X.transpose(0, 2, 1)
         latent_space = self.model_.layers[1].predict(X)
         return self.clusterer.score(latent_space)
+
+    def _fit_multi_rec_model(
+        self, autoencoder, inputs, outputs, batch_size, epochs, callbacks, optimizer
+    ):
+        import numpy as np
+        import tensorflow as tf
+        from tensorflow.keras import backend as K
+
+        train_dataset = tf.data.Dataset.from_tensor_slices((inputs, outputs))
+        train_dataset = train_dataset.shuffle(buffer_size=1024).batch(batch_size)
+
+        def layerwise_mse_loss(autoencoder, inputs, outputs):
+            def loss(y_true, y_pred):
+                # Calculate MSE for each layer in the encoder and decoder
+                mse = 0
+
+                # Encoding layers
+                _encoder_intermediate_outputs = []
+                _decoder_intermediate_outputs = []
+                for i, layer in enumerate(autoencoder.layers):
+                    if i == 0:
+                        _prop = layer(inputs)
+                    else:
+                        _prop = layer(_prop)
+                        if isinstance(layer, tf.keras.layers.Activation):
+                            # Encoder
+                            if len(_encoder_intermediate_outputs) < self.n_layers:
+                                _encoder_intermediate_outputs.append(_prop)
+
+                            # Decoder
+                            else:
+                                _decoder_intermediate_outputs.append(_prop)
+
+                # Throw away the output of encoders last layer
+                _decoder_intermediate_outputs.pop(0)
+
+                _encoder_intermediate_outputs = np.array(_encoder_intermediate_outputs)
+                _decoder_intermediate_outputs = np.array(_decoder_intermediate_outputs)
+
+                assert len(_encoder_intermediate_outputs) == len(
+                    _decoder_intermediate_outputs
+                )
+
+                mse += K.mean(
+                    K.squared(
+                        _encoder_intermediate_outputs - _decoder_intermediate_outputs
+                    )
+                )
+
+                return mse
+
+            return loss
+
+        for _ in range(epochs):
+            for step, (x_batch_train, y_batch_train) in enumerate(train_dataset):
+                with tf.GradientTape as tape:
+                    loss = layerwise_mse_loss(
+                        autoencoder=autoencoder,
+                        inputs=x_batch_train,
+                        outputs=y_batch_train,
+                    )
+
+                grads = tape.gradient(loss, autoencoder.trainable_weights)
+                self.optimizer_.apply_gradients(
+                    zip(grads, autoencoder.trainable_weights)
+                )
+
+                if step % 200 == 0:
+                    print(
+                        "Training loss (for one batch) at step %d: %.4f"
+                        % (step, float(loss))
+                    )
+                    print("Seen so far: %d samples" % ((step + 1) * batch_size))
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):


aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -61,7 +62,7 @@ class AEFCNClusterer(BaseDeepClusterer):
     verbose : boolean, default = False
         Whether to output extra information.
     loss : string, default="mean_squared_error"
-        Fit parameter for the keras model.
+        Fit parameter for the keras model. "multi_rec" for multiple mse loss.
```
hadifawaz1999, line 65: could you detail more whats the difference between both losses in the doc ?

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
         latent_space = self.model_.layers[1].predict(X)
         return self.clusterer.score(latent_space)
 
+    def _fit_multi_rec_model(
```
hadifawaz1999, line 318: even though its a private function, i think its good if we add a docstirng to it, if new contributors come along they can see what the function does in the doc

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+        inputs,
+        outputs,
+        batch_size,
+        epochs,
+    ):
+        import tensorflow as tf
+        from tensorflow.keras import backend as K
+
+        train_dataset = tf.data.Dataset.from_tensor_slices((inputs, outputs))
+        train_dataset = train_dataset.shuffle(buffer_size=1024).batch(batch_size)
```
hadifawaz1999, line 330: shouldn't the shuffle be done at each epoch iteration instead of prior to epoch looping ? because we usually shuffle at each epoch, unless this is a functionality that i dont know about and it does what am saying :) in that case never mind, if not then it should be in the loop below

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+    def _fit_multi_rec_model(
+        self,
+        autoencoder,
+        inputs,
+        outputs,
+        batch_size,
+        epochs,
+    ):
+        import tensorflow as tf
+        from tensorflow.keras import backend as K
```
hadifawaz1999, line 327: avoid using keras backend, for instead of K.mean use tf.reduce_mean etc. as long as this is in the loss not the graph, its much more efficient to use tensorflow tensor calculation instead of keras backend

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+                _decoder_intermediate_outputs = (
+                    []
+                )  # Store embeddings of each layer in the Decoder
+
+                encoder = autoencoder.layers[1]  # Returns Functional API Models.
+                decoder = autoencoder.layers[2]  # Returns Functional API Models.
+
+                # Run the models since the below given loop misses the latent space
+                # layer which doesn't contribute to the loss.
+                logits = encoder(inputs)
```
hadifawaz1999, line 354: why calling the models here ?

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+                if not (
+                    len(_encoder_intermediate_outputs)
+                    == len(_decoder_intermediate_outputs)
+                ):
+                    raise ValueError("The Auto-Encoder must be symmetric in nature.")
+
+                for enc_output, dec_output in zip(
+                    _encoder_intermediate_outputs, _decoder_intermediate_outputs
+                ):
+                    mse += K.mean(K.square(enc_output - dec_output))
```
hadifawaz1999, line 384: here it should tf not K as mentioned above

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+            num_batches = 0
+            for step, (x_batch_train, y_batch_train) in enumerate(train_dataset):
+                with tf.GradientTape() as tape:
+                    # Calculate the actual loss by calling the loss function
+                    loss_func = layerwise_mse_loss(
+                        autoencoder=autoencoder,
+                        inputs=x_batch_train,
+                        outputs=y_batch_train,
+                    )
+                    loss_value = loss_func(y_batch_train, autoencoder(x_batch_train))
```
hadifawaz1999, line 406: can u make sure here that loss_value is actually one value and not an array ? test it locally or even add an assert

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+
+                grads = tape.gradient(loss_value, autoencoder.trainable_weights)
+                self.optimizer_.apply_gradients(
+                    zip(grads, autoencoder.trainable_weights)
+                )
+
+                epoch_loss += float(loss_value)
+                num_batches += 1
+
+                if step % 200 == 0:
```
hadifawaz1999, line 416: i prefer if the logging happens after each epoch not each step, showing the avewrage loss over all steps

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +315,129 @@ def _score(self, X, y=None):
...
+        train_dataset = tf.data.Dataset.from_tensor_slices((inputs, outputs))
+        train_dataset = train_dataset.shuffle(buffer_size=1024).batch(batch_size)
+
+        if isinstance(self.optimizer_, str):
+            self.optimizer_ = tf.keras.optimizers.get(self.optimizer_)
+
+        history = {"loss": []}
+
+        def layerwise_mse_loss(autoencoder, inputs, outputs):
+            def loss(y_true, y_pred):
```
hadifawaz1999, line 338: the loss is correctly coded, however you might forgot one thing, the multi rec loss also includes what the traditional  rec loss does, which is mse between input and output of decoder, so you should add this at the end of the layer wise computation

aeon/clustering/deep_learning/tests/test_clusterer_features.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["tensorflow"], severity="none"),
+    reason="Tensorflow soft dependency not found.",
+)
+def test_multi_rec_fcn():
+    """Tests whether multi-rec loss works fine or not."""
+    X = np.random.random((100, 5, 2))
+    clst = AEFCNClusterer(n_clusters=2, n_epochs=10, loss="multi_rec")
```
hadifawaz1999, line 17: to make sure symmetry is correctly handled in the layer wise computation, can u set the n_filters here to somehting like [2,3,4] ? because the default is [128,256,128] which 1stly is computationally expensive for a test and 2ndly a bug can easilly be missed because of the repeated 128

hadifawaz1999 (changes requested):
all looks good just some minor changes, thanks !

aeon/clustering/deep_learning/_ae_fcn.py
aadya940: It was a need since the last layer in the encoder didn't have an activation hence, there was  a shape mismatch as the tensors never passed through the last layer. Basically, it is added so that propogation of tensors between encoder and decoder doesn't break.

aeon/clustering/deep_learning/_ae_fcn.py
aadya940: OK, so basically add normal "mse" to the loss?

hadifawaz1999: yes

aeon/clustering/deep_learning/_ae_fcn.py
aadya940: In the docs its done this way: 

https://www.tensorflow.org/guide/keras/writing_a_training_loop_from_scratch

aeon/clustering/deep_learning/tests/test_clusterer_features.py
```diff
@@ -0,0 +1,23 @@
...
+)
+def test_multi_rec_fcn():
+    """Tests whether multi-rec loss works fine or not."""
+    X = np.random.random((100, 5, 2))
+    clst = AEFCNClusterer(
+        n_clusters=2, n_epochs=10, n_filters=[2, 3, 4], loss="multi_rec"
+    )
+    clst.fit(X)
+    assert (
+        clst.history["loss"][0] > clst.history["loss"][1]
```
hadifawaz1999, line 22: this is not always true, especially first two epochs:p its so random at the beggining, you can assert that the loss is a list for example or its length is the same as number of epochs etc.

aeon/clustering/deep_learning/_ae_fcn.py
```diff
@@ -303,6 +318,132 @@ def _score(self, X, y=None):
...
+                    __output = _model(logits, training=True)
+                    _decoder_intermediate_outputs.append(__output)
+
+                if not (
+                    len(_encoder_intermediate_outputs)
+                    == len(_decoder_intermediate_outputs)
+                ):
+                    raise ValueError("The Auto-Encoder must be symmetric in nature.")
+
+                # # Append normal mean_squared_error
```
hadifawaz1999, line 383: i think you forgot to append the original mse here

hadifawaz1999: ah no never mind i see what you did :)

hadifawaz1999 (changes requested):
left some comments

aeon/clustering/deep_learning/tests/test_clusterer_features.py
hadifawaz1999: i would set a random state however, for the data generation and the deep mode, and compare epoch 0 with epoch 5, or even 10, just to avoid random fails in the future in CI

hadifawaz1999 (approved):

Issue #1910, opened 2024-08-05
[DOC] Add type hints for shapelet-based collection transformation.


baraline:
### Describe the issue

Other modules are increasingly using type hints in the function declaration, the shapelet-based tranformation module should do the same.

### Suggest a potential alternative/fix

Add type hints to algorithms in `aeon/transformations/collection/shapelet_based`. One PR can be opened for each algorithm if preferred, as some are quite large. 

### Additional context

If you have any question, don't hesitate to ask directly here or contact us on Slack!

baraline:
@Datadote If this one is OK for you, you can assign yourself to it !

Datadote:
Aeon-Assign bot assign @Datadote

baraline:
I don't remember if we have a bot for this, but you can do it manually on the Assignees tab on the right side, I'll do it for you this time. Happy coding, and thanks for contributing !

MatthewMiddlehurst:
There was an error in the docs that we have now fixed, but have to wait for a release for it to appear on the stable branch. Have to start with @aeon-actions-bot

MatthewMiddlehurst:
I think you need write access to manually assign yourself

Datadote:
Yeah, I don't access to manually assign myself (Tried on other issues).

I started a draft pull request with type hints, and some styling questions in #1917 . Thank you for the help

aryanpola:
Aeon-Assign bot assign @aryanpola

PR #1926, opened 2024-08-08
[ENH] add type hints to _shapelet_transform.py __init__ method

Datadote:
#### Reference Issues/PRs
- Part of #1910
- References #1454

- Add type hints to `__init__` method in _shapelet_transform.py

- New dependency added: "typing" library

#### Any other comments?
1) Did not type parallel_backend because one types is "ParallelBackendBase", and did not know how to add that
2) "contract_max_n_shapelet_samples" type -> Union[int, float] because np.inf (float type) is a default
3) "time_limit_in_minutes" type -> Union[int, float] for flexibility. Documentation had int, but default value was 0.0 originally.

### DIFF
--- a/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
+++ b/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
@@ -9,7 +9,7 @@ __all__ = ["RandomShapeletTransform"]
 import heapq
 import math
 import time
-from typing import Optional, Union
+from typing import Optional
 
 import numpy as np
 from joblib import Parallel, delayed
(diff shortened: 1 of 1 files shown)


TonyBagnall (approved):
thanks, I just checked that time limit and contact num shapelets can in fact be ints and floats and its fine

MatthewMiddlehurst:
Not a blocker, fine with that as is, but is it better to do `int or float` or just `float`?  I would say `float` implies integers are also accepted, but it is incredibly minor 🙂

Datadote:
I agree with Matthew's comment, and updated all "Union[int, float]" to "float"

MatthewMiddlehurst (approved):

Issue #1920, opened 2024-08-07
[ENH] Smoothing filters as BaseSeriesTransformers


TonyBagnall:
### Describe the feature or idea you want to propose

A while back (five years?) we compared different smoothing algorithms for time series classification. 

https://link.springer.com/chapter/10.1007/978-3-030-29859-3_5

Now, we found it did not make much difference on the UCR data, but that does not mean it is not of use. It would be good to implement some or all of these as BaseSeriesTransformers, most are pretty simple.

### Describe your proposed solution

we looked at the following, but there are probably many more. These are all parameterised, so it would be good to have an automatic way of 

- [x]  Simple Moving Average (good first issue)
- [x] Exponential smoothing (good first issue)
- [x] Gaussian filter (good first issue)
- [x] Savitzky-Golay filter
- [x] Discrete Fourier Approximation #1967 @Cyril-Meyer 
- [x] Recursive Median Sieve

### Describe alternatives you've considered, if relevant

list any others below

Cyril-Meyer:
Aeon-Assign bot assign @Cyril-Meyer

I will continue to work on this when #1967 (DFT) is merged (the next filters will be based on this implementation, so I prefer to have it validated before). I will focus on Savitzky-Golay and Recursive Median Sieve to let some good first issue for new members.
Eventually, I will finish what has not been done after some weeks.

Datadote:
I've implemented "Simple Moving Average" in PR #1927 . It is waiting for a review

Now, I'm starting on "Exponential Smoothing", and will add a draft PR.

TonyBagnall:
all done, nice one @Cyril-Meyer and @Datadote

PR #1927, opened 2024-08-08
[ENH] add BaseSeriesTransformer - Moving Average Transformer

Datadote:
#### Reference Issues/PRs
- Still in progress
- Part of #1920

- Add smoothing filter MovingAverageTransformer

- TODO

#### Any other comments?
- TODO

### DIFF
--- a/aeon/transformations/series/_moving_average.py
+++ b/aeon/transformations/series/_moving_average.py
@@ -1,0 +1,41 @@
+""" Moving average transformation """
+
+__maintainer__ = ['Datadote']
+__all__ = 'MovingAverageTransformer'
+
+import numpy as np
+
+from aeon.transformations.series.base import BaseSeriesTransformer
+
+class MovingAverageTransformer(BaseSeriesTransformer):
+    """ Filter a time series using a simple moving average.
+    """
+    
+    _tags = {
+        "capability:multivariate": True,
+        "X_inner_type": "np.ndarray",
+        "fit_is_empty": True, # TODO: what bool to set?
+    }
+
+    def __init__(
+            self,
+            window_size: int = 5,
+    ) -> None:
+        super().__init__(axis=0)
+        self.window_size = window_size
+
+    def _transform(self, X, y=None):
+        """ Transform X and return a transformed version.
+        """
+        if X.ndim == 1:
+            X = X.reshape(-1, 1)
+        csum = np.cumsum(X, axis=0)
+        csum[self.window_size:, :] = csum[self.window_size:, :] - csum[:-self.window_size, :]    
+        return csum[self.window_size - 1:, :] / self.window_size
+
+    @classmethod
+    def get_test_params(cls, parameter_set="default"):
+        """ Return testing parameter settings for the estimator.
+        """
+        params = {"window_size": 5}
+        return params


aeon/transformations/series/_moving_average.py
```diff
@@ -0,0 +1,83 @@
...
+    from aeon.transformations.series._moving_average import MovingAverageTransformer
+    X = np.array([-3, -2, -1,  0,  1,  2,  3])
+    transformer = MovingAverageTransformer(2)
+    Xt = transformer.fit_transform(X)
+    """
+    
+    _tags = {
+        "capability:multivariate": True,
+        "X_inner_type": "np.ndarray",
+        "fit_is_empty": True, # TODO: what bool to set?
```
Cyril-Meyer, line 36: If you set `fit_is_empty` to true instead of default false, the `fit` method in base class will not try to call `_fit` and stuff.
Your choice looks good to me :)

Datadote: Thanks for the reply.

aeon/transformations/series/_moving_average.py
```diff
@@ -0,0 +1,77 @@
...
+
+import numpy as np
+
+from aeon.transformations.series.base import BaseSeriesTransformer
+
+
+class MovingAverageSeriesTransformer(BaseSeriesTransformer):
+    """Filter a time series using a simple moving average.
+
+    Parameters
```
MatthewMiddlehurst, line 14: a simple description would be good in the middle here.

aeon/transformations/series/_moving_average.py
```diff
@@ -0,0 +1,77 @@
+"""Moving average transformation."""
+
+__maintainer__ = ["Datadote"]
+__all__ = "MovingAverageSeriesTransformer"
```
MatthewMiddlehurst, line 4: Can you make this a list?

MatthewMiddlehurst (commented):
Will let @TonyBagnall comment on functionality, but looks fine to me.

aeon/transformations/series/_moving_average.py
Datadote: Done

aeon/transformations/series/_moving_average.py
Datadote: I made the first line description more clear

before: """Filter a time series using a simple moving average.
to_____: """Calculate the moving average of an array of numbers.

and don't think any more description is needed for this simple class (too verbose?).

MatthewMiddlehurst: Usually we have the first line as a summary of the class (i.e. name and very basic functionality), but maybe that is all that is required in this case. Won't push for more.

aeon/transformations/series/_moving_average.py
```diff
@@ -0,0 +1,79 @@
...
+        y : ignored argument for interface compatibility
+            Additional data, e.g., labels for transformation
+
+        Returns
+        -------
+        Xt: 2D np.ndarray
+            transformed version of X
+        """
+        if X.ndim == 1:
+            X = X.reshape(-1, 1)
```
MatthewMiddlehurst, line 71: Just remembered we convert to 2d in the base class actually

https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/_base_series.py#L217

Datadote: Oh, I see. Thanks. I've removed the redundant code in most recent commit

TonyBagnall (approved):
LGTM, only thing I can think of is to check that window length is not longer than the series, but not a blocker

PR #1934, opened 2024-08-08
[ENH] Fix error messages in BaseClassifier

TonyBagnall:
size mismatch error messages were incorrect


MatthewMiddlehurst:
will conflict with #1865

PR #1943, opened 2024-08-11
[BUG] Mapped input index to class index for shapelet viz 

### LINKED ISSUE #1942: [BUG] Shapelet visualiser not enforcing class indexes
### Describe the bug

`visualize_best_shapelets_one_class` has the two following parameters: `id_example_class` and `id_example_others`. These are set to None by default. Everything works as intended for this default case, however when a user provides his own index values for either parameter the index called is from the whole dataset rather than an index of the timeseries of that class.

```python
fig = rst_rf_vis.visualize_best_shapelets_one_class(
    X_gun_test,
    y_gun_test,
    0,
    figure_options={"figsize": (18, 12), "nrows": 2, "ncols": 2},
    id_example_class = 1,
    id_example_other = 1,
)
```

### Expected results

If a user provides index 1 for both parameters then the same time series should not be called, instead the first time series from each class.

### Actual results

image
<br>
The graph essentially says the time series is from both classes.

### Versions

python: 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 15:03:56) [MSC v.1929 64 bit (AMD64)]
executable: [c:\ProgramData\miniconda3\envs\aeon\python.exe](file:///C:/ProgramData/miniconda3/envs/aeon/python.exe)

Python dependencies:
          pip: 24.0
   setuptools: 69.5.1
 scikit-learn: 1.4.2
         aeon: 0.10.0
  statsmodels: 0.14.2
        numpy: 1.26.4
        scipy: 1.12.0
       pandas: 2.0.3
   matplotlib: 3.9.1
       joblib: 1.4.2
        numba: 0.59.1
     pmdarima: 2.0.4
      tsfresh: 0.20.2

IRKnyazev:
#### Reference Issues/PRs
Fixes #1942 

This stops accidental calls to the wrong class.

I added a condition where the input param is not None to map the input to the index of time series of the given class. 

No.

### DIFF
--- a/aeon/visualisation/estimator/_shapelets.py
+++ b/aeon/visualisation/estimator/_shapelets.py
@@ -817,10 +817,11 @@ class ShapeletClassifierVisualizer:
             Number of plots to output, one per shapelet (i.e. the n_shp best shapelets
             for class_id). The default is 1.
         id_example_other : int
-            Sample ID to use for sample of other class. If None, a random one is
-            selected.
+            Sample ID to use for sample of other class. If None, a random one from that
+            class is selected.
         id_example_class : int
-            Sample ID to use for sample of class_id. If None, a random one is selected.
+            Sample ID to use for sample of class_id.If None, a random one from that
+            class is selected.
         scatter_options : dict
             Dictionnary of options passed to the scatter plot of the shapelet values.
         x_plot_options : dict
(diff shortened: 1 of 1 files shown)


baraline (approved):
The initial though was indeed to select based on the index in the whole dataset. But I agree that taking the `id_example_class-th`  sample from the sample of the class actually makes more sense and is less prone to errors.

PR #1944, opened 2024-08-11
[MNT] Make input clustering algorithm for deep leaning an estimator parameter instead of string

hadifawaz1999:
Will help accelerate training and avoid having to hard code all possibilities of clusterers


MatthewMiddlehurst (commented):
Is `clustering_params` being deprecated also? see https://www.aeon-toolkit.org/en/stable/developer_guide/deprecation.html#renaming for how to handle them

hadifawaz1999:
> Is `clustering_params` being deprecated also? see https://www.aeon-toolkit.org/en/stable/developer_guide/deprecation.html#renaming for how to handle them

yeah it is, but i already handled them https://github.com/aeon-toolkit/aeon/blob/40affaed5e493e89835a08d84c09c36a10e6f7d8/aeon/clustering/deep_learning/base.py#L124
they are all deprecated and replaced by estimator

hadifawaz1999:
btw @MatthewMiddlehurst this is not a renaming deprecation, i am deprecating 3 parameters and adding a new one that does another thing, they cant be handled as renaming

PR #1953, opened 2024-08-12
[ENH] Add support for handling file extensions in load_from_tsfile

### LINKED ISSUE #1922: [ENH] input to load_from_tsfile

the function ``load_from_tsfile`` assumes that the extension is not given and (not unreasonably, given the name) appends it as ".ts"

```python
from aeon.datasets import load_from_tsfile
full = root+classification[0]
X,y = load_from_tsfile(full_file_path_and_name=full)
```
this came up when someone gave me a file with .txt extension (but in valid data format). 

Might be nice to allow an extension and only append .ts if none is given? Currently it will try load .ts.ts if you pass full_file_path_and_name with .ts on the end. The code is a bit convoluted due to paths, but should be a simple change, just check if there is a file name extension, if not, append .ts, if there is, try with what is given

_No response_

### Additional context

_No response_

Sharathchenna:
I made changes in the data_loaders.py to check the extension of the file. If an extension is already like .txt, it will not append .ts. If there is no extension, it will append .ts extension

resolves #1922


Sharathchenna:
@MatthewMiddlehurst Hey! Hello can you please check into this code because i ran pre-commit command and made some changes but the test is still failing . Thanks!

baraline (approved):
Looks like it solve the issue, thanks for contributing ! 
I'll leave the merging to @TonyBagnall so he can confirm that's what he wanted.

TonyBagnall (approved):
great, thanks for that

TonyBagnall:
@all-contributors add @Sharathchenna for code

PR #1949, opened 2024-08-12
[DOC] Added Type Hints for _dilated_shapelet_transform 

aryanpola:
What does this implement/fix? Explain your changes.
I have added type hints for function variables and return values as described in issue.
#1910

PR #1959, opened 2024-08-13
[ENH] Added useful attributes to extracted shapelets for RDST

### LINKED ISSUE #1950: [ENH] Can't plot Random Dilated Shapelets on their extracted time point positions.
### Describe the issue

Currently trying to visualise dilated shapelets for a notebook and I want to plot 10 extracted shapelets however I can only plot them stacked upon one another. It is important to see from which part of a time series the shapelet is extracted so you need the start point of the extracted shapelet to plot it in its corresponding place. This is made possible in the shapelet transform class by accessing its 'shapelets' list attribute.

Add the shapelet start point to the shapelets class attribute in RDST in the same way as ST.

### Additional context

image
<br> a plot of extracted shapelets using ST.
<br>
image
<br>
Current attempt at making same plot but for RDST - as you can see they are all plotted from 0 rather than where they truly lie in the time series.

IRKnyazev:
#### Reference Issues/PRs
Fixes #1950 

Added a list of starting points & class values for the transformer's extracted shapelets.

No.

#### Any other comments?
I'm pretty sure this introduction doesn't harm any functionality across the module. I searched for all uses of RDST and checked for calls to RDSTClassifier._transformer.shapelets_


baraline (approved):
LGTM thanks ! Would be nice to have start index for all shapelets to visualize the generation process at some point

IRKnyazev:
I can't seem to figure out how the error relates to the code I have added

Moonzyyy:
I dmed you about the issue, someone else had it previously and it seems its just a random chance to pass it or not. Worth raising it as an issue. Haven't looked at the algorithm myself yet but id assume it most likely needs to have set a seed.

baraline:
Just noting that there will be an impact on #1949 when this goes though

baraline (approved):
I'm fine with the addition of classes, although it might not be obvious to use. For example, consider a shapelet extracted from class 0. It might be more discriminative toward class 2 due to the absence of a common pattern present in class 0 and 1 for example. 
But this is not the concern of this PR, the storing of these attributes also not likely to take much space, so LGTM !

IRKnyazev:
> I'm fine with the addition of classes, although it might not be obvious to use. For example, consider a shapelet extracted from class 0. It might be more discriminative toward class 2 due to the absence of a common pattern present in class 0 and 1 for example. 
> 

Yes I totally agree, that's exactly what made me want to use it in my notebook haha, being able to visualise shapelets and their source class should improve explainability

PR #1971, opened 2024-08-14
[BUG] Corrected shapelet plot for RDST 

### LINKED ISSUE #1970: [BUG] visualize_best_shapelets_one_class not handling RTSD correctly
### Describe the bug

For RTSD classifiers this method plots the same shapelet three times at different levels of importance.

RTSD transforms the time series space into an array of shape (n_cases, 3*n_shapelets) where 3 features relate to one shapelet. The visualize_best_shapelets_one_class method ranks the importance of each feature according to weights from the classifier, in the case of ST this works because each feature is one shapelet. This function works by calling _get_shp_importance which specially handles RDST by grouping the three features under one index for a shapelet, however this index still appears three times in the returned list. Instead the position of this index should be handled by either:
- averaging the position of the shapelets three features.
- - Returning only the best position out of the three features.

_No response_

### Expected results

Each shapelet can only be plotted once.

### Actual results

Each shapelet is plotted three times according to the importance of each of its three features.

### Versions

_No response_

IRKnyazev:
#### Reference Issues/PRs
Fixes #1970 

Ranking shapelet importance by summing coefficients of features associated to each shapelet. Also added parameter to enable worst shapelet plotting.


baraline (approved):
LGTM thanks ! :)

PR #1979, opened 2024-08-15
[ENH] Catch22 error handling for unimported pycatch22 module

Moonzyyy:

Currently when running pycatch22 within catch22 a error may occur if the user hasn't imported the module. Instead, a warning is shown that the user needs to install pycatch22 and allowing the algorithm to still run using default settings.

Warnings. Though aeon library already has the dependency installed.


aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -220,33 +221,42 @@ def _transform(self, X, y=None):
...
-                pycatch22.SP_Summaries_welch_rect_area_5_1,
-                pycatch22.SB_BinaryStats_diff_longstretch0,
-                pycatch22.SB_MotifThree_quantile_hh,
-                pycatch22.SC_FluctAnal_2_rsrangefit_50_1_logi_prop_r1,
-                pycatch22.SC_FluctAnal_2_dfa_50_1_2_logi_prop_r1,
-                pycatch22.SP_Summaries_welch_rect_centroid,
-                pycatch22.FC_LocalSimple_mean3_stderr,
-            ]
-        else:
+            try:
```
MatthewMiddlehurst, line 224: I would use this instead of try/except `_check_soft_dependencies("pycatch22", severity="none")`

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -220,33 +221,42 @@ def _transform(self, X, y=None):
...
+                    pycatch22.SP_Summaries_welch_rect_area_5_1,
+                    pycatch22.SB_BinaryStats_diff_longstretch0,
+                    pycatch22.SB_MotifThree_quantile_hh,
+                    pycatch22.SC_FluctAnal_2_rsrangefit_50_1_logi_prop_r1,
+                    pycatch22.SC_FluctAnal_2_dfa_50_1_2_logi_prop_r1,
+                    pycatch22.SP_Summaries_welch_rect_centroid,
+                    pycatch22.FC_LocalSimple_mean3_stderr,
+                ]
+            except ImportError:
+                self.use_pycatch22 = False
```
MatthewMiddlehurst, line 252: I wouldn't change this, it will cause test failures later after a revamp

aeon/transformations/collection/feature_based/_catch22.py
Moonzyyy: added soft dependency now instead of try catch

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -219,34 +221,69 @@ def _transform(self, X, y=None):
...
+                ]
+
+                use_pycatch22_transform = True
+            else:
+                warnings.warn(
+                    "pycatch22 not installed, but 'self.use_pycatch22' is set to True."
+                    "Please install pycatch22. Aeon catch22 will be used.",
+                    stacklevel=2,
+                )
+                features = [
```
MatthewMiddlehurst, line 261: Dont need to duplicate this IMO, Just define the list above and use it for both non-`pycatch22` and the backup or replace the `if not self.use_pycatch22:` line below.

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -219,34 +221,69 @@ def _transform(self, X, y=None):
...
+                    pycatch22.SC_FluctAnal_2_dfa_50_1_2_logi_prop_r1,
+                    pycatch22.SP_Summaries_welch_rect_centroid,
+                    pycatch22.FC_LocalSimple_mean3_stderr,
+                ]
+
+                use_pycatch22_transform = True
+            else:
+                warnings.warn(
+                    "pycatch22 not installed, but 'self.use_pycatch22' is set to True."
+                    "Please install pycatch22. Aeon catch22 will be used.",
```
MatthewMiddlehurst, line 258: "The aeon implementation of catch22 will be used instead."

Moonzyyy: Removed Aeon catch22 will be used and added *"The aeon implementation of catch22 will be used instead."*

aeon/transformations/collection/feature_based/_catch22.py
Moonzyyy: Defined the list above and removed duplication of lines: Feature is already defined with catch22 features but if pycatch22 an available solution and asked for will be swapped to pycatch22.

aeon/transformations/collection/feature_based/_catch22.py
MatthewMiddlehurst: I dont think this change was made

MatthewMiddlehurst:
It would be good to change/add to the testing somehow to cover this. We have CI with and without the pycatch22 install, so a test to make sure that it functions correctly without would be good.

TonyBagnall (approved):

Issue #1951, opened 2024-08-12
[ENH] Adapt ChannelSelector for regression


TonyBagnall:
### Describe the feature or idea you want to propose

the ChannelSelector class evaluates each channel independently and picks the best. It could easily be adapted for regression.

### Describe your proposed solution

adapt the class so it can take a Regressor or Classifier, then use MSE or equivalent to score

aryanpola:
Hello, I'm new to contributing to aeon, could you point me to resources that would help me to understand the issue.

aryanpola:
Aeon-Assign bot assign @aryanpola

MatthewMiddlehurst:
Bit late but I would look at the contributor guide. If you still need help, what part are you confused about?

TonyBagnall:
fixed in #1993 I think

PR #1993, opened 2024-08-20
[ENH] Added Regressor for Channel Scorer

aryanpola:
#1951 

what at does this implement/fix? Explain your changes.
I have added a channel_scorer for a regressor which evaluates through MSE.


aryanpola:
@TonyBagnall

aryanpola:
Is there any way to run these tests locally? @MatthewMiddlehurst

MatthewMiddlehurst:
Yes, it all uses the `pytest` framework. You should be able to run all tests or individual ones by configuring the command

https://docs.pytest.org/en/stable/how-to/index.html

aryanpola:
@MatthewMiddlehurst Is the code fine? please let me know if there are any changes to be made.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,28 +1,36 @@
...
-    series classifier and could easily be adapted to use forward selection or elbow
+    series estimator and could easily be adapted to use forward selection or elbow
     class methods. Approximately as described in [1]_.
 
     Parameters
     ----------
-    classifier, BaseClassifier, default = MiniROCKET
+    estimator: BaseEstimator, default: None
+        The classifier or regressor to be used for scoring channels.
+    classifier: BaseClassifier, default = MiniROCKET
```
baraline, line 33: This line can be removed

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -65,31 +73,43 @@ def _fit(self, X, y):
         if self.proportion <= 0 or self.proportion > 1:
             raise ValueError("proportion must be in the range 0-1")
 
-        if self.classifier is None:
-            self.classifier_ = RocketClassifier(
+        if self.estimator is None:
+            # Default to a classifier if no estimator is provided
+            self.estimator_ = RocketClassifier(
                 rocket_transform="minirocket", num_kernels=5000
             )
```
baraline, line 80: Would it be better to assume the default (classifier/regressor) based on the data type of y? For example, if int or str then classifier, if float then assume regression? Not sure about this one, so open to discussion!

baraline (commented):
Two minor comments. Otherwise, looking good !
One thing that we might think of is to put the metric as a parameter, but would introduce numerous changes, so might be better to keep it for a later PR.

aeon/transformations/collection/channel_selection/_channel_scorer.py
aryanpola: Wouldn't float and int have regressors?

aryanpola:
@baraline I changed the code and ran it, there is an error occurring. Can you help me with it.

aeon/transformations/collection/channel_selection/_channel_scorer.py
MatthewMiddlehurst: I think this is too complex IMO, I would keep it defaulting to classification for now. `int` can be both.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -104,5 +142,11 @@ def get_test_params(cls, parameter_set="default"):
...
         from aeon.classification import DummyClassifier
+        from aeon.regression import DummyRegressor
 
-        return {"classifier": DummyClassifier(), "proportion": 0.4}
+        if parameter_set == "default":
+            return {"estimator": DummyClassifier(), "proportion": 0.4}
+        elif parameter_set == "alternative":
+            return {"estimator": DummyRegressor(), "proportion": 0.4}
+        else:
+            raise ValueError(f"Unknown parameter_set: {parameter_set}")
```
MatthewMiddlehurst, line 152: I would revert this, we have specific sets for some purposes so I dont think this would get used. It would be better to add to `test_channel_scorer.py` to test that regression functions as intended.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,28 +1,39 @@
...
+    using an estimate of accuracy on the training data fro classifier or mean
+    squared error for regressor, then selects a proportion of the top
     channels to keep. Can be configured through the constructor to use any time
-    series classifier and could easily be adapted to use forward selection or elbow
+    series estimator and could easily be adapted to use forward selection or elbow
     class methods. Approximately as described in [1]_.
 
     Parameters
     ----------
-    classifier, BaseClassifier, default = MiniROCKET
```
MatthewMiddlehurst, line 25: The `estimator` parameter needs to be documented, not that it supports regressions it should mention what the default is for `None`.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,28 +1,39 @@
...
+from aeon.regression.base import BaseRegressor
+from aeon.regression.convolution_based._rocket_regressor import RocketRegressor
 from aeon.transformations.collection.channel_selection.base import BaseChannelSelector
 
 __maintainer__ = ["TonyBagnall"]
 __all__ = ["ChannelScorer"]
 
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
```
MatthewMiddlehurst, line 22: Not sure what this is

aeon/transformations/collection/channel_selection/_channel_scorer.py
aryanpola: @MatthewMiddlehurst So just leave it without adding the regressor part for the `get_test_params`?

aeon/transformations/collection/channel_selection/_channel_scorer.py
aryanpola: done

aeon/transformations/collection/channel_selection/_channel_scorer.py
MatthewMiddlehurst: yes

aryanpola:
@MatthewMiddlehurst Please let me know if there any more changes to be made.

MatthewMiddlehurst (commented):
Thanks for implementing my requests. Im still not sure about having the estimator dynamically chosen by default. It may be better to just keep it as it is and have the scoring function as a parameter. Thoughts @baraline @TonyBagnall?

baraline:
One option would be to have the estimator as a mandatory parameter, moving the burden of choosing the correct one depending on the task to the user. Regarding the scoring function, having it as a parameter (which can default to accuracy/mse based on the estimator type) can be really useful for cases where these two default metrics are not ideal to use.

TonyBagnall:
thanks for this, just catching up, will take a look and discuss and get back to you

TonyBagnall:
so I think it is currently over complex, but I'm not actually sure of my preferred solution. I think it best to require the estimator be passed then just fail if y is non numeric, but Im not actually sure. I'll discuss and get back to you

MatthewMiddlehurst:
I like @baraline's suggestion, the issue is this is a published method which uses ROCKET by default I believe @TonyBagnall? 

We discussed this a bit at ECML, I think the best way is to introduce a "scoring" parameter which takes a string or function defaulting to accuracy, and keep the default estimator to ROCKET classifier.

There's not really a good way to automatically to select between classification and regression, and we typically try to stay away from those types of ambiguous inferences.

baraline:
If this is indeed a published method Matthew would be best

TonyBagnall:
> I like @baraline's suggestion, the issue is this is a published method which uses ROCKET by default I believe @TonyBagnall?
> 
> We discussed this a bit at ECML, I think the best way is to introduce a "scoring" parameter which takes a string or function defaulting to accuracy, and keep the default estimator to ROCKET classifier.
> 
> There's not really a good way to automatically to select between classification and regression, and we typically try to stay away from those types of ambiguous inferences.

so its published but I dont think that should block things, I rthink @baraline's solution is probably the best on reflection, in the example I can explain the relationship to the paper later

This is assuming we are talking about this: 
"One option would be to have the estimator as a mandatory parameter, moving the burden of choosing the correct one depending on the task to the user. Regarding the scoring function, having it as a parameter (which can default to accuracy/mse based on the estimator type) can be really useful for cases where these two default metrics are not ideal to use."

so make both the estimator and the scorer and option. Sorry it became so complex @aryanpola

aryanpola:
> > I like @baraline's suggestion, the issue is this is a published method which uses ROCKET by default I believe @TonyBagnall?
> > We discussed this a bit at ECML, I think the best way is to introduce a "scoring" parameter which takes a string or function defaulting to accuracy, and keep the default estimator to ROCKET classifier.
> > There's not really a good way to automatically to select between classification and regression, and we typically try to stay away from those types of ambiguous inferences.
> 
> so its published but I dont think that should block things, I rthink @baraline's solution is probably the best on reflection, in the example I can explain the relationship to the paper later
> 
> This is assuming we are talking about this: "One option would be to have the estimator as a mandatory parameter, moving the burden of choosing the correct one depending on the task to the user. Regarding the scoring function, having it as a parameter (which can default to accuracy/mse based on the estimator type) can be really useful for cases where these two default metrics are not ideal to use."
> 
> so make both the estimator and the scorer and option. Sorry it became so complex @aryanpola

@baraline @MatthewMiddlehurst @TonyBagnall 
1. So now, if the user doesn't select any estimator it defaults to data type, depending on the data type it defaults to either a classifier or a regressor where the user's scoring function is used, else the default mse and accuracy scores.
2. If the user selects a BaseClassifier as estimator, the user's scoring function is used else default accuracy score.
3. If the user selects a BaseRegressor as estimator, the user's scoring function is used else default mse scoring.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,28 +1,47 @@
...
     Parameters
     ----------
-    classifier, BaseClassifier, default = MiniROCKET
+    estimator : BaseEstimator, optional (default=None)
+    The time series estimator used to score each channel.
+    - If `None`, the estimator is automatically selected based on the type of the
+      target variable `y`.
+        - If `y` is of type `int` or `str`, a `RocketClassifier` is used by default.
+        - If `y` is of type `float`, a `RocketRegressor` is used by default.
+
```
MatthewMiddlehurst, line 39: this shouldnt have a default anymore

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -40,16 +59,18 @@ class methods. Approximately as described in [1]_.
 
     def __init__(
         self,
-        classifier: BaseClassifier = None,
+        estimator: BaseEstimator = None,
```
MatthewMiddlehurst, line 62: Same here

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -65,31 +86,72 @@ def _fit(self, X, y):
...
+                    rocket_transform="minirocket", num_kernels=5000
+                )
+                scoring_function = (
+                    mean_squared_error
+                    if self.scoring_function is None
+                    else self.scoring_function
+                )  # noqa: E501
+                score_sign = -1  # Lower MSE is better, hence negative sign
+            else:
+                raise ValueError("y must be of type int, float, or str.")
```
MatthewMiddlehurst, line 116: This can go now

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -65,31 +86,72 @@ def _fit(self, X, y):
...
+        elif isinstance(self.estimator, BaseClassifier):
+            self.estimator_ = self.estimator.clone()
+            scoring_function = (
+                accuracy_score
+                if self.scoring_function is None
+                else self.scoring_function
+            )  # noqa: E501
+            score_sign = 1  # Higher accuracy is better, hence positive sign
```
MatthewMiddlehurst, line 124: How does `score_sign` work for input scoring functions? maybe need another parameter

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -65,31 +86,72 @@ def _fit(self, X, y):
...
+                if self.scoring_function is None
+                else self.scoring_function
+            )  # noqa: E501
+            score_sign = -1  # Lower MSE is better, hence negative sign
+        else:
             raise ValueError(
-                "parameter classifier must be None or an instance of a  "
-                "BaseClassifier."
+                "parameter estimator must be an instance of BaseClassifier, "
+                "BaseRegressor or None."
```
MatthewMiddlehurst, line 136: Remove the None part

MatthewMiddlehurst (commented):
should be no selection based on datatype and there is no default. Also none of the input parameters should be edited in `_fit`, so you may see a failure due to the new scoring parameter.

aeon/transformations/collection/channel_selection/_channel_scorer.py
aryanpola: so `score_sign` would be given by users, if not then default to 1 and -1 for classifier and regressor respectively?

aeon/transformations/collection/channel_selection/_channel_scorer.py
aryanpola: Is there an advantage in removing this?

aryanpola:
@MatthewMiddlehurst all the cases are passing locally. Adding the estimator in aeon/transformations/collection/tests/test_all_collection_transformers.py causing more errors.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,29 +1,49 @@
...
-    series classifier and could easily be adapted to use forward selection or elbow
+    series estimator and could easily be adapted to use forward selection or elbow
     class methods. Approximately as described in [1]_.
 
     Parameters
     ----------
-    classifier, BaseClassifier, default = MiniROCKET
-    proportion : float, default = 0.2
+    estimator : BaseEstimator
+    The time series estimator used to score each channel.
```
MatthewMiddlehurst, line 32: This should be indented to match the style of the parameters below.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -1,29 +1,49 @@
...
+        Scoring function used to evaluate the performance of each channel.
+        Defaults to:
+        - `accuracy_score` if using a classifier.
+        - `mean_squared_error` if using a regressor.
+
+    score_sign : float, optional (default=None)
+        The sign used to determine whether higher or lower scores are better.
+        Defaults to:
+        - 1.0 for classifiers (higher scores are better).
+        - -1.0 for regressors (lower scores are better).
```
MatthewMiddlehurst, line 44: More specifically, these defaults are because of the defaults used in `scoring_function` above. There are classification metrics where lower are better and regression ones where higher are better.

aeon/transformations/collection/channel_selection/_channel_scorer.py
```diff
@@ -65,31 +89,45 @@ def _fit(self, X, y):
         if self.proportion <= 0 or self.proportion > 1:
             raise ValueError("proportion must be in the range 0-1")
 
-        if self.classifier is None:
-            self.classifier_ = RocketClassifier(
-                rocket_transform="minirocket", num_kernels=5000
+        elif isinstance(self.estimator, BaseClassifier):
```
MatthewMiddlehurst, line 92: Think this is better as an `if` instead of an `elif` functionally does not matter because of the `raise`, but they are not really related.

MatthewMiddlehurst (commented):
Couple more comments, but I think this should be good after. Thanks for making the suggested changes, was a bit complex 🙂

aeon/transformations/collection/channel_selection/_channel_scorer.py
MatthewMiddlehurst: not reliable enough for selecting the correct type

aeon/transformations/collection/channel_selection/_channel_scorer.py
MatthewMiddlehurst: yes, match the default to the default used for the scoring function

aryanpola:
> Couple more comments, but I think this should be good after. Thanks for making the suggested changes, was a bit complex 🙂

Please let me know if there's anything more to do. Really liking the work, the community is constructive and responsive 😊.

MatthewMiddlehurst:
I am struggling to figure out why the results for RDST/RIST have changed. This is not happening in any other PR. I thought maybe it was due to this being a fork, but other PRs on the same fork seem to be fine.

aryanpola:
@MatthewMiddlehurst Maybe set random state in `ExtraTreesClassifier' which could provide a deterministic result?

baraline:
> I am struggling to figure out why the deterministic results for RDST/RIST have changed. This is not happening in any other PR. I thought maybe it was due to this being a fork, but other PRs on the same fork seem to be fine.

Sooo, from what I understand, this is an issue with the legacy numpy generator when using numba : 

> Numba supports top-level functions from the numpy.random module, but does not allow you to create individual RandomState instances. The same algorithms are used as for the standard random module, but with an independent internal state: seeding or drawing numbers from one generator won’t affect the other.

Looking at the docs of the random module:

> Numba supports top-level functions from the [random](https://docs.python.org/3/library/random.html#module-random) module, but does not allow you to create individual Random instances. A Mersenne-Twister generator is used, with a dedicated internal state. It is initialized at startup with entropy drawn from the operating system.

That last part is I think what might be getting us some trouble with numba random using legacy numpy generator as **it is initialized at startup with entropy drawn from the operating system.**

The solution would be to switch to the new random API :

> Numba supports numpy.random.Generator() objects. As of version 0.56, users can pass individual NumPy Generator objects into Numba functions and use their methods inside the functions. The same algorithms are used as NumPy for random number generation hence maintaining parity between the random number generated using NumPy and Numba under identical arguments (also the same documentation notes as NumPy Generator methods apply). The current Numba support for Generator is not thread-safe, hence we do not recommend using Generator methods in methods with parallel execution logic.

But then we'll need to check the last part, as **Generator is not thread-safe**. I don't know what this would imply though...

MatthewMiddlehurst:
My issue is more why it is only impacting this PR though.

baraline:
I remember some bug with RDST that only appeared sporadically on macOS, maybe this is kind of the same thing? Don't have a certain answer, only that the changes in this PR have no reason to impact the failing methods ...

aryanpola:
ohh! What made the test cases pass this time?

MatthewMiddlehurst (approved):

MatthewMiddlehurst:
Fixed in a another PR, still not sure why it only appeared here though.

PR #1994, opened 2024-08-20
Added type hints for rsast and sast files

aryanpola:

Added type hints for rsast and sast files under shapelet based collection transformation.

#1910

### DIFF
--- a/aeon/transformations/collection/shapelet_based/_rsast.py
+++ b/aeon/transformations/collection/shapelet_based/_rsast.py
@@ -1,3 +1,6 @@
+from typing import List as TypingList
+from typing import Optional, Union
+
 import numpy as np
 import pandas as pd
 from numba import get_num_threads, njit, prange, set_num_threads
@@ -8,7 +11,7 @@ from aeon.utils.validation import check_n_jobs
 
 
 @njit(fastmath=False)
-def _apply_kernel(ts, arr):
+def _apply_kernel(ts: np.ndarray, arr: np.ndarray) -> float:
     d_best = np.inf  # sdist
     m = ts.shape[0]
     kernel = arr[~np.isnan(arr)]  # ignore nan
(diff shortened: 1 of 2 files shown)


aryanpola:
@baraline Can you let me know if there are changes to made?

aeon/transformations/collection/shapelet_based/_rsast.py
```diff
@@ -96,11 +99,11 @@ class RSAST(BaseCollectionTransformer):
...
-        n_random_points=10,
-        len_method="both",
-        nb_inst_per_class=10,
-        seed=None,
-        n_jobs=-1,
+        n_random_points: int = 10,
+        len_method: str = "both",
+        nb_inst_per_class: int = 10,
+        seed: int = None,
+        n_jobs: int = -1,
```
baraline, line 106: Not the object of your PR, but could you modify the default value to 1 ? 
It was decided that the default value for parellel processing should be 1 to not cause issue during CI testing.

aeon/transformations/collection/shapelet_based/_rsast.py
```diff
@@ -1,3 +1,6 @@
+from typing import List as TypingList
```
baraline, line 1: In the case where we don't use numba TypedList, I think this could be left as List instead of TypingList.

aeon/transformations/collection/shapelet_based/_sast.py
```diff
@@ -88,11 +91,11 @@ class SAST(BaseCollectionTransformer):
...
-        lengths=None,
-        stride=1,
-        nb_inst_per_class=1,
-        seed=None,
-        n_jobs=-1,
+        lengths: Optional[np.ndarray] = None,
+        stride: int = 1,
+        nb_inst_per_class: int = 1,
+        seed: Optional[int] = None,
+        n_jobs: int = -1,
```
baraline, line 98: Same comment for default value to 1

baraline (changes requested):
Looking good, thanks ! Just some minor changes and some not directly related to type hints but might as well adress them while we are at it.

baraline (approved):
LGTM, thanks again! 😄

PR #1998, opened 2024-08-21
[ENH] Adds option to use raw values instead of relative ones

dguijo:

Up to now, the function `plot_boxplot_median(...)` was used to plot the relative distribution of the results achieved by the different classifiers, using the median of the results per dataset to get these relative values. 

However, given that sometimes the distribution of the raw values is preferred, this PR modifies the function to `plot_boxplot(...)`. By default (`relative=False`), the raw values are used. When set to True (`relative=True`), the relative values are plotted. 

Example of violinplot with relative values:
image

Example of violinplot with raw values:
image

None

##### For all contributions


chrisholder:
This is a great change and very useful for clustering metrics such as AMI and ARI that can be negative values. Thanks for updating this!

baraline (approved):
LGTM, this change to the boxplot function is indeed a good thing, helps to make things clearer with some metrics.

MatthewMiddlehurst (approved):

PR #2001, opened 2024-08-21
[ENH] Add a warping path series transformer 

hadifawaz1999:
After finding a warping path from elastic measures, this transformer re-aligns a series to adjust for warping corrections.
For example, on two samples of the ECG200 dataset:
image

With a dtw path as such:
image

After transforming both series (see example in docstring), we get these two new series:
image

And this is the new dtw path between them:
image


aeon/transformations/series/_warping.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    _tags = {
+        "X_inner_type": "np.ndarray",
+        "capability:multivariate": True,
+        "fit_is_empty": True,
+    }
+
+    def __init__(
+        self, series_index: int = 0, warping_path: List[Tuple[int, int]] = None
```
TonyBagnall, line 52: I think this should be required, not clear what happens if it is not passed

aeon/transformations/series/_warping.py
```diff
@@ -0,0 +1,84 @@
...
+
+        super().__init__(axis=1)
+
+    def _transform(self, X: np.ndarray, y=None) -> np.ndarray:
+        """Transform X and return a transformed version.
+
+        Parameters
+        ----------
+        X : np.ndarray
+            time series in shape (n_channels, n_timepoints)
```
TonyBagnall, line 66: series of shape

TonyBagnall (changes requested):
not sure what this is for, but seems fine except for the warping path in the  constructor, I do not think it should default to None

aeon/transformations/series/_warping.py
hadifawaz1999: yeah  makes sense

TonyBagnall (approved):

Issue #2000, opened 2024-08-21
[ENH] Collection normaliser transformer


TonyBagnall:
### Describe the feature or idea you want to propose

when writing notebooks, it would have been helpful to have a normaliser BaseCollectionTransformer that could z-normalise, standardise or min/max scale, even as a wrapper of numpy, since it needs to be done on the correct axis etc.

### Describe your proposed solution

something like 

```python
from aeon.transformations.collection import Normalise
norm = Normalise()
xt = norm.fit_transform(x)
norm = Normalise(method="min_max")
xt = norm.fit_transform(x)
```

aryanpola:
So a new class Normalise is needed in aeon.transformations.collection?

aryanpola:
@aeon-actions-bot assign @aryanpola

MatthewMiddlehurst:
Is this different from `TimeSeriesScaler` @TonyBagnall?

TonyBagnall:
done in 2005

PR #2005, opened 2024-08-23
[ENH] Add collection transformer for normalisation

aryanpola:
I have added a new file normalise in aeon\transformations\collection as requested in issue #2000.


aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+    ----------
+    X: np.ndarray shape (n_cases, n_channels, n_timepoints)
+        The training input samples.
+    y: array-like or list, default=None
+        The class values for X. If not specified, a random sample (i.e. not of the
+        same class) will be used when computing the threshold for the Shapelet
+        Occurrence feature.
+    method : str, optional (default="z_norm")
+        The normalization method to apply.
+        The normalization method to apply.
```
hadifawaz1999, line 32: duplicate sentence in docs

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+
+    """keepdims=True: Retains the reduced axes with size one in the output,
+       preserving the number of dimensions of the array."""
+
+    def _transform(self, X, y=None):
+        if self.method == "z_norm":
+            mean = np.mean(X, axis=self.axis, keepdims=True)
+            std = np.std(X, axis=self.axis, keepdims=True)
+            return (X - mean) / std
+        elif self.method == "standardize":
```
hadifawaz1999, line 55: what is the difference between z_norm  and standarize ? they seem to have the same code unless am mistaken

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+    axis : int, optional (default=2)
+        Axis along which to apply the normalization.
+    """
+
+    def __init__(self, method="z-norm", axis=2):
+        self.method = method
+        self.axis = axis
+        super().__init__()
+
+    def _fit(self, X, y=None):
```
hadifawaz1999, line 43: no need to code fit, can add it in the tag, and the tags are missing, it should be 
_tags = {
        "X_inner_type": ["np-list", "numpy3D"],
        "fit_is_empty": True,
    }

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+from aeon.transformations.collection import BaseCollectionTransformer
+
+
+class Normalise(BaseCollectionTransformer):
+    """Normaliser transformer for collections.
+
+    Parameters
+    ----------
+    X: np.ndarray shape (n_cases, n_channels, n_timepoints)
+        The training input samples.
```
hadifawaz1999, line 25: x and y are parameters in transform not init of the class, here it should only mention "method" and "axis"

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+            std = np.std(X, axis=self.axis, keepdims=True)
+            return (X - mean) / std
+        elif self.method == "standardize":
+            mean = np.mean(X, axis=self.axis, keepdims=True)
+            std = np.std(X, axis=self.axis, keepdims=True)
+            return (X - mean) / std
+        elif self.method == "min_max":
+            min_val = np.min(X, axis=self.axis, keepdims=True)
+            max_val = np.max(X, axis=self.axis, keepdims=True)
+            return (X - min_val) / (max_val - min_val)
```
hadifawaz1999, line 62: should probably handle 0 values in denominator (case where max = min)

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,64 @@
...
+    def _fit(self, X, y=None):
+        # Fit method has no operaations on class normalise
+        pass
+
+    """keepdims=True: Retains the reduced axes with size one in the output,
+       preserving the number of dimensions of the array."""
+
+    def _transform(self, X, y=None):
+        if self.method == "z_norm":
+            mean = np.mean(X, axis=self.axis, keepdims=True)
```
hadifawaz1999, line 52: would be good to handle cases where std = 0 (stationary series) in this case we dont divide by std and only normalize on the mean

hadifawaz1999 (changes requested):
Thank you for the new transformer, just some comments that i left
would be good to add a test for it , simple one to call it on random data

aeon/transformations/collection/Normalise.py
hadifawaz1999: _tags should be added before the definition of the constructor

aeon/transformations/collection/Normalise.py
aryanpola: I noticed that but left it in as the issue stated them as two different normalisation methods so chose to keep. I'll change it if you want me to.

aeon/transformations/collection/Normalise.py
aryanpola: Should I raise an error if the min and max values are the same?

aeon/transformations/collection/Normalise.py
hadifawaz1999: @TonyBagnall did you have something in mind for difference between standard norm and znorm ? i think sklearn's standard scalar does the same as done here in both

aeon/transformations/collection/Normalise.py
hadifawaz1999: not sure, so if min=max it means the series is stationary, in that case maybe divide by 1 instead of max-min, so the transformer would return 0s

aryanpola:
@hadifawaz1999 The windows 11 python 3.12 test is passing but the windows 11 python 3.11 is failing. can you please check it.

hadifawaz1999:
> @hadifawaz1999 The windows 11 python 3.12 test is passing but the windows 11 python 3.11 is failing. can you please check it.

just re-ran it now, random fail :)

aryanpola:
> > @hadifawaz1999 The windows 11 python 3.12 test is passing but the windows 11 python 3.11 is failing. can you please check it.
> 
> just re-ran it now, random fail :)

Are there any more changes needed?

hadifawaz1999:
i would like your input @TonyBagnall  on this given you made the feature request

hadifawaz1999 (changes requested):
All changes look great thanks, could you just add the transformer to api ? then we can merge it

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,80 @@
...
+class Normalise(BaseCollectionTransformer):
+    """Normaliser transformer for collections.
+
+    Parameters
+    ----------
+    method : str, optional (default="z_norm")
+        The normalization method to apply.
+        Supported methods: "z_norm", "standardize", "min_max".
+    axis : int, optional (default=2)
+        Axis along which to apply the normalization.
```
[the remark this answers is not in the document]
aryanpola: I've taken a numpy array as an example in the test for normaliser, I'm not so sure how the effect will be if there are list of numpy arrays inputed. I'll look into it and let you know.
But the current module only takes numpy arrays as input.

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,80 @@
...
+
+Classes
+-------
+Normalise : A transformer class for normalizing collections of time series data.
+BaseCollectionTransformer : A base class template for all collection transformers.
+
+The Normalise class supports several normalization methods that can be applied to
+time series data along a specified axis. It extends the BaseCollectionTransformer
+class and provides functionality to fit and transform data, ensuring consistent
+scaling across datasets.
```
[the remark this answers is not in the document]
aryanpola: Removed them

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,80 @@
...
+        Supported methods: "z_norm", "standardize", "min_max".
+    axis : int, optional (default=2)
+        Axis along which to apply the normalization.
+    """
+
+    _tags = {
+        "X_inner_type": "numpy3D",
+        "fit_is_empty": True,
+        "capability:multivariate": True,
+    }
```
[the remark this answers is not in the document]
aryanpola: same as the above, the input is only numpy arrays.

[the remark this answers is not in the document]
aryanpola: Would padding the arrays be a good idea? This would convert the different shapes to the same ones, later concatenate the arrays to form a single array to normalise, and then get back the original array.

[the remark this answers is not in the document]
MatthewMiddlehurst: Leave it for another issue IMO. Would not pad in the class itself.

aeon/transformations/collection/Normalise.py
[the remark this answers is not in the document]
MatthewMiddlehurst: The tag means that inputs are converted to a numpy array internally, users can still input a list of numpy arrays which will then be converted.

given that we use `(n_cases, n_channels, n_timepoints)` i assume `axis=1` would normalise channels and `axis=2` would normalise the series (like expected). Maybe it does not make much sense to have `axis`? i dont really know when you would want to normalise by channel.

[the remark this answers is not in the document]
aryanpola: What would the selection of axes values be based on?
Like would the resultant numpy array being normalised in a smaller range be better?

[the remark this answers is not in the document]
aryanpola: > What would the selection of axes values be based on? Like would the resultant numpy array being normalised in a smaller range be better?

All the axes values confine the resulatant values to a specific range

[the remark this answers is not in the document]
MatthewMiddlehurst: My original request was to expand the documentation, because we accept other input types than 3d numpy arrays and it may not be obvious what the parameter is doing for those (it all gets converted to 3d numpy anyway so its the same).

But it doesnt really make sense to have an `axis` parameter anyway after more thought.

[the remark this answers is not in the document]
aryanpola: I've looked at other 3d array use cases in code and fixed `axis=-1` which normalises across all the timepoints in a 3d array.

MatthewMiddlehurst (approved):

hadifawaz1999 (changes requested):
so if am not mistaken @TonyBagnall standard would mean just zero mean and not divide by std ?

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,86 @@
+"""Normalization like z-normalization, standardization and min-max scaling."""
```
TonyBagnall, line 1: could you change the file name to _normalise please?

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,86 @@
...
+    timepoints axis. For input types that are not already 3D, the transformer handles
+    the conversion internally.
+
+    Parameters
+    ----------
+    method : str, optional (default="z_norm")
+        The normalization method to apply.
+        Supported methods: "z_norm", "standardize", "min_max".
+
+        z_norm, standardize: Subtracts the mean and divides by the standard deviation
```
TonyBagnall, line 29: to mean, standardise would just remove the mean, z_norm then also divide by the standard deviation. Either way, I think its better to have a single meaning

aeon/transformations/collection/Normalise.py
```diff
@@ -0,0 +1,80 @@
...
+            The input samples to transform.
+        y: array-like or list, optional
+            The class values for X (not used in this method).
+
+        Returns
+        -------
+        X_transformed : np.ndarray
+            The normalized data.
+        """
+        if self.method in {"z_norm", "standardize"}:
```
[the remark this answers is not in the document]
TonyBagnall: I agree, see my comment above. If you want standardise, it should simply subtract the mean

TonyBagnall (changes requested):
thanks for this, just need to either change the meaning of standardise or stick with the two allowed values, its up to you

aeon/transformations/collection/Normalise.py
aryanpola: So, would there be two different methods for z-norm and standardization?
standardization returns `X - mean` and z-norm `X - mean/std_nonzero`

aryanpola:
```
INTERNALERROR> ImportError: cannot import name 'BaseCollectionTransformer' from partially initialized module 'aeon.transformations.collection' (most likely due to a circular import) (C:\aeon\aeon\transformations\collection\__init__.py)
```
@TonyBagnall can you please help with this error.

PR #2006, opened 2024-08-23
[ENH] Added useful class attributes to SAST based transforms

### LINKED ISSUE #2004: [ENH] Shapelet transforms vary in class attributes

Overarching idea is to make the shapelet transforms implemented more similarly to one another, but for now they should be made to have  the same class attributes for example shapelet transform has useful class attributes such as (shapelet information gain, shapelet length, start position the shapelet was extracted from, shapelet dimension, index of the instance the shapelet was extracted from in fit, class value of the shapelet. These are all very helpful for shapelet interpretation but aren't always present in the other transforms.

Add missing class attributes to SAST,RSAST (RDST has already been done I think). Though these attributes may not serve any purpose in the functionality of the algorithms, hence why they probably weren't added to the class, they are useful for comparing the shapelets generated by each transform,

_No response_

### Additional context

_No response_

IRKnyazev:
#### Reference Issues/PRs
Fixes #2004 

Added class attributes found in ST that are helpful for interpreting shapelets. They do not serve functionality for the classifier but improve explainability.

Implemented:
The class each subseries is from, 
the index of the original series the subseries is extracted from, 
the start pos of the subseries in the original series.


aeon/transformations/collection/shapelet_based/_rsast.py
```diff
@@ -63,11 +64,28 @@ class RSAST(BaseCollectionTransformer):
         the number of reference time series to select per class
     seed : int, default = None
         the seed of the random generator
-    classifier : sklearn compatible classifier, default = None
+    estimator : sklearn compatible classifier, default = None
         if None, a RidgeClassifierCV(alphas=np.logspace(-3, 3, 10)) is used.
```
MatthewMiddlehurst, line 68: This does not exist in the transformer by the look of things?

aeon/classification/shapelet_based/_sast.py
```diff
@@ -178,6 +178,7 @@ def plot_most_important_feature_on_ts(self, ts, feature_importance, limit: int =
         """
         import matplotlib.pyplot as plt
 
+        feature_importance = abs(feature_importance)
```
MatthewMiddlehurst, line 181: Does this make sense to do? I'm not familiar with what metric is used here.

aeon/transformations/collection/shapelet_based/_rsast.py
```diff
@@ -63,11 +64,28 @@ class RSAST(BaseCollectionTransformer):
...
+    _start_positions : list
+        The starting positions of each subsequence within the original time series.
+    _classes : list
+        The class labels associated with each subsequence.
+    _source_series: list
+        The index of the original time series in the training set from which each
+        subsequence was derived.
+    _kernels_generators_ : dict
+        A dictionary mapping class labels to the selected reference time series
+        for that class.
```
MatthewMiddlehurst, line 87: Attributes starting with _ are not public and dont need to be included here. Ending with _ means set in fit and public.

IRKnyazev: ohh I see, thanks Matthew!

aeon/classification/shapelet_based/_sast.py
IRKnyazev: There are two mentions to making feature importance abs in the SASt paper, this is specifically needed for linear classifiers. 
image
image
I could contact Michael to confirm if you would like?

aeon/transformations/collection/shapelet_based/_rsast.py
IRKnyazev: True, the docstring must have been copied across from the RSAST classifier originally. Will remove this whole bit

aeon/classification/shapelet_based/_sast.py
IRKnyazev: @baraline do you think we can remove the plot_most_important_feature_on_ts. It seems it was implemented before your shapelet viz module, and this functionality can be achieved there now instead?

baraline: It should be. Although a quick test using both modules won't hurt to ensure that they both produce the same kind of visualization.

baraline:
Hi Ivan, while memory wise the storage of these parameters would likely not cause too much trouble, I feel like these kind of optional storage should be gated behind a boolean given during initialization (something like `store_shapelet_intialization_info` ?).

I don't know if that's the goal of the PR (if not ignore it), but preferably, all the different shapelets methods should have the same name for these arguments, so we can use a common interface while inspecting them.

IRKnyazev:
> Hi Ivan, while memory wise the storage of these parameters would likely not cause too much trouble, I feel like this kind of optional storage should be gated behind a boolean given during initialization (something like `store_shapelet_intialization_info` ?).
I don't think having these extra private variables will cause confusion for users, I  think it is more straight forward having them there for accessibility rather than providing the option for storing them. More than happy to implement your idea though if you disagree. Maybe @MatthewMiddlehurst might have a preference?


> I don't know if that's the goal of the PR (if not ignore it), but preferably, all the different shapelets methods should have the same name for these arguments, so we can use a common interface while inspecting them.

Great point, I modified the start point variable name to be consistent with the other transforms!

aeon/transformations/collection/shapelet_based/_rsast.py
```diff
@@ -187,11 +194,15 @@ def _fit(self, X: np.ndarray, y: Union[np.ndarray, list]) -> "RSAST":
...
+            # Store the original indices of the sampled time series
+            original_indices = np.where(y == c)[0]
+
+            chosen_indices = self._random_state.permutation(X_c.shape[0])[:cnt]
 
             self._kernels_generators[c] = []
 
-            for rep, idx in enumerate(choosen):
+            for rep, idx in enumerate(chosen_indices):
+                original_idx = original_indices[idx]  # Get the original index
```
MatthewMiddlehurst, line 205: dont really need this line, can just add `original_indices[idx]`

MatthewMiddlehurst (commented):
Given it's private I don't think theres a harm in adding it assuming the memory footprint is not large.

I agree we should eventually unify this between the approaches if possible. Not sure how they all do it differently, I think STC is a list of lists for each shapelet.

Happy to remove the function if it is mostly duplicated by the new visualisation tools.

TonyBagnall (approved):

PR #2010, opened 2024-08-25
[ENH] Series search for similarity search module

### LINKED ISSUE #1900: [ENH] Series search for similarity search module

Following #1243, we introduced query search, which was defined as : 

- Query search : Given a query Q and a series/collection X, evaluate the similarity between Q and each admissible candidate in X.

We now want to continue with the plan that was presented to implement series search, which was defined as  :

- Series search : Given a length parameter (for now, we add techniques that don't require it later) do a query search for all admissible queries in a series/collection X. In the naive case, this is simply a broadcasting of query search, but more optimized algorithms exists for this case (e.g. STUMP/STOMP for Euclidean distance)

Implement it in the same design as query search (parameterize output and distance used in init, as well as speed-up) and by default interface it with `query_search` when no specific speed-up for `series_search` is available.

For example, use Mueen algorithm for Euclidean distance, if STAMP is not available. 

_No response_

### Additional context

_No response_

baraline:
#### Reference Issues/PRs

Fixes #1900 

After the query search, this PR introduces the series search, which is linked to the task of computing matrix profiles. 
To make this task compatible with query search (for the naive approach), I changed the return of the predict function for query search to return both index and distances to best matches.
Modified docstrings accordingly. 

For now, only the naive version (the one looping on query search, which is equivalent to STAMP from matrix profile paper 1) and STOMP (from matrix profile paper 2) algorithm for squared and euclidean distance are implemented. I'll add the different matrix profiles (e.g. SWAMP with dtw and more efficient euclidean ones) methods in a later PR.

Minimal functionality were tested given time frame for ECML, more complete test are planned in next PR.

#### Comments

I'm still not entirely satisfied with the naming of query search / series search, so that's open to discussion

##### For all contributions


baraline:
Will update notebooks tomorrow and this will be good for review ! (and fix test that fails while I'll sleep)

baraline:
Switching this back to draft as there is huge performance issue with STOMP

aeon/utils/validation/collection.py
```diff
@@ -262,7 +264,7 @@ def get_type(X):
             raise ValueError(
                 f"ERROR np.ndarray must be 2D or 3D but found " f"{X.ndim}"
             )
-    elif isinstance(X, list):  # np-list or df-list
```
MatthewMiddlehurst, line 265: Is this necessary? Needs more discussion IMO.

aeon/utils/numba/general.py
```diff
@@ -36,7 +36,7 @@
 
 import aeon.utils.numba.stats as stats
 
-AEON_NUMBA_STD_THRESHOLD = 1e-8
+AEON_NUMBA_STD_THRESHOLD = 1e-12
```
MatthewMiddlehurst, line 39: Any reason why?

MatthewMiddlehurst (commented):
Going to approve the module changes in this given the upcoming ECML tutorial, the fact the module is experimental and the next release after this is v1.0.0 with major changes.

We should have a more thorough review prior to v1.0.0 I feel, however.

Couple of comments for other changes though.

aeon/utils/validation/collection.py
baraline: Don't remember doing this one also ... I'll check if this is necessary

aeon/utils/numba/general.py
baraline: Think it's an aterfact of when I add the issue with EEG dataset, which has a very small scale and thus naturally produces std values under 1e-8.
I'm fine with staying with 1e-8 for now, it is something that would need a more thorough inspection later as you said.

baraline:
I agree that some changes would be needed before 1.0 to make thing cleaner, multiple issues are already up for this purpose.

MatthewMiddlehurst (approved):

PR #2019, opened 2024-08-31
[ENH] add BaseSeriesTransformer - ExpSmoothingSeriesTransformer

Datadote:
#### Reference Issues/PRs
- Part of #1920

- Add smoothing filter ExpSmoothingSeriesTransformer (note abbreviated Exp)
- Tests check for different alpha, and window_size values

- No new dependencies. Imports libraries: numpy, typing

#### Any other comments?
- [Arxiv paper reference](https://arxiv.org/abs/1811.00894) has an error in the exponential smoothing equation. The decaying term should be based on s_j-1, and not t_j-1 . I'm not sure what the [Springer reference](https://doi.org/10.1007/978-3-030-29859-3_5) has. I do not have access
Paper:  s_0 = t_0 and s_j = alpha * t_j + (1 - alpha) * t_j-1
Corrected: s_0 = t_0 and s_j = alpha * t_j + (1 - alpha) * s_j-1
- Allows user to use "alpha" or "window_size". If "window_size" is specified, alpha is overwritten with 2. / (window_size + 1)
- For the reference paper link, I added an arxiv link to the paper for folks without access to Springer

### DIFF
--- a/aeon/transformations/series/_exp_smoothing.py
+++ b/aeon/transformations/series/_exp_smoothing.py
@@ -13,6 +13,13 @@ from aeon.transformations.series.base import BaseSeriesTransformer
 class ExpSmoothingSeriesTransformer(BaseSeriesTransformer):
     """Filter a time series using exponential smoothing.
 
+    - Exponential smoothing (EXP) is a generalisaton of moving average smoothing that
+    assigns a decaying weight to each element rather than averaging over a window.
+    - Assume time series T = [t_0, ..., t_j], and smoothed values S = [s_0, ..., s_j]
+    - Then, s_0 = t_0 and s_j = alpha * t_j + (1 - alpha) * s_j-1
+    where 0 ≤ alpha ≤ 1. If window_size is given, alpha is overwritten, and set as
+    alpha = 2. / (window_size + 1)
+
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


aeon/transformations/series/_exp_smoothing.py
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    def __init__(
+        self, alpha: float = 0.2, window_size: Union[int, float, None] = None
+    ) -> None:
+        if not 0 <= alpha <= 1:
+            raise ValueError(f"alpha must be in range [0, 1], got {alpha}")
+        if window_size is not None and window_size <= 0:
+            raise ValueError(f"window_size must be > 0, got {window_size}")
+        super().__init__(axis=0)
```
TonyBagnall, line 70: I think this should be axis = 1, i.e. (n_channels, n_timepoints)

TonyBagnall (changes requested):
good spot on the typo, but I would prefer this to be axis = 1 as it is our standard

aeon/transformations/series/_exp_smoothing.py
Datadote: Ok, I've made axis=1, and updated the code to adjust for the input shape assumption

PR #2026, opened 2024-09-02
[ENH] extending PR #2017 to multiclass case

IRKnyazev:
Reference Issues/PRs
See also https://github.com/aeon-toolkit/aeon/issues/2016

This is extending the previous PR which missed out the multiclass case for linear classifiers, here an distance is made inversely correlated to distance and for RDST the other two features are kept positively correlated.

Any other comments?
@baraline noticed that I missed the multiclass correction and explained that coefs are one vs all for this usecase so just need to grab the class coefs and manipulate them appropriately.
This is a fresh PR resolving merge conflicts which were present in PR 2018.


aeon/visualisation/estimator/_shapelets.py
```diff
@@ -698,7 +698,6 @@ def _get_shp_importance(self, class_id):
                     class_0_coefs[:, mask] = -class_0_coefs[:, mask]
                     class_1_coefs[:, ::3] = -class_1_coefs[:, ::3]
```
baraline, line 699: Would you kindly remind me why we negate min and argmin for class 0 coefs here? Might be a bit out of loop

baraline (commented):
Quick question and we should be good to go !

aeon/visualisation/estimator/_shapelets.py
IRKnyazev: When n_classes == 1 there are two classes, linear classifiers assign positive coefficients for features correlated to class 1, so when getting the importance for class 0 we need to flip these values. It follows the same original logic implemented originally where you did coefs = -coefs :)

baraline (approved):
Thanks, and indeed the linear case is a bit weird to make out for RDST, as we're trying to explain what would likely be a non-linear behaviour with linear coefs ... This is likely why on some datasets, using non-linear classifiers with RDST gives way more accuracy than the linear one (and vice versa), guess an option would be to ensemble it ;p

Issue #2021, opened 2024-09-01
[ENH] Rework distance/matrix profile import system


baraline:
### Describe the issue

Currently, when selecting between distance or matrix profile options in `QuerySearch` or `SeriesSearch` a dict is used with as values the functions to use. This requires to import all of them at the start of the module, while only one will be used. 

### Suggest a potential alternative/fix

Make a function that only imports the necessary method at runtime.

aryanpola:
@baraline Lazy loading(dynamic importing) would be the way to solve the issue?

aryanpola:
Also, how do I classify a certain module as heavy?

baraline:
The idea would simply to import the distance profile / matrix profile functions matching the distance and speed ups paramater given during initialisation

PR #2049, opened 2024-09-11
Lazy Importing for Distance_matrix profile

aryanpola:
Implementing Lazy loading which imports packages during runtime when called in a specific function in Dictionary-based modules.
#2021


aryanpola:
@baraline I've implemented lazy loading and pre-commit passes locally but not in github actions.

baraline:
Hi, the original issue is not about dictionnary methods, I'm not sure what was your goal here ?

aryanpola:
> Hi, the original issue is not about dictionnary methods, I'm not sure what was your goal here ?

Ohh! My bad, I think it should be implemented under similarity search.

baraline:
I see where the confusion might have come from, edited the original issue

aryanpola:
I'll create a new PR for this. Deleting this one.

Issue #1991, opened 2024-08-19
[MNT] New distance compatibility tests timeout with numba disabled


MatthewMiddlehurst:
### Describe the issue

See https://github.com/aeon-toolkit/aeon/actions/runs/10445759452/job/28922165687.

These tests take too long with numba disabled. We have an extended timeout duration for these tests, but it still reaches it.

### Suggest a potential alternative/fix

Increase the efficiency of these tests and/or split them up. If not possible, disable them if numba is disabled.

TonyBagnall:
=========================== short test summary info ============================
FAILED aeon/distances/tests/test_sklearn_compatibility.py::test_multivariate[task0-3-dist14] - Failed: Timeout >1800.0s
FAILED aeon/regression/distance_based/tests/test_time_series_neighbors.py::test_knn_neighbors[twe] - Failed: Timeout >1800.0s

Additional context

MatthewMiddlehurst:
@TonyBagnall was this fixed?

TonyBagnall:
this is I believe fixed, reopenm a new issue if not

PR #2052, opened 2024-09-12
[ENH] Speed up the regression distances tests

### LINKED ISSUE #2051: [MNT] Overnight tests failing
### Describe the issue

overnights are failing due to slow tests for shape_dtw and occasionally TWE exceeding  the 1800 secs limit when numba is turned off. 

We should either reduce the data test size or even exclude them from the non numba tests. The time limit is  30 mins, we will get to the point the overnight tests take more than the night to run!

FAILED aeon/distances/tests/test_sklearn_compatibility.py::test_multivariate[task0-3-dist14] - Failed: Timeout >1800.0s
FAILED aeon/regression/distance_based/tests/test_time_series_neighbors.py::test_knn_neighbors[twe] - Failed: Timeout >1800.0s

### Additional context

_No response_

TonyBagnall:
hopefully fixes #2051 by sorting out sources of timeout when numba is off by replacing baked in data with random data

Edit, just seen #1991


dguijo (commented):
LGTM. Just one comment. Should we consider also modifying the tests that use the unit test and the Covid3M datasets as well? I think they are only 1 in test_sklearn_compatibility.py

Thanks, Tony!

TonyBagnall:
> LGTM. Just one comment. Should we consider also modifying the tests that use the unit test and the Covid3M datasets as well? I think they are only 1 in test_sklearn_compatibility.py
> 
> Thanks, Tony!

yes we should, I prefer random to loaded, but my priority atm is to get the overnights passing. I'll take a look though, should be quick

dguijo (approved):
Great! Thank you Tony! Good work!! Can merge when it passes the tests though

PR #2062, opened 2024-09-16
[ENH] Add BORF transformer

fspinna:
I am trying to add a Bag-Of-Receptive-Fields transformer as seen in `Spinnato, Francesco, et al. "A Bag of Receptive Fields for Time Series Extrinsic Predictions." arXiv preprint arXiv:2311.18029 (2023)`. It is basically a Bag-Of-Pattern transform, where a time series is transformed in a sparse matrix of counts of discretized subsequences.

This contribution introduces a dependency to the pydata sparse library (https://sparse.pydata.org/en/stable/). 

#### Any other comments?

I have a couple of issues/doubts:
1. the transformer fails 2 tests: 
```
'test_persistence_via_pickle(estimator=BORF(),datatype=EqualLengthUnivariate-Classification-numpy3D)': "FAILED: 'csr_matrix' object has no attribute 'ravel'",
'check_fit_deterministic(estimator=BORF(),datatype=EqualLengthUnivariate-Classification-numpy3D)': "FAILED: 'csr_matrix' object has no attribute 'ravel'"
```
This is because the output of the transformer is a sparse scipy matrix (https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html#scipy.sparse.csr_matrix). I don't know if this is a problem. In practice, the sparse matrix is compatible with downstream sklearn estimators/transformers.

2. the sparse package is used as a soft dependency, which is imported in the transform of `IndividualBORF`, which inherits from sklearn's BaseEstimator, TransformerMixin, and not BaseCollectionTransformer. Is this a problem?

##### For all contributions


TonyBagnall:
great, thanks for that

aeon/transformations/collection/dictionary_based/_borf.py
```diff
@@ -0,0 +1,1145 @@
...
+
+    Transforms time series data into a bag-of-receptive-fields representation [1] using
+    SAX. This transformer extracts SAX words from time series by applying sliding
+    windows, discretizing, and counting the occurrences of the SAX words in the
+    time series. The output is a sparse feature representation suitable for
+    various downstream tasks such as classification and regression.
+
+    Parameters
+    ----------
+    window_size_min : int, optional (default=4)
```
TonyBagnall, line 35: thanks for the PR, we will look into the test fails, but just on style, sorry to be annoying but we use
window_size_min : int, default=4
as our style. Could you use that?

fspinna: No problem, style is important! Let me know if it's okay now and if there are any other issues.

TonyBagnall:
hi @fspinna sorry for the delay on this, we will definitely review this week. I'm afraid there are a couple of conflicts which should be easy to resolve

TonyBagnall (commented):
thanks again, but my this is huge! A couple of questions
1. You use the sparse package. Elsewhere we use from scipy.sparse import csr_matrix. scipy is a core dependency, so it would be preferable to use this if it is fit for purpose.
2. It would also be good to have a constructor option to return the whole sparse array rather than the compressed version
3. there are two arguments for n_jobs
            n_jobs=self.n_jobs,
            n_jobs_numba=self.n_jobs_numba,
how come?
4. You have reimplemented SAX. We have a version of SAX already, would it be possible to use that? If not, see 5 below.
5. There are a lot of specific public functions which will bloat the docs, could you make the file specific functions private? things like `are_window_size_and_dilation_compatible_with_signal_length` Same with all the small sklearn transformers, better off private imo. 
6. You have implemented a hash table?

patrickzib:
> One of the contributions of the paper was a better/faster way of computing window-wise SAX. Also, this version has dilation, stride and other parameters that are not usually in SAX. So I don't believe I can use your implementation.

It might be a bit ambitious for this PR, but it would be wonderful to include a fast sliding window SAX-transform with dilation if possible.

TonyBagnall:
do we need a lower bound on the sparse package?

TonyBagnall:
re SAX, we could investigate using this version as our standard and extending it as you say @patrickzib but not in this PR :) IT would be good to make an issue on improving SAX

aeon/transformations/collection/dictionary_based/_borf.py
```diff
@@ -0,0 +1,1189 @@
...
+            dilation,
+            word_length,
+            alphabet_mean,
+            alphabet_slope,
+        ) = _extract_parameters_from_args(parameter)
+        if word_length > window_size:  # word_length cannot be greater than window_size
+            continue
+        if (
+            alphabet_slope <= 1 and word_length == 1
+        ):  # if alphabet_slope <= 1, word_length=1 is useless
```
chrisholder, line 1013: Does this comment need to exist still?

aeon/transformations/collection/dictionary_based/_borf.py
```diff
@@ -0,0 +1,1189 @@
...
+            )
+        )
+    return cleaned_parameters
+
+
+@nb.njit(cache=True)
+def _is_valid_windowing(sequence_size: int, window_size: int, dilation: int) -> bool:
+    if (
+        sequence_size < window_size * dilation
+    ):  # if window_size * dilation exceeds the length of the sequence
```
chrisholder, line 1032: Does this comment need to exist still?

aeon/transformations/collection/dictionary_based/_borf.py
```diff
@@ -0,0 +1,1189 @@
...
+        dilation = parameter["dilation"]
+        word_length = parameter["word_length"]
+        alphabet_mean = parameter["alphabet_size_mean"]
+        alphabet_slope = parameter["alphabet_size_slope"]
+        stride = parameter["stride"]
+        if not _is_valid_windowing(
+            window_size=window_size, sequence_size=max_length, dilation=dilation
+        ):
+            continue
+        # if window_size == word_length:
```
chrisholder, line 1051: Does this comment need to exist still?

chrisholder (changes requested):

fspinna:
> It might be a bit ambitious for this PR, but it would be wonderful to include a fast sliding window SAX-transform with dilation if possible.

> re SAX, we could investigate using this version as our standard and extending it as you say @patrickzib but not in this PR :) IT would be good to make an issue on improving SAX

Also, it's not a perfect replacement for standard SAX, as the window size has to be divisible by the word length in my case.

aeon/transformations/collection/dictionary_based/_borf.py
fspinna: It can be removed, but it's more of an explanatory comment. Should I remove these kinds of comments in general?

aeon/transformations/collection/dictionary_based/_borf.py
fspinna: No, I'll remove it.

TonyBagnall:
> > It might be a bit ambitious for this PR, but it would be wonderful to include a fast sliding window SAX-transform with dilation if possible.
> 
> > re SAX, we could investigate using this version as our standard and extending it as you say @patrickzib but not in this PR :) IT would be good to make an issue on improving SAX
> 
> Also, it's not a perfect replacement for standard SAX, as the window size has to be divisible by the word length in my case.

I like the idea of having a speed optimised constrained version and a more general case one, but its a project in its own right really. This looks ready to go in to me, thanks for the contribution. There are a couple of minor conflicts to resolve caused by changes elsewhere, sorry about that

TonyBagnall:
could we add your paper to this? https://www.aeon-toolkit.org/en/stable/papers_using_aeon.html

Even better, can you add your paper to this (in a separate PR)? :)

fspinna:
> do we need a lower bound on the sparse package?

I am using the latest (0.15.4)


> could we add your paper to this? https://www.aeon-toolkit.org/en/stable/papers_using_aeon.html
> 
> Even better, can you add your paper to this (in a separate PR)? :)

Sure!


> I like the idea of having a speed optimised constrained version and a more general case one, but its a project in its own right really. This looks ready to go in to me, thanks for the contribution. There are a couple of minor conflicts to resolve caused by changes elsewhere, sorry about that

Thank you! No problem, let me know if there is anything else to do!

aeon/transformations/collection/dictionary_based/_borf.py
TonyBagnall: its fine to leave. Generally we have gone for longer docstrings and fewer in line comments, but I think I actually prefer a few more inline comments for more complex constructs to aid readability

TonyBagnall:
hi @fspinna if you resolve the conflicts I think this is ready to go in

fspinna (commented):
Fixed unused comments

fspinna:
> hi @fspinna if you resolve the conflicts I think this is ready to go in

@TonyBagnall I fixed a few things (tags, doctest) but I am still failing on a few OS. Not really sure what's the issue here.

MatthewMiddlehurst:
Hi @fspinna, not too up to date with this PR but its only failing on some workflows because we split the tests up when running on PRs to save time.

I can take a look at whats failing at some point if no one else finds the time.

MatthewMiddlehurst:
Alternatively, there is a `non-deteminsitic` and `cant-pickle` tag which will skip these particular tests, but we try to only use them if it is not an easily fixable issue.

Adding it temporarily to the skipped test list is also an option, and revisiting in another PR.

TonyBagnall (approved):
LGTM, thanks for the contribution

PR #2072, opened 2024-09-18
[ENH] Make transformer files private

### LINKED ISSUE #2068: [ENH] Make all collection and series transformer files private, access through init

currently like this
image
should be like series transformers 
image

make files private, import through init

TonyBagnall:
#### Reference Issues/PRs
fixes #2068 

- Makes remaining collection transformers private files
- Removes the term Panel from all docstrings, replaced by Collection
- Removes redundant code and docstrings from BaseTransformer
- adds some tests for coverage for incorrect input of parameters


MatthewMiddlehurst (commented):
Fine with the refactor. Not really a fan of `test_wrong_paras.py`, too general. Think it would be better as class/file specific test files.

TonyBagnall:
> Fine with the refactor. Not really a fan of `test_wrong_paras.py`, too general. Think it would be better as class/file specific test files.

will result in a lot of very small test files, but ok will split it

aeon/transformations/collection/base.py
```diff
@@ -376,21 +373,3 @@ def _update(self, X, y=None):
...
-            )
-        if isinstance(y, np.ndarray) and y.ndim > 1:
-            raise TypeError(f"y must be 1-dimensional, found {y.ndim} dimensions")
-        # Check matching number of labels
-        n_labels = y.shape[0]
-        if n_cases != n_labels:
-            raise ValueError(
-                f"Mismatch in number of cases. Number in X = {n_cases} nos in y = "
-                f"{n_labels}"
-            )
```
MatthewMiddlehurst, line 396: Why?

aeon/transformations/collection/_segment.py
```diff
@@ -60,10 +60,6 @@ class IntervalSegmenter(BaseCollectionTransformer):
         and the second column giving end points of intervals
     """
 
-    _tags = {
-        "capability:unequal_length:removes": True,
-    }
```
MatthewMiddlehurst, line 65: does it not do this?

aeon/transformations/base.py
```diff
@@ -514,8 +306,9 @@ def _inverse_transform(self, X, y=None):
         X: data structure of type X_inner_type
             if X_inner_type is list, _inverse_transform must support all types in it
             Data to be transformed
-        y : Series or Panel, default=None
-            Additional data, e.g., labels for transformation
+        y : Series, default=None
+            Additional data, e.g., labels for transformation.
+        `
```
MatthewMiddlehurst, line 311: typo

aeon/transformations/collection/base.py
TonyBagnall: it is not called anywhere and is not tested

aeon/transformations/collection/_segment.py
TonyBagnall: this tag is not used anywhere and seems unnecessary to me, we can keep it if you want

TonyBagnall: put back in

aeon/transformations/collection/base.py
TonyBagnall: put it back in, needs testing though

aeon/transformations/collection/_segment.py
MatthewMiddlehurst: Used in pipelines. Would leave removing for another PR

aeon/transformations/collection/base.py
MatthewMiddlehurst: Thats pretty weird, kind of implies that we arent testing `y` for these transformers.

MatthewMiddlehurst (approved):

PR #2037, opened 2024-09-07
[ENH] KILL WITH FIRE

TonyBagnall:
#### Reference Issues/PRs

closes #37 #42 #99 #136 #240 #265 #291 #337 #341 #1388 #1491 #1629 #1446 #1156 #1734 #217 #2015

probably #599 too

#### What does this implement/fix? Explain your changes.

This is the big refactor/deprecation prior to 1.0.0, aka the project internally referred to as KILL WITH FIRE.

Removes forecasting module and datatypes module and all accompanying tests/references etc. Removes operator overloading, so called vectorisation and all output conversion from  BaseTransformer

*To Note*

1. BaseTransformer is linked into the interval trees, needs untangling to remove BaseTransformer completely, although we might want to keep it for tags

2. The registry will need some refactoring after, I'm just removing tests for now, could be the whole thing may be reworked. I think we should revisit tags

PR #1877, opened 2024-07-31
[MNT] Unit testing revamp part 4: clustering

MatthewMiddlehurst:
Follows on from #1875, part of [AEP 05](https://github.com/aeon-toolkit/aeon-admin/pull/10/files).

- Removes the current generic clustering testing and deep learning clusterer tests.
- Replaces them with a yield function in the new testing framework.

PR #2076, opened 2024-09-19
[MNT] Remove legacy testing

MatthewMiddlehurst:
Follows on from #1877 and #2037, part of [AEP 05](https://github.com/aeon-toolkit/aeon-admin/pull/10/files).

- Removes the old class based testing framework and scenario files
- Removes the current `test_all_*` files for anomaly detection, segmentation, transformers and early classification.
- Replaces them with a yield function in the new testing framework.
- Renames `test_config.py` to `testing_config.py` because `test_` files generally contain test functions which the config does not.
- Moves some of the soft dependency tests into the new framework. Non-estimator tests remain.
- Adds a bunch of new classifiers to the exclusion list.


aeon/testing/estimator_checking/_yield_anomaly_detection_checks.py
CodeLionX, line 1: Should this file not be part of the `aeon.anomaly_detection.tests` module? The base class is also defined in `aeon.anomaly_detection`.

I would assume that the actual tests reside in their respective modules and the `aeon.testing`-module "just" contains the test setup, helper methods, fixtures etc.

---

Same for all the other estimator checks

CodeLionX: Explanation: I usually just run `pytest aeon/<module>/tests` locally when working on a specific estimator or module. I'm not concerned about the tests in other modules because I won't change something there.

TonyBagnall: I think the general testing is to test base class compliance, and I think this high level testing is best centralised so that a common structure is easier to enforce and follow. Tests of specific algorithms, expected output, exception raising etc, is best done within the module

CodeLionX: Ok, I understand the reasoning. Do we have a system to selectively execute those centralized tests, then?

Example scenario: I would like to execute all the tests in `aeon.testing` just for the `aeon.anomaly.detection.DTW_MLEAD` estimator when working on this estimator (short feedback loops are crucial for my dev experience!).

MatthewMiddlehurst: `pytest -k DTW_MLEAD` or pass the class to `check_estimator`

MatthewMiddlehurst: It should be pretty simple to add a function to test a whole base class as well i.e. `BaseAnomalyDetector`

TonyBagnall (approved):
LGTM

aeon/testing/estimator_checking/_yield_anomaly_detection_checks.py
CodeLionX: Nice, let's put this in the testing documentation 👍🏼

CodeLionX (approved):
I did not go through the individual changes or checked if the checks all make sense, but the new structure is good and makes sense.

PR #2087, opened 2024-09-24
[ENH] Adding LITEMVTime deep learning classifier and rgeressor for multivariate time series

hadifawaz1999:
Recently LITE was adapted to multivariate data, LITE works with multivariate but LITEMV is specific for multivariate (so using it on unvariate makes no sense)


hadifawaz1999:
Replaced by #2255

PR #2091, opened 2024-09-24
[ENH] Issue 1641 - Matrix profile-based anomaly detectors: left STAMPi

### LINKED ISSUE #1641: [ENH] Matrix profile-based anomaly detectors: left STAMPi

The matrix profile can be computed incrementally, using e.g. STAMPi. If only the distances to all already observed windows are computed, the resulting data structure is called left matrix profile. This allows us to detect reoccurring patterns as anomalies that would otherwise not be detected because of the twin-freak problem. Add such an algorithm to aeon.

References:
- **STAMP** and **STAMPi** are described in: Chin-Chia Michael Yeh, Yan Zhu, Liudmila Ulanova, Nurjahan Begum, Yifei Ding, Hoang Anh Dau, Diego Furtado Silva, Abdullah Mueen, and Eamonn Keogh. 2016. Matrix Profile I: All Pairs Similarity Joins for Time Series: A Unifying View That Includes Motifs, Discords and Shapelets. In Proceedings of the International Conference on Data Mining (ICDM), 1317–1322. doi: 10.1109/ICDM.2016.0179.

- TimeEval implementation: https://github.com/TimeEval/TimeEval-algorithms/blob/main/left_stampi/
- tags: unsupervised, univariate
- compute left matrix profile of the input time series using stumpy as described [in their docs](https://stumpy.readthedocs.io/en/latest/Tutorial_Matrix_Profiles_For_Streaming_Data.html#bonus-section-never-update-history)
- convert left matrix profile (window-based anomaly score) to point-based anomaly score using [aeon.utils.reverse_windowing](https://github.com/aeon-toolkit/aeon/blob/63de116995a4c495aa51952ec3e53cbf62effdfd/aeon/utils/windowing.py#L118) and window_size=m

_No response_

### Additional context

_No response_

ferewi:
#### Reference Issues/PRs
Fixes #1641 

This PR implements the LeftSTAMPi Anomaly Detector based on the implementation in TimeEval (https://github.com/TimeEval/TimeEval-algorithms/blob/main/left_stampi/algorithm.py)

The Algorithm can be run in two modes.
1. Batch mode: Here the whole time series is put in at once. Internally the LeftSTAMPi algorithm is applied incrementally.
2. Stream mode: Here the Algorithm is initialized on a specified number of data points. The Matrix profile is then calculated incrementally. 

Remarks:
The batch mode is implemented in the `fit_predict` method. 
The stream mode is implemented in the `fit` and `predict`methods, where `fit`is used to init the algorithm and `predict` is used for imcremental updates. As the `predict`method only accepts `np.ndarray` as its argument, on every update the new data point, which is a scalar has to be wrapped in a one-element numpy array.  This is actually unnecessarry but avoiding this would involve a change to the interface in `BaseAnomalyDetector`.

No.

#### Any other comments?
See remarks in the implementation description.

##### For all contributions
  @all-contributors please add @ferewi for code, doc and test


aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+from aeon.utils.windowing import reverse_windowing
+
+
+class LeftSTAMPi(BaseAnomalyDetector):
+    """LeftSTAMPi anomaly detector.
+
+    LeftSTAMPi [1]_ calculates the left matrix profile of a time series,
+    which is the distance to the nearest neighbor of each subsequence
+    in the time series, in an incremental manner. The matrix profile is then
```
CodeLionX, line 19: More precise: The distance to the nearest neighbor of already observed subsequences (all preceding subsequences)

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+         - unsupervised
+
+
+    Parameters
+    ----------
+    window_size : int
+        Size of the sliding window.
+    n_init_train: int, defaul=None
+        The number of points used to init the matrix profile.
+        The discord will not be found in this segment.
```
CodeLionX, line 42: Only used in batch-case, right?

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __init__(
+        self,
+        window_size: int = 3,
+        n_init_train: int = None,
+        normalize: bool = True,
+        p: float = 2.0,
+        k: int = 1,
+    ):
+        self.mp: np.ndarray | None = None
```
CodeLionX, line 106: The `self.mp`-member is not an algorithm parameter and set during `fit`, so must be named `mp_` (sklearn and aeon convention).

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+
+        if self.k < 1 or self.k > X.shape[0] - self.window_size:
+            raise ValueError(
+                "The top `k` distances must be at least 1 and at most the length of "
+                "the time series minus the window size."
+            )
+
+    def _fit(self, X, y=None):
+        self.n_init_train = len(X)
```
CodeLionX, line 134: sklearn forbids overwriting algorithm parameters. You can either use an internal variable (`self._n_init_train`) or a local one.

CodeLionX (changes requested):
Thank you for your contribution!

I'm in favor of supporting streaming use cases, and I also like the idea of repeatedly calling `predict`. However, this violates the convention that no internal representation is updated in `predict`. We can neither use repeated calls to `fit` because it is assumed to reset the estimator on the beginning of each call.

I guess, we need to design a new API for streaming. @MatthewMiddlehurst how do you think about this?

Until we have decided on the new API, I would suggest adding leftSTAMPi only with its batch API.

ferewi:
@CodeLionX Thanks for your comments and suggestions. The failing test-suite made me aware that the approach I took for the streaming case is violating the concept of `fit`, `predict`and `fit_predict`. I'll remove the streaming case for now. An idea for the streaming API might be to have an `update` method that is allowed to modify the internal representation after the intitial fitting. 

Also, could you point me to the documentation of the sklearn and aeon conventions regarding the naming conventions (`self.mp_`vs `self.mp`), etc.? (If you have that at hand - I surely can google that myself)

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+        self._check_params(X)
+
+        if X.ndim > 1:
+            X = X.squeeze()
```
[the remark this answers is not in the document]
ferewi: Yes, I testesd this. When I pass an univariate time series of length 20 as an array of shape (20, 1), the X.squeeze() is requireed. 
Is there a better way to ave this handled by the base class?

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,184 @@
...
+        self._call_stumpi(X)
+
+        return self
+
+    def _predict(self, X: np.ndarray) -> np.ndarray:
+        if X.ndim > 1:
+            X = X.squeeze()
```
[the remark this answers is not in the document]
ferewi: please see above

aeon/anomaly_detection/tests/test_left_stampi.py
```diff
@@ -0,0 +1,360 @@
...
+
+from aeon.anomaly_detection._left_stampi import LeftSTAMPi
+from aeon.testing.data_generation import make_example_1d_numpy
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies("stumpy", severity="none"),
+    reason="required soft dependency stumpy not available",
+)
```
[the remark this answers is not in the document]
ferewi: That's right. But the library has to be installed to be mocked. Is there a standard way to avoid this?

aeon/anomaly_detection/_left_stampi.py
ferewi: I amended the sentence.

aeon/anomaly_detection/_left_stampi.py
ferewi: That's right. As there is nothing but the batch mode now, I did not add this comment. To be added when/if the streaming case gets implemented.

aeon/anomaly_detection/_left_stampi.py
ferewi: Thankss for pointing that out. I renamed the variable

aeon/anomaly_detection/_left_stampi.py
ferewi: This has been removed.

ferewi:
Still making changes to fix the failling tests. I will re-request a new review when I am done.

aeon/anomaly_detection/tests/test_left_stampi.py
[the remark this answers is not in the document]
CodeLionX: Ah, true ... you could mock/overwrite the `_call_stumpi`-method and fill the `self._mp`-member with your mocked data structures.

CodeLionX:
> Also, could you point me to the documentation of the sklearn and aeon conventions regarding the naming conventions (`self.mp_`vs `self.mp`), etc.? (If you have that at hand - I surely can google that myself)

It is sprinkled in this guide: https://scikit-learn.org/dev/developers/develop.html

MatthewMiddlehurst:
Thanks for the contribution. Feel free to ask if you have any questions regarding the failures, i.e. we have a tag for estimators which can't be pickled (usually due to dependencies outside of our control).

MatthewMiddlehurst:
Think adding to the base API should be a separate PR yeah, would need to see a proposal but if `update` or a similar method would work that sounds fine.

CodeLionX:
> Think adding to the base API should be a separate PR yeah, would need to see a proposal but if `update` or a similar method would work that sounds fine.

Another question is whether we actually want to introduce streaming algorithms in aeon. Are there already other streaming/online estimators? The current architecture is tailored to the batch-case. Maybe this is a topic for the next dev-meeting.

aeon/anomaly_detection/tests/test_left_stampi.py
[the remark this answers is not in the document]
ferewi: I found a way of mocking the entire stumpy package that works even if the package is not installed

ferewi:
Then maybe you discuss this in your next dev meeting and if you decide that you want to integrat a streaming api, I am happy to put in a proposal and another PR to implement this for the LeftSTUMPi algorithm.

ferewi:
> Thanks for the contribution. Feel free to ask if you have any questions regarding the failures, i.e. we have a tag for estimators which can't be pickled (usually due to dependencies outside of our control).

I guess I figured it out. I tagged the class with 'cant-pickle' and the tests are green now.

aeon/anomaly_detection/_left_stampi.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.fit(X[: self.n_init_train])
+
+        return self.predict(X[self.n_init_train :])
+
+    def _call_stumpi(self, X):
+        if not self._is_stumpy_loaded:
+            import stumpy
+
```
CodeLionX, line 156: Please add a call to `_check_soft_dependencies("stumpy", severity="error")` before importing to produce the correct error message when `stumpy` is not available.

CodeLionX (changes requested):
Otherwise, looks good 👍🏼

aeon/anomaly_detection/_left_stampi.py
ferewi: As I understood it, the availability of stumpy is already checked by the parent `BaseSeriesEstimator` constructor by calling `_check_estimator_deps(self)` - the package is specified as python dependency in `tags`.

This was meant to prevent importing the module twice. Happy to remove it though.

CodeLionX: Sorry, you are right `BaseSeriesEstimator` does this automatically 👍🏼 

The Python import-system, however, already [caches imported modules (in `sys.modules`)](https://docs.python.org/3/reference/import.html#the-module-cache) and guarantees that every module is imported just once. I should be safe to just import stumpy multiple times.

ferewi: Well, that's true. I removed that piece.

CodeLionX (approved):

ferewi:
@CodeLionX How is the process after the PR is approved? Do I have to do something or is this simply integrated at some point?

MatthewMiddlehurst:
Don't have to do anything, can be merged at any point really. Usually give it some time for other comments though. Not really a massive rush until it's close to release time 🙂

CodeLionX:
We will wait a bit, so that other maintainers / core decs get the chance to object; otherwise, I'll merge it in later.

CodeLionX:
@ferewi regarding the online-API, we decided in our dev-meeting that aeon will not support this in the near future. If you have a use case for streaming/online anomaly detection, we can, of course, talk about this again.

TonyBagnall:
Fantastic, thanks for this

ferewi:
Cool - I was just interested in how the usual process is :) 

Thank you for your help @CodeLionX @MatthewMiddlehurst :)

PR #2092, opened 2024-09-24
[MNT] Add typecheck workflow using `mypy`

### LINKED ISSUE #1856: [MNT] Add `pre-commit` hook for `mypy`
### Describe the issue

From July 26 2024 meeting. 

There are a few type hint errors, and no method to easily discover these. We should include some automated type checking for type hints in `pre-commit`. 

The first package that comes to mind for this is `mypy`, but it does not necessarily have to be this.

i.e.

```
- repo: https://github.com/pre-commit/mirrors-mypy
   rev: v1.0.1
   hooks:
     - id: mypy
       files: aeon/
       additional_dependencies: [ pytest ]
```

This should be set up to only be run manually for now, however. There are a lot of breakages that we do not want to do at once.

### Additional context

_No response_

MatthewMiddlehurst:
Closes #1856

Adds `mypy` as an optional workflow which can be run on PRs. The same can be run locally using `mypy aeon/` after installing, as it uses the default set up currently.

I opted to avoid using `pre-commit` for this, as it is apparently prone to issues and should really be run on the whole repo instead of only changed files. (see https://jaredkhan.com/blog/mypy-pre-commit#step-2-creating-our-own-pre-commit-hook and https://github.com/python/mypy/issues/13916)


baraline (approved):
Didn't know about the tool but itdm definitly looks like the right one for the job. Test errors seems to be due to the hpy5 issue, so LGTM!

Issue #1913, opened 2024-08-05
[BUG] Matplotlib upper bound


hadifawaz1999:
### Describe the bug

Matplotlib presented some issues with latest releases `3.9.1` and `3.9.2`, see [here](https://github.com/matplotlib/matplotlib/issues/28551)
For now #1906 fixed it by adding an upper bound, we should wait till they sort out this issue and then remove upper bound

### Expected results

CI does not break

### Actual results

CI is breaking

QuLogic:
3.9.1 was yanked, so you can remove the pin already. There is no 3.9.2 yet, but it will be out soon.

hadifawaz1999:
we still have the bound in pyproject.toml so still not fixed @TonyBagnall , @MatthewMiddlehurst any update on this ?

PR #2094, opened 2024-09-25
[MNT] Remove matplotlib upper boundAjb/pin

TonyBagnall:
in relation to #1913, see if things pass if we remove upper bound

(I must reinstall to remove this trail of old changes!)


MatthewMiddlehurst:
part of #2079

PR #2096, opened 2024-09-25
[ENH] Remove segmenter transformers, tidy up API

### LINKED ISSUE #2074: [DOC] Tidy up Transformers API

after #2068, the transformers API needs to be reorganised

_No response_

TonyBagnall:
this removes three very old transformers
IntervalSegmenter
RandomIntervalSegmenter
SlidingWindowSegmenter

these were used internally for an implementation of TimeSeriesForest that was removed a long time ago. They are based around nested pandas, are poorly documented and have no real use in aeon. Also removed a aeon.utils.datetime as no longer needed.

Also tidies up the transformers API. We can add more information to this later, but it was a bit of a mess so this is the simple, minimal version that lists all collection and series transformers, fixes #2074


aeon/transformations/series/_siv.py
```diff
@@ -18,11 +18,12 @@ class SIVSeriesTransformer(BaseSeriesTransformer):
     window_length : list of int or int, default=[3, 5, 7]
         The filter windows lenths (recommended increasing value).
 
+    implementation of a median filter and therefore runs much faster.
+
```
MatthewMiddlehurst, line 22: accidental change?

TonyBagnall: good spot

MatthewMiddlehurst (approved):

Issue #1714, opened 2024-06-21
[ENH] Add Weight Normalization layer using Tensorflow


aadya940:
### Describe the feature or idea you want to propose

Implement Weigh Normalization layer using the tensorflow core API. To be used later in the DCNN Network. 

References: [https://www.tensorflow.org/addons/api_docs/python/tfa/layers/WeightNormalization](https://www.tensorflow.org/addons/api_docs/python/tfa/layers/WeightNormalization)

hadifawaz1999:
@aadya940 do you want to add it to the dcnn and ae_dcnn networks ?

aadya940:
@hadifawaz1999 Yes sure, will handle this in a few days.

PR #2117, opened 2024-09-29
[ENH] Add Weight Normalization Keras Wrapper

aadya940:
Related to #1714.

Implements the weight normalization algorithm as a wrapper rather than a layer.


aadya940:
@hadifawaz1999

aeon/utils/tests/test_weightnorm.py
```diff
@@ -0,0 +1,49 @@
+"""Tests for the Weight Normalization layer."""
+
+import os
+
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+if _check_soft_dependencies(["tensorflow"], severity="none"):
```
MatthewMiddlehurst, line 7: I would import tensorflow inside the test rather than do this. There are a few tests with soft dependencies which conditionally skip using a decorator, I can find and link one if you need.

MatthewMiddlehurst (commented):
This was the deprecated layer right? If there's any reused code I would link to the old package and acknowledge that even if it's gone. If it's all original, then no problem.

aeon/utils/tests/test_weightnorm.py
aadya940: I'll just use `pytest.mark.skipif`.

aadya940:
> This was the deprecated layer right? If there's any reused code I would link to the old package and acknowledge that even if it's gone. If it's all original, then no problem.

The implementation is original, hence no need to link to the old package.

MatthewMiddlehurst:
Fine by me then. Will give it a bit more for @hadifawaz1999 to have a check over and put it in if theres nothing.

hadifawaz1999 (dismissed):
Great Addition !! Do you think we can use it for the dilated cnn in deep clustering ?

aadya940:
> Great Addition !! Do you think we can use it for the dilated cnn in deep clustering ?

@hadifawaz1999  Yes it should, since the online difference is a couple of zeros in the kernel. Added it in the test.

MatthewMiddlehurst (approved):

hadifawaz1999 (approved):

Issue #1952, opened 2024-08-12
[ENH] Add test that the estimator "algorithm_type" is valid


TonyBagnall:
### Describe the feature or idea you want to propose

for collection estimators we use the 
    _tags = {"algorithm_type": "distance"}
as a taxonomy of algorithms. However we dont test they are valid, as identified in #1935 

### Describe your proposed solution

add a test the type is one of a valid list, probably defined in the base class?

aryanpola:
@TonyBagnall @MatthewMiddlehurst Just creating a list of distances (like dtw, euclidean etc) in the base class and checking them to be valid in subclasses would solve the issue?

MatthewMiddlehurst:
It would be a test checking that the value of `algorithm_type` is valid i.e. `distance`, `interval`, `shapelet` etc.

aryanpola:
aeon-assign bot assign @aryanpola

MatthewMiddlehurst:
I wouldn't raise an error for now

MatthewMiddlehurst:
Above is outdated, all estimators should have the tag now.

PR #2119, opened 2024-09-30
[MNT] Algorithm type Checking

aryanpola:
Checking for valid algorithm types
#1952 

##### For all contributions


MatthewMiddlehurst (commented):
Few thoughts:

1. This doesn't need a method, just a test IMO
2. I would have this as a test for all estimators, so in `testing/estimator_checking`
3. I wouldn't raise an error for missing tags currently

MatthewMiddlehurst:
Also please do keep to the title naming structure and template as much as possible when creating PRs, it makes triage easier

MatthewMiddlehurst:
I think there may be a misunderstanding here with how we test our estimators, please feel free to ask questions if the issue is confusing. The current code essentially tests a method which does not exist.

What we want to do is check if the `algorithm_type` tag for an estimator (checking all collection estimators i.e. classifiers, regressors and clusterers) is a valid entry.

aryanpola:
> I think there may be a misunderstanding here with how we test our estimators, please feel free to ask questions if the issue is confusing. The current code essentially tests a method which does not exist.
> 
> What we want to do is check if the `algorithm_type` tag for an estimator (checking all collection estimators i.e. classifiers, regressors and clusterers) is a valid entry.

Shouldn't the code check if the input for tag `algorithm_type` is in the list of algorithm types?

MatthewMiddlehurst:
Yes that would be the test.

aryanpola:
> I think there may be a misunderstanding here with how we test our estimators, please feel free to ask questions if the issue is confusing. The current code essentially tests a method which does not exist.
> 
> What we want to do is check if the `algorithm_type` tag for an estimator (checking all collection estimators i.e. classifiers, regressors and clusterers) is a valid entry.

Would having a list of names of classifiers, regressors, and clusterers be fine?
Then, check if the algorithm type from the list is valid.

PR #2120, opened 2024-09-30
[ENH] Remove `BaseEstimator` methods

MatthewMiddlehurst:
Removes `clone_tags`, `set_params`, `get_param_defaults`, `get_param_names`, and `_get_init_signature`. `set_params` is an overload of the `sklearn` base, which seems to only add a call to `reset` for some reason. None of the other methods have usages in the package itself, except for `get_param_defaults` in a test. IMO there is not much value in retaining these for users.

This is mostly an effort to remove method bloat from the base class (and subsequently every `aeon` estimator). One of my gripes with the API is that there are a massive amount of inherited methods which I don't believe people use (i.e. https://www.aeon-toolkit.org/en/stable/api_reference/auto_generated/aeon.classification.convolution_based.Arsenal.html#aeon.classification.convolution_based.Arsenal). These are not the only removal candidates, I just dont want to do too much at once.

Adds a `random_state` param to `clone` and a `keep_attributes` param to `reset`.


TonyBagnall:
I completely agree, but could you do a full run of CI and codecov just to check?

PR #2124, opened 2024-10-01
[ENH] Convert `BaseTransformer` into an abstract class

MatthewMiddlehurst:
Only other base classes extend from `BaseTransformer` currently, and the goal is to have it remain this way in lieu of specific series and collection transformers. There is no need for it to have its own functionality, so this PR converts it into a simple abstract class for other base transformers for now.


TonyBagnall (approved):
good idea

Issue #1637, opened 2024-06-10
[ENH] Add DL anomaly detection algorithms: LSTM-AD


CodeLionX:
### Describe the feature or idea you want to propose

Add LSTM-AD as an anomaly detector to aeon. Reference paper: http://www.elen.ucl.ac.be/Proceedings/esann/esannpdf/es2015-56.pdf

### Describe your proposed solution

- TimeEval implementation
- there is no reference implementation

itsdivya1309:
Hi, could you please provide a link to the paper? The above link seems to be broken.

CodeLionX:
Sorry, the paper was moved to a "legacy"-URL 🤷🏼 : https://www.esann.org/sites/default/files/proceedings/legacy/es2015-56.pdf

This is the reference:
> Malhotra, Pankaj, Lovekesh Vig, Gautam Shroï¬€, and Puneet Agarwal. â€œLong Short Term Memory Networks for Anomaly Detection in Time Series.â€ In Proceedings of the European Symposium on Artificial Neural Networks, Computational Intelligence and Machine Learning (ESANN), Vol. 23, 2015.

PR #2125, opened 2024-10-01
[ENH] Add DL anomaly detection algorithm: LSTM-AD

itsdivya1309:
#### Reference Issues/PRs
Issue: #1637

Implemented LSTM-AD algorithm for time series anomaly detection.

No, it uses tensorflow and scipy.

### DIFF
--- a/aeon/anomaly_detection/__init__.py
+++ b/aeon/anomaly_detection/__init__.py
@@ -3,6 +3,7 @@
 __all__ = [
     "DWT_MLEAD",
     "KMeansAD",
+    "LSTM_AD",
     "MERLIN",
     "STRAY",
     "PyODAdapter",
@@ -11,6 +12,7 @@ __all__ = [
 
 from aeon.anomaly_detection._dwt_mlead import DWT_MLEAD
 from aeon.anomaly_detection._kmeans import KMeansAD
+from aeon.anomaly_detection._lstm_ad import LSTM_AD
 from aeon.anomaly_detection._merlin import MERLIN
 from aeon.anomaly_detection._pyodadapter import PyODAdapter
 from aeon.anomaly_detection._stomp import STOMP
--- a/aeon/anomaly_detection/_lstm_ad.py
+++ b/aeon/anomaly_detection/_lstm_ad.py
@@ -1,0 +1,332 @@
+"""LSTM-AD Anomaly Detector."""
+
+__all__ = ["LSTM_AD"]
+
+import numpy as np
+from scipy.stats import multivariate_normal
+from sklearn.covariance import EmpiricalCovariance
+from sklearn.metrics import fbeta_score
+from sklearn.model_selection import train_test_split
+from tensorflow.keras import layers, models
+from tensorflow.keras.callbacks import EarlyStopping
+
+from aeon.anomaly_detection.base import BaseAnomalyDetector
+
+# from aeon.utils.windowing import sliding_windows
+
+
+class LSTM_AD(BaseAnomalyDetector):
+    """LSTM-AD anomaly detector.
+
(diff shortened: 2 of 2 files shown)


hadifawaz1999 (changes requested):
Thanks for taking care of this. So given its a deep learning model, i think its better if we do the same for all deep sub modules like we do in classification and regression deep learning sub modules, meaning:

- The network would be implemented in the networks module
- Define the anomaly_detection/deep_learning module instead of putting them all together
- Define a base deep anomaly detection module like base deep classifier and base deep regressor for common functionalities like saving and loading

@CodeLionX if you prefer getting this in and then we refactor all to become the same as other modules i don't mind, as long as we set the deep anomaly detection as experimental so we can refactor without having to deprecate stuff @TonyBagnall @MatthewMiddlehurst would like your input on this too

CodeLionX:
I agree with @hadifawaz1999 and I do not see any reason to rush this in. So, if @itsdivya1309 agrees, we can first draft the base deep learning interfaces for anomaly detection before we implement the individual algorithms.

The `anomaly_detection` module is experimental, so there is no issue with deprecations right now.

itsdivya1309:
> I agree with @hadifawaz1999 and I do not see any reason to rush this in. So, if @itsdivya1309 agrees, we can first draft the base deep learning interfaces for anomaly detection before we implement the individual algorithms.
> 
> The `anomaly_detection` module is experimental, so there is no issue with deprecations right now.

I am in for the deep learning sub-module.

CodeLionX:
@itsdivya1309 Sure, happy to have somebody taking care of this! Do you want to create a draft first, and then meet with @hadifawaz1999 to discuss this draft in a dev meeting?

itsdivya1309:
> @itsdivya1309 Sure, happy to have somebody taking care of this! Do you want to create a draft first, and then meet with @hadifawaz1999 to discuss this draft in a dev meeting?

Yeah, sure, I am making the changes suggested by @hadifawaz1999 in this branch itself.

itsdivya1309:
Hi @CodeLionX @hadifawaz1999, I’ve noticed a couple of tests are failing, and I’m unable to fix them. Could you please provide insights on these test failures and how to resolve them?

hadifawaz1999:
So basically i think the best thing to do is to keep in mind that if a deep anomaly detection model is forecasting based, like the one this pr implements (if i understand correctly) then at some point it would be better to implement it at as deep learning in forecasting module (once the new forecasting module is ready) and wrap it in anomaly_deteciton/deep_learning as anomaly detector with the whole config needed.

A lot of deep anomaly detection are self-supervised models, mostly representation learning, so given we are planning to start a self supervised module (soon) then these models will implemented in self supervised module and then wrapped into anomaly detection deep learning submodule with the whole config needed

Will take a look soon on the PR to see the network and all and get back to you

itsdivya1309:
> So basically i think the best thing to do is to keep in mind that if a deep anomaly detection model is forecasting based, like the one this pr implements (if i understand correctly) then at some point it would be better to implement it at as deep learning in forecasting module (once the new forecasting module is ready) and wrap it in anomaly_deteciton/deep_learning as anomaly detector with the whole config needed.
> 
> A lot of deep anomaly detection are self-supervised models, mostly representation learning, so given we are planning to start a self supervised module (soon) then these models will implemented in self supervised module and then wrapped into anomaly detection deep learning submodule with the whole config needed
> 
> Will take a look soon on the PR to see the network and all and get back to you

Thanks for the feedback! That sounds like a solid approach. I’ll also keep the self-supervised module in mind for models focused on representation learning.

Looking forward to your review and any specific feedback on the network design. Thanks again!

TonyBagnall:
hi @itsdivya1309, this seems to be moribund, I'm closing it but feel free to reopen if you want to continue with it

PR #2137, opened 2024-10-02
[BUG] Coverage and bug fixes for utils

### LINKED ISSUE #2134: [BUG] utils/weighted_metrics.py non arguments
### Describe the bug

some legacy code in weighted_metrics allows None arguments for the weights, then tests weights.ndim, so crashes if you dont pass weights. I would KWF but its used in performance metrics

```python
def _weighted_geometric_mean(y, weights=None, axis=None):
    """Calculate weighted version of geometric mean.

    Parameters
    ----------
    y : np.ndarray
        Values to take the weighted geometric mean of.
    weights: np.ndarray
        Weights for each value in `array`. Must be same shape as `array` or
        of shape `(array.shape[0],)` if axis=0 or `(array.shape[1], ) if axis=1.
    axis : int
        The axis of `y` to apply the weights to.

    Returns
    -------
    geometric_mean : float
        Weighted geometric mean
    """
    if weights.ndim == 1:
```

```python
from aeon.utils.weighted_metrics import _weighted_geometric_mean
import numpy as np
y = np.array([1.0,2.0, 3.0])
w = _weighted_geometric_mean(y)
```

### Expected results

it should I think just require weights, otherwise whats the point?

### Actual results

```python-traceback
        geometric_mean : float
            Weighted geometric mean
        """
>       if weights.ndim == 1:
E       AttributeError: 'NoneType' object has no attribute 'ndim'

axis       = None
weights    = None
y          = array([1., 2., 3.])

..\weighted_metrics.py:34: AttributeError
```

### Versions

_No response_

### LINKED ISSUE #2135: [BUG] performance_metrics/forecasting/_functions.py unrunnable code
### Describe the bug

more legacy nonsense. The functions geometric_mean_absolute_error and geometric_mean_squared_error contain untested code 

```        output_errors = _weighted_geometric_mean(
            np.abs(errors),
            sample_weight=horizon_weight,
            axis=0,
        )
```
but the actual signature of _weighted_geometric_mean is
```def _weighted_geometric_mean(y, weights=None, axis=None):``` (see #2134)

i.e. the parameter name is wrong! doctests all skipped for some reason. 

```python
import numpy as np
from aeon.performance_metrics.forecasting import \
    geometric_mean_squared_error as gmse
y_true = np.array([3, -0.5, 2, 7, 2])
y_pred = np.array([2.5, 0.0, 2, 8, 1.25])
gmse(y_true, y_pred, horizon_weight=[0.5,0.3,0.2, .1,.1])
```

### Expected results

runs

### Actual results

```python-traceback
C:\Code\aeon\venv\Scripts\python.exe C:\Code\aeon\aeon\local\debug.py 
  File "C:\Code\aeon\aeon\local\debug.py", line 6, in <module>
    gmse(y_true, y_pred, horizon_weight=[0.5,0.3,0.2, .1,.1])
  File "C:\Code\aeon\aeon\performance_metrics\forecasting\_functions.py", line 1427, in geometric_mean_squared_error
    output_errors = _weighted_geometric_mean(
TypeError: _weighted_geometric_mean() got an unexpected keyword argument 'sample_weight'
```

### Versions

_No response_

TonyBagnall:
got a bit bigger than I expected, so will put up as draft and note changes

Fixes #2135: changes parameter name
Fixes #2134; requires weights passed as ndarray
Removes unused broken weighting functions
Adds tests for show_versions
Adds tests for sklearn functions
Makes files private with init imports: this is our general package wide policy (I think) and should be adopted where possible imo. Sub directories for utils can be changed in time


baraline:
Changes look good, but I would just like some words about the change of approach for the import of the utils module. Do we now want to have most files as private and imported direclty from utils instead of their respective files ? (E.g utils.numba... for the numba utils etc...).

TonyBagnall:
> Changes look good, but I would just like some words about the change of approach for the import of the utils module. Do we now want to have most files as private and imported direclty from utils instead of their respective files ? (E.g utils.numba... for the numba utils etc...).

I think our package wide policy was to have imports via directory init, and this is just imposing that structure. e.g. I change it when I am doing related work in a directory, and hopefully over time this will be adopted, and its better to do it now before version 1.0 to avoid deprecation :)
 
image

baraline (approved):
Agreed. LGTM then, functions from others utils files will be made importable like this when we get to work on them on some PRs !

PR #2138, opened 2024-10-02
[ENH] tag type storage and tag testing rework

MatthewMiddlehurst:
Tags rework. No renaming because that would make it even more confusing.

- Replaces the current `registry` tag descriptions with a dict in utilities
- Adds new lookup and validation utilities to go with the new dict (mainly for testing)
- Removes old tag utilities and variables from `registry`
- Remove tags: `transform-returns-same-time-index`, `skip-inverse-transform`, `X-y-must-have-same-index`, `enforce_index_type`, `y_inner_type`, `instancewise`, `transform_labels`, `remember_data` and `distribution_type`
- Removes tests for some of these tags i.e. `skip-inverse-transform`
- Updates estimator testing for tags. It will now fail if any tag is missing, does not belong in that estimator type, or has an invalid value
- Adds tag defaults to base classes which may have been previously missing (which is now important due to the above)
- Removes `ClearSkyTransformer`, which uses a lot of the above removed tags and currently skips all tests.


TonyBagnall (approved):
LGTM-KWF

PR #2139, opened 2024-10-02
[ENH] Add shift-scale distance (k-SC clusterers distance)

chrisholder:
#### Reference Issues/PRs
#2129 

#### What does this implement/fix? Explain your changes.

This PR adds the shift-scale distance that is used with the k-SC clusterer. This PR only contains the distance measure and a future PR will contain the clusterer.

#### Any other comments?

The shift-scale distance is asymmetric so I had to update all the distance tests so they work even asymmetric distances.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [x] I've added the estimator to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [x] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

##### For developers with write access
- [x] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

Issue #2129, opened 2024-10-02
[ENH] k-Spectral Centroid (k-SC) clusterer


chrisholder:
### Describe the feature or idea you want to propose

A popular clusterer in the TSCL literature is the k-Spectral Centroid (k-SC). A link to the paper can be found at: https://dl.acm.org/doi/10.1145/1935826.1935863.

### Describe your proposed solution

It would be nice to addition to aeon. I propose the lloyd's variant initially and add the incremental version (in the paper as well) at a later date.

PR #2140, opened 2024-10-03
[ENH] KSpectralCentroid clusterer

chrisholder:
#### Reference Issues/PRs
Relates to issue #2129
This PR depends on #2139

This PR implements the KSpectralCentroid (k-SC) clusterer. k-SC is similar to k-means but uses a bespoke averaging and distance measure that is invariant to scale and shift.

##### For all contributions


aeon/clustering/averaging/_shift_scale_invariant_averaging.py
```diff
@@ -0,0 +1,54 @@
...
+                initial_center, X[i], max_shift
+            )
+            optimal_shifts[i] = curr_shift
+
+    if optimal_shifts.shape[0] == 0:
+        return np.zeros((n_dims, n_timepoints))
+
+    normalized_shifts = optimal_shifts / np.sqrt(
+        np.sum(optimal_shifts**2, axis=2)
+    ).reshape(n_instances, n_dims, 1)
```
baraline, line 38: I'm missing the goal of the reshape here `optimal_shifts` is supposed to be 3D right ?

aeon/clustering/averaging/_shift_scale_invariant_averaging.py
```diff
@@ -0,0 +1,54 @@
...
+            normalized_shifts[:, d, :].T, normalized_shifts[:, d, :]
+        ) - n_instances * np.eye(n_timepoints)
+
+    new_center = np.zeros((n_dims, n_timepoints))
+    for d in range(n_dims):
+        eigvals, eigvecs = np.linalg.eig(M[d])
+        new_center[d] = np.real(eigvecs[:, np.argmax(eigvals)])
+
+        if np.sum(new_center[d]) < 0:
+            new_center[d] = -new_center[d]
```
baraline, line 52: Do you really need the matrix M and the two loops here ? It looks like there is no interactions other than storing the result of the dot product to be used in the second loop.

baraline (commented):
Minor comments mostly for my understanding, rest looks good

aeon/clustering/averaging/_shift_scale_invariant_averaging.py
chrisholder: The reason is due to how np.sqrt and np.sum change the shapes of the input.

For example:

``` 
   # Assume optimal_shift is of shape (12, 1, 10) meaning there the optimal 
   # shift for 12 time series instances has been computed

    # If we sum this array
    sum_of = np.sum(optimal_shifts**2, axis=2)
   # sum_of will be  an array of shape (12, 1). This means for each shift their total sum

   # If we sqrt then sqrt to normalise
   sqrt_of = np.sqrt(sum_of)
   # sqrt_of will be an array of shape (12, 1) meaning there are 12 different 
   # sqrts for each shift


  # If we then want to divide optimal shifts which is of shape (12, 1, 10)
  # by sqrt_of which is of shape (12, 1)
  # We first have to reshape sqrt_of so it is of shape (12, 1, 1) so it 
  # matches the shape of the optimal shifts
  reshape_sqrt = sqrt_of.reshape(n_instances, n_dims, 1)
  
  normalized_shifts = optimal_shifts / reshape_sqrt
```

aeon/clustering/averaging/_shift_scale_invariant_averaging.py
chrisholder: Yes this is a very good point I have made it into a single loop now.

aeon/clustering/averaging/_shift_scale_invariant_averaging.py
baraline: Missed the sum function my bad !

baraline (approved):
Lgtm !

PR #2150, opened 2024-10-05
[ENH] More `BaseEstimator` method and attribute changes

MatthewMiddlehurst:
- replaces `_is_fitted` and the `is_fitted` property with just an `is_fitted` attribute
- requires `_tags_dynamic` be present (I do not know why it would be gone, but there are checks for some reason)
- requires a `parameter_set` argument for `get_test_params` method
- removes `create_test_instances_and_names` method (no usages, cant see why it would be useful for users)
- removes `is_composite` method. I don't know why you would want this, but it should be a tag if it stays
- adds some overloads for `sklearn` private methods to make them more compatible with our framework (or raise a better error where there is no compatibility)

### DIFF
--- a/aeon/anomaly_detection/base.py
+++ b/aeon/anomaly_detection/base.py
@@ -82,8 +82,6 @@ class BaseAnomalyDetector(BaseSeriesEstimator, ABC):
     }
 
     def __init__(self, axis):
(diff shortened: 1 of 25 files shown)


TonyBagnall (approved):
KWF

Issue #324, opened 2023-04-24
[BUG] MacOS test for RDST Transformer


baraline:
### Describe the bug

During the integration of the RDST transformer (see #310), the output of the transformation was different than what was expected on macOS systems. 

The failing tests have been removed for now, but should re-introduced for coverage. Test which were failing were written as follows:

- [x] Confirm that the problem comes from MacOS system and not a bug in the code
- [ ] Investigate why the tests are failing on MacOS and not on other platforms

### Steps/Code to reproduce the bug

```python

import pytest
from numpy.testing import assert_array_almost_equal

from aeon.datasets import load_basic_motions, load_unit_test
from aeon.transformations.panel.dilated_shapelet_transform import (
    RandomDilatedShapeletTransform
)


def test_rdst_on_unit_test():
    """Test of ShapeletTransform on unit test data."""
    # load unit test data
    X_train, y_train = load_unit_test(split="train")
    indices = np.random.RandomState(0).choice(len(y_train), 5, replace=False)

    # fit the shapelet transform
    st = RandomDilatedShapeletTransform(max_shapelets=3, random_state=0)
    st.fit(X_train[indices], y_train[indices])

    # assert transformed data is the same
    data = st.transform(X_train[indices])
    assert_array_almost_equal(data, shapelet_transform_unit_test_data, decimal=4)


def test_rdst_on_basic_motions():
    """Test of ShapeletTransform on basic motions data."""
    # load basic motions data
    X_train, y_train = load_basic_motions(split="train")
    indices = np.random.RandomState(4).choice(len(y_train), 5, replace=False)

    # fit the shapelet transform
    st = RandomDilatedShapeletTransform(max_shapelets=3, random_state=0)
    st.fit(X_train[indices], y_train[indices])

    # assert transformed data is the same
    data = st.transform(X_train[indices])
    assert_array_almost_equal(data, shapelet_transform_basic_motions_data, decimal=4)

```

### Expected results

```python

shapelet_transform_unit_test_data = np.array(
    [
        [1.90317756, 8.0, 2.0, 2.87919021, 10.0, 3.0, 0.0, 1.0, 1.0],
        [2.16550181, 8.0, 2.0, 0.0, 10.0, 2.0, 1.52148128, 3.0, 1.0],
        [0.0, 8.0, 1.0, 3.41218663, 10.0, 2.0, 1.00243477, 1.0, 2.0],
[...]

PR #2153, opened 2024-10-06
[ENH] remove non deterministic false tag from RDST

TonyBagnall:
seeing if we can remove this tag now, see #324


baraline:
Think we need to add it to the expected result test pipeline too. I'm afraid we'll have the same issue as with the regressor tho. Somehting in the transformer (numba) is surely causing this but I really don't know what it could be.

PR #2155, opened 2024-10-06
[REF] Rename tags

MatthewMiddlehurst:
Renames tags to replace `-` with `_` and replaces the odd `:removes` capability tag names.


TonyBagnall:
Beat you to it #2152

Issue #1518, opened 2024-05-12
`aeon` v1.0.0 Checklist


MatthewMiddlehurst:
### Describe the issue

A checklist of large-scale changes to be made for an eventual version v1.0.0

- [ ] Deprecate `BaseTransformer` and remove experimental tag from `BaseSeriesTransformer`
- [ ] Replace `forecasting` and `BaseForecaster` with simplified and more maintainable version
- [ ] Merge `BaseEstimator` and `BaseObject` and consider deprecation for a few of the currently available methods
- [ ] Replace current fixture/scenario based testing with a simplified testing suite. Remove experimental tag for `testing`
- [ ] Consider removing experimental tag for other modules

TonyBagnall:
we can do these things before going to 1.0.0. Personally I think 1.0.0 should come at the end of the grant

MatthewMiddlehurst:
I think deprecating a whole module is a big enough of a change to warrant it. I don't know what would be if not. I think it's best just to get these big ones out at the same time.

chrisholder:
I think we should add I better directory structure to the distance module. Currently it's just one flat directory. 

Currently I was thinking we should have something like:

-> Elastic
-> Cross-Correlation
-> Lp (or a better name to describe non time series distances like Euclidean)
-> then either one or multiple directories for distances like dtt_sfa, sax and mpdist which I don't know how/where they would fall into

TonyBagnall:
not sure what happened to forecasting here, seems to have been removed. I put it back

PR #2175, opened 2024-10-10
[DEP] Regression coverage tests and remove deprecated CNN

TonyBagnall:
part of #1518 

picking up a few lines of coverage by testing input variants handled correctly. 

Also removes the deprecated CNNRegressor, CNNClassifier and CNNNetwork as scheduled


hadifawaz1999 (approved):
lgtm

Issue #2163, opened 2024-10-08
[DOC] Update getting started similarity search section


baraline:
### Describe the issue linked to the documentation

https://www.aeon-toolkit.org/en/latest/getting_started.html#time-series-similarity-search Is not up-to-date with the current similarity search API. It must use QuerySearch instead of TopKSimilaritySearch.

PR #2176, opened 2024-10-10
[MNT] Purging similarity search from distance as argument numba functions

### LINKED ISSUE #2071: [MNT] Similarity search and RDST test failures in CI
### Describe the issue

Similarity search and RDST are currently causing some errors in CI:

https://github.com/aeon-toolkit/aeon/actions/runs/10826946884/job/30039020141
https://github.com/aeon-toolkit/aeon/actions/runs/10863339134/job/30147372111
https://github.com/aeon-toolkit/aeon/actions/runs/10890837654/job/30220422286
https://github.com/aeon-toolkit/aeon/actions/runs/10910622966/job/30281472461
https://github.com/aeon-toolkit/aeon/actions/runs/10918273185/job/30303353216 (other failure here as well)
https://github.com/aeon-toolkit/aeon/actions/runs/10918965119/job/30305612433

In case these expire, here are some examples:

```
FAILED aeon/testing/tests/test_all_estimators.py::test_all_estimators[check_fit_updates_state(estimator=RDSTRegressor(max_shapelets=20),datatype=EqualLengthUnivariate-Regression-np-list)] - ReferenceError: underlying object has vanished
```

```
FAILED aeon/similarity_search/distance_profiles/tests/test_naive_distance.py::test_normalized_naive_distance[dtw-float64] - ReferenceError: underlying object has vanished
```

I don't see any failures in `numba` disabled tests, but there is a low sample size.

Solve the case of the missing underlying object.

### Additional context

Tests on general CI disabled by #2080

baraline:
#### Reference Issues/PRs

Fixes #2071, #2163, #2250

- Removes the naive distance and matrix profiles functions that were using a numba function (distance) as argument and some associated components in utils.
- Move some functions from RDST to general numba utils as I reuse them in similarity search to define a "naive" distance Euclidean profile to replace the old one, which was used in testing 
- Add tests for squared distance profile
- Uncomment tests that were commented because of #2071, and add some test exclusion due to issue with the estimator checking function giving wrong input shape for predict function of similarity search (3D fit, 2D predict).
- Remove some unused functions in numba utils
- Remove the fact that distance profiles were returned per channel, and then summed in `QuerySearch` estimator, now they are computed directly in the distance profile functions.
- Update similarity search image on docs
- Change `normalize` (US) to `normalise` (UK) to have the same spelling as the other files in aeon (both were sometime used in the same file).


MatthewMiddlehurst:
Will try get through it this week, but one comment before I dive in:

> Uncomment tests that were commented because of https://github.com/aeon-toolkit/aeon/issues/2071, and add some test exclusion due to issue with the estimator checking function giving wrong input shape for predict function of similarity search (3D fit, 2D predict.

It would be better to add to the testing data file rather than exclude? I can give some help if you need, but really all you need to do is add your data to the big dictionary then edit the functions below to make sure the correct key(s) is given to similarity searchers

aeon/testing/testing_config.py
```diff
@@ -78,6 +73,21 @@
     "MiniRocketRegressor": ["check_non_state_changing_method"],
     "MultiRocketRegressor": ["check_non_state_changing_method"],
     "RSTSF": ["check_non_state_changing_method"],
+    # Due to numba random issue between OS
+    "RDSTRegressor": ["check_regressor_against_expected_results"],
```
MatthewMiddlehurst, line 77: Is this still an issue even with MacOS excluded?

baraline: Yes I had failure on the CI with windows and ubuntu, MacOS gave correct results somehow if I recall correctly, problem is OS related. We could record results for different OS versions if we want to test it anyway while we try to fix it.

baraline:
I'll try to have a look at to include back the ones I excluded later this week then, I'll send you a message if I'm stuck.

aeon/testing/testing_config.py
MatthewMiddlehurst: We don't run the test on MacOS anymore since we moved over to the new runner, so if you are generating the results on a Mac it probably wont be correct.

baraline: OK, I'll try to re-generate results on ubuntu and see if we get something consistent then.

baraline:
@MatthewMiddlehurst added data generation and test for the similarity search estimators, removed the non-deterministic tags from RDST regressor and classifier and added some check against expected results for the classifier.
Only issue arise for RDST Regressor on covid3month specifically. See #2290 for details.

aeon/testing/utils/estimator_checks.py
```diff
@@ -18,15 +19,29 @@ def _run_estimator_method(estimator, method_name, datatype, split):
...
+            value = method(X=FULL_TEST_DATA_DICT[datatype][split][0])
+        else:
+            value = method()
+
+        # Similarity search return tuple as (distances, indexes)
+        if isinstance(estimator, BaseSimilaritySearch):
+            if isinstance(value, tuple):
+                return value[0]
+            else:
+                return value
```
MatthewMiddlehurst, line 42: does it fail without this? not sure what it would be checking

MatthewMiddlehurst (commented):
Should be fine to let this through after the docs are merged. I will add the testing labels just in-case.

MatthewMiddlehurst (commented):
will leave RDST for now

baraline:
Deal, i'll take a look at the conflicts and we should be good to go then.

MatthewMiddlehurst:
If you are up for it, i think we are giving in the American spelling also 🙂

aeon/testing/utils/estimator_checks.py
baraline: It would fail without this as similarity search estimators returns a tuple of two array (match_distances, match_indexes), which is not standard as other methods just return one array (with predict/transform)

MatthewMiddlehurst: I'm just wondering where in the general testing it would check for this really. It also covers transformers which have all sorts of outputs

baraline:
@MatthewMiddlehurst answred comment and fixed tests that were causing failures (I implemented these test wrong). Concerning language, I would prefer to delay it to another PR, this one is big enough.

MatthewMiddlehurst (approved):
whats the worst its going to do, break everything?

baraline:
Well it seems that it run fines on main, sooooo all good I guess ?

PR #2183, opened 2024-10-11
[ENH] Remove the return_X_y load option, improve coverage

TonyBagnall:
in preparation for removing nested_univ all together, I have removed the option to return_X_y from the loaders. It is not a useful functionality imo, and was only ever used when set to True, which was in any case the default value. It makes more sense for tabular data to add as a column, and thats what used to happen with nested_univ, but now I think it is redundant. For numpy it was simply returning the X and y as a tuple. It will do this by default if you just go 

X = load_thing()

so 
X = load_thing(return_X_y=False)

is redundant.

Also changed load_solar. The load direct from website is broken (identified because the code is uncovered) and Im not willing to spend time working it out. Added a test for macroeconomic


aeon/datasets/_data_loaders.py
```diff
@@ -337,9 +333,7 @@ def _load_saved_dataset(
     -------
     X: Data stored in specified `return_type`
         The time series data for the problem, with n instances
-    y: 1D numpy array of length n, only returned if return_X_y if True
-        The class labels for each time series instance in X
-        If return_X_y is False, y is appended to X instead.
+    y: 1D numpy array of length n
```
chrisholder, line 336: This should probably have a description still even if it's "The class labels for each time series instance in X." Also you refer later in the PR to "1D numpy array of length len(X)". I dont mine which one but should be consistent.

aeon/datasets/_data_loaders.py
```diff
@@ -337,9 +333,7 @@ def _load_saved_dataset(
...
     meta: meta data dictionary, only returned if return_meta is True
```
chrisholder, line 337: Meta should have a description as well

aeon/datasets/_single_problem_loaders.py
```diff
@@ -389,22 +356,12 @@ def load_plaid(split=None, return_X_y=True, return_type="np-list"):
...
-    ------
-    ValueError is raised if the data cannot be stored in the requested return_type.
 
     Returns
     -------
     X: list of 2D np.ndarray, one for each series.
-    y: 1D numpy array of length n, only returned if return_X_y if True
-        The class labels for each time series instance in X
-        If return_X_y is False, y is appended to X instead.
+    y: 1D numpy array of length len(X). The class labels for each time series
```
chrisholder, line 363: The additional information should be moved to the description

chrisholder (changes requested):

chrisholder (approved):

Issue #1155, opened 2024-02-08
[ENH] Remove the unused inner types for collections 


TonyBagnall:
### Describe the feature or idea you want to propose

I'm trying to reduce code bloat. We currently allow the following input and inner types 

```python
COLLECTIONS_DATA_TYPES = [
    "numpy3D",  # 3D np.ndarray of format (n_cases, n_channels, n_timepoints)
    "np-list",  # python list of 2D numpy array of length [n_cases],
    # each of shape (n_channels, n_timepoints_i)
    "df-list",  # python list of 2D pd.DataFrames of length [n_cases], each a of
    # shape (n_timepoints_i, n_channels)
    "numpy2D",  # 2D np.ndarray of shape (n_cases, n_channels*n_timepoints)
    "pd-wide",  # 2D pd.DataFrame of shape (n_cases, n_channels*n_timepoints)
    "nested_univ",  # pd.DataFrame (n_cases, n_channels) with each cell a pd.Series,
    "pd-multiindex",  # pd.DataFrame with multi-index,
]
```
we provide conversions for each, so each extra one creates a lot of code. The folllowing three are not used anywhere, and allow index confusion
    "df-list",  # python list of 2D pd.DataFrames of length [n_cases], each a of
    "pd-wide",  # 2D pd.DataFrame of shape (n_cases, n_channels*n_timepoints)
    "nested_univ",  # pd.DataFrame (n_cases, n_channels) with each cell a pd.Series,

I would like to get rid of one, two or three of these. df-list is inferior to np-list, pd-wide assumes a flattening (legacy comment) and "nested_univ" is just a hang over from the early days, no one likes it or uses it (I think keep multiindex even though its not internally used, good for conversions in base class)

Not doing this immediately, but interested in opinions 

Tony

### Describe your proposed solution

I guess deprecate any we decide to remove, with warnings during conversion and deprecate in 0.9

TonyBagnall:
consensus is to keep df-list, so after #2185 this we are supporting these data types for X_inner_type and conversions

```python
SERIES_DATA_TYPES = [
    "pd.Series",  # univariate time series of shape (n_timepoints)
    "pd.DataFrame",  # multivariate time series of shape (n_timepoints, n_channels) by
    # default or (n_channels, n_timepoints) if set by axis ==1
    "np.ndarray",  # uni/multivariate time series of shape (n_timepoints, n_channels)
    # by default or (n_channels, n_timepoints) if set by axis ==1
]

[...]

PR #2185, opened 2024-10-11
[ENH] Final purge of nested univ

### LINKED ISSUE #2184: [ENH] TSFreshFeatureExtractor returns a dataframe

found in #2183, this transform returns a dataframe and weirdly still had nested_univ input

it should return an array, but only do after #2183

TonyBagnall:
#### Reference Issues/PRs

part of #1155 
fixes #2184 

Removes the nested_univ converters and all references to it. This got a bit bigger than expected, which goes to show the need for the purge, still a couple of gotchas.

1. FastSFA: had a bodge to return a nested dataframe so it complied with tests. Removed that, testing works unless a dense representation is returned
2. TSFresh: needed a bit of a rework to stop it returning a data frame. Now it stores the names of the features retained in the transform stage in an attribute
3. Refreshed the TSFresh notebook


aeon/testing/estimator_checking/_yield_classification_checks.py
```diff
@@ -128,7 +128,7 @@ def check_classifier_against_expected_results(estimator_class):
 
 def check_classifier_tags_consistent(estimator_class):
     """Test the tag X_inner_type is consistent with capability:unequal_length."""
-    valid_types = {"np-list", "df-list", "pd-multivariate", "nested_univ"}
+    valid_types = {"np-list", "df-list", "pd-multivariate"}
```
MatthewMiddlehurst, line 131: unrelated probably but what is "pd-multivariate"? same in regression and clustering

aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -730,8 +711,7 @@ def get_test_params(cls, parameter_set="default"):
         params = {
             "word_length": 4,
             "window_size": 4,
-            "return_sparse": True,
-            "return_pandas_data_series": True,
+            "return_sparse": False,
```
MatthewMiddlehurst, line 714: What is the failure with this?

aeon/transformations/collection/feature_based/_tsfresh.py
```diff
@@ -339,6 +339,23 @@ def get_test_params(cls, parameter_set="default"):
             },
         ]
 
+    def _get_names(self):
+        """Hack to get the feature names prior to transform."""
+        from tsfresh import extract_features
+
+        X = np.random.random((2, 1, 30))
+        Xt = _from_3d_numpy_to_long(X)
+        Xt = extract_features(
```
MatthewMiddlehurst, line 348: why do we need this?

aeon/utils/validation/series.py
```diff
@@ -193,12 +193,6 @@ def _check_is_multivariate(Z, var_name="input"):
     """
     if isinstance(Z, pd.Series):
         raise ValueError(f"{var_name} must have 2 or more variables, but found 1.")
-    if isinstance(Z, pd.DataFrame):
-        nvars = Z.shape[1]
-        if nvars < 2:
-            raise ValueError(
-                f"{var_name} must have 2 or more variables, but found {nvars}."
-            )
```
MatthewMiddlehurst, line 201: this is not nested_univ i think?

MatthewMiddlehurst (commented):
Few comments

aeon/testing/estimator_checking/_yield_classification_checks.py
TonyBagnall: I have no idea! Will remove, df-list is going next anyway after this PR

aeon/datasets/_data_writers.py
```diff
@@ -33,9 +32,7 @@ def write_to_tsfile(
         Indicate if this is a regression problem, so it is correcty specified in
         the header since there is no definite way of inferring this from y
     """
-    if not (
-        isinstance(X, np.ndarray) or isinstance(X, list) or isinstance(X, pd.DataFrame)
-    ):
+    if not (isinstance(X, np.ndarray) or isinstance(X, list)):
```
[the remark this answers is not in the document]
TonyBagnall: Ill look

aeon/transformations/collection/feature_based/_tsfresh.py
TonyBagnall: because the existing test gets the features with a given name and compares to expected. You cannot extract the names in transform as you are not allowed to change state it transform. This transformer has empty fit. I couldnt work out how to get the names without doing the transform. So either remove the test or do something like this. I think its useful to store the names, for general interpretability and because as I understand it they can alter with constructor arguments, although still not sure about that. This seemed the quickest way to solve. 

```python
def test_tsfresh_extractor(default_fc_parameters):
    """Test that mean feature of TSFreshFeatureExtract is identical with sample mean."""
    X = np.random.rand(10, 1, 30)

    transformer = TSFreshFeatureExtractor(
        default_fc_parameters=default_fc_parameters, disable_progressbar=True
    )

    Xt = transformer.fit_transform(X)
    actual = Xt.filter(like="__mean", axis=1).values.ravel()
    expected = np.mean(X, axis=2).flatten()
    np.testing.assert_allclose(actual, expected)
```

aeon/datasets/_data_writers.py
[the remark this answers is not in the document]
TonyBagnall: so is_collection would be the check, but currently the writers do not support the dataframe format pd-wide. Ill create an issue, other things could be tidied up in this file, its evolved into a bit of a mess

aeon/testing/estimator_checking/_yield_classification_checks.py
TonyBagnall: I think its meant to be hierarchical

aeon/transformations/collection/dictionary_based/_sfa_fast.py
TonyBagnall: a general test fails when csr_matrix is returned, I'll run it with True to confirm and raise an issue

aeon/datasets/_data_writers.py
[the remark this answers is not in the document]
TonyBagnall: see #2208

aeon/transformations/collection/dictionary_based/_sfa_fast.py
TonyBagnall: see #2210

MatthewMiddlehurst (approved):
Would give a bit for others to take a look if they want.

PR #2187, opened 2024-10-13
[ENH] RotationForest coverage

TonyBagnall:
more make work from me for a few lines of coverage. @MatthewMiddlehurst Im not sure when this would happen, any ideas on which classifier would do this and under what circumstances? 

```python
        probas = clf.predict_proba(X_t)

        if probas.shape[1] != self.n_classes_:
            new_probas = np.zeros((probas.shape[0], self.n_classes_))
            for i, cls in enumerate(clf.classes_):
                cls_idx = self._class_dictionary[cls]
                new_probas[:, cls_idx] = probas[:, i]
            probas = new_probas
```


baraline:
I think the only thing that could look like this is the binary case where estimators produce only probas for class 1 and we want it for class 0 and 1.

Otherwise the estimators is just ignoring/adding classes which seems wrong? Unless we are in some kind of unsupervised problem such as clustering where we create clusters on the go with updates?

MatthewMiddlehurst:
Answered this in a DM, but I believe it was a workaround for bagging with very small datasets (i.e. so small you are losing classes when you select your data for the tree).

MatthewMiddlehurst:
Does this PR still need more changes?

MatthewMiddlehurst (approved):

Issue #2177, opened 2024-10-11
[MNT] Ensure base classes are abstract and adjust testing


CodeLionX:
### Describe the issue

As discussed in https://github.com/aeon-toolkit/aeon/pull/2118#discussion_r1796538246 and Slack, our current testing framework checks that every estimator class has an implementation for `_fit`/`_predict`/`_fit_predict` using

```python
>>> # Test that all anomaly detectors implement abstract predict.
>>> assert "_predict" in estimator_class.__dict__
E       AssertionError
```

This, however, just checks within the final class definition and ignores any potential implementations in the class hierarchy. Thus, in wrapper-land, we are required to overwrite these methods only to call the super-implementation, e.g.:

```python
from aeon.anomaly_detection import PyODAdapter

class IsolationForest(PyODAdapter):
    def __init__(self, window_size, stride, **params):
        from pyod.models.iforest import IForest
        ...
        super().__init__(IForest(**params), window_size, stride)

    def _fit(self, X, y=None):
        super()._fit(X, y)

    def _predict(self, X):
        return super()._predict(X)

    def _fit_predict(self, X, y=None):
        return super()._fit_predict(X, y)
```

I would like to avoid the boilerplate code and reduce the wrapper to the class-definition, documentation, and constructor.

### Suggest a potential alternative/fix

[Python's abstract base classes](https://docs.python.org/3/library/abc.html) already provide a way to enforce that abstract methods must be implemented. Otherwise, the instantiation of the class fails.

If we ensure that all our base classes inherit from `abc.ABC` and annotate their abstract methods with `abc.abstractmethod`, we can remove the testing for implementations of `_fit`/`_predict`/`_fit_predict`, and just test the instantiation of the estimator to not raise a `TypeError: Can't instantiate abstract class XYZ with abstract method _xyz`.

- [ ] ensure that all base classes inherit from `abc.ABC`
- [ ] annotate all abstract methods with `abc.abstractmethod` (at least `_predict`, `_fit`, `_fit_predict`)
- [ ] change testing framework (remove all checks for implementations of the abstract methods; add a test that instantiates an estimator and provides a proper error message)

MatthewMiddlehurst:
#2171 adds `ABC` to `BaseAeonEstimator` for the first step of this.

PR #2198, opened 2024-10-14
[ENH] Update remaining `registry` files

MatthewMiddlehurst:
- remove current contents of `registry`
- base class lists and utilities have been moved to `utils.base` and reworked into a dictionary.
- `all_estimators` has been simplified a bit and moved to `utils.discovery` (same import as sklearn)
- adds `check_tag_value` to help validating the value of singular tags


TonyBagnall (approved):
this all looks fine, but we could do with going over all base classes, make sure they have abstract constructors etc  and resolve #2177

PR #2202, opened 2024-10-14
[ENH] Implemented COPOD for anomaly detection

### LINKED ISSUE #2111: [ENH] Add PyODAdapter-implementation for COPOD

The [`PyODAdapter`](https://github.com/aeon-toolkit/aeon/blob/main/aeon/anomaly_detection/_pyodadapter.py) in aeon allows us to use any outlier detector from [PyOD](https://github.com/yzhao062/pyod), which were originally proposed for relational data, also for time series anomaly detection (TSAD). Not all detectors are equally well suited for TSAD, however. We want to represent the frequently used and competitive outlier detection techniques within the `anomaly_detection` module of aeon directly.

**Implement the [COPOD method](https://github.com/yzhao062/pyod/blob/master/pyod/models/copod.py#L52)** using the `PyODAdapter`.

- Create a new file in `aeon.anomaly_detection` for the method
- Create a new estimator class with `PyODAdapter` as the parent
- Expose the algorithm's hyperparameters as constructor arguments, create the PyOD model and pass it to the super-constructor
- Document your class
- Add tests for certain edge cases if necessary

---

Example for IsolationForest:

```python
class IsolationForest(PyODAdapter):
    """documentation ..."""
    def __init__(n_estimators: int = 100, max_samples: int | str = "auto", ..., window_size: int, stride: int):
        model = IForest(n_estimators, max_samples, ...
        super().__init__(model, window_size, stride)

    @classmethod
    def get_test_params(cls, parameter_set="default"):
        """..."""
        return {"n_estimators": 10, ...}
```

_No response_

### Additional context

_No response_

notaryanramani:
#### Reference Issues/PRs

Fixes #2111 

Implements COPOD from PyOD to be used in aeon framework along with some test cases.

No

##### For all contributions


aeon/anomaly_detection/_copod.py
```diff
@@ -0,0 +1,87 @@
...
+    _tags = {
+        "capability:multivariate": True,
+        "capability:univariate": True,
+        "capability:missing_values": False,
+        "fit_is_empty": False,
+        "python_dependencies": ["pyod"],
+    }
+
+    def __init__(self, n_jobs: int = -1, window_size: int = 10, stride: int = 1):
+        _check_soft_dependencies(*self._tags["python_dependencies"])
```
CodeLionX, line 51: Not necessary as long as the `"python_dependencies"`-tag is set correctly. The base class checks for the soft dependency, then.

aeon/anomaly_detection/_copod.py
```diff
@@ -0,0 +1,87 @@
...
+       * - Input data format
+         - univariate and multivariate
+       * - Output data format
+         - anomaly scores
+       * - Learning Type
+         - unsupervised or semi-supervised
+
+    Parameters
+    ----------
+    n_jobs : int, default=-1
```
CodeLionX, line 32:
```suggestion
    n_jobs : int, default=1
```

aeon/anomaly_detection/_copod.py
```diff
@@ -0,0 +1,87 @@
...
+
+    _tags = {
+        "capability:multivariate": True,
+        "capability:univariate": True,
+        "capability:missing_values": False,
+        "fit_is_empty": False,
+        "python_dependencies": ["pyod"],
+    }
+
+    def __init__(self, n_jobs: int = -1, window_size: int = 10, stride: int = 1):
```
CodeLionX, line 50: Please use 1 as the default for `n_jobs`. All other estimators use this default.

aeon/anomaly_detection/_copod.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict, default={}
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`.
+        """
+        return {}
```
CodeLionX, line 87: Is inherited from the base class with `{}`, so can be omitted.

CodeLionX (changes requested):
Well done. Thank you again!
I still think that the tests are excessive because we already test all cases with the `PyODAdapter`. Could you boil it down to a couple of tests specific for the `COPOD`-estimator? In addition, the current tests still refer to `iforest` 😉

aeon/anomaly_detection/_copod.py
notaryanramani: When I tried not using this function last time in `IsolationForest`, one of the test cases was failing. That's why it is also included there.

notaryanramani:
@CodeLionX I am not familiar with COPOD algorithm. Could you please give an idea what kind of tests are required?

aeon/anomaly_detection/_copod.py
notaryanramani: I think the `PyODAdapater`'s `get_test_params` returns the following dictionary:

```python
return {"pyod_model": LOF(), "window_size": 5, "stride": 1}
```

CodeLionX: Ah, you are right. I forgot. Sry, then disregard this comment!

aeon/anomaly_detection/_copod.py
CodeLionX: Strange... there should not be any input parameter checking in the constructor (`__init__`) according to the sklearn-conventions. This is instead done in `fit` (in our case `_fit`). I would consider checking soft-dependencies follows similar principles. Thus, I cannot understand why this is necessary 🤔

notaryanramani: I don't think COPOD from pyod here is an input parameter, I am importing it in the constructor itself to pass it to the base class. 

I tried not using `_check_soft_dependencies`. I got the following test case fail message:
```bash
E           AssertionError: Estimator COPOD requires soft dependencies ['pyod'] according to tags, 
but does not raise an appropriate error message on __init__, when a soft dependency is missing. 
Likely reason is that __init__ does not call super(cls).__init__, or imports super(cls).__init__ 
only after an attempted import.
```

CodeLionX: Yes, that makes sense to me now! I apologize for the confusion on my side. Your approach is correct!

CodeLionX:
> @CodeLionX I am not familiar with COPOD algorithm. Could you please give an idea what kind of tests are required?

I trust that the implementation in PyOD is correct. So, our tests would check that we integrate it in aeon correctly:

- instantiating the class should work (when pyod is installed): you can just keep one of the end-to-end tests
- the parameters of the algorithm should be correctly passed down to the PyOD model
- aeon with `window_size=1` and `stride=1` should produce the same results as calling the PyOD model directly, when I am not mistaken

notaryanramani:
Umm... sorry I was trying to rebase the branch and something wild happened 😅. Should I close this PR and open a new one?

aeon/anomaly_detection/__init__.py
```diff
@@ -8,9 +8,12 @@
     "PyODAdapter",
     "STOMP",
     "LeftSTAMPi",
+    "COPOD",
     "IsolationForest",
+    "COPOD",
```
CodeLionX, line 13: Duplicated entry

aeon/anomaly_detection/tests/test_copod.py
```diff
@@ -0,0 +1,63 @@
...
+
+    series = make_series(n_timepoints=100, return_numpy=True, random_state=0)
+    series[20:30] -= 2
+
+    # fit and predict with aeon COPOD
+    copod = COPOD(window_size=1, stride=1)
+    copod_preds = copod.fit_predict(series)
+
+    # fit and predict with PyOD COPOD
+    _series, _ = sliding_windows(series, window_size=1, stride=1, axis=0)
```
CodeLionX, line 58:
```suggestion
    _series = series.reshape(-1, 1)
```

aeon/anomaly_detection/tests/test_copod.py
```diff
@@ -0,0 +1,63 @@
...
+    not _check_soft_dependencies("pyod", severity="none"),
+    reason="required soft dependency PyOD not available",
+)
+def test_aeon_copod_with_pyod_copod():
+    """Test COPOD with PyOD COPOD."""
+    from pyod.models.copod import COPOD as PyODCOPOD
+
+    from aeon.utils.windowing import sliding_windows
+
+    series = make_series(n_timepoints=100, return_numpy=True, random_state=0)
```
CodeLionX, line 50: Please use [`aeon.testing.data_generation.make_example_1d_numpy`](https://github.com/aeon-toolkit/aeon/blob/main/aeon/testing/data_generation/_series.py#L17) instead of the legacy method.

CodeLionX (changes requested):
Looks good. Just some minor changes so that we don't need to clean it up later 👍🏼

CodeLionX (approved):

TonyBagnall:
great, thanks for this

PR #2203, opened 2024-10-15
[DOC] Add JMLR paper to readme

TonyBagnall:
adds a link and bibtex to the end of the readme, which seems to be the standard way of doing this.


patrickzib (approved):
LGTM

PR #2214, opened 2024-10-16
[ENH, BUG] Fixes issue with lower bounding in SAX for TS length not divisible by the number of segments 

patrickzib:
This fixes an issue with lower bounding in iSAX. 

If the time series length is not divisible by the number of segments, PAA is padding some segments with 0. The concrete segments to pad are determined using `np.array_split()`.

To hold lower bounding, each segment's mean must thus be multiplied by the exact number of values it represents, again using `np.array_split()`.


TonyBagnall (approved):
lgtm

PR #2215, opened 2024-10-16
[ENH] more fixes for float_

TonyBagnall:
replaces np.float_ with np.float64


CodeLionX (approved):
Search for `float_` seems to not return more than that 👍🏼

TonyBagnall:
> Search for `float_` seems to not return more than that 👍🏼

yeah I ran it under 2.0 locally. I'm optimistic it might actually be ok to upgrade to to numpy 2.0 (bold words). Running tests locally

Issue #622, opened 2023-08-01
[BUG] Some tests fail without numba


MatthewMiddlehurst:
### Describe the bug

When numba is disabled using `NUMBA_DISABLE_JIT=1` some tests begin failing. We disable numba for our code coverage test, as it does not play well together currently (https://github.com/numba/numba/issues/4268).

Failing tests: see #625 #626

Example nightly test run: 
https://github.com/aeon-toolkit/aeon/actions/runs/5721839977/job/15504019987

### Steps/Code to reproduce the bug

Run the test suite with the `NUMBA_DISABLE_JIT` environment variable set to 1.

### Expected results

No unexpected tests fail even with numba disabled.

### Actual results

Tests which normally pass start failing.

### Versions

N/A, see CI run

MatthewMiddlehurst:
See changes in #624. This skips the currently failing tests.

TonyBagnall:
having looked into #1498 and found a bug that numba was hiding, it might be good too look at these failures in more detail, so will list here

- [ ] test_all_classifiers:test_classifier_against_expected_results HIVECOTEV2 fails
- [ ] StatsForecastAutoARIMA #625  this is excluded from all tests in config if numba is not used
- [ ] test_sfa.py: test_typed_dict() excluded
- [ ] test_catch22_on_basic_motions(): excluded

TonyBagnall:
so I think both fails are caused by catch22, and by a tedious process of elimination I have isolated the problem to this function. Error happens when you comment out the njit

```python
    @staticmethod
    @njit(fastmath=True, cache=True)
    def _SB_MotifThree_quantile_hh(X):
        # Shannon entropy of two successive letters in equiprobable 3-letter
        # symbolization.
        indicies = np.argsort(X)
        bins = np.zeros(len(X))
        q1 = int(len(X) / 3)
        q2 = q1 * 2
        l1 = np.zeros(q1, dtype=np.int_)
        for i in range(q1):
            l1[i] = indicies[i]
        l2 = np.zeros(q1, dtype=np.int_)
        c1 = 0
        for i in range(q1, q2):
            bins[indicies[i]] = 1
            l2[c1] = indicies[i]
            c1 += 1
        l3 = np.zeros(len(indicies) - q2, dtype=np.int_)
        c2 = 0
        for i in range(q2, len(indicies)):
            bins[indicies[i]] = 2
            l3[c2] = indicies[i]
            c2 += 1

        found_last = False
        nsum = 0
[...]

PR #2216, opened 2024-10-16
[MNT] Make aeon numpy compatible with both =>2.0 and <2.0 

### LINKED ISSUE #2218: [ENH] Numpy 2 compatibility

it would be good to update our code to make it numpy 2.0 compatible, even if dependencies are not. See #2218

To make aeon compliant: 

- [x] Change all np.float_ to np.float64
- [x] Change all np.INF and np.Inf to np.inf
- [ ] Remove expected numerical outputs from docstrings
- [ ] Resolve difference with np.argsort
- [ ] Fix np.nan_to_num to only pass arrays

### Soft deps with potential numpy issues (deal with after)
ptsr: Implement anomaly detection functions and remove dep #2066 
tensorflow

TonyBagnall:
#### Reference Issues/PRs
closes #2218 

test numpy2 compliance over simple fixes

STRAY np.Inf -> np.inf
Catch22:  change list to array prior to call to nan_to_num
```python
        c22_array = np.array(c22_list)
        if self.replace_nans:
            c22_array = np.nan_to_num(c22_array, False, 0, 0, 0)
```
performance_metrics.forecasting._functions

there is a difference in docstring tests: output of scalars appears as np.float64(x) not just x.

I have stripped out the expected output. Under numpy 2 it expects np.float64(0.0003), but that does not work with earlier numpy versions, where just 0.0003 is expected. the numbers are exactly the same


TonyBagnall:
ok so isolated another weird difference arising in the test function `test_stomp_squared_matrix_profile`

this code
```python
import numpy as np
expected = np.array([[18.00000,9.00000,6.00000,9.00000,18.00000,33.00000],
[19.00000,6.00000,5.00000,9.00000,6.00000,5.00000]])
id_bests = np.vstack(
    np.unravel_index(np.argsort(expected.ravel()), expected.shape)
).T

print(id_bests)
```
produces different output under numpy 1.26 and numpy 2.0. I assume due to tie breaking
1.26 output
[[1 2]
 [1 5]
 [0 2]
 [1 1]
 [1 4]
 [0 1]
 [0 3]
 [1 3]
 [0 0]
 [0 4]
 [1 0]
 [0 5]]

2.0 output
[[1 5]
 [1 2]
 [0 2]
 [1 1]
 [1 4]
 [1 3]
 [0 3]
 [0 1]
 [0 4]
 [0 0]
 [1 0]
 [0 5]]

TonyBagnall:
(reminds me of the shapelet argsort issue under numba/no numba

TonyBagnall:
simplified further 

```python
expected = np.array([[18.00000,9.00000,6.00000,9.00000,18.00000,33.00000],
[19.00000,6.00000,5.00000,9.00000,6.00000,5.00000]])
x= [18.00000,9.00000,6.00000,9.00000,18.00000,33.00000,19.00000,6.00000,5.00000,
    9.00000,6.00000,5.00000]
args = np.argsort(x)
```
gives different results in args under differnt numpy versions (and previously with and without numba #622). This is a very minor issue in one test, but we do use argsort quite a lot, might be time to implement our own version ...

image

TonyBagnall:
ah they have added a "stable" argument! Problem solved :)
https://numpy.org/doc/2.0/reference/generated/numpy.argsort.html

TonyBagnall:
still some weirdness, but tbh there is little to gain from further drilling down. Simplified the test, set kind="stable" and it passes with and without numpy3

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -286,11 +286,11 @@ def _transform(self, X, y=None):
             )
             for i in range(n_cases)
         )
-
+        c22_array = np.array(c22_list)
```
MatthewMiddlehurst, line 289:
```suggestion
        
        c22_array = np.array(c22_list)
```

MatthewMiddlehurst (dismissed):
Cant say I'm a fan of just removing all the expected outputs

TonyBagnall:
> Cant say I'm a fan of just removing all the expected outputs

if you dont you cannot be backward compatible as far as I can tell

aeon/similarity_search/matrix_profiles/tests/test_stomp.py
```diff
@@ -16,8 +16,8 @@
 )
 from aeon.utils.numba.general import sliding_mean_std_one_series
 
-DATATYPES = ["float64", "int64"]
-K_VALUES = [1, 3]
+DATATYPES = ["int64", "float64"]  # DATATYPES = [ ]
+K_VALUES = [1]
```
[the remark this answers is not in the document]
TonyBagnall: because there are further issues with k == 3 I am not willing to spend further time debugging, comment can go though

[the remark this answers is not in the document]
TonyBagnall: @baraline this is your test, want to take a look?

CodeLionX:
The removed outputs are mostly in the performance metrics module. IMO, it has lower priority, so I don't care as much.

The problem is that the examples are checked using a string-comparison, right? Because `np.float64(3.2) == 3.2` should return `True`.

TonyBagnall:
> The removed outputs are mostly in the performance metrics module. IMO, it has lower priority, so I don't care as much.
> 
> The problem is that the examples are checked using a string-comparison, right? Because `np.float64(3.2) == 3.2` should return `True`.

agree, I think we skip ones we care about and just dont print of the performance metric module. I dont know how docstring runs the comparison, not really motivated to find out

aeon/similarity_search/matrix_profiles/tests/test_stomp.py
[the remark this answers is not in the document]
MatthewMiddlehurst: still seems to be an issue

aeon/similarity_search/matrix_profiles/tests/test_stomp.py
```diff
@@ -16,8 +16,8 @@
 )
 from aeon.utils.numba.general import sliding_mean_std_one_series
 
-DATATYPES = ["float64", "int64"]
-K_VALUES = [1, 3]
+DATATYPES = ["int64", "float64"]
+K_VALUES = [1]
```
CodeLionX, line 20: I agree with @MatthewMiddlehurst that this should be resolved.

CodeLionX (changes requested):

aeon/similarity_search/matrix_profiles/tests/test_stomp.py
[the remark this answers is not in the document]
TonyBagnall: its an issue with the test and argmin, I dont think it should hold up this PR

aeon/similarity_search/matrix_profiles/tests/test_stomp.py
TonyBagnall: its a thankless debugging task for no reward @CodeLionX, hopefully one of you will take it on ...

CodeLionX: Got no time for that 🤷🏼 
At least create an issue for this, then

TonyBagnall: see #2236

CodeLionX (approved):

MatthewMiddlehurst (approved):

PR #599, opened 2023-07-26
[DOC] Update data_storage and benchmarking_data notebooks

MatthewMiddlehurst:
Rewrites some of the content, fixes links and regenerates cell output for the mentioned notebooks.

Minor changes to other dataset notebooks.

PR #2233, opened 2024-10-20
[ENH] Tidy up benchmarking results loaders

MatthewMiddlehurst:
#### Reference Issues/PRs

Will probably conflict with #2222

- removes imports from `__init__`
- moves published results loaders to their own file, function contents are unchanged.
- remove "ri" from the clustering entry of `VALID_RESULT_MEASURES` because the tsc.com storage does not have rand index files.
- reworks `get_estimator_results` function and docs. By default loads all available datasets instead of a pre-set list of ucr datasets. number of resamples to load is configurable (default train/test/split). Returns a dict (classifiers) of dicts (datasets) containing the requested score or array or resampled scores.
- reworks `get_estimator_results_as_array`. same as the above, but always returns a score array as well as the dataset names to align with rows. Averaged resampled scores if requested.


MatthewMiddlehurst:
@hadifawaz1999 I think you use these functions? Don't know if you have any comments on them.

aeon/benchmarking/published_results.py
```diff
@@ -0,0 +1,288 @@
...
+    >>> plot.show()# doctest: +SKIP
+    """
+    return _get_published_results(
+        directory="Bakeoff2023",
+        classifiers=uni_classifiers_2023,
+        resamples=30,
+        suffix="_TESTFOLDS.csv",
+        default_only=default_only,
+        header="infer",
+        n_data=112,
```
hadifawaz1999, line 253: it seems to get only on the 112 datasets, no way to get it on the 142 ?

TonyBagnall:
need to rework the notebooks for this PR

aeon/benchmarking/published_results.py
MatthewMiddlehurst: Results arent uploaded i'm pretty sure, not aeon related

TonyBagnall:
maybe combine with #599

TonyBagnall (approved):

hadifawaz1999:
should we put up an issue @TonyBagnall @MatthewMiddlehurst mentioning that we need to revist this once the 142 datasets results are all online ?

PR #2237, opened 2024-10-21
[REF,ENH] Refactor test params/instance methods and disable `sklearn` metadata function creation

MatthewMiddlehurst:
- renames `get_test_params` to make it protected
- renames `create_test_instance` to make it protected
- updates docs where I found cases of the above. Removes part of the `_get_test_params` docstring because it isnt really true or relevant.
- overloads `get_metadata_routing` and  `_get_default_requests` from `BaseEstimator` to raise errors. I'm not familiar with this framework, it is disabled by default in `sklearn` and I dont believe it is compatible with any of our estimators. It currently generates a bunch of functions, this should stop that.


TonyBagnall (approved):
LGTM

TonyBagnall (approved):

PR #2242, opened 2024-10-22
[ENH] Rework `_HeterogenousMetaEstimator` and composable estimators

MatthewMiddlehurst:
This is a bit big and messy, but a lot of these composable estimators are intertwined with the mixin which is to be expected really. Just went for a tidy of the whole lot.

- renames `_HeterogenousMetaEstimator` to `_ComposableEstimatorMixin`
- reworks the class, removing a lot of the methods and functions. Centres input validation around two methods, `_check_estimators` and `_convert_estimators`
- the methods for getting and setting params/attributes should work as intended like the base class now
- remade `ClassifierChannelEnsemble` to work with the changed mixin.
- removes the deprecated `WeightedEnsembleClassifier` class
- Adds tests to verify that expected results arrays are actually being used

part of #1518


TonyBagnall:
have you checked coverage on the related classes? It was previously low
image

image

MatthewMiddlehurst:
cant check currently

TonyBagnall (approved):

Issue #2245, opened 2024-10-24
[BUG] BORF failing without numba


TonyBagnall:
### Describe the bug

the BORF transformer fails when numpy turned off, in for example, the overnight codecov tests here

https://github.com/aeon-toolkit/aeon/actions/runs/11491231670

its standard numba stuff, should be quite easy to track down

### Steps/Code to reproduce the bug

set  

Run numba-disabled codecov tests

on any PR

### Expected results

pass test

### Actual results

```python-traceback
FAILED aeon/testing/tests/test_all_estimators.py::test_all_estimators[check_fit_updates_state(estimator=BORF(),datatype=EqualLengthUnivariate-Classification-numpy3D)] - numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Untyped global name '_erfinv': Cannot determine Numba type of <class 'function'>

File "aeon/transformations/collection/dictionary_based/_borf.py", line 617:
def _ppf(x, mu=0, std=1):
    return mu + math.sqrt(2) * _erfinv(2 * x - 1) * std
    ^
FAILED aeon/testing/tests/test_all_estimators.py::test_all_estimators[check_non_state_changing_method(estimator=BORF(),datatype=EqualLengthUnivariate-Classification-numpy3D)] - numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Untyped global name '_erfinv': Cannot determine Numba type of <class 'function'>

File "aeon/transformations/collection/dictionary_based/_borf.py", line 617:
def _ppf(x, mu=0, std=1):
    return mu + math.sqrt(2) * _erfinv(2 * x - 1) * std
    ^
= 2 failed, 8705 passed, 10 skipped, 61 xfailed, 150 xpassed, 636 warnings in 1781.64s (0:29:41) =
```

Cyril-Meyer:
I've investigated this bug and here is some information I found useful :

* You can reproduce the bug easier
  * Set `NUMBA_DISABLE_JIT=1`
  * Run a copy of `_borf.py` with a call like `_get_norm_bins(16)`
* `_ppf` is `@nb.vectorize` (and its the only one in all aeon)
* If you change `_erfinv` decorator from `njit` into `vectorize`, the error disappears (*this solution looks good to me*)
  * Performance difference is negligible, here is code for test
    * vectorize = 3.88
    * njit = 3.76
```python
import time
_get_norm_bins(5)
t0 = time.time()
[...]

PR #2254, opened 2024-10-25
[BUG] BORF failing without numba

Cyril-Meyer:
#### Reference Issues/PRs

Fixe proposition for #2245.

Replace `_erfinv` decorator from `njit` into `vectorize`.
More details in https://github.com/aeon-toolkit/aeon/issues/2245#issuecomment-2438337161

No.

#### Any other comments?

Only tested locally, need to be tested for real.

Run numba-disabled codecov tests

##### For all contributions
  * Send this when change is accepted `@all-contributors add @Cyril-Meyer for bug`

### DIFF
--- a/aeon/transformations/collection/dictionary_based/_borf.py
+++ b/aeon/transformations/collection/dictionary_based/_borf.py
@@ -584,7 +584,7 @@ def _get_norm_bins(alphabet_size: int, mu=0, std=1):
     return _ppf(np.linspace(0, 1, alphabet_size + 1)[1:-1], mu, std)
 
 
-@nb.njit(fastmath=True, cache=True)
+@nb.vectorize(fastmath=True, cache=True)
 def _erfinv(x: float) -> float:
     w = -math.log((1 - x) * (1 + x))
     if w < 5:


MatthewMiddlehurst:
Is the output of the function the same with the decorator changed? Never used `vectorize`, but just changing it to njit was not good enough in my fix attempt.

Cyril-Meyer:
Thanks for adding the tag, fix seems to work, but cannot be sure that everything is fine as `pyod_model` error is here.

> Is the output of the function the same with the decorator changed?

I do not have experience with `vectorize` but I've checked that this code do the same output with or without it.
```python
for i in range(256):
    print(_get_norm_bins(i))
```

> Never used `vectorize`, but just changing it to njit was not good enough in my fix attempt.

As much as I have understood yet, vectorize is an intermediate solution between doing nothing and jit.
It allows the function to be used as a numpy `ufunc`.


I don't know if `fastmath=True, cache=True` make any difference in `vectorize` decorator.

MatthewMiddlehurst:
Seems you still have the error I had before. Other CI issues should be gone now.

Cyril-Meyer:
> Seems you still have the error I had before. Other CI issues should be gone now.

Alright, here is another try.
Instead of using more `vectorize`, lets abandon them.

1. Replaced `math.mathoperation` with `np.mathoperation`, this should not have negative impact.
2. Everything is njit and `_get_norm_bins` now create an array manually with a for loop
3. I checked performance, and it's not a disaster (from 3.76/3.88 to 4.02)
4. I checked quickly and it does not seems to change outputs from the current version.

TonyBagnall:
seems there is also a problem with the bit operators, @fspinna any ideas how to make this work without numba?

fspinna:
> seems there is also a problem with the bit operators, @fspinna any ideas how to make this work without numba?

@TonyBagnall this seems to work for me (see https://stackoverflow.com/questions/30513741/python-bit-shifting-with-numpy):

```python
import numpy as np
import numba as nb
@nb.njit(cache=True)
def _hash_function2(v):

    byte_mask = np.uint64(255)
    bs = np.uint64(v)
    x1 = (bs) & byte_mask
    x2 = (bs >> np.uint64(8)) & byte_mask
    x3 = (bs >> np.uint64(16)) & byte_mask
    x4 = (bs >> np.uint64(24)) & byte_mask
    x5 = (bs >> np.uint64(32)) & byte_mask
    x6 = (bs >> np.uint64(40)) & byte_mask
    x7 = (bs >> np.uint64(48)) & byte_mask
    x8 = (bs >> np.uint64(56)) & byte_mask

    FNV_primer = np.uint64(1099511628211)
    FNV_bias = np.uint64(14695981039346656037)
    h = FNV_bias
    h = h * FNV_primer
    h = h ^ x1
    h = h * FNV_primer
    h = h ^ x2
    h = h * FNV_primer
    h = h ^ x3
    h = h * FNV_primer
    h = h ^ x4
    h = h * FNV_primer
    h = h ^ x5
    h = h * FNV_primer
    h = h ^ x6
    h = h * FNV_primer
    h = h ^ x7
    h = h * FNV_primer
    h = h ^ x8
    return h
```

but raises warnings: `RuntimeWarning: overflow encountered in scalar multiply h = h * FNV_primer...`

MatthewMiddlehurst:
@fspinna incorporated your fix and a few other bits. Warning is a bit concerning, but the numba compiled code would have the same issue with overflows I'm pretty sure?

MatthewMiddlehurst:
@fspinna I have a feeling the numba code just does not raise a warning 🙂.  It will very happily let you go out of bounds for arrays as well, which is part of why we run these. I know regular python ints can be flexible and go beyond 64 bits, but like the numpy dtypes numba requires specific dtype sizes which can overflow.

Not sure if it requires more investigation, but will approve this for now.

MatthewMiddlehurst (approved):

TonyBagnall (approved):

PR #2255, opened 2024-10-27
[ENH] Adding LITEMV ability to LITE

hadifawaz1999:
Adding a bool parameter `use_litemv` to LITE network, LITE classifier and regressor as well as LITETime classifier and regressor, for using the specific multivariate version, [paper for LITE Multivariate](https://arxiv.org/abs/2409.02869)


MatthewMiddlehurst:
I thought we agreed to separate models in the end? Maybe I am misremembering this, did we write down notes?

hadifawaz1999:
> I thought we agreed to separate models in the end? Maybe I am misremembering this, did we write down notes?

maybe i miss understood, and that we wont separate them but instead add an option for it

hadifawaz1999:
> I thought we agreed to separate models in the end? Maybe I am misremembering this, did we write down notes?

after the meeting i was actually convinced that making it an option is better.

MatthewMiddlehurst:
Possible I am just mistaken. This seems fine to me if it is for you. I would double-check the online API that the description includes both

hadifawaz1999:
> Possible I am just mistaken. This seems fine to me if it is for you. I would double-check the online API that the description includes both

how to check online api on a pr ? never tried it:p

MatthewMiddlehurst:
click on details on the read the docs test. Just a link to the website generated for this PR

TonyBagnall (dismissed):
looks good, but could you add tests for this and the other options
https://app.codecov.io/gh/aeon-toolkit/aeon/blob/main/aeon%2Fclassification%2Fdeep_learning%2F_lite_time.py

image

TonyBagnall (approved):

PR #2256, opened 2024-10-27
[MNT] Remove python 3.12 constraint for deep learners

hadifawaz1999:
trying to remove constraint of python 3.12 for testing deep learners, fixing as well an old issue related to Encoder network removing the need of lamda layer


TonyBagnall:
good stuff, is there an issue related to the old issue with encoders? If so, could you tag it?

aeon/networks/base.py
```diff
@@ -4,23 +4,18 @@
...
-    _check_soft_dependencies,
-)
+from aeon.utils.validation._dependencies import _check_soft_dependencies
 
 
 class BaseDeepLearningNetwork(ABC):
     """Abstract base class for deep learning networks."""
 
-    def __init__(self, soft_dependencies="tensorflow", python_version="<3.12"):
+    def __init__(self, soft_dependencies="tensorflow"):
```
MatthewMiddlehurst, line 13: I would keep this but just default it to `None` or somthing

aeon/networks/tests/test_all_networks.py
```diff
@@ -21,12 +18,10 @@
 def test_network_config(network):
     """Tests if the config dictionary of classes is correctly configured."""
     assert "python_dependencies" in network._config.keys()
-    assert "python_version" in network._config.keys()
```
MatthewMiddlehurst, line 24: same with all this stuff

hadifawaz1999: done

aeon/networks/base.py
hadifawaz1999: done added <=3.12 by default

MatthewMiddlehurst (approved):
Made a few changes which look like they were left over. PR looks fine to me now, but feel free to change if i made any mistakes above.

hadifawaz1999:
> Made a few changes which look like they were left over. PR looks fine to me now, but feel free to change if i made any mistakes above.

thanks some i forgot probably to add when reverting the python version, the encoder thing is not actually necessary as it is the same as the one of base class which is why i removed it, but does not hurt

hadifawaz1999:
@MatthewMiddlehurst seems <=3.12 should be <3.13 because we actually use 3.12.7 ! will fix that thanks for the changes CI wouldnt have detected them

hadifawaz1999:
pushed the changes to <3.13 and added the config in every network, even if it inherits the base class config, better to make it clear for each one, to be consistent as AE based ones we re-do the config because the structure element is different

MatthewMiddlehurst (approved):

PR #2271, opened 2024-10-29
[DOC] Remove glossary

TonyBagnall:
removes the glossary.md in docs directory. As far as I can tell this is only linked from getting_started.md and it contains a lot of redundant and misleading definitions, such as "Panel" and annotation. I think it best to just get rid, seems an odd thing to have, I guess scikit have one ...


MatthewMiddlehurst:
could you do a search for "term" to make sure there arent any remaining references if we do this. i.e. :term:\`Glossary <n_jobs>\`

TonyBagnall:
> could you do a search for "term" to make sure there arent any remaining references if we do this. i.e. :term:`Glossary <n_jobs>`

already did "As far as I can tell this is only linked from getting_started.md" as far as I cant tell is because I do not trust pycharm find that much

TonyBagnall:
> Glossary <n_jobs>

there is a reference to the scikit glossary here
"""Check `n_jobs` parameter according to the scikit-learn convention.

    https://scikit-learn.org/stable/glossary.html#term-n_jobs

is that the one you meant?

MatthewMiddlehurst (approved):

PR #2273, opened 2024-10-30
[DOC,ENH] `base` docs and testing

MatthewMiddlehurst:
Testing and documentation for base classes. Refactors a few usages of `get_class_tag` where it is unnecessary and a couple of other minor refactors which explode the file change count.


aeon/base/_base.py
```diff
@@ -221,7 +226,7 @@ def get_tags(self):
         collected_tags.update(self._tags_dynamic)
         return deepcopy(collected_tags)
 
-    def get_tag(self, tag_name, tag_value_default=None, raise_error=True):
+    def get_tag(self, tag_name, raise_error=True, tag_value_default=None):
```
TonyBagnall, line 229: these are the default arguments and not needed

aeon/testing/mock_estimators/_mock_classifiers.py
```diff
@@ -41,26 +43,27 @@ class MockClassifierFullTags(MockClassifierPredictProba):
...
+    """Mock classifier for testing fit/predict with multiple parameters.
 
     Parameters
     ----------
     return_ones : bool, default=False
         If True, predict ones, else zeros.
     """
 
-    def __init__(self, return_ones=False):
+    def __init__(self, return_ones=False, value=50):
```
TonyBagnall, line 55: why does it have a parameter value now?

TonyBagnall (approved):
LGTM

aeon/testing/mock_estimators/_mock_classifiers.py
MatthewMiddlehurst: just a useless param to help with testing

aeon/base/_base.py
MatthewMiddlehurst: little confused, are there calls to the method which input the default?

TonyBagnall: yes comment in wrong place! There was one call with explicit default parameters, but it doenst matter

Issue #1036, opened 2024-01-12
[ENH] Adapt `plot_series` to accept numpy arrays and list of numpy


hadifawaz1999:
### Describe the feature or idea you want to propose

For now the function `plot_series` accepts only pd.Series, which is not the best approach because if someone works with numpy than the code will crash. 

### Describe your proposed solution

It should be adapted to work with 3d numpy, 2d numpy and list of 2d numpy

MatthewMiddlehurst:
I think `plot_series` only work with a singular time series, though that may be wrong. May need a separate functions for multi-dimensional series/collections.

hadifawaz1999:
from what i understand in the docs @MatthewMiddlehurst it can work on a collection

Cyril-Meyer:
> from what i understand in the docs @MatthewMiddlehurst it can work on a collection

It does :

```python
y1 = load_airline()
y2 = y1 ** 0.99
y3 = y1 ** 0.97

fig, ax = plot_series(y1, y2, y3)
plt.show()
```

![image](https://github.com/user-attachments/assets/4266e0e0-8d3b-47dd-8ebd-8a309399e07e)

> It should be adapted to work with 3d numpy, 2d numpy and list of 2d numpy

Working with 1d numpy array should be fine, we can add a check like `if numpy and 1d: x=pd.Series(x)` :

```python
fig, ax = plot_series(pd.Series(np.array(y1)), pd.Series(np.array(y2)), pd.Series(np.array(y3)))
plt.show()
```

![image](https://github.com/user-attachments/assets/45ff4fd7-1a0d-4270-b020-2da9fabf3d7c)

Working with 2d numpy array should also be fine, we can add a check add a conversion function : 
```python
def pd_s_to_np_2d(series):
    index_values = series.index.values
    series_values = series.values
    array_2d = np.column_stack((index_values, series_values))
    return array_2d

def np_2d_to_pd_s(array_2d):
    index_values = array_2d[:, 0]
    series_values = array_2d[:, 1]
    series = pd.Series(series_values, index=index_values)
    return series

fig, ax = plot_series(np_2d_to_pd_s(pd_s_to_np_2d(y1)), np_2d_to_pd_s(pd_s_to_np_2d(y2)), np_2d_to_pd_s(pd_s_to_np_2d(y3)))
plt.show()

```

![image](https://github.com/user-attachments/assets/208086a3-466f-4a75-95f2-0a1b4d9de2df)

@hadifawaz1999 
[...]

PR #2279, opened 2024-10-31
[ENH] Rework series plotting to work with numpy arrays

### LINKED ISSUE #2278: [ENH] refactor visualisation/series/_series.py functions

plot functions for series are all based around pd.Series and pd.DataFrame and do not work with numpy arrays. Should refactor to at least work with numpy and also to default to wide format for multivariate

plot_series
plot_lags
plot_correlations
plot_spectrogram

protected functions
_plot_interval
_check_colors

these all require pd.series, should all just work by allowing numpy and internally converting

_No response_

### Additional context

_No response_

TonyBagnall:
Fixes #2278 

the two univariate series plot functions  plot_lag and plot_correlation now work with np.ndarray as well as pd.Series. 

plot_series is the only one that takes multivariate. It was configured to plot either a single series or a list of series, thus making it in effect a duplicate of plot_collection. I have changed this so that it can now only plot a (possibly multivariate) series of valid type, e.g. np.ndarray, pd.Series or pd.DataFrame of shape `(n_channels, n_timepoints)`. It will not work with a collection. 

I have removed testing that was really in effect testing the legacy function check_y.


aeon/visualisation/series/_series.py
```diff
@@ -66,13 +72,27 @@ def plot_series(
     from matplotlib.cbook import flatten
     from matplotlib.ticker import FuncFormatter, MaxNLocator
 
+    # Convert single multivariate series in wide format to list of series
+    if isinstance(series, pd.DataFrame):
+        series = [row for _, row in series.iterrows()]
+    elif isinstance(series, pd.Series):
+        series = [series]
+    elif isinstance(series, np.ndarray):
```
baraline, line 80: In this check, don't you need to check if series is not 3d array ?

aeon/visualisation/series/tests/test_series_plotting.py
```diff
@@ -59,64 +44,51 @@ def test_plot_series():
...
-            _plot_series(series_to_plot)
-    else:
-        with pytest.raises(ValueError, match="input must be univariate"):
-            _plot_series(series_to_plot)
+    series_to_plot = "This is a string"
+    with pytest.raises((TypeError), match="found type: <class 'str'>"):
+        plot_series(series_to_plot)
+    series_to_plot = (pd.DataFrame({"y1": y_airline, "y2": y_airline}),)
+    with pytest.raises(ValueError, match="input must be univariate"):
+        plot_series(series_to_plot)
```
baraline, line 92: If 3d is not allowed could bé good to add a check here

aeon/transformations/series/_dobin.py
```diff
@@ -64,18 +64,18 @@ class Dobin(BaseSeriesTransformer):
     >>> import numpy as np
     >>> import pandas as pd
     >>> from aeon.datasets import load_uschange
-    >>> _, X = load_uschange()
-    >>> scaler = MinMaxScaler()
-    >>> X = scaler.fit_transform(X)
+    >>> X = load_uschange()
+    >>> min = MinMaxScaler()
```
baraline, line 68: Minor but I would call it minmax rather than just min for clarity

baraline (commented):
Minor comment and a check question, otherwise lgtm

MatthewMiddlehurst:
conflicted

aeon/visualisation/series/_series.py
TonyBagnall: its a plot function for a series, but yes a check function its a valid type is a good idea

TonyBagnall:
so its a little more complex than I thought, this function was being used to print lists of pd.Series with different axis, thus allowing two series to have different offsets. AS I have redone it, it is for a single series and I think lists of series should be a separate function.

hadifawaz1999:
mentioning [this](https://github.com/aeon-toolkit/aeon/issues/1036) here as it is related

aeon/visualisation/series/_series.py
```diff
@@ -60,19 +62,39 @@ def plot_series(
...
+    s = []
     for y in series:
-        check_y(y, allow_index_names=True)
-
-    series = list(series)
+        if isinstance(y, np.ndarray):
+            y = y.squeeze()
+            y = pd.Series(y)
+        s.append(y)
+    series = s
```
baraline, line 95: This felt weird has squeezing shouldn't have any effect here and this code could be replaced by calling pd.series in the if condition for ndarrays above. Otherwise for series/dataframes it does nothing I think ?

baraline (approved):
Think it is good to only accept single series in this function, so I agree with the changes. Rationalizing the index type checks also looks better. 
Just one non-blocking comment that doesn't impact correctness, feel free to address it or not.

aeon/visualisation/series/_series.py
TonyBagnall: goof spot is left over from when a list/collection could be passed, so it was previously ok to pass a list of numpy.

TonyBagnall:
> mentioning [this](https://github.com/aeon-toolkit/aeon/issues/1036) here as it is related

good spot, I think this fixes #1036, list of numpy would be with plot_collection

chrisholder (approved):

PR #2287, opened 2024-10-31
[DOC] TSC notebook

TonyBagnall:
a few minor corrections


baraline:
The phrasing is a bit weird here and got me confused :
`
We use 3D numpy even if the data is univariate: all classifiers work with shape (n_cases, n_timepoints), this can get confused with single multivariate time series, which are of shape (n_channels, n_timepoints).
`
Did you mean something like this? 
`
We use 3D numpy even if the data is univariate: even though a univariate classifier could work using a 2D array of shape (n_cases, n_timepoints), this 2D shape can get confused with single multivariate time series, which are of shape (n_channels, n_timepoints). Hence, to differentiate both cases, we enforce the 3D format (n_cases, n_channels, n_timepoints) to avoid any confusion.
`

hadifawaz1999 (approved):
LGTM, would be nice to add a deep learning section in the notebook as well at some point, i can raise an issue for it once this is in

PR #2305, opened 2024-11-05
[BUG] Fixed subsampling in highly imbalances datasets giving subsamples with only a single class

### LINKED ISSUE #1726: [BUG] HC2 sporadic CI failure
### Describe the issue

See https://github.com/aeon-toolkit/aeon/actions/runs/9676429469/job/26695950355

Appears to be related to TDE `fit_predict`.

### Additional context

_No response_

ferewi:
#### Reference Issues/PRs
Fixes #1726 

Added new attribute 'max_subsamples' to subsample multiple times in case of unbalanced datasets giving subsamples with only one class. This caused TDE and HC2 to fail.
If a subsample with only a sinlge class is found, subsampling is repeated until the number of classes in the subsample is >1 or until max_subsamples is reached. In the latter case an AttributeError is raised.

No

#### Any other comments?
In the Issue I proposed two ideas how to fix this and I went for a modified option 1 that does not lead to potentially infinite loops as it is more than 3 times faster compared to option 2 (StratifiedShuffleSplit).

### PR checklist
@all-contributors add @ferewi for code

##### For all contributions

does not apply

does not apply


MatthewMiddlehurst (commented):
Is there a situation where you wouldn't want to continue until there is a viable subsample? I would just use a while loop and continue until you do. 

The only time this will backfire is if there is only 1 class in the input, but we should catch that higher up in input verification IMO.

ferewi:
I was a bit caught up last week hence my late reply. @MatthewMiddlehurst I think your're right, the new parameter is actually not needed. I was thinking that given a large unbalanced dataset, the chance for drawing a subsample that contains just one class is high, but given the subsample size of 70% this chance is around 0.5 in the worst case. So just resampling until a valid sample is obtained should be fine.

MatthewMiddlehurst (approved):

MatthewMiddlehurst:
LGTM, thanks.

MatthewMiddlehurst:
@all-contributors add @ferewi for bug

PR #2316, opened 2024-11-07
[ENH] Add DisjointCNN classifier and Regressor

### LINKED ISSUE #1902: [ENH] Implement Disjoint-CNN deep learning for classification and regression and adding mutlvariate specificity tag

Having specific multivariate deep learning models is a good thing, i like to have Disjoint-CNN, Monash's multivariate CNN model [1] into aeon, its implementation is already quite clear and easy and is in tensorflow [here](https://github.com/Navidfoumani/Disjoint-CNN)

[1] Foumani, Seyed Navid Mohammadi, Chang Wei Tan, and Mahsa Salehi. "Disjoint-cnn for multivariate time series classification." 2021 International Conference on Data Mining Workshops (ICDMW). IEEE, 2021.

- The DisjointCNNNetwork class should be added to the network class but parametrized to number of layers, number of filters, kernel size per layer. HINT: look on how already FCN ResNet and all other CNN networks are already implemented

- Adding The DisjointCNNClassifier and DisjointCNNRegressor classes to deep classification/regression models

- Add a tag for the networks module, deep classification and regression modules called "multivariate_univariate_duality" or something like that: 
- 1. if this flag is True, means that it works fine with both
- 2. if its false, it means it was proposed for multivariate, it can technically work on univariate but doesnt make sense, case example: DisjointCNN

_No response_

### Additional context

i already have a code ready for it, adding this issue to remember the tagging thing, assigning myself

hadifawaz1999:
Fixes #1902 

Check list:

- [x] Add network
- [x] add test for network for coverage
- [x] add classifier
- [x] add regressor

Also doing a huge refactor on docs for deep learning while at it, old flaw in docstrings fixed now

### DIFF
--- a/aeon/classification/deep_learning/__init__.py
+++ b/aeon/classification/deep_learning/__init__.py
@@ -12,8 +12,10 @@ __all__ = [
     "TapNetClassifier",
     "LITETimeClassifier",
     "IndividualLITEClassifier",
+    "DisjointCNNClassifier",
 ]
 from aeon.classification.deep_learning._cnn import TimeCNNClassifier
+from aeon.classification.deep_learning._disjoint_cnn import DisjointCNNClassifier
 from aeon.classification.deep_learning._encoder import EncoderClassifier
 from aeon.classification.deep_learning._fcn import FCNClassifier
 from aeon.classification.deep_learning._inception_time import (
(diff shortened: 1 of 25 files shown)


TonyBagnall (approved):
nice work! It may have been better to split the reformatting and rewording, all of which looks good, from the introduction of a new estimator, on which I cant really comment, but LGTM

Issue #1886, opened 2024-08-01
[ENH] n_jobs/_n_jobs, parameter in classifiers


TonyBagnall:
the use of n_jobs/_n_jobs  in CollectionEstimators is confusing and could be tidied up.

See https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/_base_collection.py#L92 where `_n_jobs` is defined. This is not really enforced in any way currently.

baraline:
May be related to #1797

lucifer4073:
Hi, is this issue resolved ?

MatthewMiddlehurst:
No i don't think so.

lucifer4073:
> the use of n_jobs/_n_jobs in CollectionEstimators is confusing and could be tidied up.
> 
> See https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/_base_collection.py#L92 where `_n_jobs` is defined. This is not really enforced in any way currently.

Interest to work with this one. I plan to clarify the usage of _n_jobs which will be overwritten by n_jobs if defined in child class.

PR #2317, opened 2024-11-07
[MNT] Improve multithreading testing

MatthewMiddlehurst:
Adds tests for multithreading and `n_jobs` the test which runs multithreaded estimators is not enabled anywhere currently. Adds tags and changes defaults in a bunch of estimators.


aeon/testing/estimator_checking/_yield_multithreading_checks.py
```diff
@@ -0,0 +1,115 @@
...
+    if can_thread and MULTITHREAD_TESTING:
+        # test class instances
+        for i, estimator in enumerate(estimator_instances):
+            # test all data types
+            for datatype in datatypes[i]:
+                yield partial(
+                    check_estimator_multithreading,
+                    estimator=estimator,
+                    datatype=datatype,
+                )
```
SebastianSchmidl, line 34: Have you run this locally? Does it succeed for all our estimators?

aeon/testing/estimator_checking/_yield_multithreading_checks.py
```diff
@@ -0,0 +1,115 @@
...
+    # fit multithreaded estimator
+    _run_estimator_method(mt_estimator, "fit", datatype, "train")
+
+    # check n_jobs_ attribute is set
+    assert mt_estimator.n_jobs_ == n_jobs, (
+        f"Multithreaded estimator {mt_estimator} does not store n_jobs_ "
+        f"attribute correctly. Expected {n_jobs}, got {mt_estimator.n_jobs_}."
+        f"It is recommended to use the check_n_jobs function to set n_jobs_ and use"
+        f"this for any multithreading."
+    )
```
SebastianSchmidl, line 101: I guess, this would fail for the current implementations of the AD module. I think they are not using `check_n_jobs` because they just pass the parameter down to PyOD.
Can you check this?

SebastianSchmidl (approved):
I have not checked if you got all the estimators to set their tag `"capability:multithreading"`, otherwise looks good 👍🏼

SebastianSchmidl:
Maybe we can add an option to the aeon actions bot to run these tests on a PR?

MatthewMiddlehurst:
That is the plan eventually. Yeah, this would fail on more than just the anomaly detectors currently, but similar to the type checking stuff It is more about making the framework available for when someone is a bit more motivated to tackle multithreading.

The main stuff I wanted to make sure is tested right now are the parameters and tags so we don't have to do a massive refactor down the line if anyone does ever take it on 🙂. Getting rid of that one commented out test is also a nice bonus.

SebastianSchmidl:
Yes, not of high priority, and this already improves consistency considerably! 👍🏼 

I'd like to see some way to track the progress and open issues for this topic. I did not locate an issue tracking the work for improving the multithreading. Maybe, you could create an issue and add the already identified issues there?

This also seems related: #1886

MatthewMiddlehurst:
Yep no problem I can create an issue for this. 

#1886 is more surrounding these lines IMO (https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/_base_collection.py#L92) which is definitely related but good as its own issue

PR #2319, opened 2024-11-08
[ENH] Move performance metrics to benchmarking

TonyBagnall:
I dont feel strongly about this, its ok for them to remain where they are if anyone does, but on balance I favour keeping the root as slim as possible and encapsulating where we can, performance_metrics seems better in benchmarking to me. 

update: pycharm refactoring is so bad!


TonyBagnall:
refactoring too tedious

PR #2321, opened 2024-11-08
[BUG] Fix bug_report.yml template show_versions

SebastianSchmidl:
The `show_versions()`-utility is in the package `aeon` or `aeon.utils._maint` and cannot be imported from `aeon.utils`. This fixes the template to import from `aeon`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -80,7 +80,7 @@ body:
       Please run the following code snippet and paste the output here.
 
       ```python
-      from aeon.utils import show_versions; show_versions()
+      from aeon import show_versions; show_versions()
       ```
     placeholder: |
       <details>


TonyBagnall (approved):

TonyBagnall (changes requested):
actually just tried this and the function seems to still be in utils?

TonyBagnall:
show_versions returns a string, so it should be

```python
from aeon.utils import show_versions
print(show_versions())
```
I think there is some legacy bug here, there is an argument to return string, default to false, but it is not used
```python

def show_versions(as_str: bool = False) -> Union[str, None]:
    """Print useful debugging information.

    Parameters
    ----------
    as_str : bool, default=False
        If True, return the output as a string instead of printing.

    Returns
    -------
    str or None
        The output string if `as_str` is True, otherwise None.

    Examples
    --------
    >>> from aeon.utils import show_versions
    >>> vers = show_versions(as_str=True)
    """
    sys_info = {
        "python": sys.version.replace("\n", " "),
        "executable": sys.executable,
        "machine": platform.platform(),
    }
    str = "\nSystem:"
    for k, stat in sys_info.items():
        str = f"{str}\n{k:>10}: {stat}"

    deps_info = {"aeon": __version__}
    for modname in deps:
        try:
            deps_info[modname] = version(modname)
        except PackageNotFoundError:
            deps_info[modname] = None
    str = f"{str}\nPython dependencies:"
    for k, stat in deps_info.items():
        str = f"{str}\n{k:>13}: {stat}"  # noqa: T001, T201
    return str
```

CodeLionX:
Ah, @MatthewMiddlehurst changed that in #2167. For all published versions (thus all `pip install`s of aeon), however, the issue template instructions do not work. I doubt that our users use the current `main`.

MatthewMiddlehurst (approved):
yeah fair point. overlooked that.

TonyBagnall (approved):

PR #2326, opened 2024-11-08
[ENH] Whole-series anomaly detection

### LINKED ISSUE #2662: [ENH] Base class for whole-series (collection) anomaly detection

While the primary goal of our anomaly detection (AD) module is to detect anomalous points or subseries in single time series, there is also a task for whole-series AD. In whole-series AD you have a collection of series, and the goal is to select the label the series which are anomalous. This can be similar to rare-class classification in some cases, but there are also unsupervised methods.

Implement framework for whole-series anomaly detection as a submodule of `anomaly_detection`. This includes a base class extending from `BaseCollectionEstimator`.

I do not believe this is large (or different) enough to warrant its own module. An AD submodule seems like the correct way to start this. 

### Additional context

_No response_

MatthewMiddlehurst:
Closes #2662

Base class for "whole-series" anomaly detection where you have a collection of series, and the goal is to select the label the series which are anomalous.


TonyBagnall:
could you clarify how this differs from imbalanced classification?

SebastianSchmidl:
@TonyBagnall 
In theory, whole-series AD is an unsupervised task. It neither is clustering because we are not interested in certain groups of time series, but rather in the outlying ones. If the detector requires a training step, I would also just classify it as imbalanced classification.

aeon/anomaly_detection/whole_series/base.py
```diff
@@ -0,0 +1,199 @@
...
+        Must match the number of cases in X.
+        """
+        if not isinstance(y, (pd.Series, np.ndarray)):
+            raise TypeError(
+                f"y must be a np.array or a pd.Series, but found type: {type(y)}"
+            )
+        if isinstance(y, np.ndarray) and y.ndim > 1:
+            raise TypeError(f"y must be 1-dimensional, found {y.ndim} dimensions")
+
+        if not all([x == 0 or x == 1 for x in y]):
```
SebastianSchmidl, line 184: `not np.bitwise_or(y == 0, y == 1).all()` would probably be faster. You already checked for `pd.Series` or `np.ndarray` above, so this should be supported by the types. You could also move the conversion to numpy from L197f before this check.

SebastianSchmidl (commented):
Should we allow supervised usage for whole-series detection?
As I have mentioned in my reply to Tony's comment, I would argue that whole-series AD should be unsupervised. Otherwise, we could just use any time series classifier from aeon for this imbalanced binary-classification task.
Maybe, we should include Patrick M.'s and Andreas' view here as well. E.g., their [ROCKAD-algorithm](https://link.springer.com/chapter/10.1007/978-3-031-30047-9_33) is semi-supervised.

If we focus on unsupervised methods here, I would just have a `fit_predict`-method in the base class.

pattplatt:
> It is a recent change and based off a taxonomy. See #2663. I would keep everything inheriting from the collection base class in the `whole-series` module for now. No issue with additional subpackages from there is required.

> Yes that would diverge from how we design things at the moment. All the current anomaly detectors inherit from a base class which solely handles single series input, while this new base class inherits from one which deals with collections of series. I think eventually we may have a base class which handles collections in `fit` then single series in `predict`/`transform` etc, but I do not see us allowing single classes which switch between them.
> 
> We have algorithms which can be used for both classification and regression, but they are always separated into their own module. Similar case here IMO. If there is shared functionality, it would be best to have that in a separate file then call it from both implementations. If you can reduce from whole-series to pointwise then just calling the whole-series class/functions from the pointwise one sounds fine as well.
> 
> Always open for discussion, but unlikely to change soon IMO.

Ok then we create a new PR to move the current time point based ROCKAD back to /distance_based/_rockad.py and add the whole-series  to /whole_series/_rockad.py? If shared functionality can be put in a seperate file where could this file be located?

SebastianSchmidl:
## Regarding supervision

OK, I get the points for including all three cases (unsupervised, semi-supervised, and supervised) for whole series AD, despite duplication to classification. This mimics the learning types of TSAD then as well.

## Regarding folder structure

The current module folder is the root for all TSAD algorithms that are grouped into algorithm families (distance, encoding, distribution, etc.). The `whole_series` subfolder marks a new category of algorithms. Because we do not have many algorithms for whole series AD, there is no need for extra sub-subfolders. I would just put whole series ROCKAD in `anomaly_detection/whole_series/_rockad.py`.

@MatthewMiddlehurst, maybe we should have two subfolders in `anomaly_detection`: `series` (current TSAD stuff) and `collection` (the whole series AD estimators) and then branch-off from there?

```
aeon/anomaly_detection
├── series
│   ├── base.py
│   ├── tests
│   │   └── [...]
│   ├── distance_based
│   │   ├── _kmeans.py
│   │   └── [...]
│   └── [...]
├── collection
│   ├── base.py
│   ├── tests
│   │   └── [...]
│   ├── _rockad.py
│   └── [...]
└── __init__.py
```

Common functionality could then live in the base folder or in a separate one (e.g., `aeon/anomaly_detection/_common` or similar).

MatthewMiddlehurst:
Ok will try formatting like the above and see if there are any issues. Will also move the current ROCKAD implementation.

aeon/anomaly_detection/collection/base.py
```diff
@@ -0,0 +1,248 @@
...
+            "capability:unequal_length")`` is False, they cannot handle unequal
+            length input. In both situations, a ``ValueError`` is raised if X has a
+            characteristic that the estimator does not have the capability for is
+            passed.
+
+        Returns
+        -------
+        predictions : np.ndarray
+            1D np.array of float, of shape (n_cases) - predicted anomalies
+            indices correspond to instance indices in X
```
SebastianSchmidl, line 149: Are they required to be predictions (0/1) or are also anomaly scores (in [0, 1]) allowed?

aeon/anomaly_detection/collection/base.py
```diff
@@ -0,0 +1,248 @@
...
+            whether the corresponding subsequence is anomalous or its anomaly score.
+        """
+        if self.get_tag("requires_y"):
+            if y is None:
+                raise ValueError("Tag requires_y is true, but fit called with y=None")
+
+        # reset estimator at the start of fit
+        self.reset()
+
+        X = self._preprocess_series(X, axis, True)
```
SebastianSchmidl, line 195:
```suggestion
        X = self._preprocess_series(X, axis, store_metadata=True)
```
?

aeon/anomaly_detection/collection/base.py
```diff
@@ -0,0 +1,248 @@
...
+        Must match the number of cases in X.
+        """
+        if not isinstance(y, (pd.Series, np.ndarray)):
+            raise TypeError(
+                f"y must be a np.array or a pd.Series, but found type: {type(y)}"
+            )
+        if isinstance(y, np.ndarray) and y.ndim > 1:
+            raise TypeError(f"y must be 1-dimensional, found {y.ndim} dimensions")
+
+        if not all([x == 0 or x == 1 for x in y]):
```
SebastianSchmidl, line 233: Same as above: `not np.bitwise_or(y == 0, y == 1).all()` should be faster

aeon/anomaly_detection/collection/base.py
```diff
@@ -0,0 +1,248 @@
...
+        y : one of aeon.base._base_series.VALID_SERIES_INPUT_TYPES, default=None
+            The target values for the time series.
+            A valid aeon time series data structure. See
+            aeon.base._base_series.VALID_SERIES_INPUT_TYPES for aeon supported types.
+        axis : int, default=1
+            The time point axis of the input series if it is 2D. If ``axis==0``, it is
+            assumed each column is a time series and each row is a time point. i.e. the
+            shape of the data is ``(n_timepoints, n_channels)``. ``axis==1`` indicates
+            the time series are in rows, i.e. the shape of the data is
+            ``(n_channels, n_timepoints)``.
```
SebastianSchmidl, line 180: Why is this interface different from `predict` and `fit`? Here, you allow specifying an `axis` and refer to the standard aeon data types.

aeon/anomaly_detection/series/distance_based/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+]
+
+from aeon.anomaly_detection.series.distance_based._cblof import CBLOF
+from aeon.anomaly_detection.series.distance_based._kmeans import KMeansAD
+from aeon.anomaly_detection.series.distance_based._left_stampi import LeftSTAMPi
+from aeon.anomaly_detection.series.distance_based._lof import LOF
+from aeon.anomaly_detection.series.distance_based._merlin import MERLIN
+from aeon.anomaly_detection.series.distance_based._rockad import ROCKAD
+from aeon.anomaly_detection.series.distance_based._stomp import STOMP
+from aeon.anomaly_detection.series.outlier_detection._one_class_svm import OneClassSVM
```
SebastianSchmidl, line 21: ❔

aeon/anomaly_detection/series/distance_based/_left_stampi.py
```diff
@@ -6,11 +6,11 @@
 
 import numpy as np
 
-from aeon.anomaly_detection.base import BaseAnomalyDetector
+from aeon.anomaly_detection.series.base import BaseSeriesAnomalyDetector
 from aeon.utils.windowing import reverse_windowing
 
 
-class LeftSTAMPi(BaseAnomalyDetector):
+class LeftSTAMPi(BaseSeriesAnomalyDetector):
```
SebastianSchmidl, line 13: Example import path not updated

aeon/anomaly_detection/series/distance_based/_stomp.py
```diff
@@ -7,11 +7,11 @@
 
 import numpy as np
 
-from aeon.anomaly_detection.base import BaseAnomalyDetector
+from aeon.anomaly_detection.series.base import BaseSeriesAnomalyDetector
 from aeon.utils.windowing import reverse_windowing
 
 
-class STOMP(BaseAnomalyDetector):
+class STOMP(BaseSeriesAnomalyDetector):
```
SebastianSchmidl, line 14: Example import path not updated

aeon/anomaly_detection/series/outlier_detection/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Time Series Outlier Detection."""
+
+__all__ = [
+    "IsolationForest",
+    "STRAY",
+]
+
+from aeon.anomaly_detection.series.outlier_detection._iforest import IsolationForest
+from aeon.anomaly_detection.series.outlier_detection._stray import STRAY
```
SebastianSchmidl, line 9: OneClassSVM is missing

aeon/utils/base/_register.py
```diff
@@ -32,26 +33,34 @@
...
 VALID_ESTIMATOR_BASES = {
     k: BASE_CLASS_REGISTER[k]
     for k in BASE_CLASS_REGISTER.keys()
-    - {"estimator", "collection-estimator", "series-estimator", "transformer"}
+    - {
+        "estimator",
+        "collection-estimator",
+        "series-estimator",
+        "transformer",
+        "anomaly-detector",
```
SebastianSchmidl, line 64: Shouldn't this rather be one of `series-anomaly-detector` or `collection-anomaly-detector`? The `AnomalyDetectorBase` is not meant to be inherited, right?

aeon/utils/tags/tests/test_discovery.py
```diff
@@ -42,8 +42,8 @@ def test_all_tags_for_estimator_anomaly_detection():
     assert "capability:contractable" not in tags
 
     assert tags == all_tags_for_estimator(MockAnomalyDetector)
-    assert tags == all_tags_for_estimator(BaseAnomalyDetector)
-    assert tags == all_tags_for_estimator("anomaly-detector")
+    assert tags == all_tags_for_estimator(BaseSeriesAnomalyDetector)
+    assert tags == all_tags_for_estimator("series-anomaly-detector")
```
SebastianSchmidl, line 46: Shouldn't there be something similar for `BaseCollectionAnomalyDetector`? Same for `aeon/utils/tests/test_discovery.py`

docs/api_reference/anomaly_detection.rst
```diff
@@ -3,25 +3,38 @@
...
+
+.. currentmodule:: aeon.anomaly_detection.collection
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseCollectionAnomalyDetector
+    ClassificationAdapter
+    OutlierDetectionAdapter
```
SebastianSchmidl, line 29:
```suggestion
    ClassificationAdapter
    OutlierDetectionAdapter
    BaseCollectionAnomalyDetector
```

The other bases are also at the bottom

SebastianSchmidl (changes requested):

SebastianSchmidl:
> Thanks for the review, will take a look and reply. Before that I will try get #2652 in though, as it may be a pain to do so after merging this.

Sure thing. #2652 is, however, still failing CI.

aeon/anomaly_detection/collection/base.py
MatthewMiddlehurst: scores also

aeon/testing/mock_estimators/_mock_anomaly_detectors.py
```diff
@@ -10,10 +10,10 @@
 
 import numpy as np
 
-from aeon.anomaly_detection.base import BaseAnomalyDetector
+from aeon.anomaly_detection.series.base import BaseSeriesAnomalyDetector
 
 
-class MockAnomalyDetector(BaseAnomalyDetector):
+class MockAnomalyDetector(BaseSeriesAnomalyDetector):
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Not for anything here but would be good to include eventually. Can do it here but pretty large already.

aeon/utils/tags/tests/test_discovery.py
MatthewMiddlehurst: Don't think we need one for every module, but would do it as a separate issue if so. Every module is included in the estimator discovery testing i believe, as a test parameter.

aeon/anomaly_detection/series/distance_based/_left_stampi.py
MatthewMiddlehurst: Fixed. I hate the examples that skip tests 🙂

MatthewMiddlehurst:
Merged the other AD stuff from main and applied most of the review suggestions. Mostly changes to testing from the merge.

aeon/anomaly_detection/series/distance_based/_left_stampi.py
SebastianSchmidl: I can relate 😓

aeon/utils/tags/tests/test_discovery.py
SebastianSchmidl: Ok

aeon/utils/base/_register.py
SebastianSchmidl: Is this fixed in the current version?

SebastianSchmidl (changes requested):

aeon/utils/base/_register.py
SebastianSchmidl: Ah, now I get it! This is the set of non-valid bases to inherit from, which is removed from the set of all base classes (`BASE_CLASS_REGISTER`). 👍🏼

PR #2327, opened 2024-11-08
[ENH] show versions option to print

### LINKED ISSUE #2323: [BUG] show_version() always returns a string
### Describe the bug

The function has an argument to either print or return a string, but always returns a string. See #2321 

``python
def show_versions(as_str: bool = False) -> Union[str, None]:
    """Print useful debugging information.

    Parameters
    ----------
    as_str : bool, default=False
        If True, return the output as a string instead of printing.

    Returns
    -------
    str or None
        The output string if `as_str` is True, otherwise None.

    Examples
    --------
    >>> from aeon.utils import show_versions
    >>> vers = show_versions(as_str=True)
    """
    sys_info = {
        "python": sys.version.replace("\n", " "),
        "executable": sys.executable,
        "machine": platform.platform(),
    }
    str = "\nSystem:"
    for k, stat in sys_info.items():
        str = f"{str}\n{k:>10}: {stat}"

    deps_info = {"aeon": __version__}
    for modname in deps:
        try:
            deps_info[modname] = version(modname)
        except PackageNotFoundError:
            deps_info[modname] = None
    str = f"{str}\nPython dependencies:"
    for k, stat in deps_info.items():
        str = f"{str}\n{k:>13}: {stat}"  # noqa: T001, T201
    return str
```

```python
from aeon.utils import show_versions
show_versions()
```

### Expected results

by default, it should print and above should do this
```python
from aeon.utils import show_versions
print(show_versions())
```

### Actual results

Nothing prints as it always returns a string

### Versions

_No response_

TonyBagnall:
fix #2323

### DIFF
--- a/aeon/utils/show_versions.py
+++ b/aeon/utils/show_versions.py
@@ -59,7 +59,7 @@ def show_versions(as_str: bool = False) -> Union[str, None]:
             deps_info[modname] = None
     str = f"{str}\nPython dependencies:"
     for k, stat in deps_info.items():
-        str = f"{str}\n{k:>13}: {stat}"  # noqa: T001, T201
+        str = f"{str}\n{k:>13}: {stat}"
     if as_str:
         return str
     print(str)  # noqa: T201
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst (dismissed):
Could you try removing the noqa's above the print? think i forgot the finishing touches on this one 🙂

TonyBagnall:
> Could you try removing the noqa's above the print? think i forgot the finishing touches on this one 🙂

I did spend 2 secs wondering about that :) Its gone

MatthewMiddlehurst (approved):

PR #2330, opened 2024-11-09
[MNT] Kill TapNet with fire

hadifawaz1999:
our implementation of tapnet is wrong, i prefer not re-work and simply restart from scratch at some point, finding errors line by line will consume more energy and time than starting from scratch as i recall lots of it were wrongly coded. Prefer purging it before 1.0

Issues i found in it:

- implementation of classifier and regressor not inline with all others, needs a huge refactor to be able to save/load, use always best model, random state usage
- usage of lambda layers which is no longer allowed, such as for Encoder we had to change it
- vary much bits of code that are not used just declared like : "conv_1_models" which is a declared keras sequential model but not used, "euclidean_dist" too
- the convolution operations are not applied as the paper is saying it should, the original way is doing RDP then one convolution block on each Group, concatenate them and share the rest of the network, the current implementation does the whole network on each Group independently. The usage of keras-self attention should be stopped as many new ways in keras exist now to do it, as keras self attention is no longer reliable [repository archive](https://github.com/CyberZHG/keras-self-attention)
- RDP implementation is not efficient with tf.gather, making custom layers with no trainable parameters would be more appropriate

For more information on TapNet, the [original paper](https://xuczhang.github.io/papers/aaai20_tapnet_full.pdf)


TonyBagnall:
we have discussed this off and on for a while so it makes sense to do this now before version 1.0

TonyBagnall:
removing the soft dependency on keras-self attention is a big plus too

TonyBagnall (approved):

hadifawaz1999:
TODO: remove tapenet from notebook results loaders

TonyBagnall (approved):

PR #2338, opened 2024-11-11
[ENH] Add and Validate `n_layers`, `n_units`, `activation` & `dropout_rate` kwargs to MLPNetwork

### LINKED ISSUE #2337: [ENH] Parametrize MLP Network, classifier and regressor

all networks in aeon are parameterized, on number of layers and parameters per layers, default values as the ones in their published associated paper.
would be nice to have this option for mlp too, handling list inputs as for fcn and others

simply do for mlp network like for fcn network, and then add the arguments to the classifier and regresdor that uses the network

_No response_

### Additional context

_No response_

aadya940:
Closes #2337 .


aeon/networks/_mlp.py
```diff
@@ -24,9 +37,56 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
     def __init__(
         self,
+        n_layers: int = 3,
+        n_units: Union[int, list[int]] = 200,
+        activation: Union[str, list[str]] = "relu",
+        dropout_rate: Union[int, list[int]] = None,
     ):
         super().__init__()
 
+        self._n_layers = n_layers
```
hadifawaz1999, line 47: better to declare all of them as self before defining the internal versions, so self.activation = activation and then define self._activation

aeon/networks/_mlp.py
```diff
@@ -24,9 +37,56 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
+
+        if isinstance(activation, str):
```
hadifawaz1999, line 49: better to declare all of them as self before defining the internal versions, so self.activation = activation and then define self._activation

aeon/networks/_mlp.py
```diff
@@ -24,9 +37,56 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
+            self._activation = [activation] * self._n_layers
+        elif isinstance(activation, list):
+            assert (
+                len(activation) == self._n_layers
+            ), "There should be an `activation` function associated with each layer."
+            assert all(
+                isinstance(a, str) for a in activation
+            ), "Activation must be a list of strings."
+            assert (
```
hadifawaz1999, line 58: asserting the len() twice

aeon/networks/_mlp.py
hadifawaz1999, line 47: no need to define private for this one with underscore as it wont change

aeon/networks/_mlp.py
```diff
@@ -24,9 +37,56 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
+                len(activation) == n_layers
+            ), "Activation list length must match number of layers."
+            self._activation = activation
+
+        if dropout_rate is None:
+            self._dropout_rate = [0.2] * self._n_layers
```
hadifawaz1999, line 64: the default value of dropout is not 0.2 in the original implementation (check main branch), its 0.1 then 0.2 then 0.2

aeon/networks/_mlp.py
```diff
@@ -11,6 +13,17 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
 
+    Parameters
+    ----------
+    n_layers : int, optional (default=3)
+        The number of dense layers in the MLP.
+    n_units : Union[int, list[int]], optional (default=200)
+        Number of units in each dense layer.
+    activation : Union[str, list[str]], optional (default='relu')
+        Activation function(s) for each dense layer.
+    dropout_rate : Union[int, float, list[Union[int, float]]], optional (default=None)
```
hadifawaz1999, line 24: i would obligate float always for dropout_rate and say that is should be between 0 and 1

aeon/networks/_mlp.py
```diff
@@ -24,9 +37,56 @@ class MLPNetwork(BaseDeepLearningNetwork):
...
+        elif isinstance(dropout_rate, (int, float)):
+            self._dropout_rate = [float(dropout_rate)] * self._n_layers
+        elif isinstance(dropout_rate, list):
+            assert (
+                len(dropout_rate) == self._n_layers
+            ), "There should be a `dropout_rate` associated with each layer."
+            assert all(
+                isinstance(d, (int, float)) for d in dropout_rate
+            ), "Dropout rates must be int or float."
+            assert (
```
hadifawaz1999, line 74: asserting on len twice

aeon/networks/_mlp.py
```diff
@@ -42,19 +102,18 @@ def build_network(self, input_shape, **kwargs):
...
-        # flattened because multivariate should be on same axis
         input_layer = keras.layers.Input(input_shape)
         input_layer_flattened = keras.layers.Flatten()(input_layer)
 
-        layer_1 = keras.layers.Dropout(0.1)(input_layer_flattened)
-        layer_1 = keras.layers.Dense(500, activation="relu")(layer_1)
-
-        layer_2 = keras.layers.Dropout(0.2)(layer_1)
-        layer_2 = keras.layers.Dense(500, activation="relu")(layer_2)
+        x = keras.layers.Dropout(self._dropout_rate[0])(input_layer_flattened)
```
hadifawaz1999, line 108: better to include this block in the loop
where each block in the loop will be one dropout layer followed by a dense layer

aeon/networks/_mlp.py
```diff
@@ -42,19 +102,18 @@ def build_network(self, input_shape, **kwargs):
...
-        layer_3 = keras.layers.Dropout(0.2)(layer_2)
-        layer_3 = keras.layers.Dense(500, activation="relu")(layer_3)
+        for idx in range(1, self._n_layers):
+            x = keras.layers.Dropout(self._dropout_rate[idx])(x)
+            x = keras.layers.Dense(
+                self._n_units[idx], activation=self._activation[idx]
+            )(x)
 
-        output_layer = keras.layers.Dropout(0.3)(layer_3)
+        output_layer = keras.layers.Dropout(0.3)(x)
```
hadifawaz1999, line 117: i would define a parameter called dropout_last default to 0.3, and document it in details for that this dropout layer is applied at the end

hadifawaz1999 (changes requested):
Thx for taking care of it.
In general:

- All parameters should be defined as private in constructor before starting anything
- the assertions and all the list checking should only be in build_network method not constructor, this will avoid causing issues on CI (check for example how fcn network is implemented)
- Also given we are parametrizing the network, the associated classifier and regressor should be parametrized as such (also docs) and use them when calling the network
- I left some other comments to check

aeon/networks/_mlp.py
aadya940: How to extrapolate it to `n_layers`? Should I do the follows?
```
self._dropout_rate = [0.1].extend([0.2] * self._n_layers - 1)
```

hadifawaz1999 (changes requested):
Looks great, just the original default number of units is actually 500 not 200, once this is changed we can get this in

aeon/classification/deep_learning/_mlp.py
```diff
@@ -21,6 +21,17 @@ class MLPClassifier(BaseDeepClassifier):
...
     ----------
+    n_layers : int, optional (default=3)
+        The number of dense layers in the MLP.
+    n_units : Union[int, List[int]], optional (default=200)
+        Number of units in each dense layer.
+    activation : Union[str, List[str]], optional (default='relu')
+        Activation function(s) for each dense layer.
+    dropout_rate : Union[float, List[Union[int, float]]], optional (default=None)
+        Dropout rate(s) for each dense layer. If None, a default rate of 0.2 is used.
+        Dropout rate(s) are typically a number in the interval [0, 1].
```
dschrempf, line 32: As far as I can see, if `dropout_rate` is `None`, it will be set to

```
            self._dropout_rate = [0.1]
            self._dropout_rate.extend([0.2] * (self.n_layers - 1))
```
I think we should adapt this info.

hadifawaz1999 (approved):

PR #2349, opened 2024-11-13
[MNT] `testing` remaining cleanup

MatthewMiddlehurst:
Cleans up some utility and data loading files from `testing`. Leaves `make_series` as that is being dealt with in #2310.

After this and #2310, the only remaining thing to do for the testing module is review the current checks.


TonyBagnall (approved):
LGTM

PR #2352, opened 2024-11-14
[ENH] Collection and series base tidy

MatthewMiddlehurst:
- makes `__init__` abstracts for the classes in `base`
- renames the base series data arrays to contain series in the var name
- check for soft dependencies in `BaseAeonEstimator` rather than in the base classes below
- remove `self.fit_time_` from some collection estimators and the collection base. This was not very consistently used.
- tidy up docs a bit


aeon/base/_base.py
```diff
@@ -44,12 +46,15 @@ class BaseAeonEstimator(BaseEstimator, ABC):
...
     }
 
+    @abstractmethod
     def __init__(self):
         self.is_fitted = False  # flag to indicate if fit has been called
         self._tags_dynamic = dict()  # storage for dynamic tags
 
         super().__init__()
 
+        _check_estimator_deps(self)
```
SebastianSchmidl, line 56: I did not know that this was possible, but seems reasonable for our use case.

More details in the _Notes_-box at the end of https://docs.python.org/3/library/abc.html#abc.abstractmethod

aeon/transformations/collection/base.py
```diff
@@ -45,8 +45,6 @@ class BaseCollectionTransformer(BaseCollectionEstimator, BaseTransformer):
     }
 
     def __init__(self):
-        self._estimator_type = "transformer"
-
         super().__init__()
```
SebastianSchmidl, line 48: Non-abstract?

The `BaseSeriesTransformer.__init__` is also non-bastract?

SebastianSchmidl (changes requested):

aeon/transformations/collection/base.py
MatthewMiddlehurst: Made the remaining base classes abstract

SebastianSchmidl (approved):

TonyBagnall (approved):

chrisholder (approved):

chrisholder (approved):

PR #2353, opened 2024-11-14
[REF] Refactor `performance_metrics` into `benchmarking.metrics` and use a local version of `sphinx-remove-toctrees`

MatthewMiddlehurst:
Moves performance metric and refactors the segmentation metrics into a single file.

Updates the benchmarking API file. While doing this I noticed `sphinx-remove-toctrees` had broke, See https://github.com/executablebooks/sphinx-remove-toctrees/issues/9. This PR introduces a fixed version locally.

### DIFF
--- a/docs/api_reference/benchmarking.rst
+++ b/docs/api_reference/benchmarking.rst
@@ -123,7 +123,7 @@ Segmentation
 Stats
 -----
 
-.. currentmodule:: aeon.benchmarking.metrics.stats
+.. currentmodule:: aeon.benchmarking.stats
 
 .. autosummary::
     :toctree: auto_generated/


docs/api_reference.md
```diff
@@ -20,7 +20,6 @@ api_reference/data_format
 api_reference/datasets
 api_reference/distances
 api_reference/networks
-api_reference/performance_metrics
```
SebastianSchmidl, line 23: Where is the documentation now?

SebastianSchmidl (changes requested):
I did not check whether there are other places still referencing the old metrics. Should be fine, though.

Are the metrics part of the public API? If yes, we should include them in the API reference.

docs/api_reference.md
MatthewMiddlehurst: So far with these big module changes I have waited for it to settle down a bit before going into the docs, but this is the last change I wanted to make to benchmarking really.

Will update it under `benchmarking` and make sure the rest of the module is there as well

SebastianSchmidl: Makes sense, thank you!

docs/api_reference/benchmarking.rst
```diff
@@ -3,14 +3,119 @@
 Benchmarking
 ============
 
-.. automodule:: aeon.benchmarking
-    :no-members:
-    :no-inherited-members:
+The ``aeon.benchmarking`` module contains tools and utilities for comparing and
+evaluating and time series models, loading stored results, and calculating performance
+metrics for a variety of tasks.
```
SebastianSchmidl, line 8: This sentence has probably too many `and`s. I don't get it.

docs/api_reference/benchmarking.rst
```diff
@@ -3,14 +3,119 @@
...
+
+.. currentmodule:: aeon.benchmarking.metrics.segmentation
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: function.rst
+
+    count_error
+    hausdorff_error
+    prediction_ratio
```
SebastianSchmidl, line 121: What about the `aeon.benchmarking.metrics.stats` module containing
- check_friedman
- nemenyi_test
- wilcoxon_test

IMO, those are not metrics. I would move them to `aeon.benchmarking.stats` and also add them to the documentation here.

SebastianSchmidl (changes requested):
Thanks for fixing the sphynx-plugin. I think the PR is still fine in size.

docs/api_reference/benchmarking.rst
MatthewMiddlehurst: good catch. made the suggested change

SebastianSchmidl (approved):

PR #2361, opened 2024-11-16
[ENH] fix the zero_method of the wilcoxon sign rank test to "wilcox" in mcm

### LINKED ISSUE #2358: [BUG] Different p-values in scatter and MCM
### Describe the bug

both these plotting tools show a one sided p-value for wilcoxon sign rank test, but in or lates experiments these are different, possibly from calling different functions. Needs investigation

plot both with same data

### Expected results

same p vals

### Actual results

image
image

### Versions

_No response_

TonyBagnall:
fixes #2358 sets the zero_method in mcm to "wilcox" to match scatter and cd diagrams


hadifawaz1999 (approved):
Thanks for the fix !

PR #2376, opened 2024-11-21
[ENH] Added ROCKAD anomaly detector to aeon

pattplatt:

Implemented the ROCKAD anomaly detector.

##### For all contributions


SebastianSchmidl:
The CI tests are still failing. Can you check why this is the case?

pattplatt:
@SebastianSchmidl should be better now

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,301 @@
...
+    n_inf_cols_ : list
+        List of feature columns containing infinite values, identified during fitting.
+    """
+
+    _tags = {
+        "capability:univariate": True,
+        "capability:multivariate": True,
+        "capability:missing_values": False,
+        "fit_is_empty": False,
+    }
```
baraline, line 72: You will need `capability:multithreading=True` in the tags as you have a n_jobs `parameter`

baraline (changes requested):
Small change needed for a softdep test. For the other ones that are still failing, it look a bit sneaky. Found this issue that could be relevant, https://github.com/scikit-learn/scikit-learn/issues/27499

baraline:
The issue might be caused by an overflow, see steps to reproduce :
```python
from sklearn.preprocessing import PowerTransformer
transformer = PowerTransformer()
transformer.fit([[23.96], [23.98], [23.99]])
```
Gives
```python-traceback
C:\Users\AntoineGUILLAUME\miniconda3\envs\watson\Lib\site-packages\sklearn\preprocessing\_data.py:3438: RuntimeWarning: overflow encountered in power
  out[pos] = (np.power(x[pos] + 1, lmbda) - 1) / lmbda
...
BracketError: The algorithm terminated without finding a valid bracket. Consider trying different initial points.
```
Changing the inputs slightly will cause the error to either appear or disappear, not familiar with the transformer, but the value lambda used in the power function seems to cause the overflow on some cases. Issue seems to be happening independently of the sklearn version.

pattplatt:
> The issue is caused by an overflow, see steps to reproduce :
> 
> ```python
> from sklearn.preprocessing import PowerTransformer
> transformer = PowerTransformer()
> transformer.fit([[23.96], [23.98], [23.99]])
> ```
> 
> Gives
> 
> ```
> RuntimeWarning: overflow encountered in power: out[pos] = (np.power(x[pos] + 1, lmbda) - 1) / lmbda
> ...
> BracketError: The algorithm terminated without finding a valid bracket. Consider trying different initial points.
> ```
> 
> Changing the inputs slightly will cause the error to either appear or disappear, not familiar with the transformer, but the value lambda used in the power function seems to cause the overflow. The issue seems to be happening independently of the sklearn version.

Thanks for the suggestions 👍, after testing a little I think the main reason seems to be the window size and stride applied to the input data which leads to the final bracket error.  I will add a fallback and a warning if this occurs.

SebastianSchmidl:
Could we restrict the `window_size` and `stride` parameters in some way to prevent the error? E.g. limit window_size to >5, <=0.5*n_timepoints or similar?

pattplatt:
> Could we restrict the `window_size` and `stride` parameters in some way to prevent the error? E.g. limit window_size to >5, <=0.5*n_timepoints or similar?

As I implemented it now, power transform gets deactivated dynamically only if it leads to the error. I also added a ValueError if  n_neighbors are larger than the count of windows. The soft-deps test passes now. What do you think, is this sollution also fine?

baraline:
If we decide to go with the current implementation, keeping an open issue with the context, the reproduction steps, and the related issues as backlog so we can keep an eye on it. Because as I understand it, the issue is not caused by our usage of the PowerTransformer.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+            n_kernels=self.n_kernels,
+            normalise=self.normalise,
+            n_jobs=self.n_jobs,
+            random_state=self.random_state,
+        )
+
+        Xt = self.rocket_transformer_.fit_transform(X)
+
+        Xt = Xt.astype("float64")
+        self.Xtp = None  # X: values, t: (rocket) transformed, p: power transformed
```
SebastianSchmidl, line 154: `self.Xtp` is not used outside this function. Instead, `self.Xtp_df` is. `self.Xtp` can be removed, and a local variable be used in the try-except-clause.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+        Xtp_scaled = None
+
+        if self._use_power_transform is True:
+
+            inf_columns = self.Xtp_df.columns[np.isinf(self.Xtp_df).any(axis=0)]
+            self.Xtp_df = self.Xtp_df.drop(columns=inf_columns)
+            self.inf_columns_ = inf_columns
+            self.scaler_ = StandardScaler()
+            Xtp_scaled = self.scaler_.fit_transform(self.Xtp_df)
```
SebastianSchmidl, line 183: Here, you use the `StandardScaler` independent of the input parameter `normalise`. Is this intended? If yes, please improve the parameter documentation.

I am also confused by first using `PowerTransformer(standardize=False)` and then standardizing manually afterward. Why not use `PowerTransformer(standardize=True)`?

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+            # Bootstrap Aggregation
+            Xtp_scaled_sample = resample(
+                Xtp_scaled,
+                replace=True,
+                n_samples=None,
+                random_state=self.random_state + idx_estimator,
+                stratify=None,
+            )
+
+            Xtp_scaled_sample = Xtp_scaled_sample
```
SebastianSchmidl, line 206: Noop, please remove.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+            inf_columns = self.Xtp_df.columns[np.isinf(self.Xtp_df).any(axis=0)]
+            self.Xtp_df = self.Xtp_df.drop(columns=inf_columns)
+            self.inf_columns_ = inf_columns
+            self.scaler_ = StandardScaler()
+            Xtp_scaled = self.scaler_.fit_transform(self.Xtp_df)
+        else:
+            Xtp_scaled = self.Xtp_df.astype(np.float64)
+
+        self.list_baggers = []
```
SebastianSchmidl, line 187: This should be a fitted member: `self.list_baggers_`

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+        for idx_estimator in range(self.n_estimators):
+            # Initialize estimator
+            self.estimator = NearestNeighbors(
+                n_neighbors=self.n_neighbors,
+                n_jobs=self.n_jobs,
+                metric=self.metric,
+                algorithm="kd_tree",
+            )
```
SebastianSchmidl, line 196: The `self.estimator` member gets overwritten and only stores the last estimator. I think this should be a local variable instead.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+        """
+        y_scores = np.zeros((len(X), self.n_estimators))
+        # Transform into rocket feature space
+        Xt = self.rocket_transformer_.transform(X)
+        Xtp_scaled = None
+
+        if self._use_power_transform is True:
+            # Power Transform using yeo-johnson
+            Xtp = self.power_transformer_.transform(Xt)
```
SebastianSchmidl, line 266: Is this guaranteed not to raise the bracket-error?

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+    # Replace infinite values
+    infinite_mask = np.isinf(df)
+    if infinite_mask.any().any():
+        df[infinite_mask] = inf_replacement
+
+    # Replace NaN values
+    if df.isna().any().any():
+        df = df.fillna(nan_replacement)
+
+    return df
```
SebastianSchmidl, line 326: If this is used only within the `_predict_proba` function, you can replace it with the way easier to maintain statement:

```python
df[~np.isfinite(df)] = 0.0
```

aeon/anomaly_detection/tests/test_rockad.py
```diff
@@ -0,0 +1,67 @@
...
+    """Test ROCKAD incorrect input."""
+    rng = check_random_state(seed=2)
+    series = rng.normal(size=(100,))
+
+    with pytest.raises(ValueError, match="The window size must be at least 1"):
+        ad = ROCKAD(window_size=0)
+        ad.fit_predict(series)
+    with pytest.raises(ValueError, match="The stride must be at least 1"):
+        ad = ROCKAD(stride=0)
+        ad.fit_predict(series)
```
SebastianSchmidl, line 67: Please add tests for the following:

- test of the `kneighbors`-check
- force of bracket error and test if the recovery strategy kicks in

SebastianSchmidl (changes requested):
Thank you for this contribution. Looks good already. I left some minor code improvements below.

Should we try to get rid of Pandas within the implementation? All things should be possible with NumPy. @baraline your take on this?

pattplatt:
> If we decide to go with the current implementation, keeping an open issue with the context, the reproduction steps, and the related issues as backlog so we can keep an eye on it. Because as I understand it, the issue is not caused by the way we use the PowerTransformer right ? But if we can show that the error is bounded between some window size and stride values, we could enforce it.

As it turned out it was a problem with the way it was implemented. I now transpose the data before the transform is applied and now the error is gone, so I think its fine now.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def __init__(
+        self,
+        n_estimators=10,
+        n_kernels=100,
+        normalise=True,
+        n_neighbors=5,
+        n_jobs=1,
+        metric="euclidean",
+        power_transform=True,
```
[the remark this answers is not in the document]
pattplatt: You are right, I set the default to False

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def _inner_fit(self, X: np.ndarray) -> None:
+        self._use_power_transform = self.power_transform
+        if X.shape[0] < self.n_neighbors:
+            raise ValueError(
+                f"Window count ({X.shape[0]}) has to be larger than "
+                f"n_neighbors ({self.n_neighbors})."
+                "Please choose a smaller n_neighbors value or increase window count "
+                "by choosing a smaller window size or larger stride."
+            )
```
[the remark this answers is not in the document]
pattplatt: The problem is that the check has to be done after the data is transformed into windows, _check_params is called before. Should we move _check_params to be called after windowing in this case?

aeon/anomaly_detection/_rockad.py
pattplatt: It was the cause, I implemented a transpose and the error is gone, so I also removed the fallback, seems to be cleaner now.

aeon/anomaly_detection/_rockad.py
pattplatt: removed the function

aeon/anomaly_detection/_rockad.py
pattplatt: Yes this makes not much sense as PowerTransformer(standardize) seems to do the same, I removed StandardScaler completely and set PowerTransformer(standardize) True as default

pattplatt:
> Thank you for this contribution. Looks good already. I left some minor code improvements below.
> 
> Should we try to get rid of Pandas within the implementation? All things should be possible with NumPy. @baraline your take on this?

I removed all Pandas dependencies on the latest commit

baraline:
Thanks I'll try to have a more throughout look at it this evening, but it looks fine at a glance

aeon/anomaly_detection/_rockad.py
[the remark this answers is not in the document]
SebastianSchmidl: Can't we compute the expected window count by `(n_timepoints - window_size - padding_length) / stride + 1` and then compare this against `n_neighbors`?

But you could also keep it here, just add a comment that this check operates on the windows.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+        self.stride = stride
+        self.random_state = random_state
+
+        self.rocket_transformer_: Optional[Rocket] = None
+        self.estimator_: Optional[NearestNeighbors] = None
+        self.power_transformer_: Optional[PowerTransformer] = None
+        self.inf_columns_: Optional[np.ndarray] = None
+
+        if power_transform:
+            self.power_transformer_ = PowerTransformer()
```
SebastianSchmidl, line 107: Members with `_` at the end should be set only from the `fit()` function.

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+            )
+
+        self.rocket_transformer_ = Rocket(
+            n_kernels=self.n_kernels,
+            normalise=self.normalise,
+            n_jobs=self.n_jobs,
+            random_state=self.random_state,
+        )
+
+        Xt = self.rocket_transformer_.fit_transform(X)
```
SebastianSchmidl, line 149: It is really difficult to track shapes. Am I right with this?

```suggestion
        # X: (n_windows, window_size)
        Xt = self.rocket_transformer_.fit_transform(X)
        # XT: (n_cases, n_kernels)
```

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+
+        if power_transform:
+            self.power_transformer_ = PowerTransformer()
+
+    def _fit(self, X: np.ndarray, y: Optional[np.ndarray] = None) -> "ROCKAD":
+        self._check_params(X)
+
+        _X, _ = sliding_windows(
+            X, window_size=self.window_size, stride=self.stride, axis=0
+        )
```
SebastianSchmidl, line 114:
```suggestion
        # X: (n_timepoints, 1) because __init__(axis==0)
        _X, _ = sliding_windows(
            X, window_size=self.window_size, stride=self.stride, axis=0
        )
        # _X: (n_windows, window_size)
```

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+            n_jobs=self.n_jobs,
+            random_state=self.random_state,
+        )
+
+        Xt = self.rocket_transformer_.fit_transform(X)
+
+        Xt = Xt.astype("float64")
+
+        if self.power_transformer_ is not None:
+            Xtp = self.power_transformer_.fit_transform(Xt.T)
```
SebastianSchmidl, line 154: `PowerTransformer` expects `(n_samples, n_features)` and `Xt` is in `(n_cases, n_kernels)`. If we transpose, we get `n_samples = n_kernels` and `n_features = n_cases`. This does not sound right to me. Did I miss something?

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+        Step size for moving the sliding window over the time series data.
+    random_state : int, default=42
+        Random seed for reproducibility.
+
+    Attributes
+    ----------
+    rocket_transformer_ : Optional[Rocket]
+        Instance of the ROCKET transformer used to extract features, set after fitting.
+    estimator_ : Optional[NearestNeighbors]
+        k-NN estimator used for anomaly scoring, set after fitting.
```
SebastianSchmidl, line 60: This should be `list_baggers_ : Optional[List[NearestNeighbors]`

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+        self.n_neighbors = n_neighbors
+        self.n_jobs = n_jobs
+        self.metric = metric
+        self.power_transform = power_transform
+        self.window_size = window_size
+        self.stride = stride
+        self.random_state = random_state
+
+        self.rocket_transformer_: Optional[Rocket] = None
+        self.estimator_: Optional[NearestNeighbors] = None
```
SebastianSchmidl, line 102: Unused

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,274 @@
...
+
+        # Drop columns identified by self.inf_columns_
+        if self.inf_columns_ is not None and len(self.inf_columns_) > 0:
+            mask = np.ones(Xtp.shape[1], dtype=bool)
+            mask[self.inf_columns_] = False
+            Xtp = Xtp[:, mask]
+
+            Xtp[~np.isfinite(Xtp)] = 0.0
+
+            Xtp = Xtp.astype(np.float64).to_numpy()
```
SebastianSchmidl, line 256: This should probably be one level deeper within the `if self.power_transformer_ is not None`-branch, right? Otherwise, `Xtp` might not be defined.

SebastianSchmidl (changes requested):
I could not follow the shapes of the different data representations and check whether transposing the input to the power transformer is actually correct. IMO the transpose is wrong.

aeon/anomaly_detection/_rockad.py
baraline: This is how I understand it too, transposing is fixing the bug but output of the PowerTransformer will not be what is expected. IMO If we can't figure out the bug, the solution of having a try catch with warning has you previously did works fine.

aeon/anomaly_detection/_rockad.py
pattplatt: I think XT should be: (n_windows, n_kernels*2), Rocket producer for each feature 2 values but the windows remain. Or is n_cases just another term for window?

SebastianSchmidl: Ok, `(n_windows, n_kernels*2)` makes sense, and `n_cases` = `n_windows` 👍🏼

SebastianSchmidl (approved):

aeon/anomaly_detection/_rockad.py
SebastianSchmidl: Is this still in there?

pattplatt: Sorry, thought I already removed this

SebastianSchmidl (approved):
LGTM, let's give @baraline some time to also review

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,280 @@
...
+                self.inf_columns_ = inf_columns
+            except Exception:
+                warnings.warn(
+                    "Power Transform failed and thus has been disabled. "
+                    "Try increasing the window size.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self.power_transformer_ = None
+                Xtp = Xt
```
baraline, line 170: I would add a warning about the inf_columns, changing the shape of the input data (here n_features) should not be done without at least some kind of feedback to the user IMO. When do we get these inf values btw ? They are different from the bug with the power transformer ?

baraline (changes requested):
LGTM but one small comment, sorry for the delay, don't hesitate to ping me on slack next time !

aeon/anomaly_detection/_rockad.py
pattplatt: No the inf values are not different from the bug as it seems. I quickly tested if they still occur and it seems because the power transform gets deactivated when the error occurs, the "inf values case" also does not occur anymore. Thus I think, we maybee could also remove the code in fit() as well as _predict_proba(). What do you think?

SebastianSchmidl: If the `inf`s are just a product of the `PowerTransformer`-bug, then we should, IMHO, also remove this code.

baraline: Agreed, this is just confusing if these values only occur due to the supposed bug of PowerTransformer. In that case, we should get rid of the inf values thing.

baraline (dismissed):
LGTM, thanks ! :)

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,261 @@
...
+        "fit_is_empty": False,
+    }
+
+    def __init__(
+        self,
+        n_estimators=10,
+        n_kernels=100,
+        normalise=False,
+        n_neighbors=5,
+        n_jobs=1,
```
MatthewMiddlehurst, line 80: We usually have n_jobs at the bottom

aeon/anomaly_detection/_rockad.py
```diff
@@ -0,0 +1,261 @@
...
+        metric="euclidean",
+        power_transform=True,
+        window_size: int = 10,
+        stride: int = 1,
+        random_state=42,
+    ):
+        super().__init__(axis=0)
```
MatthewMiddlehurst, line 87: This is best put as low down as possible in init to reduce the risk of overwriting attributes set higher up

SebastianSchmidl (approved):

PR #2377, opened 2024-11-21
[MNT] Update and consolidate general estimator checks

MatthewMiddlehurst:
Final part of [AEP 05](https://github.com/aeon-toolkit/aeon-admin/pull/10/files). As far as I know.

For the most part the all-estimator checks were mostly good, but could use some tidying. Some tests i merged to reduce `fit` calls and to reduce redundancy.

- `check_fit_updates_state` now also checks cloning, to avoid having to fit just to check clone resets `is_fitted`
- `check_set_params` had two tests essentially with a bit of redundancy, these have been merged
- `check_constructor` has the most confusing changes probably. It should mostly the same except all kw params are tested and all `callable` parameters are now allowed as defaults (matching sklearn)


aeon/testing/estimator_checking/_yield_regression_checks.py
```diff
@@ -163,11 +160,6 @@ def check_regressor_overrides_and_tags(estimator_class):
         else:  # must be a list
             assert any([t in valid_unequal_types for t in X_inner_type])
 
-    # Must have at least one set to True
-    multi = estimator_class.get_class_tag(tag_name="capability:multivariate")
-    uni = estimator_class.get_class_tag(tag_name="capability:univariate")
-    assert multi or uni
-
```
TonyBagnall, line 170: is this tested elsewhere or just not needed?

TonyBagnall (approved):

aeon/testing/estimator_checking/_yield_regression_checks.py
MatthewMiddlehurst: moved to all estimators instead of just regression

Issue #2132, opened 2024-10-02
[ENH] R-Clustering clusterer


chrisholder:
### Describe the feature or idea you want to propose

The R-Clustering clusterer is a recently proposed TSCL algorithm using random convolutional kernels (Rocket) with a dimensionality reduction  algorithm with k-means to achieve state-of-the-art. The paper that proposes R-Clustering is here: https://link.springer.com/article/10.1007/s10618-024-01018-x

### Describe your proposed solution

It would make a nice addition to aeon

PR #2382, opened 2024-11-22
[ENH] Added R-Clustering clusterer to aeon for issue #2132

Ramana-Raja:
##### For all contributions

### DIFF
--- a/aeon/clustering/_r_cluster.py
+++ b/aeon/clustering/_r_cluster.py
@@ -1,0 +1,323 @@
+import numpy as np
+from numpy.random import RandomState
+
+from typing import Optional, Union
+from sklearn.cluster import KMeans
+from sklearn.decomposition import PCA
+from sklearn.preprocessing import StandardScaler
+
+from aeon.clustering.base import BaseClusterer
+
+from numba import njit, prange
+
+
+class RCluster(BaseClusterer):
+    """Time series R Clustering implementation .
+
+       Adapted from the implementation used in [1]_
+
+       Parameters
+       ----------
+       num_kernels : int , default = 84
+            The number of convolutional kernels used to transform the input time series
+            These kernels are fixed and pre-defined (not random) and are optimized for computational speed and
+            feature diversity
+
+       max_dilations_per_kernel : int , default = 32
+            The maximum number of dilation rates applied to each kernel
+            Dilations control the spacing of the kernel's receptive field over the time series,
+            capturing patterns at varying scales
+
+       num_features : int , default = 500
+            The number of features extracted per kernel after applying the transformation
+
+       num_cluster : int , default = 8
+            The number of clusters used
+
+       n_init : int , default = 10
+            The number of times the clustering algorithm (e.g., KMeans) will run with different centroid seeds
+            to avoid poor local optima
+
+       max_iter: int, default=300
+            Maximum number of iterations of the k-means algorithm for a single
+            run.
+       random_state: int or np.random.RandomState instance or None, default=None
+            Determines random number generation for centroid initialization.
+       Notes
+       -----
+       Adapted from the implementation from source code
+       https://github.com/jorgemarcoes/R-Clustering/blob/main/R_Clustering_on_UCR_Archive.ipynb
+
+       References
+       ----------
+       .. [1]  Time series clustering with random convolutional kernels
+       https://link.springer.com/article/10.1007/s10618-024-01018-x
+       """
+    def __init__(self,
+                 num_features=500,
+                 num_kernels=84,
+                 max_dilations_per_kernel=32,
+                 n_clusters=8,
+                 n_init=10,
+                 random_state: Optional[Union[int, RandomState]] = None,
+                 max_iter=300):
+        self.num_features = num_features
+        self.num_kernels = num_kernels
+        self.max_dilations_per_kernel = max_dilations_per_kernel
+        self.num_cluster = n_clusters
+        self.n_init = n_init
+        self.random_state = random_state
+        self.max_iter = max_iter
+
+    @staticmethod
+    @njit("float32[:](float32[:,:],int32[:],int32[:],float32[:])", fastmath=True, parallel=False, cache=True)
+    def __fit_biases(X, dilations, num_features_per_dilation, quantiles):
+
+        num_examples, input_length = X.shape
+
+        # equivalent to:
+        # >>> from itertools import combinations
+        # >>> indices = np.array([_ for _ in combinations(np.arange(9), 3)], dtype = np.int32)
+        ###MODIFICATION
+        indices = np.array((
+            1, 3, 6, 1, 2, 7, 1, 2, 3, 0, 2, 3, 1, 4, 5, 0, 1, 3, 3, 5, 6, 0,
+            1, 2, 2, 5, 8, 1, 3, 7, 0, 1, 8, 4, 6, 7, 0, 1, 4, 3, 4, 6, 0, 4,
+            5, 2, 6, 7, 5, 6, 7, 0, 1, 6, 4, 5, 7, 4, 7, 8, 1, 6, 8, 0, 2, 6,
+            5, 6, 8, 2, 5, 7, 0, 1, 7, 0, 7, 8, 0, 3, 5, 0, 3, 7, 2, 3, 8, 2,
+            3, 4, 1, 4, 6, 3, 4, 5, 0, 3, 8, 4, 5, 8, 0, 4, 6, 1, 4, 8, 6, 7,
+            8, 4, 6, 8, 0, 3, 4, 1, 3, 4, 1, 5, 7, 1, 4, 7, 1, 2, 8, 0, 6, 7,
+            1, 6, 7, 1, 3, 5, 0, 1, 5, 0, 4, 8, 4, 5, 6, 0, 2, 5, 3, 5, 7, 0,
+            2, 4, 2, 6, 8, 2, 3, 7, 2, 5, 6, 2, 4, 8, 0, 2, 7, 3, 6, 8, 2, 3,
+            6, 3, 7, 8, 0, 5, 8, 1, 2, 6, 2, 3, 5, 1, 5, 8, 3, 6, 7, 3, 4, 7,
+            0, 4, 7, 3, 5, 8, 2, 4, 5, 1, 2, 5, 2, 7, 8, 2, 4, 6, 0, 5, 6, 3,
+            4, 8, 0, 6, 8, 2, 4, 7, 0, 2, 8, 0, 3, 6, 5, 7, 8, 1, 5, 6, 1, 2,
+            4, 0, 5, 7, 1, 3, 8, 1, 7, 8
+        ), dtype=np.int32).reshape(84, 3)
+
+        num_kernels = len(indices)
+        num_dilations = len(dilations)
+
+        num_features = num_kernels * np.sum(num_features_per_dilation)
+
+        biases = np.zeros(num_features, dtype=np.float32)
+
+        feature_index_start = 0
+
+        for dilation_index in range(num_dilations):
+
+            dilation = dilations[dilation_index]
+            padding = ((9 - 1) * dilation) // 2
+
+            num_features_this_dilation = num_features_per_dilation[dilation_index]
+
+            for kernel_index in range(num_kernels):
+
+                feature_index_end = feature_index_start + num_features_this_dilation
+
+                _X = X[np.random.randint(num_examples)]
+
+                A = -_X  # A = alpha * X = -X
+                G = _X + _X + _X  # G = gamma * X = 3X
+
+                C_alpha = np.zeros(input_length, dtype=np.float32)
+                C_alpha[:] = A
+
+                C_gamma = np.zeros((9, input_length), dtype=np.float32)
+                C_gamma[9 // 2] = G
+
+                start = dilation
+                end = input_length - padding
+
+                for gamma_index in range(9 // 2):
+                    C_alpha[-end:] = C_alpha[-end:] + A[:end]
+                    C_gamma[gamma_index, -end:] = G[:end]
+
+                    end += dilation
+
+                for gamma_index in range(9 // 2 + 1, 9):
+                    C_alpha[:-start] = C_alpha[:-start] + A[start:]
+                    C_gamma[gamma_index, :-start] = G[start:]
+
+                    start += dilation
+
+                index_0, index_1, index_2 = indices[kernel_index]
+
+                C = C_alpha + C_gamma[index_0] + C_gamma[index_1] + C_gamma[index_2]
+
+                biases[feature_index_start:feature_index_end] = np.quantile(C, quantiles[
+                                                                               feature_index_start:feature_index_end])
+
+                feature_index_start = feature_index_end
+
+        return biases
+
+    def __fit_dilations(self,input_length, num_features, max_dilations_per_kernel):
+
+        num_kernels = 84
+
+        num_features_per_kernel = num_features // num_kernels
+        true_max_dilations_per_kernel = min(num_features_per_kernel, max_dilations_per_kernel)
+        multiplier = num_features_per_kernel / true_max_dilations_per_kernel
+
+        max_exponent = np.log2((input_length - 1) / (9 - 1))
+        dilations, num_features_per_dilation = \
+            np.unique(np.logspace(0, max_exponent, true_max_dilations_per_kernel, base=2).astype(np.int32),
+                      return_counts=True)
+        num_features_per_dilation = (num_features_per_dilation * multiplier).astype(np.int32)  # this is a vector
+
+        remainder = num_features_per_kernel - np.sum(num_features_per_dilation)
+        i = 0
+        while remainder > 0:
+            num_features_per_dilation[i] += 1
+            remainder -= 1
+            i = (i + 1) % len(num_features_per_dilation)
+
+        return dilations, num_features_per_dilation
+
+    def __quantiles(self,n):
+        return np.array([(_ * ((np.sqrt(5) + 1) / 2)) % 1 for _ in range(1, n + 1)], dtype=np.float32)
+
+    def __fit_rocket(self,X):
+
+        _, input_length = X.shape
+
+
+        dilations, num_features_per_dilation = self.__fit_dilations(input_length,
+                                                                    self.num_features,
+                                                                    self.max_dilations_per_kernel)
+
+        num_features_per_kernel = np.sum(num_features_per_dilation)
+
+        quantiles = self.__quantiles(self.num_kernels * num_features_per_kernel)
+
+        ###MODIFICATION
+        quantiles = np.random.permutation(quantiles)
+
+        biases = self.__fit_biases(X, dilations, num_features_per_dilation, quantiles)
+
+        return dilations, num_features_per_dilation, biases
+
+    def __transform(self,X, parameters):
+
+        num_examples, input_length = X.shape
+
+        dilations, num_features_per_dilation, biases = parameters
+
+        # equivalent to:
+        # >>> from itertools import combinations
+        # >>> indices = np.array([_ for _ in combinations(np.arange(9), 3)], dtype = np.int32)
+        indices = np.array((
+            1, 3, 6, 1, 2, 7, 1, 2, 3, 0, 2, 3, 1, 4, 5, 0, 1, 3, 3, 5, 6, 0,
+            1, 2, 2, 5, 8, 1, 3, 7, 0, 1, 8, 4, 6, 7, 0, 1, 4, 3, 4, 6, 0, 4,
+            5, 2, 6, 7, 5, 6, 7, 0, 1, 6, 4, 5, 7, 4, 7, 8, 1, 6, 8, 0, 2, 6,
+            5, 6, 8, 2, 5, 7, 0, 1, 7, 0, 7, 8, 0, 3, 5, 0, 3, 7, 2, 3, 8, 2,
+            3, 4, 1, 4, 6, 3, 4, 5, 0, 3, 8, 4, 5, 8, 0, 4, 6, 1, 4, 8, 6, 7,
+            8, 4, 6, 8, 0, 3, 4, 1, 3, 4, 1, 5, 7, 1, 4, 7, 1, 2, 8, 0, 6, 7,
+            1, 6, 7, 1, 3, 5, 0, 1, 5, 0, 4, 8, 4, 5, 6, 0, 2, 5, 3, 5, 7, 0,
+            2, 4, 2, 6, 8, 2, 3, 7, 2, 5, 6, 2, 4, 8, 0, 2, 7, 3, 6, 8, 2, 3,
+            6, 3, 7, 8, 0, 5, 8, 1, 2, 6, 2, 3, 5, 1, 5, 8, 3, 6, 7, 3, 4, 7,
+            0, 4, 7, 3, 5, 8, 2, 4, 5, 1, 2, 5, 2, 7, 8, 2, 4, 6, 0, 5, 6, 3,
+            4, 8, 0, 6, 8, 2, 4, 7, 0, 2, 8, 0, 3, 6, 5, 7, 8, 1, 5, 6, 1, 2,
+            4, 0, 5, 7, 1, 3, 8, 1, 7, 8
+        ), dtype=np.int32).reshape(84, 3)
+
+        num_kernels = len(indices)
+        num_dilations = len(dilations)
+
+        num_features = num_kernels * np.sum(num_features_per_dilation)
+
+        features = np.zeros((num_examples, num_features), dtype=np.float32)
+
+        for example_index in prange(num_examples):
+
+            _X = X[example_index]
+
+            A = -_X  # A = alpha * X = -X
+            G = _X + _X + _X  # G = gamma * X = 3X
+
+            feature_index_start = 0
+
+            for dilation_index in range(num_dilations):
+
+                _padding0 = dilation_index % 2
+
+                dilation = dilations[dilation_index]
+                padding = ((9 - 1) * dilation) // 2
+
+                num_features_this_dilation = num_features_per_dilation[dilation_index]
+
+                C_alpha = np.zeros(input_length, dtype=np.float32)
+                C_alpha[:] = A
+
+                C_gamma = np.zeros((9, input_length), dtype=np.float32)
+                C_gamma[9 // 2] = G
+
+                start = dilation
+                end = input_length - padding
+
+                for gamma_index in range(9 // 2):
+                    C_alpha[-end:] = C_alpha[-end:] + A[:end]
+                    C_gamma[gamma_index, -end:] = G[:end]
+
+                    end += dilation
+
+                for gamma_index in range(9 // 2 + 1, 9):
+                    C_alpha[:-start] = C_alpha[:-start] + A[start:]
+                    C_gamma[gamma_index, :-start] = G[start:]
+
+                    start += dilation
+
+                for kernel_index in range(num_kernels):
+
+                    feature_index_end = feature_index_start + num_features_this_dilation
+
+                    _padding1 = (_padding0 + kernel_index) % 2
+
+                    index_0, index_1, index_2 = indices[kernel_index]
+
+                    C = C_alpha + C_gamma[index_0] + C_gamma[index_1] + C_gamma[index_2]
+
+                    if _padding1 == 0:
+                        for feature_count in range(num_features_this_dilation):
+                            features[example_index, feature_index_start + feature_count] = ((C > biases[feature_index_start + feature_count]).astype(float)).mean()
+                    else:
+                        for feature_count in range(num_features_this_dilation):
+                            features[example_index, feature_index_start + feature_count] =((C[padding:-padding] > biases[feature_index_start + feature_count]).astype(float)).mean()
+
+
+                    feature_index_start = feature_index_end
+
+        return features
+
+    def _fit(self,X,y=None):
+        parameters = self.__fit_rocket(X=X)
+        transformed_data = self.__transform(X=X, parameters=parameters)
+
+        sc = StandardScaler()
+        X_std = sc.fit_transform(transformed_data)
+
+        pca = PCA().fit(X_std)
+
+        self.optimal_dimensions = np.argmax(pca.explained_variance_ratio_ < 0.01)
+
+        pca_optimal = PCA(n_components=self.optimal_dimensions)
+        transformed_data_pca = pca_optimal.fit_transform(X_std)
+
+        self._r_cluster = KMeans(
+            n_clusters=self.num_cluster,
+            n_init=self.n_init,
+            random_state=self.random_state,
+            max_iter=self.max_iter)
+        self._r_cluster.fit(transformed_data_pca)
+
+    def _predict(self, X, y=None) -> np.ndarray:
+
+        parameters = self.__fit_rocket(X=X)
+        transformed_data = self.__transform(X=X, parameters=parameters)
+        sc = StandardScaler()
+        X_std = sc.fit_transform(transformed_data)
+
+        pca_optimal = PCA(n_components=self.optimal_dimensions)
+        transformed_data_pca = pca_optimal.fit_transform(X_std)
+
+        return self._r_cluster.predict(transformed_data_pca)


TonyBagnall:
hi, thanks for this but if we include this clusterer we want it to use our version of Rocket transformers which are optimised for numba

Ramana-Raja:
> hi, thanks for this but if we include this clusterer we want it to use our version of Rocket transformers which are optimised for numba

sure, I will try to reimplement it and use aeon Rocket transformers

Ramana-Raja:
> Hi @Ramana-Raja thanks for this PR! Just looked at your code and it looks like it follow the original. However, the original code is a reimplementation of MINIROCKET. We don't really want to have another copy of MINIROCKET in the repository to maintain. See: https://www.aeon-toolkit.org/en/stable/examples/transformations/minirocket.html
> 
> When we were thinking about implementing R-Clustering we wanted to try find a way to use MINIROCKET (or at least some of its functions). I haven't looked into seeing how possible this is. Would it be possible to use MINIROCKET or some of its functions in this implementation?

Just to confirm, you’re asking if MINIROCKET or its functions are being used here? I’ve already updated the code in my PR to use the existing MINIROCKET functions from the repository. Let me know if there’s anything else you’d like me to tweak

chrisholder:
> Author

R-Clustering uses MiniRocket with 3 or so small modifications (They comment in the original #modification for their changes). I want to try reuse the current implementation and functions as much as possible. This may *NOT* be possible but I would like to try reuse as much code as possible.

Ramana-Raja:
> > Just to confirm, you’re asking if MINIROCKET or its functions are being used here? I’ve already updated the code in my PR to use the existing MINIROCKET functions. Let me know if there’s anything else you’d like me to tweak
> 
> R-Clustering uses MiniRocket with 3 or so small modifications (They comment in the original #modification for their changes). I want to try reuse the current implementation and functions as much as possible. This may _NOT_ be possible but I would like to try reuse as much code as possible.

Hi, I have tired to reuse as much code as possible but I noticed that some of the tests are failing. Do you have any ideas on what might be causing the issues? Any recommendations on things I should check or approaches to try?

MatthewMiddlehurst:
@chrisholder @TonyBagnall Tests seem to be passing here and changes made, good to take a look when you have the time.

Ramana-Raja:
Hi, @TonyBagnall @chrisholder I've implemented all the required changes and ensured that the tests are passing successfully. Could you please review the updates and let me know if there are any additional adjustments or improvements needed?

chrisholder:
Hi @Ramana-Raja thanks for working on this! Sorry it's taking me a while to look at this. I will looking at this tomorrow (friday) and get back to you! Hopefully we can try merge it over this weekend/early next week!

Ramana-Raja:
> Hi @Ramana-Raja thanks for working on this! Sorry it's taking me a while to look at this. I will look at this tomorrow (friday) and get back to you! Hopefully we can try merge it over this weekend/early next week!

No problem at all, and thank you for the update! Please take your time to review.

chrisholder (commented):
Really good job reusing as much of the existing minirocket as possible! Most of my comments are just to do with standardising the code with the existing aeon code. Great work!

In addition to the comments I've left it would also be nice to have some tests that test equivalence to the original r-clustering code. I would just generate some test data with aeon and then pass it through the original and save the labels and then write a test similar to:

```

original_labs = [1, 0, 0, 0, 1, 1, 1, 0, 1, 2]

def test_r_cluster():
        X = make_example_3d_numpy(10, 1, 10, return_y=False)
        cluster = RCluster(n_clusters=3, n_jobs=1)
        cluster.fit(X)
        labs = cluster.labels_
       assert np.array_equal(labs, original_labs)
```

In addition we should move r-cluster inside cluster/aeon/feature_based/_r_cluster.py

Add it to the init of the feature_based. Feel free to keep it in the top level init as well.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+    Adapted from the implementation used in [1]_
+
+    Parameters
+    ----------
+    num_kernels : int , default = 84
+         The number of convolutional kernels used to transform the input time series
+         These kernels are fixed and pre-defined (not random) and are
+         optimized for computational speed and
+         feature diversity
+
```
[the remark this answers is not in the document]
Ramana-Raja: sure, i will do it

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+        "capability:missing_values": False,
+    }
+
+    def __init__(
+        self,
+        n_kernels=84,
+        max_dilations_per_kernel=32,
+        n_clusters=8,
+        n_init=10,
+        random_state=42,
```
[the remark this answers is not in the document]
Ramana-Raja: sure i will do it

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+         over the time series,capturing patterns at varying scales
+
+    num_features : int , default = 500
+         The number of features extracted per kernel after applying
+         the transformation
+
+    num_cluster : int , default = 8
+         The number of clusters used
+
+    n_init : int , default = 10
```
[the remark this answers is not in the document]
Ramana-Raja: ok, i will take reference from the code

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+        self._r_cluster = KMeans(
+            n_clusters=self.n_clusters,
+            n_init=self.n_init,
+            random_state=self.random_state,
+            max_iter=self.max_iter,
+        )
+        super().__init__()
+
+    def _get_parameterised_data(self, X):
+        np.random.seed(self.random_state)
```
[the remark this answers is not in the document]
Ramana-Raja: got it

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+    def __init__(
+        self,
+        n_kernels=84,
+        max_dilations_per_kernel=32,
+        n_clusters=8,
+        n_init=10,
+        random_state=42,
+        max_iter=300,
+        n_jobs=1,
+    ):
```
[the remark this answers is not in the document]
Ramana-Raja: If the random state is set to None during training and testing, the results can vary because the randomness isn’t fixed. This means things like labels_ and predictions on the same training data might not match every time you run it.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+        )
+
+        num_features_per_kernel = np.sum(num_features_per_dilation)
+
+        quantiles = _quantiles(self.n_kernels * num_features_per_kernel)
+
+        # MODIFICATION
+        quantiles = np.random.permutation(quantiles)
+
+        n_dilations = len(dilations)
```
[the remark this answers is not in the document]
Ramana-Raja: The aeon minirocket.fit_biases function is designed to handle both multivariate and univariate data. However, the version used in the original implementation was modified to work only with univariate data, as aeon minirocket.fit_biases requires additional parameters for multivariate input. I used the same logic as in the original aeon minirocket.static_fit, which supports both types of data. Therefore, this implementation should function just like the original one and handle the data in the same way.

aeon/clustering/_r_cluster.py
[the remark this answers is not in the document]
chrisholder: Yes this is fine we want it to be different every time unless a random_state is set. So for testing just set random_state=1 or something similar.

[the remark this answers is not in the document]
Ramana-Raja: ok, sure I will do it.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+            np.array([_], dtype=np.int32),
+            dilations,
+            num_features_per_dilation,
+            biases,
+        )
+
+    def _get_transformed_data(self, X, parameters):
+        X = X.astype(np.float32)
+        _, n_channels, n_timepoints = X.shape
+        prev_threads = get_num_threads()
```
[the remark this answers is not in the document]
Ramana-Raja: Sure, I’ve implemented it.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+        if self.n_jobs < 1 or self.n_jobs > multiprocessing.cpu_count():
+            n_jobs = multiprocessing.cpu_count()
+        else:
+            n_jobs = self.n_jobs
+        set_num_threads(n_jobs)
+        if n_channels == 1:
+            X = X.squeeze(1)
+            X_ = _static_transform_uni(X, parameters, self.indices)
+        else:
+            raise ValueError(
```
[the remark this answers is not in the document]
Ramana-Raja: Done.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+
+        dilations, num_features_per_dilation = _fit_dilations(
+            n_timepoints, self.n_kernels, self.max_dilations_per_kernel
+        )
+
+        num_features_per_kernel = np.sum(num_features_per_dilation)
+
+        quantiles = _quantiles(self.n_kernels * num_features_per_kernel)
+
+        # MODIFICATION
```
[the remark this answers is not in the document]
Ramana-Raja: Done.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+
+    num_cluster : int , default = 8
+         The number of clusters used
+
+    n_init : int , default = 10
+         The number of times the clustering algorithm (e.g., KMeans) will
+         run with different centroid seeds
+         to avoid poor local optima
+
+    max_iter: int, default=300
```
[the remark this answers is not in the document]
Ramana-Raja: Done.

aeon/clustering/_r_cluster.py
```diff
@@ -0,0 +1,469 @@
...
+                3,
+                8,
+                1,
+                7,
+                8,
+            ),
+            dtype=np.int32,
+        ).reshape(84, 3)
+        self.is_fitted = False
+        self._r_cluster = KMeans(
```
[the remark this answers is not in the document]
Ramana-Raja: Using this is causing some test issues. Can you help me understand why it's happening?

chrisholder:
@MatthewMiddlehurst can you have a look at this error I'm unsure why it is happening. How did you get this to work with the other feature based clusterers?

Ramana-Raja:
Hi @chrisholder, I discovered that the issue was caused by how self._random_state = check_random_state(self.random_state) was being handled. Removing it resolved the problem, though I'm not entirely sure why. Perhaps @MatthewMiddlehurst  could provide some insights. Additionally, I removed the estimator as an input for the class, as the research paper only utilized KMeans.

chrisholder:
Looks good - Final thing before approving have you checked equivalence to the original. It's doesnt have to be completely 100% the same (due to random state). However, we should check for 10 or so datasets (which you can load from aeon.datasets) that the scores are similar.

Ramana-Raja:
> Looks good - Final thing before approving have you checked equivalence to the original. It's doesnt have to be completely 100% the same (due to random state). However, we should check for 10 or so datasets (which you can load from aeon.datasets) that the scores are similar.
> 
> This doesn't need to be an actual test in aeon but if you could run it locally and post the adjusted rand index using sklearn (https://scikit-learn.org/stable/modules/generated/sklearn.metrics.adjusted_rand_score.html) the we can check this version matches the original!

I tested the ARI scores between the new RCluster and the original RCluster on the experimental dataset provided in the original RCluster code. Here are the results:
ari_scores

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+        set_num_threads(n_jobs)
+        return X
+
+    def _get_transformed_data(self, X, parameters):
+        prev_threads = get_num_threads()
+        X = self.check_params(X)
+        X = X.squeeze(1)
+        X_ = _static_transform_uni(X, parameters, self.indices)
+        set_num_threads(prev_threads)
+        return X_
```
MatthewMiddlehurst, line 414: Is the `n_jobs` stuff because of the private ROCKET functions? Instead of transforming the data to 2D use the tag `"X_inner_type": "numpy3D",` if you can.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+from aeon.clustering.base import BaseClusterer
+from aeon.transformations.collection.convolution_based._minirocket import (
+    _fit_biases,
+    _fit_dilations,
+    _quantiles,
+    _static_transform_uni,
+)
+
+
+class RCluster(BaseClusterer):
```
MatthewMiddlehurst, line 18: Is the name not RClustering?

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+        "capability:missing_values": False,
+    }
+
+    def __init__(
+        self,
+        n_kernels=84,
+        max_dilations_per_kernel=32,
+        n_clusters=8,
+        random_state=None,
+        n_jobs=1,
```
MatthewMiddlehurst, line 72: put `random_state` and `n_jobs` at the end, ordering in the docstring should match the ordering here.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+from aeon.transformations.collection.convolution_based._minirocket import (
+    _fit_biases,
+    _fit_dilations,
+    _quantiles,
+    _static_transform_uni,
+)
+
+
+class RCluster(BaseClusterer):
+    """Time series R Clustering implementation .
```
MatthewMiddlehurst, line 19: R-Clustering?

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+        self.is_fitted = False
+        self.estimator = KMeans(
+            n_clusters=self.n_clusters,
+            random_state=self.random_state,
+            n_init=self.n_init,
+        )
+        super().__init__()
+
+    def _get_parameterised_data(self, X):
+        np.random.seed(self.random_state)
```
MatthewMiddlehurst, line 349: I may have missed an earlier version of this, but you should use a RandomState object instead of setting the seed like this.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+        X = X.astype(np.float32)
```
MatthewMiddlehurst, line 350: Do we do this elsewhere or is this what the original does? just wondering why its necessary

MatthewMiddlehurst (changes requested):
Not ready sorry.

Please add some code comments explaining the use of private functions from another estimator and the multiple PCAs if you can. Don't usually request it but it is pretty confusing as is.

Class attributes set in `fit` and `fit_predict` (self.) should start with _ (protected/private) or end with _ (public and should be documented in the docstring under Attributes)

Ramana-Raja:
Hi @MatthewMiddlehurst , could you help me understand why this error is occurring? Specifically, the error states that the number of components in the principal components is greater than the minimum of the number of samples and features, even though I’ve accounted for this case with an exception

Ramana-Raja:
@chrisholder While reviewing the code, I identified an architectural issue. The R cluster implementation cannot have separate fit and predict methods because it relies on PCA for dimensionality reduction. For example, if PCA determines the optimal dimension during training to be 13, and we attempt to predict on test data with fewer than 13 dimensions, it results in an error. Even if we create a new PCA and apply fit_transform on the test data, we would need to retrain KMeans, which is what I did. However, I don't believe this is an optimal solution. Could you suggest a better approach to handle this scenario?

TonyBagnall:
> @chrisholder While reviewing the code, I identified an architectural issue. The R cluster implementation cannot have separate fit and predict methods because it relies on PCA for dimensionality reduction. For example, if PCA determines the optimal dimension during training to be 13, and we attempt to predict on test data with fewer than 13 dimensions, it results in an error. Even if we create a new PCA and apply fit_transform on the test data, we would need to retrain KMeans, which is what I did. However, I don't believe this is an optimal solution. Could you suggest a better approach to handle this scenario?

PCA has a separate fit and transform step, so you can PCA fit_transform in fit, save the transform then just PCA transform in transform?

Ramana-Raja:
> > @chrisholder While reviewing the code, I identified an architectural issue. The R cluster implementation cannot have separate fit and predict methods because it relies on PCA for dimensionality reduction. For example, if PCA determines the optimal dimension during training to be 13, and we attempt to predict on test data with fewer than 13 dimensions, it results in an error. Even if we create a new PCA and apply fit_transform on the test data, we would need to retrain KMeans, which is what I did. However, I don't believe this is an optimal solution. Could you suggest a better approach to handle this scenario?
> 
> PCA has a separate fit and transform step, so you can PCA fit_transform in fit, save the transform then just PCA transform in transform?

If we transform test data(while predicting that is predicting this test data without fitting it using _predict method) with a number of features less than the n_components of PCA, it will cause an error, So we cannot predict test data having number of features less than that of n_components of PCA.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,485 @@
...
+        params : dict or list of dict, default={}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        """
+        return {
+            "n_kernels": 84,
+            "max_dilations_per_kernel": 32,
+            "n_clusters": 8,
+            "random_state": 1,
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Setting other parameters to reduce runtime in testing is fine, just don't set random_state and n_clusters

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,509 @@
...
+                n_clusters=self.n_clusters,
+                random_state=self.random_state,
+                n_init=self.n_init,
+            )
+            return estimator.fit_predict(transformed_data_pca)
+        else:
+            transformed_data_pca = self.pca.fit_transform(X_std)
+            return self.estimator.predict(transformed_data_pca)
+
+    def _fit_predict(self, X, y=None) -> np.ndarray:
```
MatthewMiddlehurst, line 466: _fit_predict does not seem necessary here as it does the same as the base class default. I would remove it TBH

aeon/clustering/feature_based/tests/test_r_cluster.py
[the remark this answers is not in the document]
MatthewMiddlehurst: The new test does not seem very useful. Ideally it should be a problem the clusterer can accurately cluster. See the data generators in `testing` which will be cleaner than the current X/Y lines.

aeon/clustering/feature_based/_r_cluster.py
MatthewMiddlehurst: Still seems incorrect.

aeon/clustering/feature_based/_r_cluster.py
MatthewMiddlehurst: This has completely changed instead of just adding a -, new one is incorrect IMO

aeon/clustering/feature_based/_r_cluster.py
MatthewMiddlehurst: There are still calls to np.random

MatthewMiddlehurst:
Please respond to the reviews in text instead of just resolving them unless it is a simple change. Feel free to ask for clarification. There are still issues IMO.

> If we transform test data(while predicting that is predicting this test data without fitting it using _predict method) with a number of features less than the n_components of PCA, it will cause an error, So we cannot predict test data having number of features less than that of n_components of PCA.

Why would there be less features in predict?

Ramana-Raja:
> Please respond to the reviews in text instead of just resolving them unless it is a simple change. Feel free to ask for clarification. There are still issues IMO.
> 
> > If we transform test data(while predicting that is predicting this test data without fitting it using _predict method) with a number of features less than the n_components of PCA, it will cause an error, So we cannot predict test data having number of features less than that of n_components of PCA.
> 
> Why would there be less features in predict?

Considering the case where we train the clusterer on a large dataset but only need to make predictions on a smaller dataset with fewer features

MatthewMiddlehurst:
Bit confused on what you mean by fewer features. The algorithm explicitly does not allow unequal length series in the tags. 

Also, this appears to be done after the ROCKET transform, so I don't think it would have a different number of features either way?

Ramana-Raja:
> Bit confused on what you mean by fewer features. The algorithm explicitly does not allow unequal length series in the tags.
> 
> Also, this appears to be done after the ROCKET transform, so I don't think it would have a different number of features either way?

I was referring to the data after it has been transformed by ROCKET. If the number of features ends up being lower than PCA's, it can cause some issues. I fixed it by training a new KMeans model, but do you have any suggestions for a better method?

MatthewMiddlehurst:
ROCKET should not be producing a different amount of features. No fitting should be going on in predict here I feel.

Ramana-Raja:
> ROCKET should not be producing a different amount of features. No fitting should be going on in predict here I feel.

Removing the fitting step in the predict function,results in the following error: ValueError: n_components=9 must be between 0 and min(n_samples, n_features)=5 with svd_solver='full'. I thought this could be resolved by fitting a new KMeans model. However, if you have a better approach, could you please share it

MatthewMiddlehurst:
I dont really have the capacity to debug this currently. It still looks like there are transforms being fit in predict, which again does not look correct to me. If ROCKET is producing different size feature vectors that is incorrect.

Ramana-Raja:
@MatthewMiddlehurst I've resolved the PCA issue, and all test cases are now passing. I also added random_state in test case default param, similar to other clustering models in Aeon, to fix the test case error where estimator.labels_ was not matching estimator.fit(data). If you have some time, could you review the code and let me know if any improvements are needed?

Ramana-Raja:
> @MatthewMiddlehurst I've resolved the PCA issue, and all test cases are now passing. I also added random_state in test case default param, similar to other clustering models in Aeon, to fix the test case error where estimator.labels_ was not matching estimator.fit(data). If you have some time, could you review the code and let me know if any improvements are needed?

Hi @MatthewMiddlehurst , just checking in, kindly following up on this PR when you have a moment.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,568 @@
...
+        np.ndarray
+            Processed input data with float32 type.
+        """
+        X = X.astype(np.float32)
+        if self.n_jobs < 1 or self.n_jobs > multiprocessing.cpu_count():
+            n_jobs = multiprocessing.cpu_count()
+        else:
+            n_jobs = self.n_jobs
+        set_num_threads(n_jobs)
+        return X
```
MatthewMiddlehurst, line 429: I do not think this is required. Use the `check_n_jobs` utility. If you are setting numba threads, make sure to set it back to the original when done.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,568 @@
...
+        parameters = self._get_parameterised_data(X)
+
+        transformed_data = self._get_transformed_data(X=X, parameters=parameters)
+        X_std = self.scaler.transform(transformed_data)
+        pca = self.pca
+        transformed_data_pca = pca.transform(X_std)
+
+        return self.estimator.predict(transformed_data_pca)
+
+    def _fit_predict(self, X, y=None) -> np.ndarray:
```
MatthewMiddlehurst, line 511: As mentioned this is not needed. Can remove the whole method.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,568 @@
...
+        Returns
+        -------
+        params : dict or list of dict, default={}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        """
+        return {
+            "n_clusters": 2,
+            "random_state": 1,
```
MatthewMiddlehurst, line 567: Do not set random state in the test parameters.

aeon/clustering/feature_based/tests/test_r_cluster.py
```diff
@@ -0,0 +1,177 @@
+"""Test For RCluster."""
+
+import numpy as np
+from sklearn import metrics
+
+from aeon.clustering.feature_based._r_cluster import RClusterer
+from aeon.datasets import load_gunpoint
+
+X_ = [
```
MatthewMiddlehurst, line 9: If this is randomly generated data use the data generated utility in the individual test instead. If it is not what is the source?

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,568 @@
...
+        random_state=None,
+        n_jobs=1,
+    ):
+        self.num_features = num_features
+        self.n_init = n_init
+        self.n_jobs = n_jobs
+        self.n_kernels = n_kernels
+        self.max_dilations_per_kernel = max_dilations_per_kernel
+        self.n_clusters = n_clusters
+        self.random_state = random_state
```
MatthewMiddlehurst, line 82: order these the same as they are above

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,568 @@
...
+        self.indices = np.array(
+            (
```
MatthewMiddlehurst, line 84: I would not put this in init, if you can try to put it at the bottom as it is needlessly long and wont be used anywhere else.

MatthewMiddlehurst (changes requested):
Hi, this is a complex PR and the project is currently very busy so it is unlikely this will be in soon. I have left a few comments but I don't imagine they will be the last. 

I see you linked some results above at some point, but This seems to be for just one of the estimators? Not sure if that is this or the original code. One of the things I am going to ask for before merging is a comparison for both this and the original so that is also something you can do.

Ramana-Raja:
> Hi, this is a complex PR and the project is currently very busy so it is unlikely this will be in soon. I have left a few comments but I don't imagine they will be the last.
> 
> I see you linked some results above at some point, but This seems to be for just one of the estimators? Not sure if that is this or the original code. One of the things I am going to ask for before merging is a comparison for both this and the original so that is also something you can do.

The results were comparing how accurate or similar the original implementation is to this one. btw, thanks for taking the time to review this.

aeon/clustering/feature_based/_r_cluster.py
MatthewMiddlehurst: what changed? this looks the same.

MatthewMiddlehurst:
I see the image you posted, but it only has one column of ARI scores. I'm not sure what those scores are for, the clusters produced by your or produced by the original. We would want them for both so we can compare.

Test failure seems legit. `_labels` and predict output should be the same.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,534 @@
...
+        parameters = self._get_parameterised_data(X)
+
+        transformed_data = self._get_transformed_data(X=X, parameters=parameters)
+        X_std = self.scaler.transform(transformed_data)
+        pca = self.pca
+        transformed_data_pca = pca.transform(X_std)
+
+        return self.estimator.predict(transformed_data_pca)
+
+    def set_indices(self):
```
MatthewMiddlehurst, line 256: This does not have to be a public method.  I was thinking more just a private function at the very bottom so no one has to scroll past it.

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,534 @@
...
+        Returns
+        -------
+        labels : np.ndarray
+            Array of cluster labels for each time series.
+        """
+        parameters = self._get_parameterised_data(X)
+
+        transformed_data = self._get_transformed_data(X=X, parameters=parameters)
+        X_std = self.scaler.transform(transformed_data)
+        pca = self.pca
```
MatthewMiddlehurst, line 251: this is unnecessary

aeon/clustering/feature_based/_r_cluster.py
```diff
@@ -0,0 +1,534 @@
...
+        X_std = self.scaler.fit_transform(transformed_data)
+
+        pca = PCA().fit(X_std)
+        optimal_dimensions = np.argmax(pca.explained_variance_ratio_ < 0.01)
+
+        self.pca = PCA(n_components=optimal_dimensions, random_state=self.random_state)
+        self.pca.fit(X_std)
+        transformed_data_pca = self.pca.transform(X_std)
+
+        self.estimator = KMeans(
```
MatthewMiddlehurst, line 223: attributes set in fit should be protected starting in _ or end in _

Ramana-Raja:
> I see the image you posted, but it only has one column of ARI scores. I'm not sure what those scores are for, the clusters produced by your or produced by the original. We would want them for both so we can compare.
> 
> Test failure seems legit. `_labels` and predict output should be the same.

The ARI scores are calculated between the output produced by this implementation and the original one. The test cases only fail when random_state is not provided,as you can see, the estimator was passed without a random_state

image

MatthewMiddlehurst:
It should not fail with no random_state, or at least we should know why and that it is unsolvable.

So you took the cluster predictions from both and used those to calculate ARI? That is not how you evaluate these algorithms if so.

Ramana-Raja:
> It should not fail with no random_state, or at least we should know why and that it is unsolvable.
> 
> So you took the cluster predictions from both and used those to calculate ARI? That is not how you evaluate these algorithms if so.

without specifying a random state, the transformed data( from _get_transformed_data) will be different even when using the same input, which results in differences between the predicted and labels_. I thought ARI is typically used to assess the similarity between two clustering outputs , such as between the original model and our implementation. However, if you'd prefer that I evaluate the similarity of each model's output against the true Y values instead, I'm happy to do that.

MatthewMiddlehurst:
> without specifying a random state, the transformed data (from _get_transformed_data) will be different even when using the same input, which results in differences between the predicted and labels (as we are calling _get_transformed_data in both fit and predict).

Yes why is this happening.

> I thought ARI is typically used to assess the similarity between two clustering outputs, such as between the original model and our implementation. However, if you'd prefer that I evaluate the similarity of each model's output against the true Y values instead, I'm happy to do that.

Yes please do that. I am more interested on performance against the labels. Your previous results do show that there are large differences between the clusterers for some datasets it looks like?

Ramana-Raja:
> > without specifying a random state, the transformed data (from _get_transformed_data) will be different even when using the same input, which results in differences between the predicted and labels (as we are calling _get_transformed_data in both fit and predict).
> 
> Yes why is this happening.

_get_parameterised_data uses
```python
quantiles = random_state.permutation(quantiles)
```
so without setting a random state, the quantiles might change in fit and predict. Similarly, _fit_biases
```python
biases = _fit_biases(
            X,
            n_channels_per_combination,
            channel_indices,
            dilations,
            num_features_per_dilation,
            quantiles,
            self.indices,
            self.random_state,
        )
```
 also depends on the random state—so without it, the parameters can end up different even for the same input

> 
> > I thought ARI is typically used to assess the similarity between two clustering outputs, such as between the original model and our implementation. However, if you'd prefer that I evaluate the similarity of each model's output against the true Y values instead, I'm happy to do that.
> 
> Yes please do that. I am more interested on performance against the labels. Your previous results do show that there are large differences between the clusterers for some datasets it looks like?
Here is the result:
image

It also aligns with the original results: https://github.com/jorgemarcoes/R-Clustering/blob/main/results/benchmark_UCR_results.csv

MatthewMiddlehurst:
is `_get_parameterised_data` generating the rocket kernels? If so why are we generating new kernels in `predict`?

Ramana-Raja:
> is `_get_parameterised_data` generating the rocket kernels? If so why are we generating new kernels in `predict`?

Since those kernels depend on the input data, I figured it made sense to generate them based on the given data. The original source code only had fit_predict, so that’s what led me to think this way. Do you think it’s reasonable to use the same parameters that is used during fitting for predict too?

Ramana-Raja:
> > is `_get_parameterised_data` generating the rocket kernels? If so why are we generating new kernels in `predict`?
> 
> Since those kernels depend on the input data, I figured it made sense to generate them based on the given data for the predict function. The original source code only had fit_predict, so that’s what led me to think this way. Do you think it’s reasonable to use the same parameters that is used during fitting for predict too?

MatthewMiddlehurst:
No, a new kernel is a completely new feature essentially. The feature set you are creating in predict is completely different from the one you are generating in fit.

Ramana-Raja:
> No, a new kernel is a completely new feature essentially. The feature set you are creating in predict is completely different from the one you are generating in fit.

Should I use the same parameters created in fit for predict too?

Ramana-Raja:
@MatthewMiddlehurst I’ve updated the predict function to utilize the parameters from fit, and it’s now passing all the test cases. Feel free to take a look when you get a chance.

TonyBagnall:
hi @Ramana-Raja I'm converting this to draft unless you object, it would be nice to have but is a long way behind main

PR #2397, opened 2024-11-25
[DOC] Update API webpages to include new classes and functions

MatthewMiddlehurst:
Title, just 4 files for now. Makes the estimator base classes private, this can be easily undone if we want them in the API.


baraline (approved):
I think it makes sense to have these classes as private, as they shouldn't be necessary from a user point of view, and it can avoid issues with relative imports while using them for new estimators.

TonyBagnall (approved):

PR #2399, opened 2024-11-25
[DOC] Regression and segmentation api

MatthewMiddlehurst:
Regression and segmentation api pages and regression compose fix


baraline (approved):
Making a base section for the base classes makes things alot cleaner when looking at the page for estimators, lgtm !

Issue #2066, opened 2024-09-17
[ENH] Implement range-based precision and recall within aeon (replacing prts package)


SebastianSchmidl:
### Describe the feature or idea you want to propose

The package [`prts`](https://github.com/compml/prts) is used in the performance evaluation module for anomaly detection as a soft-dependency. However, there have not been any changes to it for 3 years, the maintainers have not reacted to [my PR](https://github.com/CompML/PRTS/pull/77), and the package was not implemented by the original authors of the method/paper.

### Describe your proposed solution

We should implement the metrics within aeon directly:

- [x] implement `ts_fscore`, `ts_precision`, `ts_recall` in `aeon/benchmarking/metrics/anomaly_detection/range_metrics`  
  ➡ addressed in PR #2400
- [x] Update tests for range metrics: The tests should use fixtures instead of the `prts` implementations.
  ➡ #2576  
  ➡ addressed in #2672
- [x] Refactorings, e.g. fix method signatures to ensure compatibility with our other metrics and move tests from `tests/test_metrics.py` to `tests/test_range_metrics.py` ...  
  ➡ #2780
  ➡ addressed in #2781 
- [x] improve Python docs and replace old metric documentation with the new metrics' docs  
  ➡ #2767 (obsolete because the old metric API is deprecated in #2780)
- [x] The functions in _continuous.py need to be updated to use the new implementations  
  ➡ #2785
- [ ] Remove `prts` dependency and old metric implementations. **Requires 1.2.0 release!** 
 ➡ #2576 

Target module: [`aeon.benchmarking.metrics.anomaly_detection`](https://github.com/aeon-toolkit/aeon/tree/main/aeon/performance_metrics/anomaly_detection)
Original publication:

> Tatbul, Nesime, Tae Jun Lee, Stan Zdonik, Mejbah Alam, and Justin Gottschlich. "Precision and Recall for Time Series." In Proceedings of the International Conference on Neural Information Processing Systems (NeurIPS), 1920–30. 2018.
> http://papers.nips.cc/paper/7462-precision-and-recall-for-time-series.pdf.

### Describe alternatives you've considered, if relevant

not relevant

aryanpola:
@aeon-actions-bot assign @aryanpola

aryanpola:
@CodeLionX So, the true anomaly ranges would use the timepoints to create a range to compare with the algorithm-detected anomaly range. Maybe later find an intersection between them.

CodeLionX:
[...]

PR #2400, opened 2024-11-26
[ENH] Replace `prts` metrics

aryanpola:
#### Reference Issues/PRs
Addresses #2066 

Implementation of Precision, Recall and F1-score metrics.

##### For all contributions


aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,281 @@
...
+__all__ = ["ts_precision", "ts_recall", "ts_fscore"]
+
+
+def __init__(self, bias="flat", alpha=0.0, gamma=None):
+    assert gamma in ["reciprocal", "one", "udf_gamma"], "Invalid gamma type"
+    assert bias in ["flat", "front", "middle", "back"], "Invalid bias type"
+
+    self.bias = bias
+    self.alpha = alpha
+    self.gamma = gamma
```
MatthewMiddlehurst, line 12: why does this exist?

aeon/benchmarking/metrics/anomaly_detection/tests/test_metrics.py
MatthewMiddlehurst, line 1: This does not contain pytest tests like those used elsewhere in the package.. Look at other test files or the documentation

MatthewMiddlehurst (commented):
Some of these functions look like they should be private/protected. In the tests please run both the new and current functions to ensure the output is the same (using pytest)

aeon/benchmarking/metrics/anomaly_detection/tests/test_metrics.py
aryanpola: I'll change the format of the tests, but the previous test cases have an array input whereas the range_metrics will have sets inside arrays (y_pred set, y_real set).

For example:
In range_metrics : [(1, 2), (7, 8)], [(3, 4), (9, 10)] -> [y_pred, y_real]
In _binary:     y_pred = np.array([0, 1, 1, 0])
                     y_true = np.array([0, 1, 0, 0])

aryanpola:
> Some of these functions look like they should be private/protected. In the tests please run both the new and current functions to ensure the output is the same (using pytest)

How do I declare a function private/protected, have never done that before.

aryanpola:
Changes made from the previous commit:
1. Added a function for flattening lists of lists into a list.
2. Removes overlapping functions.
3. Changed the metric calculations from averaging(w.r.t cardinality) to global(w.r.t overlapping positions) calculation of real and pred ranges.

aryanpola:
@MatthewMiddlehurst Please let me know if the test cases are fine, I can separate them into different functions if necessary.

TonyBagnall:
thanks for this, its really helpful. We will look again next week

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,222 @@
+"""Calculate Precision, Recall, and F1-Score for time series anomaly detection."""
+
+maintainer = []
+__all__ = ["ts_precision", "ts_recall", "ts_fscore"]
+
+
+def flatten_ranges(ranges):
```
TonyBagnall, line 7: could you make flatten_ranges, calculate_bias and gamma_select protected functions (_flatten_ranges, _calculate_bias and _gamma_select)?

TonyBagnall (changes requested):
lgtm, just make the functions protected and its good to go imo

aryanpola:
> lgtm, just make the functions protected and its good to go imo

Done. Thanks a bunch for reviewing!

SebastianSchmidl (commented):
Thank you for your valuable contribution! The code looks good, and I especially appreciate your effort to compare the results of your implementation to the existing one. Which code did you use to generate the fixtures?

However, I am not 100% happy with the existing state:

- The code lacks information and details in the documentation, e.g.
  - There is no reference to the original paper: http://papers.nips.cc/paper/7462-precision-and-recall-for-time-series.pdf
  - There is no description, how the `ts_precision` or the `ts_recall` are calculated. What is "Global Precision"?
  - It is unclear, how the parameters influence the calculations. E.g., for `ts_precision`, the combination `bias_type="udf_gamma" && udf_gamma!=0` (I guess this is how one would want to set the existence reward weight) is not a good idea. Because precision by definition emphasizes on prediction quality, there is no need for an existence reward and the Gamma should always be set to 0.
  - From the documentation, it is unclear, how the inputs `y_pred` and `y_real` should look like. How are the "real (actual) ranges" defined: in this case, we require a list of tuples with the begin and end indices. Is the end index inclusive or exclusive? How does the nested list work?
- The other AD performance metrics use bit masks as input. The new functions are incompatible. I guess, we can merge this without properly integrating the functions yet. Then, we can look at how to transform the bit masks to index ranges in a separate PR, when we remove the `prts`-package.
- The tests are all using the same parameter configuration. We should also test other parameters to see if the code is compatible with the original authors' one.

MatthewMiddlehurst:
I would like someone to independently verify the output of old and new are the same, given that the the expected results are hard coded.

Alternatively, if you could provide a code snippet running both functions to help us verify that would be helpful (this would not go in the repo).

aryanpola:
@SebastianSchmidl @MatthewMiddlehurst Do you want me to add point based metrics in a different file? I'll alter code to range based in this file.

SebastianSchmidl:
Hi @aryanpola,

I do not get what you mean by "point based metrics". We are interested in the metrics described in the paper by Tatbul et al.: "Precision and Recall for Time Series" at NeurIPS 2018, linked in the issue. The paper introduces "range-based recall" in Sect. 4.1 and "range-based precision" in Sect. 4.2. Those are the metrics, I want to have in aeon. They can, then, also be combined for a "range-based F1"-score.

As I understand it, the repository https://github.com/IntelLabs/TSAD-Evaluator was created by the paper's authors and also links to this paper. So, I assume that it contains the implementation for the metrics. But I have not looked into the code in details, so I cannot tell you, which portion they describe in the paper. Excuse any possible confusions caused by me not explaining my view on the code earlier.

Can you identify the relevant code for "range-based recall" and "range-based precision" in https://github.com/IntelLabs/TSAD-Evaluator?

aryanpola:
> Can you identify the relevant code for "range-based recall" and "range-based precision" in https://github.com/IntelLabs/TSAD-Evaluator?

Yes yes, it's just that I got carried away trying to implement classic metrics (obviously not what the issue stated).
I've updated the code with the necessary changes and added more test cases which should cover everything.

aryanpola:
Also, value errors will rise when `gamma = udf_gamma` in precision which wouldn't affect the precision calculation.

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+            return 1.0 / udf_gamma
+        else:
+            raise ValueError("udf_gamma must be provided for 'udf_gamma' gamma type.")
+    else:
+        raise ValueError(
+            "Invalid gamma type. Choose from ['one', 'reciprocal', 'udf_gamma']."
+        )
+
+
+def calculate_overlap_reward_precision(pred_range, overlap_set, bias_type):
```
MatthewMiddlehurst, line 106: I dont think this should be public

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        bias_value = _calculate_bias(i, length, bias_type)
+        max_value += bias_value
+
+        if global_position in overlap_set:
+            my_value += bias_value
+
+    return my_value / max_value if max_value > 0 else 0.0
+
+
+def calculate_overlap_reward_recall(real_range, overlap_set, bias_type):
```
MatthewMiddlehurst, line 140: this as well

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+    series.
+
+    Parameters
+    ----------
+    y_pred : list of tuples or list of lists of tuples
+        The predicted anomaly ranges.
+        - Each tuple represents a range (start, end) of the anomaly where
+          start is starting index (inclusive) and end is ending index (inclusive).
+        - If y_pred is in the format of list of lists, they will be flattened into a \
+          single list of tuples bringing it to the above format.
```
MatthewMiddlehurst, line 189: why would it be a list of lists?

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+
+    References
+    ----------
+    .. [1] Tatbul, Nesime, Tae Jun Lee, Stan Zdonik, Mejbah Alam,and Justin Gottschlich.
+       "Precision and Recall for Time Series." 32nd Conference on Neural Information
+       Processing Systems (NeurIPS 2018), Montréal, Canada.
+       http://papers.nips.cc/paper/7462-precision-and-recall-for-time-series.pdf
+    """
+    if udf_gamma is not None:
+        raise ValueError("`udf_gamma` is not applicable for precision calculations.")
```
MatthewMiddlehurst, line 221: this is not documented

aeon/benchmarking/metrics/anomaly_detection/range_metrics.py
```diff
@@ -0,0 +1,390 @@
...
+        - Each tuple represents a range (start, end) of the anomaly where
+          start is starting index (inclusive) and end is ending index (inclusive).
+        - If y_real is in the format of list of lists, they will be flattened into a
+          single list of tuples bringing it to the above format.
+    bias_type : str
+        Type of bias to apply. Should be one of ["flat", "front", "middle", "back"].
+        (default: "flat")
+    gamma : str
+        Cardinality type. Should be one of ["reciprocal", "one"].
+        (default: "one")
```
MatthewMiddlehurst, line 201: This way of displaying default does not match our other documentation, it should be i.e. `str, default="one"`

MatthewMiddlehurst (commented):
Its fine to use bits of documentation from the other function when applicable. I'll leave the input type call to Sebastian but bit odd to change it IMO.

Again i would like the output of the current functions used in testing for now just to verify that the output is the same.

MatthewMiddlehurst:
I'm not quite sure what you mean. In case there is confusion of the original issue, the end goal is to remove the other version of these functions eventually.

aryanpola:
> I'm not quite sure what you mean. In case there is confusion of the original issue, the end goal is to remove the other version of these functions eventually.

The implementation supports range metrics (ts_precision, ts_recall, ts_fscore) and removes the dependency on `prts` package.

aryanpola:
Also, there are some changes to be made:
1. Separate parameters -> alpha, bias to p_alpha, r_alpha, p_bias, r_bias.
2. Remove `udf_gamma` from precision. Currently, it only raises errors.

I'll update in some time.

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl changes from my side are done, please review when free and suggest changes if any :)

MatthewMiddlehurst:
> The implementation supports range metrics (ts_precision, ts_recall, ts_fscore) and removes the dependency on prts package.

I may just be missing something here. Why would there be repetition in documentation if you use documentation from the functions we are removing and replacing with these. The `prts` functions are still present and `prts` is still a dependency for the package. The replacement functions do not take the same input. I would disagree with the old ones being removed in this PR currently, given the ones you have implemented are different and unverified.

Please:
- Describe why the new implementations have deviated from the original wrappers in terms of input https://github.com/aeon-toolkit/aeon/blob/main/aeon/benchmarking/metrics/anomaly_detection/_binary.py
- Write tests comparing the old and new versions to ensure they have the same output.

aryanpola:
> I may just be missing something here. Why would there be repetition in documentation if you use documentation from the functions we are removing and replacing. 

I meant for the functions precision and recall, they have the same docstrings.

aryanpola:
The already existing implementation (_binary.py) takes inputs as binary values (arrays). The one we are trying to replace with, takes inputs as list of tuples. So the changes would be to convert the list of tuples to binary arrays for compatibility already existing functions?

MatthewMiddlehurst:
It would be fine to accept both and convert to either internally.

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl I've added a function to check if the inputs are binary, if yes, they get converted to ranges based tuples for further calulations. Also added the existing test case from `test_ad_metrics.py`.

aryanpola:
A bit of help needed with the errors.

aryanpola:
@MatthewMiddlehurst Thanks for the help!

SebastianSchmidl:
> Write tests comparing the old and new versions to ensure they have the same output. [previous comment](https://github.com/aeon-toolkit/aeon/pull/2400#issuecomment-2612527988)

This is still missing, right?

aryanpola:
> > Write tests comparing the old and new versions to ensure they have the same output. [previous comment](https://github.com/aeon-toolkit/aeon/pull/2400#issuecomment-2612527988)
> 
> This is still missing, right?

The updated test compares the results of the previous version (_binary.py) and the new version (range_metrics).

MatthewMiddlehurst:
Do you know why the new functions have different parameters from the old ones? Don't know if this was explained previously.

aryanpola:
> Do you know why the new functions have different parameters from the old ones? Don't know if this was explained previously.

1. I've added the beta parameter in fscore function, which follows the same logic from prts package.
2. I also added the udf_gamma function which allows the user to alter the input for gamma parameter.
3. Since alpha has no role in precision calculation, took the liberty to not include it in the list of parameters.

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl Do you want me to convert all the test cases in a format that compares the output from prts and current implementation?
example in test_metrics.py (line 298)

SebastianSchmidl:
If I understand you correctly, yes. The new implementation should produce the same results as the old implementation for all cases.

Please excuse that I am slow to respond and cannot spend much time here. I will be busy with other obligations until the middle of summer this year.

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl  , I’ve updated the test cases so that they now compare the current implementation with the prts package for all parameter configurations.

SebastianSchmidl:
Please add a check for the soft-dependency "prts" to those test where you call into the existing range-based metrics, similar to the existing tests:

https://github.com/aeon-toolkit/aeon/blob/8770d2f33b416ebbf62d51bfe4336b93cce4e809/aeon/benchmarking/metrics/anomaly_detection/tests/test_ad_metrics.py#L228-L232

aryanpola:
> Please add a check for the soft-dependency "prts" to those test where you call into the existing range-based metrics to make the CI happy; similar to the existing tests:
> 
> https://github.com/aeon-toolkit/aeon/blob/8770d2f33b416ebbf62d51bfe4336b93cce4e809/aeon/benchmarking/metrics/anomaly_detection/tests/test_ad_metrics.py#L228-L232

Done. Let me know if there's anything else to be changed.

MatthewMiddlehurst (dismissed):
Thanks a lot for addressing my concerns. I'm happy for this to go in the with testing done.

I would not close the issue yet, the next step is to remove the old functions and `prts` as a dependency. Unfortunately, this would involve changing a lot of the testing you have done here again, but the main purpose of the output comparison is to smooth the transition anyway.

SebastianSchmidl (changes requested):
Just some small documentation changes

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl 
Further changes to be made:
1. Remove `prts` from all_extras
2. Update tests to compare the implementation output with hard-coded values  (already done in previous commits).
3. Remove the older functions, which include `aeon\benchmarking\metrics\anomaly_detection\_continuous.py` and `aeon\benchmarking\metrics\anomaly_detection\_binary.py`

SebastianSchmidl:
> @MatthewMiddlehurst @SebastianSchmidl Further changes to be made:
> 
> 1. Remove `prts` from all_extras
> 2. Update tests to compare the implementation output with hard-coded values  (already done in previous commits).
> 3. Remove the older function, which includes `aeon\benchmarking\metrics\anomaly_detection\_binary.py`. `aeon\benchmarking\metrics\anomaly_detection\_continuous.py` still uses prts as a dependency. What should be done with that?

The functions in `_continuous.py` can use the new implementation within aeon.

> 
> This should be everything?

Could you add this information to the corresponding issue? This should be addressed in separate PRs.

aryanpola:
> > @MatthewMiddlehurst @SebastianSchmidl Further changes to be made:
> > 
> > 1. Remove `prts` from all_extras
> > 2. Update tests to compare the implementation output with hard-coded values  (already done in previous commits).
> > 3. Remove the older function, which includes `aeon\benchmarking\metrics\anomaly_detection\_binary.py`. `aeon\benchmarking\metrics\anomaly_detection\_continuous.py` still uses prts as a dependency. What should be done with that?
> 
> The functions in `_continuous.py` can use the new implementation within aeon.
> 
> > This should be everything?
> 
> Could you add this information to the corresponding issue? This should be addressed in separate PRs.

3 different PR's for 3 mentioned changes?

aryanpola:
@MatthewMiddlehurst Please merge this PR so I can pull the changes and update the test cases to hard-coded ones.

MatthewMiddlehurst (approved):

aryanpola:
@MatthewMiddlehurst @SebastianSchmidl Anything else to do here? 
I tried to create a new PR with a branch within recall branch, this brings all the previous commits from recall branch creating conflicts.  I'll update the test cases once this gets merged.

PR #2404, opened 2024-11-26
[DOC] Imports and api for visualisation and transformers

MatthewMiddlehurst:
This is probably not perfect, but this is getting very boring 🙂. distances and datasets to go (and then the issues that will surely pop up from the PRs).


baraline (approved):
A bit out of loop about the issues you are encountering but this LGTM nonetheless

PR #2408, opened 2024-11-27
[ENH] Refactor BinSegSegmenter to BinSegmenter

TonyBagnall:
BinSeg is short for BinSegmenter, so the name makes no sense


baraline (approved):
The renaming makes sense and all occurrences seem to have been dealt with

PR #2409, opened 2024-11-27
[ENH] Remove test exclusions

TonyBagnall:
new testing identified estimators that changed state in predict or transform. This was due to saving data in predict/transform to self.some_variable. Changed a couple to not store this data. If storing is required, it should be done in fit


aeon/segmentation/base.py
```diff
@@ -116,7 +116,7 @@ def fit(self, X, y=None, axis=1):
         self.reset()
         if axis is None:  # If none given, assume it is correct.
             axis = self.axis
-        X = self._preprocess_series(X, axis, True)
```
MatthewMiddlehurst, line 119: This should be fine for `fit`? Think thats the metadata param but may be wrong

MatthewMiddlehurst (commented):
good chunk removed

aeon/segmentation/base.py
TonyBagnall: it is fine to store for fit, and  would prefer it, but iirc we dont store in collections? I'll check

TonyBagnall: reverted to True

MatthewMiddlehurst (approved):

PR #2410, opened 2024-11-27
[ENH] Implement a function for direct forecasting

TonyBagnall:
implements a simple function to conduct direct forecasting, i.e. make a number forecasts using a forecaster trained on different horizons

at the moment I see no reason to make this a wrapper class, a simple function seems much more python appropriate, but perhaps at a later date we may do so 

when the tags PR is in, we can add a check of capabilities here


aeon/forecasting/__init__.py
```diff
@@ -5,8 +5,10 @@
     "BaseForecaster",
     "RegressionForecaster",
     "ETSForecaster",
+    "direct_forecasting",
```
MatthewMiddlehurst: Think maybe we should have a comment here to divide the functions from the classes. Agree it should be importable from here but not really a mix we do elsewhere.

aeon/forecasting/_direct.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+    if not isinstance(forecaster, BaseForecaster):
+        raise TypeError(
+            "forecaster attribute must be an instance of BaseForecaster or its "
+            "subclass."
+        )
+
+    preds = np.zeros(steps_ahead)
+    for i in range(1, steps_ahead + 1):
+        f = deepcopy(forecaster)
```
MatthewMiddlehurst: probably better to use https://github.com/aeon-toolkit/aeon/blob/main/aeon/base/_base.py#L489

aeon/forecasting/_direct.py
```diff
@@ -0,0 +1,59 @@
...
+    Implements the direct strategy, cloning the given estimator and setting its horizon.
+
+    .. code-block:: python
+
+        preds = []
+        for h in 1 to steps_ahead:
+            model = clone(forecaster)
+            model.horizon = h
+            model.fit(y)
+            preds.append(model.predict())
```
MatthewMiddlehurst: Is this necessary? people can just click on the button to view the code. Better to just give a brief explanation IMO, its not so complicated.

examples/forecasting/img/direct.png
MatthewMiddlehurst, line 1: is window the correct parameter?

aeon/forecasting/__init__.py
TonyBagnall: imo naming conventions make it obvious, a comment of '#functions seems redundant to me, as it is clearly a function. Also I will probably combine them into one file of functions at some point, so this will change then

examples/forecasting/img/direct.png
TonyBagnall: good spot!

aeon/forecasting/_direct.py
TonyBagnall: removed

aeon/forecasting/_direct.py
TonyBagnall: it should be public if it is meant to be generally used

TonyBagnall:
changing to draft to see what they look like as base class methods

TonyBagnall:
replaced by #2899

Issue #2415, opened 2024-11-28
[MNT] `pytest` ubuntu workflows using too much memory


MatthewMiddlehurst:
### Describe the issue

Something is causing our `pytest` runs to crash with unhelpful errors i.e. `The operation was canceled.` and `The hosted runner: GitHub Actions 12 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.`

As suggested by the second error, this is likely an out of memory issue as adding 5GB of swap space memory in #2413 resolve these crashes.

This only appears under certain circumstances though:
- Periodic tests (https://github.com/aeon-toolkit/aeon/actions/runs/12071259094/job/33662613284) - PASS
- Regular PR testing (https://github.com/aeon-toolkit/aeon/actions/runs/12053699859/job/33610809333) - PASS
- PR testing with all tests with numba cache  (https://github.com/aeon-toolkit/aeon/pull/2408) - PASS
- PR/`main` testing with all tests, cache disabled (same as release testing) (https://github.com/aeon-toolkit/aeon/actions/runs/12066354265) - FAIL

There seems to be some strange interaction with the `numba` cache action, despite loading nothing in the periodic test running it causes tests to pass, while not running it at all causes a failure.

Just increasing the swap space is not a great solution to this, but works temporarily.

### Suggest a potential alternative/fix

Investigate the differences in testing setups, and figure out what is causing the OOM error if that is the case.

### Additional context

I have only seen this on Linux runners, not macOS or Windows. `ubuntu-20.04` seems to produce this less often. Related to #1162 possibly.

TonyBagnall:
@MatthewMiddlehurst is this still an issue? Not seen anything like it, I'll close this but if it crops up again, revisit

PR #2416, opened 2024-11-28
[MNT] Swapped tensorflow and pytorch to install only CPU version

chrisholder:
#### Reference Issues/PRs
#2415

This PR adds a custom github action when installing all_extras on ubuntu. The custom action installs tensorflow-cpu and pytorch cpu instead of the "normal" version which comes with all the GPU code. By installing the CPU version it saves considerable install time but in addition saves a considerable amount of RAM usage during the test running (as the package sizes are dramatically smaller).  This should help to prevent issues causing the CI to crash due to exceeding memory limits.

##### For all contributions


chrisholder:
Do not merge this yet need to get rid of the additional periodic test which I turned on just for this branch.

.github/workflows/periodic_test_run.yml
```diff
@@ -0,0 +1,225 @@
+name: Periodic Tests
+
```
chrisholder, line 2: This will be removed I just added it to get a periodic running on my branch

.github/workflows/periodic_tests.yml
```diff
@@ -1,161 +1,32 @@
-name: Periodic Tests
+name: Periodic test pytest
```
chrisholder, line 1: Will revert before mergine

.github/workflows/periodic_test_run.yml
```diff
@@ -0,0 +1,63 @@
+name: Periodic test pytest
```
chrisholder, line 1: This file will be removed before merging into main. Just for testing.

chrisholder:
> 1. why only linux?
> 2. doesn't this still install tensorflow and torch through `all_extras`?
> 3. wouldnt we want to still install the gpu versions somewhere in the testing workflows?

**why only linux?**
By default windows and macos doesn't install the gpu version so you only need to opt out on linux. There docs say:

Windows:
`Caution: TensorFlow 2.10 was the last TensorFlow release that supported GPU on native-Windows. Starting with TensorFlow 2.11, you will need to install [TensorFlow in WSL2](https://tensorflow.org/install/pip#windows-%5Bwsl2%5D), or install tensorflow-cpu and, optionally, try the [TensorFlow-DirectML-Plugin](https://github.com/microsoft/tensorflow-directml-plugin#tensorflow-directml-plugin-)`

Macos:
`Currently there is no official GPU support for running TensorFlow on MacOS. The following instructions are for running on CPU.`

The same is true for Torch macos doesn't support GPU, and windows you have to specify additional parameters to get the GPU version (which we don't currently do).

**doesn't this still install tensorflow and torch through `all_extras`?**
I believe if you install the dependency (i.e. tensorflow/torch) on a specific version and then run pip install .[all_extras] it won't reinstall them as they already are installed (and pip won't try change the version). While tensorflow-cpu has a different name it resolves to tensorflow so pip doesn't know the difference. I checked the logs here: https://github.com/aeon-toolkit/aeon/actions/runs/12073657794/job/33670204314?pr=2416#step:5:346 and it seems that it doesn't try reinstall the dependency.

**wouldnt we want to still install the gpu versions somewhere in the testing workflows?**
I don't think so? The code that we run in the CI will be the exact same. Tensorflow-cpu is literally just the same code but all the extra gpu "stuff" is excluded from the build. Unless we're able to run the CI with a GPU then I don't think there is a point.

MatthewMiddlehurst:
ok

chrisholder:
Ready for review @MatthewMiddlehurst

.github/actions/cpu_all_extras/action.yml
```diff
@@ -0,0 +1,35 @@
...
+        max_attempts: 3
+        command: python -m pip install tensorflow-cpu
+
+    - name: Install CPU PyTorch
+      if: runner.os == 'Linux'
+      uses: nick-fields/retry@v3
+      with:
+        timeout_minutes: 30
+        max_attempts: 3
+        command: python -m pip install torch --index-url https://download.pytorch.org/whl/cpu
```
MatthewMiddlehurst, line 27: dont we only want to do this with all_extras?

MatthewMiddlehurst (approved):
lgtm

Issue #1742, opened 2024-07-01
[ENH] Replace SFA with SFAFast in REDCOMETS


zy18811:
### Describe the feature or idea you want to propose

Replace instances of [SFA](https://github.com/aeon-toolkit/aeon/blob/708db2e76914d78c1d36044d73b47ad23941996a/aeon/transformations/collection/dictionary_based/_sfa.py#L34-L1085) with [SFAFast](https://github.com/aeon-toolkit/aeon/blob/708db2e76914d78c1d36044d73b47ad23941996a/aeon/transformations/collection/dictionary_based/_sfa_fast.py#L47-L676) in [REDCOMETS](https://github.com/aeon-toolkit/aeon/blob/708db2e76914d78c1d36044d73b47ad23941996a/aeon/classification/dictionary_based/_redcomets.py#L27-L638) to improve speed as discussed in [PR#1722](https://github.com/aeon-toolkit/aeon/pull/1722#issuecomment-2198540654).

### Describe your proposed solution

Replace instances of SFA with SFAFast, updating class methods used where appropriate.

PR #2418, opened 2024-11-30
[ENH] Replace SFA with SFAFast in REDCOMETS

itsdivya1309:
#### Reference Issues/PRs
Closes: #1742 

Replace SFA with the faster version, SFAFast.

None

##### For all contributions

### DIFF
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -18,7 +18,7 @@ from sklearn.utils import check_random_state
 
 from aeon.classification.base import BaseClassifier
 from aeon.transformations.collection import Normalizer
-from aeon.transformations.collection.dictionary_based import SAX, SFA
+from aeon.transformations.collection.dictionary_based import SAX, SFAFast
 from aeon.utils.validation._dependencies import _check_soft_dependencies
 
 
@@ -62,7 +62,7 @@ class REDCOMETS(BaseClassifier):
 
     See Also
     --------
-    SAX, SFA
+    SAX, SFA, SFAFast
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst (commented):
Hi, could you do a small evaluation on how this impacts the words output and the overall accuracy/speed of the classifier? Some of the smaller UCR datasets should be fine.

May be on interest @zy18811. Don't need to do so if they are fine with the change, but wouldn't want to make a big change without either them knowing or evaluating.

patrickzib:
Hi, very cool. I am a bit surprised that there are no changes required on the test cases for REDCOMETS. 

@MatthewMiddlehurst is there a way to check, if the `test_redcomets.py` was were re-evaluated?

zy18811:
@itsdivya1309 thanks for doing this PR. 

My (very brief) look suggests that the word outputs are identical between SFA and SFAFast in the case of the whole series transform that REDCOMETS uses so absolutely no difference in accuracy (hence the old test cases still pass). 

It seems to give a small speedup most of the time, but is sometimes slower  - not sure if this is additonal numba overhead or something else. It could do with a little bit more thorough of an analysis to confirm SFAFast is an improvement overall wrt. speed. 

I also remember last time I looked at doing this there were some issues with particular alphabet sizes and word lengths on some datasets due to how SFA/SFAFast stores the words and numba having a max integer size. I'll have a look and see if I can replicate any of this - but I can see that @patrickzib has made a number of improvements to SFAFast since I last looked so might be fixed anyway.

MatthewMiddlehurst:
@zy18811 the class will still be limited in sizes due to the 64 bit limit for words required in the functions. The original does not have this.

zy18811:
@MatthewMiddlehurst

I've done some quick analysis, 40 datasets 10 resamples, and it looks like SFA Fast has **identical accuracy with a speedup**. 

sfa_fast_acc
sfa_fast_time

Regarding the numba errors discussed above, I think I ran into the issues when experimenting with different alphabet size ranges but the current range should have no problems I believe. 


Overall, PR looks good to me - thanks @itsdivya1309.

TonyBagnall (approved):
LGTM, thanks. Wonder if the same can be done with TDE?

patrickzib:
> I've done some quick analysis, 40 datasets 10 resamples, and it looks like SFA Fast has **identical accuracy with a speedup**.

That is very cool!

MatthewMiddlehurst:
Thanks both for implementing and evaluating. Does not seem like much downside, merging.

PR #2426, opened 2024-12-05
[GOV] Infrastructure workgroup lead and voting ambiguity fix

MatthewMiddlehurst:
Adds myself as lead to the infrastructure workgroup and clears up some voting ambiguity brought I'm in the governance document previously.

After this is merged we will go down from 4 GitHub owners to 2.


baraline (approved):
Considering the vote results and the two week minimum delay for voting, the change to majority of voting from the majority of core dev is reasonable, as inactive core dev blocking a vote could be an issue.

PR #2427, opened 2024-12-05
[ENh] Forecasting tests

TonyBagnall:
should be 100% coverage for base and regressor, slight restructuring of ETS means still some coverage required, but will do it in a later PR


MatthewMiddlehurst (approved):
fine for now but I feel a lot of it will become redundant when the general testing it worked on.

PR #2461, opened 2024-12-16
[BUG, ENH] SFA fix: Std-Normalization, as used in BOSS and WEASEL models, is potentially harmful for lower bounding

patrickzib:
This fixes an issue with std normalization in SFA, as is commonly used for BOSS and WEASEL models.

For Lower Bounding computations this however adds a potential violation of the lower bounding distance. Thus, I remove it, if whole series matching is applied.

This PR further adds:
- a module `SFAWhole`, which simplifies usage of SFA, for lower bounding in whole series, by correctly setting default parameters and only exposing needed parameters for lower bounding.
- parallel computation of the DFT using `scipy.fft.rfft(X, axis=1, workers=-1)`
- option to use sampling for learning BINs when using whole series.


aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -254,6 +267,15 @@ def _fit_transform(self, X, y=None):
         # self.word_bits = self.word_length_actual * self.letter_bits
         X = X.squeeze(1)
 
+        # subsample the samples
+        if self.sampling_factor:
+            sampled_indices = np.random.choice(
```
baraline, line 272: Is numpy seeded before? Best would be to use the new random generators

aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -807,21 +855,24 @@ def _fast_fourier_transform(X, norm, dft_length, inverse_sqrt_win_size):
...
-    dft /= stds.reshape(-1, 1)
     dft *= inverse_sqrt_win_size
 
+    # apply z-normalization
+    if norm_std:
+        stds = np.zeros(len(X))
+        for i in range(len(stds)):
+            stds[i] = np.std(X[i])
+        # stds = np.std(X, axis=1)  # not available in numba
+        stds = np.where(stds < 1e-8, 1, stds)
```
baraline, line 873: We have a constant for that case defined in

baraline (commented):
Fast review on "looks" and not algorithm correctness which I'll trust you with !

aeon/transformations/collection/dictionary_based/_sfa_fast.py
patrickzib: Thanks! I forgot to seed it. I have added seeding fro numpy. 

What do you mean by new random generators?

aeon/transformations/collection/dictionary_based/_sfa_fast.py
patrickzib: Thanks! I did not know about these. I have imported the constant and used it.

patrickzib:
> Fast review on "looks" and not algorithm correctness which I'll trust you with !

Thank you, and happy new year :)

aeon/transformations/collection/dictionary_based/_sfa_fast.py
baraline: I mean the [Random Generator API](https://numpy.org/doc/2.1/reference/random/generator.html) (i.e. not using np.random.XXXX)

patrickzib: I see. I have changed the code.

baraline:
Quick question before approval, are you supposed to get the same output from SFA vs SFA fast ? If so did you test it ?

patrickzib:
> Quick question before approval, are you supposed to get the same output from SFA vs SFA fast ? If so did you test it ?

Good question. For the same parameterization is should be. I have extended the test case in `test_symbolic_mindist.py.test_sfa_mindist` to check this.

baraline (approved):
LGTM !

PR #2463, opened 2024-12-18
[MNT] Set upper bound on matplotlib version

chrisholder:

The CI started breaking today on python version 3.9: https://github.com/aeon-toolkit/aeon/actions/runs/12390736933/job/34586409381.

@patrickzib did some investigating and found Matplotlib stopped supporting python 3.9 today as per their docs: https://github.com/matplotlib/matplotlib/blob/main/doc/devel/min_dep_policy.rst

As such the latest version fails to install on 3.9. As we still support 3.9 we have to set a upper bound on the version so that users still on 3.9 don't find their install breaking. This PR adds a upper bounding to matplotlib.

##### For all contributions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,9 @@ dependencies = [
 all_extras = [
     "esig>=0.9.7; platform_system != 'Darwin' and python_version < '3.11'",
     "imbalanced-learn",
-    "matplotlib>=3.3.2",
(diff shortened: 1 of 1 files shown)


patrickzib (changes requested):
I think, it should be: `esig>=0.9.7,<1.0.0;`

patrickzib (dismissed):
LGTM :)

patrickzib:
Maybe we do not need to bound matlab after all?

PR #2465, opened 2024-12-19
[ENH] Added IDK² and s-IDK² Anomaly Detector To Aeon

### LINKED ISSUE #2114: [ENH] Implement (s-)DIK² anomaly detector

IDK² and s-IDK² (sliding window) can be used for anomalous subsequence detection.

Implement both variants of IDK² directly in aeon:

- Paper: Isolation Distribution Kernel base paper: https://dl.acm.org/doi/10.1145/3394486.3403062, IDK for TSAD:https://www.vldb.org/pvldb/vol15/p2321-liu.pdf
- Code: https://github.com/IsolationKernel/Codes/tree/main/IDK/TS

_No response_

### Additional context

_No response_

Ramana-Raja:
closes #2114


aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    def __init__(
+        self,
+        psi1,
+        psi2,
+        width,
+        t=100,
+        sliding=False,
+        random_state=None,
```
SebastianSchmidl, line 75: Please add type hints (https://www.aeon-toolkit.org/en/latest/developer_guide/adding_typehints.html [still WIP]).

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,205 @@
...
+    psi1 : int
+         Number of samples randomly selected in each iteration for the feature
+         map matrix.
+    psi2 : int
+         Number of samples used for the second-stage feature map embedding.
+    width : int
+         Size of the sliding or fixed-width window for anomaly detection.
+    t : int, default=100
+         Number of iterations (time steps) for random sampling to construct
+         feature maps.
```
SebastianSchmidl, line 39: `t` is not very descriptive. I would call it `max_iter` similar on how sklearn does it, e.g. for [kmeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html)

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,205 @@
...
+        return onepoint_matrix
+
+    def __IDK(self, X, psi, t=100):
+        point_fm_list = self.__IK_inne_fm(X=X, psi=psi, t=t)
+        feature_mean_map = np.mean(point_fm_list, axis=0)
+        return np.dot(point_fm_list, feature_mean_map) / t
+
+    def _IDK_T(self, X):
+        np.random.seed(self.random_state)
+        random.seed(self.random_state)
```
SebastianSchmidl, line 121: Please do not use `random` at all and do not seed the global random number generator of numpy. Instead use a `Generator`-instance, see [NumPy RNG guide](https://numpy.org/doc/stable/reference/random/index.html#quick-start)

SebastianSchmidl (changes requested):
I just had a very brief look at the code because I am short on time. I will have another look next year after the Christmas break. Meanwhile, the code quality can be improved and tests need to be added.

- The tests are missing! Please add reasonable tests for the new algorithm. How do you make sure that it produces the same results as the original implementation?
- The code mixes standard python (e.g. `random`, `range`, ...) with numpy (e.g. `np.random`, `np.arange`, ...). Please mostly rely on numpy for better code quality and performance.
- Why are most methods in capital letters and start with dunders (`__`)? If this is a reference to the original implementation, please add the original name as a comment and use our [coding standard](https://www.aeon-toolkit.org/en/latest/developer_guide/coding_standards.html).

Thank you for taking the time and contributing to aeon!

aeon/anomaly_detection/_idk.py
Ramana-Raja: added type hints to the function as requested

aeon/anomaly_detection/_idk.py
Ramana-Raja: Done

aeon/anomaly_detection/_idk.py
Ramana-Raja: "I utilized 't' as it was implemented this way in the original code

Ramana-Raja:
> I just had a very brief look at the code because I am short on time. I will have another look next year after the Christmas break. Meanwhile, the code quality can be improved and tests need to be added.
> 
> * The tests are missing! Please add reasonable tests for the new algorithm. How do you make sure that it produces the same results as the original implementation?
> * The code mixes standard python (e.g. `random`, `range`, ...) with numpy (e.g. `np.random`, `np.arange`, ...). Please mostly rely on numpy for better code quality and performance.
> * Why are most methods in capital letters and start with dunders (`__`)? If this is a reference to the original implementation, please add the original name as a comment and use our [coding standard](https://www.aeon-toolkit.org/en/latest/developer_guide/coding_standards.html).
> 
> Thank you for taking the time and contributing to aeon!

Hello, I recently added a new test case, but it seems to be causing errors in testing. The test case runs successfully in my local environment, so I'm unsure why the issue arises. Could you assist me in identifying the cause?

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,214 @@
...
+        "capability:univariate": True,
+        "capability:multivariate": False,
+        "capability:missing_values": False,
+    }
+
+    def __init__(
+        self,
+        psi1: int,
+        psi2: int,
+        width: int,
```
MatthewMiddlehurst, line 88: Are there no default values for these? This should only be done if necessary

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,214 @@
...
+        params : dict
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        """
+        return {
+            "psi1": 8,
+            "psi2": 2,
+            "width": 1,
+            "random_state": 42,
```
MatthewMiddlehurst, line 213: Don't set the random_state for testing parameters

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,214 @@
...
+        t: int = 100,
+        sliding: bool = False,
+        random_state: int = None,
+    ) -> None:
+        self.psi1 = psi1
+        self.psi2 = psi2
+        self.width = width
+        self.t = t
+        self.sliding = sliding
+        self.random_state = np.random.default_rng(random_state)
```
MatthewMiddlehurst, line 98: self.random_state should just be random_state, attributes of the same name as parameters just store said parameters with no changes. This is causing your error also.

aeon/anomaly_detection/_idk.py
Ramana-Raja: When I remove random_state from this part of the code, one of the test cases fails, displaying the following issue: 'Running predict after fit twice with test parameters gives different results.'

aeon/anomaly_detection/_idk.py
Ramana-Raja: Added default values

aeon/anomaly_detection/_idk.py
Ramana-Raja: Done, thanks for the help.

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,212 @@
...
+
+    .. list-table:: Capabilities
+       :stub-columns: 1
+
+       * - Input data format
+         - univariate
+       * - Output data format
+         - anomaly scores
+       * - Learning Type
+         - unsupervised
```
SebastianSchmidl, line 22: We removed this table in favor of an automatically generated capabilities table in a recent PR #2468.

aeon/anomaly_detection/tests/test_idk.py
```diff
@@ -0,0 +1,24 @@
+"""Tests for the IDK Class."""
+
+import numpy as np
+
+from aeon.anomaly_detection import IDK
+
+
+def test_idk_univariate():
```
SebastianSchmidl, line 8: We want to make sure that our implementation is correct. Can you add tests that compare your implementation to the original authors' one?

You already linked to the reference implementation: https://github.com/IsolationKernel/Codes/tree/main/IDK/TS
You can execute the reference implementation on some sample data and store the results. Then, you could test this implementation on the same input data to produce the same results up to a certain precision.

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,212 @@
...
+        sliding: bool = False,
+        random_state: int = None,
+    ) -> None:
+        self.psi1 = psi1
+        self.psi2 = psi2
+        self.width = width
+        self.t = t
+        self.sliding = sliding
+        self.random_state = random_state
+        self.rng = np.random.default_rng(random_state)
```
SebastianSchmidl, line 99: If you instantiate the RNG once in the object, it cannot be reused and repeated calls to `predict` with the same input yield different results. This is also the source of the CI failure!

To fix this, use `self.random_state` in `_fit` and `_predict` to create a new, seeded RNG instance.

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,212 @@
...
+        for time in range(self.t):
+            sample_indices = self.rng.choice(X.shape[0], size=self.psi1, replace=False)
+            sample = X[sample_indices, :]
+            tem1 = np.dot(np.square(X), np.ones(sample.T.shape))
+            tem2 = np.dot(np.ones(X.shape), np.square(sample.T))
+            point2sample = tem1 + tem2 - 2 * np.dot(X, sample.T)
+
+            sample2sample = point2sample[sample_indices, :]
+            row, col = np.diag_indices_from(sample2sample)
+            sample2sample[row, col] = 99999999
```
SebastianSchmidl, line 146: Maybe, it is cleaner to set this to `np.nan` and use `np.nanmin` later on. This communicates your intent to ignore the main diagonal better.

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,212 @@
...
+
+            min_point2sample_index = np.argmin(point2sample, axis=1)
+            min_dist_point2sample = min_point2sample_index + time * psi
+            point2sample_value = point2sample[
+                range(len(onepoint_matrix)), min_point2sample_index
+            ]
+            ind = point2sample_value < radius_list[min_point2sample_index]
+            onepoint_matrix[ind, min_dist_point2sample[ind]] = 1
+
+        return onepoint_matrix
```
SebastianSchmidl, line 125: This code looks duplicated to `_idk_t`. Can you merge the duplicates together into one function?

aeon/anomaly_detection/tests/test_idk.py
```diff
@@ -0,0 +1,24 @@
...
+
+    ad = IDK(psi1=8, psi2=2, width=1, random_state=2)
+    pred = ad.fit_predict(series)
+    ad_sliding = IDK(psi1=16, psi2=4, width=10, sliding=True, random_state=1)
+    pred_sliding = ad_sliding.fit_predict(series)
+
+    assert pred.shape == (100,)
+    assert pred.dtype == np.float64
+    assert 50 <= np.argmax(pred) <= 58
+    assert pred_sliding.shape == (91,)
```
SebastianSchmidl, line 22: The result should have the same shape as the input in both cases!

SebastianSchmidl (changes requested):

aeon/anomaly_detection/_idk.py
SebastianSchmidl: The test error is different from this issue.

aeon/anomaly_detection/tests/test_idk.py
Ramana-Raja: The width parameter changes the way the model generates its output, since its 10 the output length 91(91 windows - each size of 10)

aeon/anomaly_detection/_idk.py
Ramana-Raja: Done

aeon/anomaly_detection/_idk.py
Ramana-Raja: Done

aeon/anomaly_detection/_idk.py
Ramana-Raja: Thanks for the help, updated the code.

aeon/anomaly_detection/tests/test_idk.py
Ramana-Raja: Added new case for this.

aeon/anomaly_detection/_idk.py
Ramana-Raja: done

aeon/anomaly_detection/tests/test_idk.py
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    )
+    y = [0.52333333, 0.19, 0.52333333]
+    ad = IDK(psi1=8, psi2=2, width=1, random_state=2)
+    pred = ad.fit_predict(series)
+    ad_sliding = IDK(psi1=16, psi2=4, width=10, sliding=True, random_state=1)
+    pred_sliding = ad_sliding.fit_predict(series)
+    ad_2 = IDK(psi1=4, psi2=2, width=3, t=10)
+    pred2 = ad_2.fit_predict(series1)
+    mae = np.mean(np.abs(y - pred2))
```
SebastianSchmidl, line 33: If you compare two arrays, for which you are not sure that they match equally, you can use the methods from `numpy.testing`:, e.g.
- https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose
- https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html

aeon/anomaly_detection/tests/test_idk.py
SebastianSchmidl: Please put the test case in a new function.

SebastianSchmidl (changes requested):
The testing is still too weak.
TBH, I cannot follow the details of the code, so I think that we need a second person to check whether this implementation produces the same results as the implementation by the original authors of IDK.

aeon/anomaly_detection/_idk.py
```diff
@@ -83,9 +73,9 @@ class IDK(BaseAnomalyDetector):
...
         self,
-        psi1: int,
-        psi2: int,
-        width: int,
+        psi1: int = 8,
+        psi2: int = 2,
+        width: int = 1,
         t: int = 100,
         sliding: bool = False,
         random_state: int = None,
```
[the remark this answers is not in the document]
Ramana-Raja: Done

aeon/anomaly_detection/tests/test_idk.py
Ramana-Raja: Changed mae to asser_allclose

aeon/anomaly_detection/tests/test_idk.py
Ramana-Raja: I used aeon.utils.windowing.reverse_windowing to make output shape same as input when sliding is set to True.

SebastianSchmidl: Usage of `reverse_windowing` seems fine to me 👍🏼

MatthewMiddlehurst:
Are there any other implementations or results we can use to compare output? Accuracy to the original algorithm(s) is important. I am not familiar with this area, but having incorrect implementations is generally damaging. Code could be fine, but not a good idea to merge until someone if confident that it is.

Ramana-Raja:
> Are there any other implementations or results we can use to compare output? Accuracy to the original algorithm(s) is important. I am not familiar with this area, but having incorrect implementations is generally damaging. Code could be fine, but not a good idea to merge until someone if confident that it is.

The expected results array for the last test case was generated using the real algorithm I used for comparison with this anomaly detector. Let me know if you'd like me to run any additional tests

MatthewMiddlehurst:
@SebastianSchmidl any suggestion on how to best evaluate this? Does not have to be as rigorous as a publication of course just enough to give us the idea the output is similar enough.

SebastianSchmidl:
> @SebastianSchmidl any suggestion on how to best evaluate this? Does not have to be as rigorous as a publication of course just enough to give us the idea the output is similar enough.

How do you test the implementations in the other modules?

I would compare this implementation against the original code: Probably, I would run the original code in a Docker image and capture the output for some test cases (including edge cases). Then, we can use the same input and output as fixtures for the tests in aeon, and use `np.testing.assert_close` et al. to ignore rounding problems.

Unfortunately, I need to focus on other stuff until mid-summer. So, I cannot say if or when I would be able to do this.

MatthewMiddlehurst:
@SebastianSchmidl For classification we would just say run both on a decent number of UCR datasets and provide the accuracy results for all those datasets. 

No rush on you doing anything here, the onus is on the code contributor to prove that output is equivalent IMO unless they are trusted by the community or the algorithm author/similar. If you can get around to doing it eventually thats great, but there is no obligation to do anything at all really.

Ramana-Raja:
> np.testing.assert_close

sorry for the delay, been caught up with exams. I ran the anomaly detection using np.random.normal 100 times, each with a sample size of 100(used np.testing.assert_allclose ,mean and variance),I also added random_state to the original code since not using it can lead to a lot of variation in the results, here is the data and results:

data/code:
image
results:
image

MatthewMiddlehurst:
This does not really help us determine whether the implementation is accurate to the original algorithm. We would be looking for a comparison against the original implemented (or published results) using actual datasets and evaluation metrics.

Ramana-Raja:
> This does not really help us determine whether the implementation is accurate to the original algorithm. We would be looking for a comparison against the original implemented (or published results) using actual datasets and evaluation metrics.

I used AUC to evaluate both the original model and AEON, and they produced identical results. This is expected since I set a random state for the original model, ensuring consistency. The scores are relatively low because I used only a width of 1, which otherwise reduces the output size of the original model. Here are the results:

code(dataset used : kdd_tsad_135,ecg_diff_count_3):
image
output:
image

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+
+        subsequence_fm_list = (cumsum[self.width :] - cumsum[: -self.width]) / float(
+            self.width
+        )
+
+        return self._idk(X=subsequence_fm_list, psi=self.psi2, t=self.t, rng=rng)
+
+    def _predict(self, X):
+        rng = np.random.default_rng(self.random_state)
+        if self.sliding or self.width > 1:
```
MatthewMiddlehurst, line 180: why `self.width > 1`? This does not seem to match the docstring where it mentioned this used for both the sliding and fixed window.

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+        sample2sample = point2sample[sample_indices, :]
+        row, col = np.diag_indices_from(sample2sample)
+        sample2sample[row, col] = np.nan
+
+        radius_list = np.nanmin(sample2sample, axis=1)
+        min_dist_point2sample = np.argmin(point2sample, axis=1)
+
+        return point2sample, radius_list, min_dist_point2sample
+
+    def _ik_inne_fm(self, X, psi, t, rng):
```
MatthewMiddlehurst, line 109: This seems to be a typo in the function name?

MatthewMiddlehurst (commented):
Hi, I'm a little concerned about the licensing the the original code. Please ensure this is an adaptation, and there are not significant chunks of copy and pasted code if any are currently.

MatthewMiddlehurst:
That is a good base for evaluating but please clean up the code examples before you post, it is a little difficult to parse. Why have you limited it to a width of 1, that seems a bit odd for windowing?

Ramana-Raja:
> That is a good base for evaluating but please clean up the code examples before you post, it is a little difficult to parse. Why have you limited it to a width of 1, that seems a bit odd for windowing?

I made that change because, in the original model, the output size is reduced based on the window size. However, the Aeon model uses a reverse window to restore the original output size. I didn’t use the exact same reverse window since I felt it would change too much of the original implementation. But if you'd prefer that approach, I'm happy to make the adjustment and reevaluate the test again .

Ramana-Raja:
> That is a good base for evaluating but please clean up the code examples before you post, it is a little difficult to parse. Why have you limited it to a width of 1, that seems a bit odd for windowing?

I made that change because, in the original model, the output size is reduced based on the window size. However, the Aeon model uses a reverse window to restore the original output size. I didn’t use the exact same reverse window since I felt it would change too much of the original implementation. But if you'd prefer that approach, I'm happy to reevaluate the test again.

aeon/anomaly_detection/_idk.py
Ramana-Raja: @SebastianSchmidl mentioned that the output should be same size as input, if the width becomes greater than 1 it reduces the size of output so i thought i would add reverse window for that too.

aeon/anomaly_detection/_idk.py
Ramana-Raja: changed function names

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+            onepoint_matrix[ind, min_dist_point2sample[ind]] = 1
+
+        return onepoint_matrix
+
+    def _idk(self, X, psi, t, rng):
+        point_fm_list = self._ik_inne_fm(X=X, psi=psi, t=t, rng=rng)
+        feature_mean_map = np.mean(point_fm_list, axis=0)
+        return np.dot(point_fm_list, feature_mean_map) / t
+
+    def _idk_t(self, X, rng):
```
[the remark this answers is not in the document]
Ramana-Raja: done

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+    Notes
+    -----
+    This implementation is inspired by the Isolation Distributional Kernel (IDK)
+    approach as detailed in [1]_.
+    The code is adapted from the open-source repository [2]_.
+
+    References
+    ----------
+    [1] Isolation Distributional Kernel: A New Tool for Kernel-Based Anomaly Detection.
+         DOI: https://dl.acm.org/doi/10.1145/3394486.3403062
```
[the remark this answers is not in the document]
Ramana-Raja: done

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+    """IDK² and s-IDK² anomaly detector.
+
+    The Isolation Distributional Kernel (IDK) is a data-dependent kernel for efficient
+    anomaly detection, improving accuracy without explicit learning. Its extension,
+    IDK², simplifies group anomaly detection, outperforming traditional methods in
+    speed and effectiveness.
+
+    Parameters
+    ----------
+    psi1 : int
```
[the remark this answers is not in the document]
Ramana-Raja: done

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
...
+         but increase the runtime.
+    sliding : bool, default=False
+         Determines whether a sliding window approach is used for anomaly detection.
+         If True, the model computes scores for overlapping windows across the
+         time series,
+         providing more detailed anomaly scores at each step. If False, the
+         model processes
+         the data in fixed-width segments, offering faster computation at the
+         cost of granularity.
+    random_state : int, Random state or None, default=None
```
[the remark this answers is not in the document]
Ramana-Raja: done

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
+from typing import Optional
+
+import numpy as np
+
+from aeon.anomaly_detection.base import BaseAnomalyDetector
+from aeon.utils.windowing import reverse_windowing
+
+
+class IDK(BaseAnomalyDetector):
```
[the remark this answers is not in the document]
Ramana-Raja: changed name to IDK2

aeon/anomaly_detection/_idk.py
```diff
@@ -0,0 +1,213 @@
+from typing import Optional
```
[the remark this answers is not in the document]
Ramana-Raja: added

aeon/anomaly_detection/_idk.py
MatthewMiddlehurst: Hi that is correct, it should be the same length as the input.

MatthewMiddlehurst:
I wouldn't change the tests given we want it to work with width>3, I don't think this will be merged until that is resolved somehow.

I don't know enough about the algorithm and the windowing function to help with that unfortunately, does using stride and padding_length not work?

Ramana-Raja:
> > I wouldn't change the tests given we want it to work with width>3, I don't think this will be merged until that is resolved somehow.
> > I don't know enough about the algorithm and the windowing function to help with that unfortunately, does using stride and padding_length not work?
> 
> Hi, this Aeon implementation works just as well as the original implementation, with one key difference: while the original model reduces the input size, Aeon attempts to restore it. This approach works well with sliding enabled, but without sliding, the output becomes too small for reverse windowing. Even if we apply reverse windowing, the result is just repeated numbers. So, I figured it’s best to return the output as is when sliding is set to false.
> 
> (used width=2) image

Hi @SebastianSchmidl , just wanted to follow up on this PR when you get a chance.

Ramana-Raja:
Hi @SebastianSchmidl , I’ve added all the changes we talked about in the call. Instead of np.repeat, I used reverse_windowing. The reason is that with x.shape = 10 and width = 3, np.repeat gives an output shape of 9 since the original shape is only 3. Let me know if you'd like me to try a different approach.

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    def __init__(
+        self,
+        psi1: int = 8,
+        psi2: int = 2,
+        width: int = 1,
+        t: int = 100,
+        sliding: bool = False,
+        random_state: Optional[int] = None,
```
SebastianSchmidl, line 94: The documentation states that also `np.random.RandomState` can be passed, which one is correct?

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
+"""IDK2 anomaly detector."""
+
```
SebastianSchmidl, line 2: Please include the  `__mainainer__` and `__all__` properties.

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+         IsolationKernel/Codes: IDK Implementation for Time Series Data
+         URL: https://github.com/IsolationKernel/Codes/tree/main/IDK/TS
+
+    References
+    ----------
+    [1]_ Kai Ming Ting, Bi-Cun Xu, Takashi Washio, Zhi-Hua Zhou (2020)
+         'Isolation Distributional Kernel: A New Tool for Kernel-Based
+          Anomaly Detection',
+          DOI: https://dl.acm.org/doi/10.1145/3394486.3403062
+
```
SebastianSchmidl, line 78: Two examples of how to use this estimator (IDK² and s-IDK²) would be nice.

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+        params : dict
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        """
+        return {
+            "psi1": 8,
+            "psi2": 2,
+            "width": 1,
+        }
```
SebastianSchmidl, line 236: @MatthewMiddlehurst should we define two parameter sets, one for IDK² and one for s-IDK² (sliding window variant), here?

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+
+            for i in range(X.shape[0]):
+                if (
+                    point2sample[i][min_dist_point2sample[i]]
+                    < radius_list[min_dist_point2sample[i]]
+                ):
+                    onepoint_matrix[i][time] = (
+                        min_dist_point2sample[i] + time * self.psi1
+                    )
+                    featuremap_count[(int)(i / self.width)][
```
SebastianSchmidl, line 161:
```suggestion
                    featuremap_count[int(i / self.width)][
```
looks a bit cleaner (also in other places)

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+                        onepoint_matrix[i][time]
+                    ] += 1
+
+        for i in range(window_num):
+            featuremap_count[i] /= self.width
+        isextra = X.shape[0] - (int)(X.shape[0] / self.width) * self.width
+        if isextra > 0:
+            featuremap_count[-1] /= isextra
+        if isextra > 0:
```
SebastianSchmidl, line 170:
```suggestion
```

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+            featuremap_count = np.delete(
+                featuremap_count, [featuremap_count.shape[0] - 1], axis=0
+            )
```
SebastianSchmidl, line 173: This would delete the entry that you just modified with `featuremap_count[-1] /= isextra` in L169. Please check the code again, looks fishy.

aeon/anomaly_detection/distribution_based/tests/test_idk.py
```diff
@@ -0,0 +1,75 @@
...
+    series = rng.normal(size=(100,))
+    series[50:58] -= 5
+
+    ad = IDK2(psi1=8, psi2=2, width=7, random_state=2)
+    pred = ad.fit_predict(series)
+
+    assert ad.original_output_.shape == (14,)
+    assert pred.shape == (100,)
+    assert pred.dtype == np.float64
+    assert 77 <= np.argmax(pred) <= 80
```
SebastianSchmidl, line 35: If the anomaly is at 50:58, the peak in the anomaly score should also be there. Otherwise, this test does not make any sense!

aeon/anomaly_detection/distribution_based/tests/test_idk.py
```diff
@@ -0,0 +1,75 @@
...
+def test_idk_univariate_basic_greater_width():
+    """Test IDK on basic univariate data with width greater than 1."""
+    rng = np.random.default_rng(seed=2)
+    series = rng.normal(size=(100,))
+    series[50:58] -= 5
+
+    ad = IDK2(psi1=8, psi2=2, width=7, random_state=2)
+    pred = ad.fit_predict(series)
+
+    assert ad.original_output_.shape == (14,)
```
SebastianSchmidl, line 32: You create ` ⌈100/7⌉ = 15` windows (by `window_num = int(np.ceil(X.shape[0] / self.width))`), so I would also expect 15 anomaly scores. What is happening here?

aeon/anomaly_detection/distribution_based/tests/test_idk.py
```diff
@@ -0,0 +1,75 @@
...
+    rng = np.random.default_rng(seed=2)
+    series = rng.normal(size=(100,))
+    series[50:58] -= 5
+
+    ad_sliding = IDK2(psi1=16, psi2=4, width=10, sliding=True, random_state=1)
+    pred_sliding = ad_sliding.fit_predict(series)
+
+    assert pred_sliding.shape == (100,)
+    assert pred_sliding.dtype == np.float64
+    assert 60 <= np.argmax(pred_sliding) <= 80
```
SebastianSchmidl, line 49: same here: anomaly is between 50 and 58

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+
+        subsequence_fm_list = (cumsum[self.width :] - cumsum[: -self.width]) / float(
+            self.width
+        )
+
+        return self._compute_idk_score(
+            X=subsequence_fm_list, psi=self.psi2, t=self.t, rng=rng
+        )
+
+    def _predict(self, X):
```
SebastianSchmidl, line 194: Please validate your estimator's parameters against the input in `_predict`. Example:

https://github.com/aeon-toolkit/aeon/blob/89c300932e7ce6246301a6ac0201ab51b8a7b936/aeon/anomaly_detection/distance_based/_kmeans.py#L126

SebastianSchmidl (changes requested):
I'd like the output of your implementation verified against the original implementation to verify that the output is the same:
Please generate some test cases and compare the results of your implementation to the original implementation (by the paper authors).

aeon/anomaly_detection/distribution_based/tests/test_idk.py
Ramana-Raja: The last window is being deleted by 
```python
if isextra > 0:
    featuremap_count = np.delete(featuremap_count, [featuremap_count.shape[0] - 1], axis=0)
```
I just followed the original implementation:
https://github.com/IsolationKernel/Codes/blob/main/IDK/TS/IDK_T.py

aeon/anomaly_detection/distribution_based/tests/test_idk.py
Ramana-Raja: The peak anomaly scores fall within those values when psi1 and psi2 are sufficiently high or when the dataset is large enough,as shown in the updated test cases.

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Fixed

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Fixed

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Fixed

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Done, I've added two examples

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Done, I've __maintainer__ and __all__

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+         'Isolation Distributional Kernel: A New Tool for Kernel-Based
+          Anomaly Detection',
+          DOI: https://dl.acm.org/doi/10.1145/3394486.3403062
+
+    """
+
+    _tags = {
+        "capability:univariate": True,
+        "capability:multivariate": False,
+        "capability:missing_values": False,
```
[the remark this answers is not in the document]
Ramana-Raja: Added

aeon/anomaly_detection/distribution_based/_idk.py
```diff
@@ -0,0 +1,236 @@
...
+    IDK², simplifies group anomaly detection, outperforming traditional methods in
+    speed and effectiveness.This implementation is inspired by the Isolation
+    Distributional Kernel (IDK) approach as detailed in Kai Ming Ting, Bi-Cun Xu,
+    Takashi Washio, Zhi-Hua Zhou (2020) [1]_.
+
+    This Anomaly Detector assumes the input time series is stationary,
+    so trends should be removed prior to detection. IDK² is recommended
+    for periodic time series,while s-IDK² is better suited for non-periodic
+    cases.The attribute `original_output_` stores the raw anomaly scores before
+    reverse-windowing is applied when width > 1 only.
```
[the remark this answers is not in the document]
Ramana-Raja: Added

aeon/anomaly_detection/distribution_based/_idk.py
Ramana-Raja: Fixed and updated docs

aeon/anomaly_detection/distribution_based/_idk.py
MatthewMiddlehurst: If you want both to be covered by the general testing then sure. If the differences are minor enough that it can be covered in the test file go with that. I would add a new set if there is a chance the output shape could change or any other jank stuff.

SebastianSchmidl: Yes, that is my concern. Then, let us include s-IDK² in the general testing by adding a second parameter set.

TonyBagnall:
@Ramana-Raja I'm converting this to draft, feel free to change if you want to come back to it

Issue #1806, opened 2024-07-15
[DOC] Add examples for QuerySearch in SimilaritySearch


TonyBagnall:
### Describe the issue linked to the documentation

just having a play with similarity search, some examples in the docstring would be helpful.

baraline:
Nice spot, missed it ! I recommend the notebooks in the meantime.

baraline:
Hi, the issue will be made irrelevant one #2473 is done, as QuerySearch estimator will be removed. I'll probably create a bunch of good first issues afterward to fill the gaps with what I won't cover in the PR. 
Thanks for your interest !

PR #2472, opened 2024-12-26
adds examples for querysearch

naxatra2:
Adds #1806 
I have added some examples in the docstring for query search


baraline:
Hi ! Thanks for being willing to contribute to aeon. 

That's totally due to a lack of communication from me, but this module is being reworked, and I didn't update the issues to reflect that. I've put up PR #2473 to reflect the upcoming changes. 

There will be similar things to do as you did here once #2473 is in, so what I can propose to you is that we close this one, and let the examples to be done so you can contribute. Otherwise, I'm sure there is other things that we could use right now, don't hesitate to chat on slack !

Issue #2236, opened 2024-10-21
[BUG] Test fail with numpy2 for stomp_squared_matrix_profile


TonyBagnall:
### Describe the bug

in the module similarity_searc/matrix_profiles the test function `test_stomp_squared_matrix_profile` fails with numpy 2.0 if this parameter is 

K_VALUES = [1,3]

this is caused by the np.argsort method, which by default uses the unstable quicksort which means occasional differences in ties with the two versions. 
```python

        id_bests = np.vstack(
            np.unravel_index(
                np.argsort(expected.ravel(), kind="stable"), expected.shape
            )
        ).T
```

The case K_VALUES = [1] is fixed by making this call stable, but not at all sure about case with 3. See https://github.com/aeon-toolkit/aeon/pull/2216 for more details

### Expected results

set K_VALUES = [1,3] and run with numpy <2.0 (passes) and numpy>=2.0 (fails)

### Actual results

```python-traceback
aeon\similarity_search\matrix_profiles\tests\test_stomp.py:45 (test_stomp_squared_matrix_profile[3-int64])
dtype = 'int64', k = 3

    @pytest.mark.parametrize("dtype", DATATYPES)
    @pytest.mark.parametrize("k", K_VALUES)
    def test_stomp_squared_matrix_profile(dtype, k):
        """Test naive series search."""
        X = np.asarray(
            [[[1, 2, 3, 4, 5, 6, 7, 8]], [[1, 2, 4, 4, 5, 6, 5, 4]]], dtype=dtype
        )
    
        S = np.asarray([[3, 4, 5, 4, 3, 4, 5, 3, 2, 4, 5]], dtype=dtype)
        L = 3
        mask = np.ones((X.shape[0], X.shape[2] - L + 1), dtype=bool)
        distance = get_distance_function("squared")
        mp, ip = stomp_squared_matrix_profile(X, S, L, mask, k=k)
        for i in range(S.shape[-1] - L + 1):
            q = S[:, i : i + L]
    
            expected = np.array(
                [
                    [distance(q, X[j, :, _i : _i + L]) for _i in range(X.shape[-1] - L + 1)]
                    for j in range(X.shape[0])
                ]
            )
            id_bests = np.vstack(
                np.unravel_index(
                    np.argsort(expected.ravel(), kind="stable"), expected.shape
                )
            ).T
    
            for j in range(k):
                assert_almost_equal(mp[i][j], expected[id_bests[j, 0], id_bests[j, 1]])
[...]

Issue #2028, opened 2024-09-02
[DOC] Update similarity search notebook 


baraline:
### Describe the issue linked to the documentation

With the addition of Series search in #1900, the theoretical (deep dive) and benchmark notebook should be updated to include the new estimator. An extension of the similarity search notebook to show the use of the different parameters is also needed.

It would also be nice to vary the examples instead of just using GunPoint...

Issue #2020, opened 2024-09-01
[ENH] Rethink the mask parameter in similarity search


baraline:
### Describe the issue

Currently, a boolean mask is created to indicate which part of the input time series dataset is to be ignored from the computations if the query used in the search is included in the input time series dataset. 

The issue is that the computations linked to these areas are still made, they are only removed from the result array before returning it.

### Suggest a potential alternative/fix

Rethink the mask parameter to be able to skip computation that won't be used instead of ignoring them later on.

vedpawar2254:
@aeon-actions-bot assign @vedpawar2254

MatthewMiddlehurst:
@vedpawar2254 please don't take on multiple issues at once unless the others are close to completion.

vedpawar2254:
Hi, @MatthewMiddlehurst I completed the work and sebastian has reviewed it, I'll update the PR tonight

MatthewMiddlehurst:
@vedpawar2254 you assigned https://github.com/aeon-toolkit/aeon/issues/2081

vedpawar2254:
Yes @MatthewMiddlehurst , I am working on that rn but it's kind of getting the better of so I though I'll try something relatively easier so I can start on that one again with a fresh mind, i'll unassign myself if you want

PR #2473, opened 2024-12-26
[MNT, ENH, DOC] Rework similarity search

### LINKED ISSUE #2341: [ENH] IndexSearch class with attimo or LL algorithm

Implement the IndexSearch class for similarity search. I use the following as a paper trail of which hash/index method could be used as a first use case for this class :

- [attimo](https://www.vldb.org/pvldb/vol15/p3841-ceccarello.pdf) 
- [LL](https://epubs.siam.org/doi/epdf/10.1137/1.9781611975673.16)
- https://drops.dagstuhl.de/storage/00lipics/lipics-vol077-socg2017/LIPIcs.SoCG.2017.37/LIPIcs.SoCG.2017.37.pdf (similar to LL ?)

For the IndexSearch class, we will use the BaseSimilaritySearch class to provide a common interface with QuerySearch and SeriesSearch class. 

Following papers above, the following code base exists and could serve as a base for implementation : 

- [attimo](https://github.com/Cecca/attimo) 
- [LL](https://github.com/xiaoshengli/LL)

_No response_

### Additional context

The choice of wether to go for altimo or LL will depend on how complex it is to implement them, as I'd prefer to go with something simple for the first IndexSearch case.

baraline:
#### Reference Issues/PRs
Fixes #2341, #2236, #2028, #2020, #1806, #2475, #2538

The previous structure for similarity search was not in line with the structure we would expect considering other aeon modules, the lack of distinct base classes for some tasks, as well as the initial design choice (due to the lack of practical experience with using and expanding the module) lead to some really complex code when working on #2341 to make everything work together. Further expanding the module would have made thing worse.

To make the module more flexible and comprehensible, the following rework is proposed in this PR (AEP to be updated acordingly):

The module structure is now : 

```
|-similarity_search
|------series
|------------neighbors #NN on subsequences 
|------------motifs #Motif extraction on subsequences methods
|------collection
|------------neighbors # series methods for subsequence NN adapted to collection case or Approximate NN on whole series
|------------motifs #Series methods adapted to collection case
```
Base classes are `BaseSimilaritySearch, BaseSeriesSimilaritySearch, BaseCollectionSimilaritySearch`
Implemented estimators are :

- (series/neighbors) MassSNN : subsequence nearest neighbors and distance profile computation
- (series/neighbors) DummySNN : brute force subsequence nearest neighbors
- (series/motifs) StompMotifs : top k motifs extraction (supports motifs pairs, k-motif or r-motifs)
- (collection/neighbors) RandomProjectionIndexANN: Approximate nearest neighbors on whole series using a random projection LSH method.

The sufix of the estimators (SNN/ANN/Motifs) remains an open discussion, not sure it's the right way to go.

I removed the support for collections for Stomp and Mass for now to focus on the "expected and well known" use cases, I'll make them in another PR.

All similarity search estimators now use fit/predict interface, with predict returning two arrays (NN/Motifs indexes, and NN/Motifs distances).

No.

#### Any other comments?

As this is still a WIP, I would love some inputs on the structure (notably from @patrickzib !) to make the module more future-proof to future additions and easier to use.

### TODO list :
- [x] Finish to include testing suite for base estimators in the testing module for the `SubsequenceSearch` part and fix them
- [x] Implement LSH index as a simple first case for `BaseCollectionSimilaritySearch`
- [x] Implement tests for base classes and estimators
- [x] Update API docs / doc pages 
- [x] Update notebooks
- [x] Check docstrings
- [x] Cleanup TODOs in the code
- [x] updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.


patrickzib:
Thank you very much for working on this. 

Some thoughts:

> * Focus the module on two distinct tasks : `find_neighbors` and `find_motifs` for all type for similarity search estimators. Similarly to the `fit`/ `predict` interface we already know well, here, we first `fit` and then either `find_motifs` or `find_neighbors` ("predict" keyword don't make much sense here). We give a collection to use as database in `fit`, and a single series in `find_neighbors` or `find_motifs` to use as query for the search.

That is an interesting problem. Here is my view:

For *whole series similarity search* fit requires a dataset of time series of equal length, and find_neighbors would get one or many query series of this length.

For *subsequence similarity search* fit requires a single time series, and find_neighbors commonly gets a single query sequence which length is shorter than the single series length. It would be fine however, to extend it to multiple short sequences.

There is only one *whole series consensus motif search* paper, which would be the use case of whole matching and motif discovery. The input to fit would be the whole dataset, and find_motif has no input series X. not sure, what an input series X should trigger.

Most papers solve the problem of *motif discovery* in a single long time series, defined as subsequences of the time series. Here, fit gets a single series, and find_motif has no input series X.

> * Distinguish between two kind of similarity search tasks with the two submodules, `SeriesSearch` and `SubsequencesSearch`. The `SubsequencesSearch` focuses on tasks for which the goal is to find motifs or neighbors in subsequences of time series (e.g. Matrix Profiles, Motiflets, etc.). the `SeriesSearch` focuses on task using whole series (e.g. Indexes such as LSH, iSAX, etc.)


> * Have base classes for families of method to limit code duplication (e.g. `BaseMatrixProfile`, and STOMP, where most existing code was ported), so the we can focus on implementing the computational logic when adding new estimators.

What is the difference between BaseMatrixProfile and STOMP?

At least for Motiflets, we cannot use STOMP/MP, as it only gives a 1-NN profile, but we need k-NN profiles. Same problem would be the case, if you want to solve  k-nearest neighbors similarity search.




> #### Questions to answers for motif search :
> * Do we want to make providing `X` optional in `find_motifs` ? Providing X means that we search for subsequences in X that are motifs in the collection given in `fit`. Not providing X would mean that we search for motifs in the collection given in `fit` only. I think it would make sense to make it optional, but would love some comment from people actually doing motif discovery.
> 

I think that X is not meaningful for motif discovery.

> #### Does your contribution introduce a new dependency? If yes, which one?
> No.
> 
> #### Any other comments?
> As this is still a WIP, I would love some inputs on the structure (notably from @patrickzib !) to make the module more future-proof to future additions and easier to use.
> 
> ### TODO list :
> * [x]  Finish to include testing suite for base estimators in the testing module for the `SubsequenceSearch` part and fix them
> * [ ]  Implement LSH index as a simple first case for `BaseIndexSearch`
> * [ ]  Implement tests for `SeriesSearch` base class and estimators
> * [ ]  Update API docs / doc pages
> * [ ]  Update notebooks
> * [ ]  Check docstrings
> * [ ]  Cleanup TODOs in the code
> * [ ]  updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

baraline:
Thanks for the inputs @patrickzib😄

> For *whole series similarity search* fit requires a dataset of time series of equal length, and find_neighbors would get one or many query series of this length.

Completely in line with this, but what about the case of unequal length series, with, for example, elastic distance measures? Wouldn't that be a plausible use case?

> For *subsequence similarity search* fit requires a single time series, and find_neighbors commonly gets a single query sequence which length is shorter than the single series length. It would be fine however, to extend it to multiple short sequences.

For this case, I'm defining a length parameter during ``__init__``, and accept a 2D subsequence of shape ``(n_channels, length)`` for ``find_neighbors``, ensuring that ``length`` is not bigger than the length of any series given in `fit`. However, any reason why you think we should restrain ourselves to a single series for `fit` ? 

> There is only one whole series consensus motif search paper, which would be the use case of whole matching and motif discovery. The input to fit would be the whole dataset, and find_motif has no input series X. not sure, what an input series X should trigger.

This is the tricky one for me too. I'm not sure how giving `X` during `find_motifs` on whole series would fit any use case.

> Most papers solve the problem of motif discovery in a single long time series, defined as subsequences of the time series. Here, fit gets a single series, and find_motif has no input series X.

I've been kinda frustrated by this limitation for practical use cases, wouldn't it be fine to loop on series of a collection with the motif discovery methods and then merge the results ? That's how I implemented STOMP for now for example. For each subsequence in `X`, it computes the distance profile to all series in a collection, and keep the top `k` among all of them (also storing the sample ID and timepoint ID).

> What is the difference between BaseMatrixProfile and STOMP?
> At least for Motiflets, we cannot use STOMP/MP, as it only gives a 1-NN profile, but we need k-NN profiles. Same problem would be the case, if you want to solve k-nearest neighbors similarity search.

`BaseMatrixProfile` is simply a base class that defines abstract `compute_matrix_profile` and `compute_distance_profile` methods to be implemented by child classes such as STOMP and the likes (STUMP, etc...). 

The logic for finding neighbors / motifs is then handled in the `BaseSubsequenceSearch`.
As stated above, I already extended STOMP to support k-NN profiles for collections (multivariate and unequal length compatible).

I suppose that in this context, motiflets would either inherit from `BaseMatrixProfile` if you need to implement methods like  `compute_matrix_profile` and `compute_distance_profile`. Otherwise, It would inherit from `BaseSubsequenceSearch` and make its own methods to answer the `find_neighbors`/`find_motifs` tasks. (I would need to read the paper again!)

Note that it's possible to simply raise a "NotImplementedError" or something similar if a method only supports neighbors or motifs search.

My goal here is to find a base class structure that enables us to move most common code to there and focus on the computational optimisations of each method in the child classes.

patrickzib:
> Completely in line with this, but what about the case of unequal length series, with, for example, elastic distance measures? Wouldn't that be a plausible use case? (all whole series estimators don't have to support it)

Sure. I  did not think of this. 

> > For _subsequence similarity search_ fit requires a single time series, and find_neighbors commonly gets a single query sequence which length is shorter than the single series length. It would be fine however, to extend it to multiple short sequences.
> 
> For this case, I'm defining a length parameter during `__init__`, and accept a 2D subsequence of shape `(n_channels, length)` for `find_neighbors`, ensuring that `length` is not bigger than the length of any series given in `fit`. However, any reason why you think we should restrain ourselves to a single series for `fit` ?

Simplicity :) But I agree that you could have multiple series in fit, too - this would mimic the Shapelet use case, I suppose?

> > Most papers solve the problem of motif discovery in a single long time series, defined as subsequences of the time series. Here, fit gets a single series, and find_motif has no input series X.
> 
> I've been kinda frustrated by this limitation for practical use cases, wouldn't it be fine to loop on series of a collection with the motif discovery methods and then merge the results ? That's how I implemented STOMP for now for example. For each subsequence in `X`, it computes the distance profile to all series in a collection, and keep the top `k` among all of them (also storing the sample ID and timepoint ID).

Sorry, yes, that is what the authors refer to as consensus motif:
https://www.cs.ucr.edu/~eamonn/consensus_Motif_ICDM_Long_version.pdf


> `BaseMatrixProfile` (which inherit `BaseSubsequenceSearch`) is simply a base class that defines abstract `compute_matrix_profile` and `compute_distance_profile` methods to be implemented by child classes such as STOMP and the likes (STUMP, etc...). 

I see. I personally do not like to use the terms matrix-profile for simple k-NN distances or k-NN indices though. It was a brilliant re-framing of EK, such that all 1-NN algorithms are now suddenly an instance of matrix profile. Yet, the concept is much older.

> As stated above, I already extended STOMP to support k-NN profiles for collections (multivariate and unequal length compatible).

Great. 


> > I think that X is not meaningful for motif discovery.
> 
> In the context of motif search in a single series I agree, but wouldn't there be some interest when dealing with a collection ? For example find motifs in the collection at the condition that they are similar to a subsequence in X ? (This is pure speculation)

I would not say that this is impossible, but I have not seen it. :)

baraline:
> I see. I personally do not like to use the terms matrix-profile for simple k-NN distances or k-NN indices though. It was a brilliant re-framing of EK, such that all 1-NN algorithms are now suddenly an instance of matrix profile. Yet, the concept is much older.

I'm not against the idea of a different naming, especially if methods labelled differently from MPs would fit in the base class without any change of parameter/interface. Would you have any proposal? Something like `BaseNeighborhoodSearch` ?

patrickzib:
> I'm not against the idea of a different naming, especially if methods labelled differently from MPs would fit in the base class without much change of parameter/interface. Would you have any proposal? Something like `BaseNeighborhoodSearch` ?

In sklearn it is simply `NearestNeighbors` ? :) And it returns indices and distances. 

https://scikit-learn.org/1.5/modules/neighbors.html

baraline:
I'll leave the implementation of `k-motiflets` for later as a separate estimator so I can focus on getting the testing, tags and docs right for this PR to have the structure for the module set.

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+class MassSNN(BaseSeriesSimilaritySearch):
+    """
+    Estimator to compute the subsequences nearest neighbors using MASS _[1].
+
+    Parameters
+    ----------
+    length : int
+        The length of the subsequences to use for the search.
+    normalize : bool
+        Wheter the subsequences should be z-normalized.
```
patrickzib, line 32: Whether

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+        ----------
+        X : np.ndarray, shape=(n_channels, length)
+            Subsequence we want to find neighbors for.
+        k : int
+            The number of neighbors to return.
+        dist_threshold : float
+            The maximum allowed distance of a candidate subsequence of X_ to X
+            for the candidate to be considered as a neighbor.
+        allow_trivial_matches: bool, optional
+            Wheter a neighbors of a match to a query can be also considered as matches
```
patrickzib, line 82: whether

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+        if parameter_set == "default":
+            params = {"length": 20}
+        else:
+            raise NotImplementedError(
+                f"The parameter set {parameter_set} is not yet implemented"
+            )
+        return params
+
+
+@njit(cache=True, fastmath=True)
```
patrickzib, line 199: parallel=True missing

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+def _squared_distance_profile(QT, T, Q):
+    """
+    Compute squared distance profile between query subsequence and a single time series.
```
patrickzib, line 202: It is Euclidean Distance, right?

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+
+        distance_profile += _qsum + _QT[k]
+        distance_profile[0] += _sum
+        for i in prange(1, profile_length):
+            _sum += T[k, i + (query_length - 1)] ** 2 - T[k, i - 1] ** 2
+            distance_profile[i] += _sum
+    return distance_profile
+
+
+@njit(cache=True, fastmath=True)
```
patrickzib, line 243: parallel=true missing

patrickzib (changes requested):
Sorry, only a very brief review. I would have to test the code to give a better review, but I will be leaving for vacation...

aeon/similarity_search/_base.py
```diff
@@ -0,0 +1,73 @@
...
+    def predict(
+        self,
+        X: Union[np.ndarray, None] = None,
+    ):
+        """
+        Predict method.
+
+        Parameters
+        ----------
+        X : 2D np.array of shape ``(n_cases, n_timepoints)``
```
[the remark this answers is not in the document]
baraline: I just didn't find much use case for allowing 3D time series in the predict method. For example, when querying an index, you give a single time series, the same when you perform a query search with MASS, or when computing A/B MPs. 
Allowing 3D could be of interest to perform batch “predictions”, which could make sense in the indexing case, but not so much in the data mining task.

If you have some examples in mind of other input types that would be useful to support, this is not set in stone.

(None is when we perform a analysis on the data given during fit without any querying)

aeon/similarity_search/collection/_base.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @final
+    def predict(self, X, **kwargs):
+        """
+        Predict function.
+
+        Parameters
+        ----------
+        X : np.ndarray, shape = (n_cases, n_channels, n_tiempoints)
+            Collections of series to predict on.
```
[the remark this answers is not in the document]
baraline: Do you mean different from `BaseSimilaritySearch` fit ? 

I convert single series to 3D `(1, n_channels, n_timepoints)`, but I didn't yet add support for unequal length as there was no estimator using it.

aeon/similarity_search/collection/motifs/__init__.py
```diff
@@ -0,0 +1 @@
+"""Motif search for time series collection."""
```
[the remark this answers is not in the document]
baraline: I was aiming for something more generic than consensus motifs, simply motif discovery on a collection of time series instead on a single series. I didn't want to bind the module to one terminology. But consensus motif would be the first thing to implement in there yes !

aeon/similarity_search/collection/neighbors/_rp_cosine_lsh.py
```diff
@@ -0,0 +1,289 @@
...
+from numba import get_num_threads, njit, prange, set_num_threads
+
+from aeon.similarity_search.collection._base import BaseCollectionSimilaritySearch
+from aeon.utils.numba.general import z_normalise_series_3d
+
+
+@njit(cache=True)
+def _hamming_dist(X, Y):
+    d = 0
+    for i in prange(X.shape[0]):
```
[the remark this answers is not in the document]
baraline: I don't think it is, just a habbit, I will test it to be sure !

aeon/similarity_search/collection/neighbors/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Neighbors search for time series collection."""
```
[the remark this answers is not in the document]
baraline: My idea here is that we would support both whole and subsequence search. But there might be a need for a clearer distinction between the two I agree.

For whole series estimators (example the LSH index estimator by design made for whole series), a generic wrapper to extract subsequences from the input collection would make such estimators compatible with subsequence search.

For subsequences search estimators, I am yet to find a method that is not a loop on single series nn estimators (e.g. MASS) for each series of the collection.

So maybe a submodule to split whole series/subsequence nearest neighbour search ?

aeon/similarity_search/collection/neighbors/_rp_cosine_lsh.py
```diff
@@ -0,0 +1,289 @@
...
+
+import numpy as np
+from numba import get_num_threads, njit, prange, set_num_threads
+
+from aeon.similarity_search.collection._base import BaseCollectionSimilaritySearch
+from aeon.utils.numba.general import z_normalise_series_3d
+
+
+@njit(cache=True)
+def _hamming_dist(X, Y):
```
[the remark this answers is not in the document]
baraline: I use it with boolean inputs indeed here as i'm using ^ operator which is bitwise, so yeah naming could be more precise

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,146 @@
...
+        Raises
+        ------
+        TypeError
+            If the input X array is not 2D raise an error.
+
+        Returns
+        -------
+        self
+        """
+        self.reset()
```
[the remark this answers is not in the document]
baraline: I think `_preprocess_series` should check for that, will confirm

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,146 @@
...
+            if hasattr(self, "length"):
+                max_timepoints -= self.length
+            if X_index >= max_timepoints or X_index < 0:
+                raise ValueError(
+                    "The value of X_index cannot exced the number "
+                    "of timepoint in series given during fit. Expected a value "
+                    f"between [0, {max_timepoints - 1}] but got {X_index}"
+                )
+
+    def _check_predict_series_format(self, X):
```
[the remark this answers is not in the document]
baraline: Mostly for cheking the number of channels between fit/predict. I think I could call `_preprocess_series` also here for the ndims/type.

aeon/similarity_search/series/_commons.py
```diff
@@ -0,0 +1,221 @@
...
+        ``length`` the size of the query used to compute the distance profiles.
+    k : int
+        Number of best matches to return
+    threshold : float
+        A threshold on the distances of the best matches. To be returned, a candidate
+        must have a distance bellow this threshold. This can reduce the number of
+        returned matches to be bellow ``k``
+    allow_trivial_matches : bool
+        Wheter to allow returning matches that are in the same neighborhood.
+    exclusion_size : int
```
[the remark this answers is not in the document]
baraline: I use the size to define the zone later on :
```python 
min(
    top_k_indexes[_current_k] + exclusion_size,
    len(dist_profile),
)
max(top_k_indexes[_current_k] - exclusion_size, 0)
```
or is the exclusion zone parameter referring to this +/- directly in the litterature?

aeon/similarity_search/series/neighbors/_mass.py
baraline: We cannot parallelise MASS (I think ?), I used pranges out of habbit, will check if it changes anything.

aeon/similarity_search/series/neighbors/_mass.py
baraline: Yes, missing from documentation

aeon/similarity_search/series/neighbors/_mass.py
baraline: Same comment for MASS parallelisation.

aeon/similarity_search/series/neighbors/_mass.py
```diff
@@ -0,0 +1,295 @@
...
+            if Q_is_constant[k] and Sub_is_constant[k]:
+                _val = 0
+            # One Constant case
+            elif Q_is_constant[k] or Sub_is_constant[k]:
+                _val = query_length
+            else:
+                denom = query_length * Q_stds[k] * T_stds[k, i]
+
+                p = (QT[k, i] - query_length * (Q_means[k] * T_means[k, i])) / denom
+                p = min(p, 1.0)
```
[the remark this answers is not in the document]
baraline: I don't remember, that's copied from previous code.

I'll try to find again why, i'm sure there is a good reason

aeon/similarity_search/collection/neighbors/__init__.py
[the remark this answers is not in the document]
baraline: Another way to put it is that the module is split by input types (single series and collections) rather than by task. Then, submodules like `neighbors` implement estimators that perform these tasks for the given input type.

If an estimator works for both cases (`series` and `collection`), we would only implement it in `collection` and convert single series into collections.

aeon/similarity_search/collection/neighbors/_rp_cosine_lsh.py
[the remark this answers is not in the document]
baraline: It is equivalent, so I've reverted back to range

aeon/similarity_search/series/_base.py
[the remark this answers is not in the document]
baraline: `_preprocess_series` perform this check

aeon/similarity_search/series/_base.py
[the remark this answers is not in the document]
baraline: `_preprocess_series` is also called during predict, so I adjusted the check

aeon/similarity_search/series/neighbors/_mass.py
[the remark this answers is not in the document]
baraline: Found it in all other MASS codes, but not in the paper, so I'm not 100% sure

docs/getting_started.md
```diff
@@ -21,8 +21,9 @@ classical techniques for the following learning tasks:
 - [**Clustering**](api_reference/clustering), where a collection of time series without
   any labels are used to train a model to label cases
   ([more details](examples/clustering/clustering.ipynb)).
-- [**Similarity search**](api_reference/similarity_search), where the goal is to evaluate
-  the similarity between a query time series and a collection of other longer time series
+- [**Similarity search**](api_reference/similarity_search), where the goal is to find
+  time series motifs or nearest neighbors in an efficient way for either single series
+  or collections.
```
MatthewMiddlehurst, line 26: There is an example on this page which is now outdated.

aeon/utils/tags/_tags.py
```diff
@@ -155,7 +155,7 @@ class : identifier for the base class of objects this tag applies to
         "values?",
     },
     "input_data_type": {
-        "class": "transformer",
+        "class": ["transformer", "similarity-search"],
```
MatthewMiddlehurst, line 158: I would update the description this is needed for more than transformers now

aeon/utils/numba/general.py
```diff
@@ -451,15 +454,56 @@ def get_subsequence_with_mean_std(
     return values, means, stds
 
 
+@njit(cache=True, fastmath=True, parallel=True)
+def compute_mean_stds_collection_parallel(X):
```
MatthewMiddlehurst, line 458: Thanks for updating the list at the top, I think this one is still missing though? Alternatively should be protected

MatthewMiddlehurst: Not required here but helpful: updating the utils API page where relevant

aeon/utils/discovery.py
```diff
@@ -92,6 +92,7 @@ def all_estimators(
         # ignore test modules and base classes
         "base",
         "tests",
+        "similarity_search"
```
MatthewMiddlehurst, line 95: I wouldn't think we want to ignore these. All of these are essentially empty of estimators or for testing only.

docs/api_reference/utils.rst
```diff
@@ -87,7 +87,6 @@ Mock Estimators
     MockUnivariateSeriesTransformer
     MockMultivariateSeriesTransformer
     MockSeriesTransformerNoFit
-    MockSimilaritySearch
```
MatthewMiddlehurst, line 90: This is missing the two new ones

aeon/testing/mock_estimators/__init__.py
```diff
@@ -29,8 +29,6 @@
     "MockUnivariateSeriesTransformer",
     "MockMultivariateSeriesTransformer",
     "MockSeriesTransformerNoFit",
-    # similarity search
-    "MockSimilaritySearch",
```
MatthewMiddlehurst, line 33: Missing the two new ones

MatthewMiddlehurst (commented):
Some comments for the bits outside the actual module itself. I will have a look at the rest as well, but generally trust you and Patrick agree on something sensible together. Let me know if any of this contradicts a previous review.

aeon/similarity_search/collection/_base.py
```diff
@@ -0,0 +1,120 @@
...
+from abc import abstractmethod
+from typing import final
+
+import numpy as np
+
+from aeon.base import BaseCollectionEstimator
+from aeon.similarity_search._base import BaseSimilaritySearch
+
+
+class BaseCollectionSimilaritySearch(BaseCollectionEstimator, BaseSimilaritySearch):
```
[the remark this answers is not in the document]
patrickzib: Maybe add typical use cases to the description of the template classes?

aeon/similarity_search/series/motifs/_stomp.py
```diff
@@ -0,0 +1,498 @@
...
+            ``allow_trivial_matches`` is set to False. For a given timestamp,
+            the exclusion zone starts from
+            :math:`id_timestamp - length//exclusion_factor` and end at
+            :math:`id_timestamp + length//exclusion_factor`.
+        inverse_distance : bool
+            If True, the matching will be made on the inverse of the distance, and thus,
+            the farther neighbors will be returned instead of the closest ones.
+        motif_extraction_method : str
+            A string indicating the methodology to use to extract the top motifs.
+            Available methods are "r_motifs" and "k_motifs". "r_motifs" means we rank
```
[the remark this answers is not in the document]
baraline: Improved docstring to clear confusion

aeon/similarity_search/series/motifs/_stomp.py
```diff
@@ -0,0 +1,498 @@
...
+        motif_size : int
+            The number of subsequences in a motif. Default is 1, meaning we extract
+            motif pairs (the query and its best match).
+        dist_threshold : float
+           The maximum allowed distance of a candidate subsequence of X to a query
+           subsequence from X_ for the candidate to be considered as a neighbor.
+        inverse_distance : bool
+            If True, the matching will be made on the inverse of the distance, and thus,
+            the worst matches to the query will be returned instead of the best ones.
+        exclusion_factor : float, default=1.
```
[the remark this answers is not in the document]
baraline: Replaced by multiplication instead of division in all occurences

aeon/similarity_search/collection/tests/test_base.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_input_shape_fit_predict_collection():
+    """Test input shapes."""
+    estimator = MockCollectionSimilaritySearch()
+    # dummy data to pass to fit when testing predict/predict_proba
+    X_3D_uni = make_example_3d_numpy(n_channels=1, return_y=False)
+    X_3D_multi = make_example_3d_numpy(n_channels=2, return_y=False)
+    X_2D_uni = make_example_2d_numpy_series(n_channels=1)
+    X_2D_multi = make_example_2d_numpy_series(n_channels=2)
+    X_1D = make_example_1d_numpy()
```
MatthewMiddlehurst, line 25: I would use the test data dict in testing. There are a lot more potential inputs than this IMO.

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Returns
+        -------
+        self
+        """
+        self.reset()
+        X = self._preprocess_series(X, self.axis, True)
+        # Store minimum number of n_timepoints for unequal length collections
+        self.n_channels_ = X.shape[0]
+        self.n_timepoints_ = X.shape[1]
```
MatthewMiddlehurst, line 58: We also have a `self.metadata_` here.

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,142 @@
...
+        self.X_ = X
```
MatthewMiddlehurst, line 59: This is likely going to be awful for pickling unless it is actually required for every approach

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,142 @@
...
+        ----------
+        X : np.ndarray, shape = (n_channels, n_timepoints)
+            A series to be used in predict.
+
+        """
+        if self.n_channels_ != X.shape[0]:
+            raise ValueError(
+                f"Expected X to have {self.n_channels_} channels but"
+                f" got {X.shape[0]} channels."
+            )
```
MatthewMiddlehurst, line 142: I'd say use `_check_shape` again but it looks like we dont have one

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,142 @@
...
+    def _check_X_index(self, X_index: int):
+        """
+        Check wheter a X_index parameter is correctly formated and is admissible.
+
+        Parameters
+        ----------
+        X_index : int
+            Index of a timestamp in X_.
+
+        """
```
MatthewMiddlehurst, line 113: Like `self.X_` I dont think this should be here unless its used by everything, in which case the parameter should be more integrated into the base class.

aeon/similarity_search/series/_base.py
```diff
@@ -0,0 +1,142 @@
...
+        distances : np.ndarray, shape = (k)
+            Distance of the matches to each series
+
+        """
+        self._check_is_fitted()
+        if X is not None:
+            X = self._preprocess_series(X, self.axis, False)
+            self._check_predict_series_format(X)
+        else:
+            X = self.X_
```
MatthewMiddlehurst, line 97: If you have to save a model you dont really want to save the training data in most cases which this forces. If you want to keep the optimisation have a `fit_predict` probably.

MatthewMiddlehurst (commented):
Base class comments

aeon/similarity_search/collection/_base.py
```diff
@@ -0,0 +1,125 @@
...
+        TypeError
+            If the input X array is not 3D raise an error.
+
+        Returns
+        -------
+        self
+        """
+        self.reset()
+        X = self._preprocess_collection(X)
+        # Store minimum number of n_timepoints for unequal length collections
```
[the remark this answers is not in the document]
baraline: Rather something I forgot here, I'm not sure if that kind of operation is not best left to child classes when the base class support both cases.

aeon/similarity_search/collection/_base.py
```diff
@@ -0,0 +1,125 @@
...
+                )
+        else:
+            raise TypeError(
+                "Expected a 3D np.ndarray in predict but found" f" {type(X)}."
+            )
+        if self.n_channels_ != X[0].shape[0]:
+            raise ValueError(
+                f"Expected X to have {self.n_channels_} channels but"
+                f" got {X[0].shape[0]} channels."
+            )
```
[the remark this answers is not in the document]
baraline: Agreed on the collection/2d side, but I don't think preprocess collections checks for consistency between number of channels between fit and predict ?

aeon/similarity_search/collection/tests/test_base.py
baraline: Switched to use _get_datatypes_for_estimator and FULL_TEST_DATA_DICT

aeon/similarity_search/series/_base.py
baraline: Likely need to keep X_ for anything as you need to compare the fit series to the predict series, I'm not sure there is an alternative to storing it if that what you meant.

aeon/similarity_search/series/_base.py
baraline: added timepoint index with axis as it is not stored in metadata

aeon/similarity_search/series/_base.py
baraline: You mean to ensure same number of channels between fit and predict as in collection right ? Can raise an issue so we add the check to BaseAeon later

aeon/similarity_search/series/_base.py
baraline: It will be used for nearly all series/neighbors estimators, the X_ is similar to storing the fit data for a KNN, I don't think there is a way around it in this context

baraline: The solution would be to have a base class for series/neighbors estimators for the _check_X_index function I guess ?

aeon/similarity_search/series/_base.py
baraline: removed that and made a fit_predict

baraline:
Based on your comment @MatthewMiddlehurst, I think another set of base class for neighbours / motifs estimator is necessary to not have weird thing in the series & collection base classes. 
I wanted to avoid creating too many bases classes, but it seems necessary to have "cleaner" ones.
I'll put it in place and then you can check back it that feels better

baraline:
Tried to make additional base classes, but this made the whole module structure feel too overcomplicated. Reverted and instead factored some functions away from the base class and made a fit_predict for stomp to avoid the weird X=None possibility in predict. Open to suggestions if you have some

aeon/similarity_search/series/_base.py
MatthewMiddlehurst: Yeah should raise an issue or you can just add it here. Not sure why its in one base class and not the other.

MatthewMiddlehurst:
Its all experimental. Better to just get it in when clear issues are resolved then iterate.

MatthewMiddlehurst:
It is also possible I am misunderstanding the goal a bit. If you want to do something that fits a collection and predicts on a series, we probably need a new base class for that. The collection and series bases are not really built for it. Not saying has to be done in this PR, but probably where we want to end up.

aeon/similarity_search/series/_base.py
baraline: Raised issue #2768

aeon/similarity_search/series/_base.py
```diff
@@ -53,9 +56,9 @@ def fit(
         """
         self.reset()
         X = self._preprocess_series(X, self.axis, True)
-        # Store minimum number of n_timepoints for unequal length collections
-        self.n_channels_ = X.shape[0]
-        self.n_timepoints_ = X.shape[1]
+        self.n_channels_ = self.metadata_["n_channels"]
+        timepoint_idx = 1 if self.axis == 1 else 0
+        self.n_timepoints_ = X.shape[timepoint_idx]
```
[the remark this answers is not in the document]
baraline: I'm can't seem to find a n_timepoints in the metadata stored by BaseSeriesEstimator, so I guess this also need an issue, raised #2769

baraline:
I think we need to explore this a bit further down the road when more estimators are added to see if it would really fit a use case. 
This could also be solved with a simple wrapper like `predict_series` calling `self.predict([X])`, but let's give it some thought first agree.

aeon/similarity_search/collection/_base.py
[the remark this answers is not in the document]
baraline: Added reference to this in #2768, but maybe this function should be integrated in the base class somehow and not left as a standalone. We can discuss that on the issue.

MatthewMiddlehurst (approved):
This is too big to reasonably review just looking at the code and text. Generally I'm fine with the changes, and the module is experimental so stuff can be modified and removed as required. No one has raised any major objections in the duration of this PR.

In the future I would like to see some generic testing added to make sure these adhere so some structure.

baraline:
Agreed, will resume work on this module during summer when things cool down, and testing will definitly be an important part of the work

PR #2517, opened 2025-01-25
[ENH] adjust test for non numpy output

TonyBagnall:
currently, the test function `check_fit_deterministic` and `check_persistence_via_pickle` assume that the output from any non state changing method (i.e. `predict`, `transform` or `predict_proba` is an array here
```
            assert_array_almost_equal(
                output,
                results[i],
                err_msg=f"Running {method} after fit twice with test "
                f"parameters gives different results.",
            )
```
we have and will have estimators that return dictionaries or tuples, so this seems artificial. This PR adds a method to test dictionaries and tuples too.  Return types are actually more complex

    Valid data structures are:
    1. float: returns a single value (e.g. forecasting)
    2. numpy array of floats: stores an equal length collection or series (default)
    3.  numpy array of booleans: (MERLIN)
    4. dict: a histogram of counts, usually of discrete series (e.g. SFA)
    5. pd.DataFrame: series stored in dataframe (e.g. Dobin)
    6. list: stores possibly unequal length series in a format 2-4
    7. tuple: stores two or more series/collections in a format 2-4 (e.g. imbalance
    transformers)

then find out we already have a function to test this, so switch to use deep_equals

Note that the Hydra transformer currently returns a pytorch Tensor. I have changed this to a numpy array, as I dont think we should mix and match arrays, but if anyone objects can just adapt the testing to change it


chrisholder (approved):
LGTM

PR #2526, opened 2025-02-03
[MNT] stop segmenters changing state in predict

TonyBagnall:
several segmenters failed the test that predict does not change the state. This PR fixes that. All have fit_is_empty = True. The choice is to not store data or to introduce a fit method. For these three

    "FLUSSSegmenter": ["check_non_state_changing_method"],
    "InformationGainSegmenter": ["check_non_state_changing_method"],
    "GreedyGaussianSegmenter": ["check_non_state_changing_method"],
just stopped storing data not used anywhere. 
edit: will leave clasp, used in notebooks


baraline (commented):
(Long time I looked at segmentation) For GG and IGTS changes should be OK, but for Fluss, if you change the `X` you give in predict it won't change the prediction, it feels weird. Couldn't you just remove the self storing of `found_cps` and the others to avoid the state change in predict ?

TonyBagnall:
> (Long time I looked at segmentation) For GG and IGTS changes should be OK, but for Fluss, if you change the `X` you give in predict it won't change the prediction, it feels weird. Couldn't you just remove the self storing of `found_cps` and the others to avoid the state change in predict ?

long time since I looked at this PR :) I just reverted FLUSS for now, will do it for the other two.

baraline (approved):
LGTM then !

PR #2528, opened 2025-02-05
[MNT] Fix random state deep clustering checking test

hadifawaz1999:
This is the right way of testing it, should close #1904


baraline (approved):
Testing logic LGTM !

Issue #2497, opened 2025-01-14
[ENH] Coverage for the networks module


TonyBagnall:
### Describe the feature or idea you want to propose

networks has a lot of parameters that are not covered in testing, it would be good if they were esp the ae ones @aadya940

### Describe your proposed solution

add tests for network parameters

Nikitas100:
Hello, I would like to contribute by adding tests for the network parameters. Could you please provide more details on which parameters specifically require coverage, or should I proceed with reviewing the networks module to identify them? Any guidance or specifications would be appreciated.

MatthewMiddlehurst:
While this does not cover every single case and item we want tested, an east place to start is our test coverage https://app.codecov.io/gh/aeon-toolkit/aeon

Nikitas100:
Hi @MatthewMiddlehurst ,

I noticed that the tests don't fully cover 10-11 files in the networks module. Would you prefer I start with specific files or parameters, or should I gradually work on covering all uncovered areas?

Thanks!

aadya940:
@Nikitas100 you can start with any file, typically divide the tasks to 1 network per pull request. :)

lucifer4073:
Hi  @Nikitas100,

There seems to be a decent number of files which need a higher test coverage.  It might be a good idea to collaborate on addressing this issue together.

lucifer4073:
Hi @Nikitas100 ,

If you are not currently working in this issue, I will be processing with the rest of the files.

MatthewMiddlehurst:
Good catch, I have closed that one as this seems more recently active. @hadifawaz1999 is it fine to work on the clustering ones now?

MatthewMiddlehurst:
Is this complete? Can still open PRs at least.

TonyBagnall:
its at 93% now, above package level of 92%, so closing this

PR #2537, opened 2025-02-12
[ENH] Test coverage for MLP Network improved

shinymack:
#### Reference Issues/PRs
Contributes to #2497 

Adds test coverage for _mlp.py

No

#### Any other comments?
Achieves 96% coverage on _mlp.py, checked using 
```pytest --cov=aeon.networks aeon/networks/ ```

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you after the PR has been merged.

### DIFF
--- a/aeon/networks/tests/test_mlp.py
+++ b/aeon/networks/tests/test_mlp.py
@@ -1,0 +1,179 @@
+"""Tests for the MLPNetwork Model."""
+
+import pytest
+
+from aeon.networks import MLPNetwork
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["tensorflow"], severity="none"),
+    reason="Tensorflow soft dependency unavailable.",
+)
+@pytest.mark.parametrize(
+    "n_layers, n_units, activation",
+    [
+        (3, 500, "relu"),
+        (5, [256, 128, 128, 64, 32], "sigmoid"),
+        (2, 128, ["tanh", "relu"]),
+    ],
+)
+def test_mlp_initialization(n_layers, n_units, activation):
+    """Test whether MLPNetwork initializes correctly with different configurations."""
+    from tensorflow.keras.layers import Dense, Dropout, Flatten, InputLayer
+    from tensorflow.keras.models import Model
+
+    mlp = MLPNetwork(n_layers=n_layers, n_units=n_units, activation=activation)
+    input_layer, output_layer = mlp.build_network((1000, 5))
+
+    # Wrap in a Model to access internal layers
+    model = Model(inputs=input_layer, outputs=output_layer)
+    layers = model.layers
+
+    assert isinstance(layers[0], InputLayer), "Expected first layer to be InputLayer"
+
+    assert isinstance(layers[1], Flatten), "Expected second layer to be Flatten"
+
+    # Check dropout and dense layers ordering
+    for i in range(n_layers):
+        dropout_layer = layers[2 + 2 * i]  # Dropout before Dense
+        dense_layer = layers[3 + 2 * i]  # Dense comes after Dropout
+
+        assert isinstance(
+            dropout_layer, Dropout
+        ), f"Expected Dropout at index {2 + 2 * i}"
+        assert isinstance(dense_layer, Dense), f"Expected Dense at index {3 + 2 * i}"
+
+        # Assert activation function
+        expected_activation = (
+            activation[i] if isinstance(activation, list) else activation
+        )
+        assert dense_layer.activation.__name__ == expected_activation, (
+            f"Expected activation {expected_activation}, "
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: aeon/networks/tests/test_mlp.py [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+        (5, [256, 128, 128, 64, 32], "sigmoid"),
+        (2, 128, ["tanh", "relu"]),
+    ],
+)
+def test_mlp_initialization(n_layers, n_units, activation):
+    """Test whether MLPNetwork initializes correctly with different configurations."""
+    mlp = MLPNetwork(n_layers=n_layers, n_units=n_units, activation=activation)
+    input_layer, output_layer = mlp.build_network((1000, 5))
+    assert input_layer is not None
+    assert output_layer is not None
```
baraline, line 26: Same comment as in another PR, the tests should explicitly test for the expected object type rather than none.
Additionally, testing that those objects have the expected number of layers/units/the correct activation function would be nice to really cover the estimator.

baraline (changes requested):
The first comment applies for all tests.

aeon/networks/tests/test_mlp.py
shinymack: Thanks for the feedback! I've updated the tests to explicitly check for the expected object type instead of just None. Additionally, I’ve added assertions to verify the number of layers, units, and activation functions where applicable to ensure better coverage of the estimator. Let me know if you have any further suggestions!

shinymack:
@baraline @hadifawaz1999, I've updated the tests to explicitly assert the expected layer types and verify the model parameters. Initially, isinstance(input_layer, InputLayer) wasn't working because input_layer is a KerasTensor, not a direct layer instance. To resolve this, I accessed _keras_history[0], which gives the actual layer. Let me know if this approach aligns with what you were expecting.

Also, I ensured TensorFlow is treated as a soft dependency by moving imports inside the test functions, so the CI should now pass without issues. Please review the changes and let me know if any further modifications are needed.

Thanks for your time!

baraline:
Thanks for the changes, I'm not very familiar with TensorFlow, so I'll let Ali confirm about the use of `_keras_history[0]`, but about the parameters, I was thinking more like what you did here `output_layer_actual.rate == dropout_last`.

The goal is to test that we've built a the network that matches the parameter that we define in `__init__` (i.e that we don't forget to give those parameters when building the network), hence testing directly from the estimator (e.g. `mlp.dropout_last == dropout_last`) is not what I had in mind, but more like what you did for `output_layer_actual.rate == dropout_last`.

hadifawaz1999:
> Thanks for the changes, I'm not very familiar with TensorFlow, so I'll let Ali confirm about the use of `_keras_history[0]`, but about the parameters, I was thinking more like what you did here `output_layer_actual.rate == dropout_last`.
> 
> The goal is to test that we've built a the network that matches the parameter that we define in `__init__` (i.e that we don't forget to give those parameters when building the network), hence testing directly from the estimator (e.g. `mlp.dropout_last == dropout_last`) is not what I had in mind, but more like what you did for `output_layer_actual.rate == dropout_last`.

I would prefer not using ```_keras_history``` as it is unknown if it will still exists in new versions to come, so i would instead do as @baraline is proposing and check if the internal parameters of the layers are correct, and if they are none or not, also their output shpaes

shinymack:
@baraline @hadifawaz1999 Thanks for the feedback! I’ve updated the tests to avoid using _keras_history and instead wrapped the returned input_layer and output_layer in a Model. This way, I can directly inspect model.layers to verify the structure and parameters, making the checks more reliable and future-proof. Let me know if this looks good or if there’s anything else you’d suggest!

hadifawaz1999 (changes requested):
Thanks for taking care of this, but could you explicity say "mlp" in the test function names ? this way we are sure of the test association

shinymack:
> Thanks for taking care of this, but could you explicity say "mlp" in the test function names ? this way we are sure of the test association

Updated the test function names to include 'mlp'. Let me know if anything else is needed!

baraline (approved):
LGTM, thanks ! 
Small note, next time, try to avoid using force pushing, it prevents us from seeing the changes made between each commit correctly when reviewing.

hadifawaz1999 (approved):
LGTM thanks

hadifawaz1999:
> LGTM, thanks ! Small note, next time, try to avoid using force pushing, it prevents us from seeing the changes made between each commit correctly when reviewing.

I agree

shinymack:
Got it! I'll keep that in mind and avoid force pushing from next time. Thanks for your time and feedback!

shinymack:
@all-contributors please add @shinymack for code

Issue #1797, opened 2024-07-14
[MNT] Add joblib backend option and set default to all parallelized estimators


baraline:
### Describe the issue

Some of the estimators that use joblib for parallelization use process-based backend, while other use threads-based backend. Ideally, we want this to be a parameter tunable by the users. To this end, we want to include a `joblib_backend` parameter which would default to `threading`, and use this parameter to set the joblib `backend` parameter during `Parallel` calls.

TonyBagnall:
I remember we looked at this in some detail when debugging some rocket features, was there some interaction with its use of prange? I cant remember!

```python

def _static_transform_uni(X, parameters, indices):
    """Transform a 2D collection of univariate time series.

    Implemented separately to the multivariate version for numba efficiency reasons.
    See issue #1778.
    """
    n_cases, n_timepoints = X.shape
    (
        _,
        _,
        dilations,
        n_features_per_dilation,
        biases,
    ) = parameters
    n_kernels = len(indices)
    n_dilations = len(dilations)
    f = n_kernels * np.sum(n_features_per_dilation)
    features = np.zeros((n_cases, f), dtype=np.float32)
    for i in prange(n_cases):
```

CodeLionX:
I think adding this parameter to so many APIs would clutter the parameter lists. What are your thoughts on using [`joblib.parallel_config`](https://joblib.readthedocs.io/en/latest/generated/joblib.parallel_config.html#joblib.parallel_config)?
This means that all aeon code does not specify any parallel backend in their calls to `joblib.Parallel` (if it does not require a certain backend) and the user can choose its backend by wrapping calls within:

```python
with joblib.parallel_config(backend="loky"):
   # aeon code
```

The same is actually possible for `n_jobs` etc.

baraline:
True, forgot about this option ! That or we set a global variable such as `AEON_JOBLIB_BACKEND` and document its usage

CodeLionX:
A custom env var `AEON_JOBLIB_BACKEND` IMO just makes sense if it is also used for other parallel stuff besides joblib. Otherwise, we can use the existing (documented and potentially already known) joblib-facilities.

MatthewMiddlehurst:
Sounds like a good idea, would have to be documented though. Not sure about removing `n_jobs`.

baraline:
[...]

PR #2545, opened 2025-02-17
[ENH] Distance module n_jobs support

chrisholder:
#### Reference Issues/PRs
#1886 #1797

This PR adds parallelism to the distance module. Specifically to pairwise distance computations.

To do this the n_job parameter can now be specified to any pairwise distance (aside mpdist which was excluded due to complexity) and it will use that many threads.

This PR just updates the distance module. In a follow up PR I will update all our models that use pairwise_distance and n_jobs to make use of this. 

##### For all contributions

### DIFF
--- a/aeon/distances/_distance.py
+++ b/aeon/distances/_distance.py
@@ -1,5 +1,6 @@
 __maintainer__ = []
 
+import warnings
 from enum import Enum
 from typing import Any, Callable, Optional, TypedDict, Union
 
@@ -173,6 +174,7 @@ def pairwise_distance(
     y: Optional[np.ndarray] = None,
     method: Union[str, DistanceFunction, None] = None,
     symmetric: bool = True,
+    n_jobs: int = 1,
     **kwargs: Unpack[DistanceKwargs],
 ) -> np.ndarray:
     """Compute the pairwise distance matrix between two time series.
@@ -197,6 +199,10 @@ def pairwise_distance(
         function is provided as the "method" parameter, then it will compute an
         asymmetric distance matrix, and the entire matrix (including both upper and
         lower triangles) is returned.
+    n_jobs : int, default=1
+        The number of jobs to run in parallel. If -1, then the number of jobs is set
+        to the number of CPU cores. If 1, then the function is executed in a single
+        thread. If greater than 1, then the function is executed in parallel.
     kwargs : Any
         Extra arguments for distance. Refer to each distance documentation for a list of
         possible arguments.
@@ -240,11 +246,13 @@ def pairwise_distance(
            [ 48.]])
     """
     if method in PAIRWISE_DISTANCE:
-        return DISTANCES_DICT[method]["pairwise_distance"](x, y, **kwargs)
+        return DISTANCES_DICT[method]["pairwise_distance"](
+            x, y, n_jobs=n_jobs, **kwargs
+        )
     elif isinstance(method, Callable):
         if y is None and not symmetric:
-            return _custom_func_pairwise(x, x, method, **kwargs)
-        return _custom_func_pairwise(x, y, method, **kwargs)
+            return _custom_func_pairwise(x, x, method, n_jobs=n_jobs, **kwargs)
+        return _custom_func_pairwise(x, y, method, n_jobs=n_jobs, **kwargs)
     else:
         raise ValueError("Method must be one of the supported strings or a callable")
 
@@ -253,10 +261,21 @@ def _custom_func_pairwise(
     X: Optional[Union[np.ndarray, list[np.ndarray]]],
     y: Optional[Union[np.ndarray, list[np.ndarray]]] = None,
     dist_func: Union[DistanceFunction, None] = None,
+    n_jobs: int = 1,
     **kwargs: Unpack[DistanceKwargs],
 ) -> np.ndarray:
     if dist_func is None:
         raise ValueError("dist_func must be a callable")
+
+    if n_jobs != 1:
+        warnings.warn(
+            "You are using a custom distance function with n_jobs > 1. "
+            "Aeon does not support parallelization for custom distance "
+            "functions. If it is an existing aeon distance try using the "
+            "string name instead.",
+            UserWarning,
+            stacklevel=2,
+        )
 
     multivariate_conversion = _is_numpy_list_multivariate(X, y)
     X, _ = _convert_collection_to_numba_list(X, "X", multivariate_conversion)
--- a/aeon/distances/_mpdist.py
+++ b/aeon/distances/_mpdist.py
@@ -1,5 +1,6 @@
 """Matrix Profile Distances."""
 
+import warnings
 from typing import Optional, Union
 
 import numpy as np
@@ -287,6 +288,7 @@ def mp_pairwise_distance(
     X: Union[np.ndarray, list[np.ndarray]],
     y: Optional[Union[np.ndarray, list[np.ndarray]]] = None,
     m: int = 0,
+    **kwargs,
 ) -> np.ndarray:
     """Compute the mpdist pairwise distance between a set of time series.
 
@@ -339,13 +341,19 @@ def mp_pairwise_distance(
            [2.82842712],
            [2.82842712]])
     """
+    if "n_jobs" in kwargs:
+        warnings.warn(
+            "n_jobs is not supported for the mpdist distance method and will be "
+            "ignored.",
+            UserWarning,
+            stacklevel=2,
+        )
+    if m == 0:
+        m = int(X.shape[2] / 4)
     multivariate_conversion = _is_numpy_list_multivariate(X, y)
     _X, unequal_length = _convert_collection_to_numba_list(
         X, "X", multivariate_conversion
     )
-
-    if m == 0:
-        m = int(_X.shape[2] / 4)
 
     if y is None:
         return _mpdist_pairwise_distance_single(_X, m)
--- a/aeon/distances/_sbd.py
+++ b/aeon/distances/_sbd.py
@@ -2,14 +2,16 @@
 
 __maintainer__ = ["SebastianSchmidl"]
 
+import warnings
 from typing import Optional, Union
 
 import numpy as np
-from numba import njit, objmode
+from numba import njit, objmode, prange, set_num_threads
 from numba.typed import List as NumbaList
 from scipy.signal import correlate
 
 from aeon.utils.conversion._convert_collection import _convert_collection_to_numba_list
+from aeon.utils.validation import check_n_jobs
 from aeon.utils.validation.collection import _is_numpy_list_multivariate
 
 
@@ -117,6 +119,8 @@ def sbd_pairwise_distance(
     X: Union[np.ndarray, list[np.ndarray]],
     y: Optional[Union[np.ndarray, list[np.ndarray]]] = None,
     standardize: bool = True,
+    n_jobs: int = 1,
+    **kwargs,
 ) -> np.ndarray:
     """
     Compute the shape-based distance (SBD) between all pairs of time series.
@@ -138,6 +142,10 @@ def sbd_pairwise_distance(
     standardize : bool, default=True
         Apply z-score to both input time series for standardization before
         computing the distance. This makes SBD scaling invariant. Default is True.
+    n_jobs : int, default=1
+        The number of jobs to run in parallel. If -1, then the number of jobs is set
+        to the number of CPU cores. If 1, then the function is executed in a single
+        thread. If greater than 1, then the function is executed in parallel.
 
     Returns
     -------
@@ -188,6 +196,17 @@ def sbd_pairwise_distance(
            [0.36754447, 0.        , 0.29289322],
            [0.5527864 , 0.29289322, 0.        ]])
     """
+    n_jobs = check_n_jobs(n_jobs)
+    set_num_threads(n_jobs)
+    if n_jobs > 1:
+        warnings.warn(
+            "You have set n_jobs > 1. For this distance function "
+            "unless your data is very large (> 10000 time series), it is "
+            "recommended to use n_jobs=1. If this function is slower than "
+            "expected try setting n_jobs=1.",
+            UserWarning,
+            stacklevel=2,
+        )
     multivariate_conversion = _is_numpy_list_multivariate(X, y)
     _X, _ = _convert_collection_to_numba_list(X, "", multivariate_conversion)
 
@@ -199,14 +218,14 @@ def sbd_pairwise_distance(
     return _sbd_pairwise_distance(_X, _y, standardize)
 
 
-@njit(cache=True, fastmath=True)
+@njit(cache=True, fastmath=True, parallel=True)
 def _sbd_pairwise_distance_single(
     x: NumbaList[np.ndarray], standardize: bool
 ) -> np.ndarray:
     n_cases = len(x)
     distances = np.zeros((n_cases, n_cases))
 
-    for i in range(n_cases):
+    for i in prange(n_cases):
         for j in range(i + 1, n_cases):
             distances[i, j] = sbd_distance(x[i], x[j], standardize)
             distances[j, i] = distances[i, j]
@@ -214,7 +233,7 @@ def _sbd_pairwise_distance_single(
     return distances
 
 
-@njit(cache=True, fastmath=True)
+@njit(cache=True, fastmath=True, parallel=True)
 def _sbd_pairwise_distance(
     x: NumbaList[np.ndarray], y: NumbaList[np.ndarray], standardize: bool
 ) -> np.ndarray:
@@ -222,7 +241,7 @@ def _sbd_pairwise_distance(
     m_cases = len(y)
     distances = np.zeros((n_cases, m_cases))
 
-    for i in range(n_cases):
+    for i in prange(n_cases):
         for j in range(m_cases):
             distances[i, j] = sbd_distance(x[i], y[j], standardize)
     return distances
--- a/aeon/distances/pointwise/_euclidean.py
+++ b/aeon/distances/pointwise/_euclidean.py
@@ -1,9 +1,10 @@
 __maintainer__ = []
 
+import warnings
 from typing import Optional, Union
 
 import numpy as np
-from numba import njit
+from numba import njit, prange, set_num_threads
 from numba.typed import List as NumbaList
 
 from aeon.distances.pointwise._squared import (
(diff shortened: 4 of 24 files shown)


aeon/distances/_distance.py
```diff
@@ -253,11 +261,22 @@ def _custom_func_pairwise(
...
 ) -> np.ndarray:
     if dist_func is None:
         raise ValueError("dist_func must be a callable")
 
+    if n_jobs != 1:
+        warnings.warn(
+            "You are using a custom distance function with n_jobs > 1. "
+            "Aeon does not support parallelization for custom distance "
+            "functions. If it is an existing aeon distance try using the "
+            "string name instead.",
```
baraline, line 275: I guess this is for simplicity in this PR, but this could be added with joblib in `_custom_pairwise_distance` right ?

baraline (commented):
LGTM! The only thing that would be missing from the PR to fix the issue is to include `n_jobs` in the distance params in our `KNeighborsTimeSeriesClassifier` and we would be good to go

aeon/distances/_distance.py
chrisholder: I was going to follow up adding n_jobs to classifiers and clusterers in a follow up PR. Some of the classifiers don't use pairwise distance functions atm so may require more than just a one line change and I didn't want to overcomplicate this PR.

Regarding custom I did try briefly with joblib but was getting some weird results (as in timing results) and wasn't sure why. Something we can revist.

baraline (approved):

aeon/distances/_distance.py
```diff
@@ -253,11 +261,22 @@ def _custom_func_pairwise(
...
+    n_jobs: int = 1,
     **kwargs: Unpack[DistanceKwargs],
 ) -> np.ndarray:
     if dist_func is None:
         raise ValueError("dist_func must be a callable")
 
+    if n_jobs != 1:
+        warnings.warn(
+            "You are using a custom distance function with n_jobs > 1. "
+            "Aeon does not support parallelization for custom distance "
```
MatthewMiddlehurst, line 273: No capital in aeon

aeon/distances/_sbd.py
```diff
@@ -188,6 +196,17 @@ def sbd_pairwise_distance(
            [0.36754447, 0.        , 0.29289322],
            [0.5527864 , 0.29289322, 0.        ]])
     """
+    n_jobs = check_n_jobs(n_jobs)
+    set_num_threads(n_jobs)
```
MatthewMiddlehurst, line 200: I feel like this should set it back to the original value after?

Maybe its possible to use a decorator in these cases similar to https://github.com/aeon-toolkit/aeon/blob/v1.0.0/aeon/testing/utils/output_suppression.py

MatthewMiddlehurst (commented):
Would be nice to benchmark and see if this actually speeds it up.

baraline:
Can confirm that this does indeed provide some speed ups : 
```python
import numpy as np
from aeon.distances import euclidean_pairwise_distance

X = np.random.rand(1000, 1000)
Y = np.random.rand(1000, 1000)
euclidean_pairwise_distance(X, Y)

%timeit euclidean_pairwise_distance(X, Y)
%timeit euclidean_pairwise_distance(X, Y, n_jobs=-1)
```
```
199 ms ± 7.05 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
82.7 ms ± 1.19 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Seconds Matthew's comment to revert the number of threads back to previously setted value (using get_num_threads).

chrisholder:
With distance measures like Euclidean I found unless you have a huge number of very long time series threading is almost never going to be faster just due to the cost of spawning a thread. This is why I added the warning as I can't really see a scenario where running pairwise euclidean distance thread will ever achieve a meaningful speed up.

However, for something like DTW I get huge speed ups:

```
    X = make_example_3d_numpy(10000, 1, 100, return_y=False, random_state=42)
    start = time.time()
    pairwise_distance(X, method="dtw", n_jobs=1)
    print(f"1 Job Time: {time.time() - start}")

    start = time.time()
    pairwise_distance(X, method="dtw", n_jobs=14)
    print(f"14 Job Time: {time.time() - start}")
```
```
Time: 1620.196809053421
Time: 255.28647780418396
```

chrisholder:
I followed your suggestion @MatthewMiddlehurst and defined a decorator to handle setting and resetting threading.

aeon/utils/_threading.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        NUMBA_ENV_THREADS = os.environ.get("NUMBA_NUM_THREADS")
+
+        if NUMBA_ENV_THREADS is not None and NUMBA_ENV_THREADS.isdigit():
+            original_thread_count = int(NUMBA_ENV_THREADS)
+        else:
+            original_thread_count = num_threads_default()
```
chrisholder: I've had to change this from get_num_threads() as suggested above because apparently the number of threads Python has available to it and the number of thread Numba has available are not the same. What this meant is numba was throwing errors in the CI when I used get_num_threads as the default. Therefore, I took the "num_threads_default" from the numba source code which is used to determine the number of threads to use if one isnt provided inside numba.

I would have imported it from numba however they defined it as a nested function. This means I can't access it via a import statement and therefore I have copied it over.

aeon/distances/_sbd.py
chrisholder: I have created a decorator called "threaded" located in aeon/utils/_threading.py.

```
def threaded(func: Callable) -> Callable:
    @functools.wraps(func)
    def wrapper(*args: Any, **kwargs: Any) -> Any:
        NUMBA_ENV_THREADS = os.environ.get("NUMBA_NUM_THREADS")

        if NUMBA_ENV_THREADS is not None and NUMBA_ENV_THREADS.isdigit():
            original_thread_count = int(NUMBA_ENV_THREADS)
        else:
            original_thread_count = num_threads_default()

        n_jobs = None
        if "n_jobs" in kwargs:
            n_jobs = kwargs["n_jobs"]
        else:
            sig = inspect.signature(func)
            param_names = list(sig.parameters.keys())

            if "n_jobs" in param_names:
                n_jobs_index = param_names.index("n_jobs")
                if n_jobs_index < len(args):
                    n_jobs = args[n_jobs_index]
                else:
                    default = sig.parameters["n_jobs"].default
                    n_jobs = default if default is not inspect.Parameter.empty else None

            if n_jobs is None and args and hasattr(args[0], "n_jobs"):
                n_jobs = args[0].n_jobs

        adjusted_n_jobs = check_n_jobs(n_jobs)
        set_num_threads(adjusted_n_jobs)

        try:
            result = func(*args, **kwargs)
            return result
        finally:
            set_num_threads(original_thread_count)

    return wrapper
```

This function gets the users default number of numba threads (using the exact same logic numba does to determine it), stores it and then sets the specified n_jobs as the available thread for numba. The function specified is then called and after it has run the threads are restored to the default. An example usage is:

```
@threaded
def dtw_pairwise_distance(
    X: Union[np.ndarray, list[np.ndarray]],
    y: Optional[Union[np.ndarray, list[np.ndarray]]] = None,
    window: Optional[float] = None,
    itakura_max_slope: Optional[float] = None,
    n_jobs: int = 1,
    **kwargs,
) -> np.ndarray:
    multivariate_conversion = _is_numpy_list_multivariate(X, y)
    _X, unequal_length = _convert_collection_to_numba_list(
        X, "X", multivariate_conversion
    )

    if y is None:
        # To self
        return _dtw_pairwise_distance(_X, window, itakura_max_slope, unequal_length)
    _y, unequal_length = _convert_collection_to_numba_list(
        y, "y", multivariate_conversion
    )
    return _dtw_from_multiple_to_multiple_distance(
        _X, _y, window, itakura_max_slope, unequal_length
    )
```

TonyBagnall (commented):
LGTM but will wait to see if @baraline or @MatthewMiddlehurst want to comment

aeon/distances/_distance.py
```diff
@@ -244,45 +251,71 @@ def pairwise_distance(
...
-            distances[j, i] = distances[i, j]
+            distances[j, i] = distances[i, j]  # Mirror for symmetry
+    else:
+        results = Parallel(n_jobs=n_jobs)(
+            delayed(compute_single_distance)(i, j) for i, j in indices
+        )
+
+        for i, j, dist in results:
+            distances[i, j] = dist
+            distances[j, i] = dist  # Mirror for symmetry
```
MatthewMiddlehurst: Is splitting this necessary? If the overhead from just using Parallel always is negligible, it would be a lot cleaner. Same with the other function.s

MatthewMiddlehurst: Just to confirm this uses Joblib and the rest uses numba? Any reason we can't check for n_jobs in the input and use that?

aeon/distances/_mpdist.py
MatthewMiddlehurst, line 1: Dont care enough about this code but these seem like large changes. local import seems a bit unnecessary.

aeon/distances/_sbd.py
```diff
@@ -188,6 +197,15 @@ def sbd_pairwise_distance(
...
     """
+    if n_jobs > 1:
+        warnings.warn(
+            "You have set n_jobs > 1. For this distance function "
+            "unless your data has a large number of time points, it is "
+            "recommended to use n_jobs=1. If this function is slower than "
+            "expected try setting n_jobs=1.",
+            UserWarning,
+            stacklevel=2,
+        )
```
MatthewMiddlehurst: Is it specifically bad for this distance? this may get annoying quick if you call it multiple times. May be better to only print once per length and/or if < a certain length.

aeon/distances/pointwise/_euclidean.py
```diff
@@ -69,9 +71,12 @@ def _univariate_euclidean_distance(x: np.ndarray, y: np.ndarray) -> float:
     return np.sqrt(_univariate_squared_distance(x, y))
 
 
+@threaded
 def euclidean_pairwise_distance(
     X: Union[np.ndarray, list[np.ndarray]],
     y: Optional[Union[np.ndarray, list[np.ndarray]]] = None,
+    n_jobs: int = 1,
+    **kwargs,
```
MatthewMiddlehurst: Why do they all have kwargs now? is this used anywhere?

aeon/distances/pointwise/_manhattan.py
```diff
@@ -133,6 +142,16 @@ def manhattan_pairwise_distance(
...
     """
+    if n_jobs > 1:
+        warnings.warn(
+            "You have set n_jobs > 1. For this distance function "
+            "unless your data has a large number of time points, it is "
+            "recommended to use n_jobs=1. If this function is slower than "
+            "expected try setting n_jobs=1.",
+            UserWarning,
+            stacklevel=2,
+        )
```
MatthewMiddlehurst: Ah nvm there are other ones. Same point as before though. "this distance" is not very specific.

aeon/utils/_threading.py
MatthewMiddlehurst: I'm a bit confused, I thought get_num_threads was a numba function? You would think it also reads and resolves the env variable as well.

aeon/utils/_threading.py
```diff
@@ -0,0 +1,71 @@
...
+            if "n_jobs" in param_names:
+                n_jobs_index = param_names.index("n_jobs")
+                if n_jobs_index < len(args):
+                    n_jobs = args[n_jobs_index]
+                else:
+                    default = sig.parameters["n_jobs"].default
+                    n_jobs = default if default is not inspect.Parameter.empty else None
+
+            if n_jobs is None and args and hasattr(args[0], "n_jobs"):
+                n_jobs = args[0].n_jobs
```
MatthewMiddlehurst: Could you explain a bit more whats going on here and maybe leave a comment. Is this just to handle both kwargs and args?

aeon/utils/_threading.py
```diff
@@ -0,0 +1,71 @@
...
+
+    cpu_count = os.cpu_count()
+    if cpu_count is not None:
+        return max(1, cpu_count)
+
+    return 1
+
+
+def threaded(func: Callable) -> Callable:
+    """Set thread count based on n_jobs parameter and restore it afterward.
```
MatthewMiddlehurst: This seems specific to numba and should probably be named to reflect that and go in the numba utilities.

aeon/utils/_threading.py
```diff
@@ -0,0 +1,71 @@
...
+import os
+from typing import Any, Callable
+
+from numba import set_num_threads
+
+from aeon.utils.validation import check_n_jobs
+
+
+def num_threads_default():
+    try:
```
MatthewMiddlehurst: should maybe be private. Add a docstring explaining where it is from if we keep it.

MatthewMiddlehurst (changes requested):

aeon/distances/_mpdist.py
TonyBagnall: MPdist is going to be deprecated soon

aeon/utils/_threading.py
chrisholder: Yea no idea why. I just tried it again to make sure I had properly tried it for some reason it doesnt work. Here is a CI run showing it fails to restore the correct number of threads when using get_num_threads() from numba: https://github.com/aeon-toolkit/aeon/actions/runs/16756664926/job/47441015560?pr=2545

aeon/distances/_distance.py
chrisholder: I ran some basic benchmarking to check and while the performance difference is measurable (about 0.5 seconds for time series datasets of shape (1000, 10, 1000)) it isnt significant enough that anyone would actually notice. Therefore I have made the suggested change above.

aeon/distances/_sbd.py
chrisholder: I also spoke with Tony about this and so I have added the warning to the docstrings of distances that have this issue and removed the warnings in code all together.

aeon/utils/_threading.py
chrisholder: Yes this section inspects the function passed and finds out if n_jobs is presents in args, kwargs or if it's a object with the attribute n_jobs. I have inline commented the whole function now here is this specific section:

```python3
        # 2) Extract the requested n_jobs value
        n_jobs = None

        # A) If passed explicitly as a kwarg, use that
        if "n_jobs" in kwargs:
            n_jobs = kwargs["n_jobs"]
        else:
            # B) Otherwise, inspect the signature to see which positional argument
            #    corresponds to n_jobs, if any.
            sig = inspect.signature(func)
            params = list(sig.parameters.keys())
            if "n_jobs" in params:
                idx = params.index("n_jobs")
                if idx < len(args):
                    # If n_jobs is passed as a positional argument, use that value
                    n_jobs = args[idx]
                else:
                    # Use the function’s declared default if one exists
                    default = sig.parameters["n_jobs"].default
                    if default is not inspect.Parameter.empty:
                        n_jobs = default
            # C) Otherwise, check to see if n_jobs is a method on an object with a
            # .n_jobs attr.
            elif n_jobs is None and args and hasattr(args[0], "n_jobs"):
                n_jobs = getattr(args[0], "n_jobs")
```

aeon/utils/_threading.py
chrisholder: Agreed this is now a numba only function that sets n_jobs for numba functions only. I have moved the function to aeon/utils/numba/_threading.py and updated the docstring to reflect this.

aeon/utils/_threading.py
chrisholder: Made this private and add a small docstring explaining what it is and where it is from

chrisholder:
I've had to disable the CanonicalIntervalForestRegressor "check_regressor_against_expected_results" test. I believe this has started to fail due to instability with numba in the module. I'm not really sure at all how this PR is impacting this test my best guess is switching the number of threads is causing unstable behaviour in tests that occur after certain distance tests run. 

I have also noticed that two other regressors RDST and RIST Regressor also have excluded tests in the aeon/testing/testing_config.py EXCLUDED_TESTS so I would assume this is a similar issue. I have had this problem both with numba caches on and off. I think this is a deeper issue with the regression module itself.

I could change the expected results as it seems to consistently now resolve to 0.14 instead of 0.16 (fourth result in the expected array) however, I don't feel comfortable changing this not knowing if this is a bug or not.

I don't think I can fix this myself so instead after this PR is merged I'll open an issue documenting the bug so we can resolve it in the future. Here is a CI run: https://github.com/aeon-toolkit/aeon/actions/runs/16759105314/job/47458932638?pr=2545

```
FAILED aeon/testing/tests/test_all_estimators.py::test_all_estimators[check_regressor_against_expected_results(estimator_class=CanonicalIntervalForestRegressor,data_name=CardanoSentiment,data_loader=load_cardano_sentiment)] - AssertionError: 
Arrays are not almost equal to 2 decimals
Failed to reproduce results for CanonicalIntervalForestRegressor on CardanoSentiment
Mismatched elements: 1 / 10 (10%)
Max absolute difference among violations: 0.02247
Max relative difference among violations: 0.1605
 ACTUAL: array([0.28, 0.15, 0.33, 0.16, 0.19, 0.22, 0.15, 0.22, 0.13, 0.2 ])
 DESIRED: array([0.28, 0.15, 0.33, 0.14, 0.19, 0.22, 0.15, 0.23, 0.12, 0.2 ])
= 1 failed, 5841 passed, 64 skipped, 25 xfailed, 12 xpassed, 44707 warnings in 902.86s (0:15:02) =
```

aeon/distances/pointwise/_euclidean.py
chrisholder: Good spot this is unnecessary I have removed this.

aeon/distances/_distance.py
chrisholder: I don't understand the second part of this question. The reason for the split is because the numba functions run much faster using numba prange rather than Joblib. Therefore, only when a custom distance function is passed, or the distance isn't numba ocompiled, the joblib is used.

TonyBagnall: do they really run faster with prange? Didnt know that

MatthewMiddlehurst: I meant can we check the custom function for an n_jobs but i suppose we dont really care about it enough to optimise that much.

MatthewMiddlehurst (commented):
in testing config we have a variable called MULTITHREAD_TESTING, can you use that for tests which play with threads, It runs it on a workflow without pytest-xdist (the tool we use to thread testing)

chrisholder:
Ill make that change now

Issue #809, opened 2023-10-09
[DOC] Inconsistent double tick quotes in docstrings


zy18811:
### Describe the issue linked to the documentation

Single ticks render as italics while double ticks render as code in the api docs, e.g. _int_ v.s. ``int``. Various methods use this inconsistently, example from the [HIVECOTE 2.0 docs](https://www.aeon-toolkit.org/en/latest/api_reference/auto_generated/aeon.classification.hybrid.HIVECOTEV2.html#aeon.classification.hybrid.HIVECOTEV2) below:

Docstring for ``random_state`` uses single ticks (italics) for all references to code, i.e. _int_, _RandomState_, and _None_.

![image](https://github.com/aeon-toolkit/aeon/assets/74973887/9a512b1f-aa48-4360-9dcf-236b24a2ecd8)

Docstring  for ``n_jobs`` uses double ticks for ``-1`` to represent it referring to code.

![image](https://github.com/aeon-toolkit/aeon/assets/74973887/4e14a3a4-d0e6-4c00-a3b8-da195583f011)

### Suggest a potential alternative/fix

Select preferred approach out of single and double ticks for referring to code in docstrings and apply consistently throughout __aeon__.

TonyBagnall:
I think double ticks is preferable

adityagh006:
Hi @TonyBagnall, I'd like to work on this issue. Could you please assign it to me?

MatthewMiddlehurst:
Please read our contributor guide. It would be a bit much to do this for the whole package, why not select a module then post here that you will be working on it?

Val-2608:
Hi ,I'd like to work on resolving any inconsistencies in the documentation for the regression module.

AnaghDeshpande:
Hi, @MatthewMiddlehurst i would like to work on anomaly_detection

NasheetaFarzana:
Hello @MatthewMiddlehurst @TonyBagnall.   I would like to work on this project.  But as a beginner I need your guidance for how to start  contributing  ( mainly improving the documentation and docstrings ) in the code.

MatthewMiddlehurst:
Hi @NasheetaFarzana that depends on how experienced you are really. I would look at our contributing guide on the website. As i mentioned on Slack for stuff such as how to use Git, GitHub, or learning Python we are not the place unfortunetly.

nMaax:
[...]

PR #2574, opened 2025-02-26
[DOC]resolved inconsistencies of double ticks in the regression module #809

Val-2608:
#### Reference Issues/PRs
#809 

This PR standardizes the usage of double backticks for referring to code elements in docstrings across the aeon library. Previously, there was inconsistency, with some docstrings using single ticks, which rendered as italics instead of code formatting in the API documentation.

No new dependencies introduced.

##### For all contributions


MatthewMiddlehurst:
Hi, there is way too much going on here. Please keep to the original issue, and keep other docs improvements in a separate PR if you believe they should be made.

 For now just stick to singular ` instead of adding new ones, I disagree with a lot of your additions. You only have to do this for public docstrings, this is not required for anything private not on the webpage (starting with _).

MatthewMiddlehurst:
This is also causing the tests to fail, please take a look at the output and resolve after editing.

Issue #2593, opened 2025-03-06
[ENH] Better activation function for CNN multiclass classification


KowshikB03:
In classification/deep_learning/cnn.py, by default the activation function used is _sigmoid_ which is good choice for binary classification. However, _sigmoid_ function is not the best activation function for multi-class classification. I experimented with 2 aeon multiclass datasets - osuleaf and basic_motions. CNN with _sigmoid_ activation produced descent accuracy and loss, however when CNN with _softmax_ activation clearly outperforms with higher accuracy and lower loss. Below is the comparison output,

**Osuleaf-Softmax**

> Epoch 1/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 2s 5ms/step - accuracy: 0.2781 - loss: 0.1350
> Epoch 2/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 4ms/step - accuracy: 0.4493 - loss: 0.1154
> Epoch 3/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 4ms/step - accuracy: 0.4875 - loss: 0.1097
> Epoch 4/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 4ms/step - accuracy: 0.5134 - loss: 0.1065
> Epoch 5/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 5ms/step - accuracy: 0.5273 - loss: 0.1044
> Epoch 6/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 6ms/step - accuracy: 0.5437 - loss: 0.1029
> Epoch 7/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 7ms/step - accuracy: 0.5541 - loss: 0.1018
> Epoch 8/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 4ms/step - accuracy: 0.5662 - loss: 0.1009
> Epoch 9/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 1s 4ms/step - accuracy: 0.5579 - loss: 0.1000
> Epoch 10/50
> 111/111 ━━━━━━━━━━━━━━━━━━━━ 0s 4ms/step - accuracy: 0.5583 - loss: 0.0992
> Epoch 11/50
[...]

PR #2594, opened 2025-03-06
[ENH] Softmax activation function added

KowshikB03:
Fixes Issue #2593

### DIFF
--- a/aeon/classification/deep_learning/_cnn.py
+++ b/aeon/classification/deep_learning/_cnn.py
@@ -125,7 +125,7 @@ class TimeCNNClassifier(BaseDeepClassifier):
         kernel_size=7,
         n_filters=None,
         avg_pool_size=3,
-        activation="sigmoid",
+        activation=None,
         padding="valid",
         strides=1,
         strides_pooling=None,
@@ -187,7 +187,7 @@ class TimeCNNClassifier(BaseDeepClassifier):
             activation=self.activation,
             padding=self.padding,
             strides=self.strides,
-            strides_pooling=self.strides_pooling,
+            # strides_pooling=self.strides_pooling,
             dilation_rate=self.dilation_rate,
             use_bias=self.use_bias,
         )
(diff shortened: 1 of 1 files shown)


hadifawaz1999:
Thank you for this but i left a comment on the corresponding issue on how this is not a bug

MatthewMiddlehurst:
Hi, I agree with Ali here if these are the published defaults then I don't really want to change them unless there is a publication showing it is better.

@hadifawaz1999 a warning or documentation change should be fine though if you think this would produce better results in certain scenarios? If not or it's unknown, keep it as it is. Up to you really.

hadifawaz1999:
not sure how worth it it would be, its not really the best model at the end of the day, with or without softmax, so would just keep it as is and not add confusion to users

MatthewMiddlehurst:
ok will leave it up for a bit for others to comment but likely declined

KowshikB03:
In the comparison of multi-class datasets using the OsuLeaf dataset, the model with Softmax activation shows steady improvement in accuracy. Although it didn’t achieve the highest accuracy within 50 epochs, it may improve further with 100 epochs. On the other hand, the model with Sigmoid activation stabilizes around 0.67–0.68 after 40 epochs, with no consistent improvement in accuracy or loss reduction. This was the same pattern in basic_motions dataset training. The main point is that a simple CNN doesn't always give better accuracy without fine-tuning. Starting with better default functions makes fine-tuning easier, helping us reach 95+ accuracy instead of just 85+ for multi-class single-label datasets.

Adding articles from stanford, wandbai and few more that explain theoretical reasons why softmax is better form multi-class single-label datasets.

1. https://web.stanford.edu/~nanbhas/blog/sigmoid-softmax/
2. https://wandb.ai/ayush-thakur/dl-question-bank/reports/Comparing-Sigmoid-MSE-With-Softmax-Cross-Entropy-for-Image-Classification--VmlldzoyMDA3ODQ
3. https://glassboxmedicine.com/2019/05/26/classification-sigmoid-vs-softmax/

Maybe we can train for further more epochs like 200 for more convincing results. I am open for your suggestions.

hadifawaz1999:
Just to be clear, every model in aeon uses softmax activation in the classification/deep_learning module, as we do in our research field. TimeCNN is an exception because this is how it was proposed in the original paper. It is not the best model as models such as FCN ResNet InceptionTime and LITETime are much more performant, so even configuring TimeCNN in a different way won't outperform the state-of-the-art. Moreover, this is not really the interest of aeon as the goal is to implement existing models that are proposed in the literature in a most optimized and organized way

What could be a possible additional information is a sentence in the TimeCNNClassifier documentation that mentions explicitly that this model uses sigmoid activation and mse loss function. Closing this PR and it's related issue for now, feel free to PR for the additional sentence in the docstring of the classifier

KowshikB03:
Thank you for the clarification. 

I understand the Aeon interests, and appreciate your insights. I will create a new PR to update the TimeCNNClassifier documentation accordingly. We can close this PR

PR #2613, opened 2025-03-11
[ENH] Adds support for distances that are asymmetric but supports unequal length

tanishy7777:
Distances like DTW with global invariance are asymmetric but allow unequal length time series.
For such cases the `_validate_distance_result` function in `aeon/distances/tests/test_distances.py` fails.
This PR fixes this but modifying the self calling behavior of the function by making sure that
it calls itself with swapped x and y only when the function is symmetric.

##### For all contributions


chrisholder (approved):
Good spot on this. LGTM!

PR #2630, opened 2025-03-16
[MNT] Remove REDCOMETs from testing exclusion list

MatthewMiddlehurst:
Fixes a bug introduced in the last REDCOMETs PR. Between its creation and merge there was an update to SFA fast which broke everything.


baraline (approved):
Lgtm !

PR #2653, opened 2025-03-19
[MNT] Add codespell support (config, workflow to detect/not fix) and make it fix a "few" typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

This is just a draft: due to large number of typos detected, so before tuning up config and skips, I decided first to check with you either you would be interested/willing to eventually review/accept such a PR.  But if you spot false positive -- alert or introduce into pyproject.toml config for codespell and let's work together to make aeon "typos free"!

TODOs
- ...
- [ ] run round of interactive fixes for ambiguous typos --  I didn't do yet


aeon/benchmarking/results_loaders.py
```diff
@@ -21,7 +21,7 @@
 VALID_RESULT_MEASURES = {
     "classification": ["accuracy", "auroc", "balacc", "f1", "logloss"],
     "clustering": ["clacc", "ami", "ari", "mi"],
-    "regression": ["mse", "mae", "r2", "mape", "rmse"],
+    "regression": ["mse", "mae", "r2", "map", "rmse"],
```
MatthewMiddlehurst, line 24: This is an example of a change that would break things.

aeon/networks/tests/test_ae_resnet.py
```diff
@@ -22,13 +22,13 @@ def test_ae_res_unit_activation(
     latent_space_dim, n_residual_blocks, activation, n_filters
 ):
     """Test whether AEResNetNetwork initializes correctly with different activations."""
-    aer = AEResNetNetwork(
+    are = AEResNetNetwork(
```
MatthewMiddlehurst, line 25: This is an acronym

MatthewMiddlehurst (commented):
So in principle I would be fine with having this and helping. There are a few concerns regarding automation that would mean I would not want it editing my code on its own (currently, can discuss), but making PRs or suggestions is fine. Left 2 comments for some bad changes, not had a full look through.

Is it necessary to have both a workflow and pre-commit checks? Interested to heal about some of the configuration also.

aeon/networks/tests/test_ae_resnet.py
yarikoptic: coolio -- that is an instance where I would just skip all `aer` hits. Adding it to overall skips.

yarikoptic:
@MatthewMiddlehurst thanks for the review!

> There are a few concerns that would mean I would not want it editing my code/docs on its own, but making PRs or suggestions is fine, and I could see it being helpful.

this would not do any changes -- it would just detect and annotate like at https://github.com/aeon-toolkit/aeon/pull/2653/files#diff-55dd3e1949bcb682be343ce038cc8afa221db3fd5d3096eee878870ca6b34543 you can now see 
image

which you would not get if you have only pre-commit.  Up to you to either leave a dedicated CI as well, or just resort to pre-commit.  I personally prefer a dedicated one for this and other reasons  such as immediately becomes clear that likely culprit is just a typo ;)

MatthewMiddlehurst:
Hi, @yarikoptic I went ahead and did the above. I think It would fit better into our current setup going through pre-commit. Can you just do the annotations without the other action? Would be helpful if you could take a look at the setup for any issues. Don't have to go through all the typos 🙂

MatthewMiddlehurst:
Also, I am a fan of skipping `.git` don't think we want to skip `.github`, `.git*` will do both I believe, tried a few ways to do one but not the other but didn't work. Not sure if you could help with that @yarikoptic

yarikoptic:
> Hi, @yarikoptic I went ahead and did the above. I think It would fit better into our current setup going through pre-commit. Can you just do the annotations without the other action? Would be helpful if you could take a look at the setup for any issues. Don't have to go through all the typos 🙂

I am not sure that annotation action alone, without codespell one, would work.  Could we push some TEMP commit with a typo and see if it gets properly annotated?  (and then drop that commit and force-push; let me know if you need help on that) I think there is overall little to no harm to have additional codespell step.  

> Also, I am a fan of skipping `.git` don't think we want to skip `.github`, `.git*` will do both I believe, tried a few ways to do one but not the other but didn't work. Not sure if you could help with that @yarikoptic

replace `.git*` with just `.git` in [skip](https://github.com/aeon-toolkit/aeon/pull/2653/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R193), or am I missing something ?

MatthewMiddlehurst:
No, not missing anything. I must have missed the simplest option while trying to fix that 🙂. Agreed, no harm in having an additional step/failing check for the annotations.

I am fine with this now but will let other developers comment if they want. Thank you for putting this up and for the help @yarikoptic.

pyproject.toml
```diff
@@ -186,3 +186,10 @@ filterwarnings = [
     "ignore:numpy.dtype size changed",
     "ignore:numpy.ufunc size changed",
 ]
+
+[tool.codespell]
+ignore-words = ".github/utilities/codespell_ignore_words.txt"
+ignore-regex = '(^\s*"image/\S+": ".*|Binay|Lamba|Mapp|Hart|Sch"afer|SOM|COO)'
```
SebastianSchmidl, line 192: Why do we ignore "Binay", "Lamba", "Mapp", and "Hart"?

SebastianSchmidl (commented):
Really nice tool to improve our code and documentation quality! I like it! Just a small comment regarding the excludes.

pyproject.toml
MatthewMiddlehurst: They are names, I do not remember if it from the contributor file or if various files references. Could probably exclude the contributors file later if it starts getting long or annoying. As its regex it will only ignore them if all letters are in the same upper/lower case.

yarikoptic: FWIW: yes, `ignore-regex` is the only way to ignore "case sensitive" with codespell.  the `ignore-words` are case insensitive.  Let me know if you would like to change things... ~~we already accumulated merge conflicts so ideally I might need to rebase/redo fixing to avoid messing with merge conflicts altogether.  But let's decide on how to approach those names if there is desire to change~~

yarikoptic: oops, just now I mentioned that there many commits beyond mine so likely we just merge/resolve once again, but IMHO would be good to finalize this PR soonish so conflicts do not keep coming

PR #2654, opened 2025-03-20
[ENH] Collection conversion cleanup and `df-list` fix

MatthewMiddlehurst:
- Changes `df-list` to use the (n_channels, n_timepoints) format. I thought we already agreed to change this and had done so in other places. I see this more of a fix than a new change.
- Adds more testing to collection datatype conversions. 
- Tidies up and makes some fixes for collection conversions based on the `df-list` changes and new testing catching stuff
- Changes collection validation for `df-list` and `pd-multiindex` in places


TonyBagnall:
I think we should just remove DF-List, there is no internal use case

MatthewMiddlehurst:
It is used as an in-between I believe, so not in this PR at least.

TonyBagnall (approved):
lgtm

PR #2668, opened 2025-03-20
[MNT] Use MacOS for examples/ workflow

### LINKED ISSUE #2665: [MNT] Use MacOS for `examples/` workflow
### Describe the issue

MacOS is generally faster for our CI, we should try running our examples workflow using it and see if there is a speed up.

Swap over to MacOS runners in the example CI. Evaluate whether this is faster than the current Linux runner.

### Additional context

See https://github.com/aeon-toolkit/aeon/pull/2390, these are the three files that will have to change (at least)

shinymack:
#### Reference Issues/PRs
Fixes #2665

- Changed the os in the workflows for `pr_examples` and `periodic_tests` to `macos-latest`.
- Changed the workflow to install `bash 5.x` before running `run_examples.sh`
as lastpipe isnt available in `bash 3.2` which is in the github runner image for `macos-latest` (https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md).
- Also added a step to install `boost` and `esig` before the `run_examples.sh` as `esig` is not installed on Darwin since its excluded in `pyproject.toml`.

No

##### For all contributions

### DIFF
--- a/.github/utilities/run_examples.sh
+++ b/.github/utilities/run_examples.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/opt/homebrew/bin/bash
 
 # Script to run all example notebooks.
 set -euxo pipefail
--- a/.github/utilities/run_examples.sh
+++ b/.github/utilities/run_examples.sh
@@ -32,7 +32,7 @@ notebooks=()
 runtimes=()
 
 # Loop over all notebooks in the examples directory.
-find "examples/" -name "*.ipynb" -print0 |
+find "examples" -name "*.ipynb" -print0 |
   while IFS= read -r -d "" notebook; do
     # Skip notebooks in the excluded list.
     if printf "%s\0" "${excluded[@]}" | grep -Fxqz -- "$notebook"; then
--- a/.github/utilities/run_examples.sh
+++ b/.github/utilities/run_examples.sh
@@ -23,6 +23,7 @@ if [ "$1" = true ]; then
     "examples/classification/shapelet_based.ipynb"
     "examples/classification/convolution_based.ipynb"
     "examples/similarity_search/code_speed.ipynb"
+    "examples/transformations/signature_method.ipynb"
 
   )
 fi
--- a/.github/workflows/periodic_tests.yml
+++ b/.github/workflows/periodic_tests.yml
@@ -46,12 +46,17 @@ jobs:
           extra_args: --all-files
 
   run-notebook-examples:
-    runs-on: ubuntu-24.04
+    runs-on: macos-latest
 
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Install bash 5.x
+        run: |
+          brew install bash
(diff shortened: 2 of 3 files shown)


shinymack:
Hi @MatthewMiddlehurst, Now the PR-Examples workflow runs in ~10 mins with macOS, which is better than Ubuntu's ~12 to 15 mins. Is this alright or should i adjust anything else?

.github/workflows/periodic_tests.yml
```diff
@@ -46,12 +46,17 @@ jobs:
           extra_args: --all-files
 
   run-notebook-examples:
-    runs-on: ubuntu-24.04
+    runs-on: macos-latest
```
MatthewMiddlehurst, line 49: use a fixed version. See the pytest runners.

.github/utilities/run_examples.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/opt/homebrew/bin/bash
```
MatthewMiddlehurst, line 1: Will this make it harder to run the script locally? If you need to update bash is there a better way to do so.

.github/workflows/periodic_tests.yml
```diff
@@ -69,6 +74,9 @@ jobs:
         with:
           additional_extras: "dev,binder"
 
+      - name: Install esig on macOS
+        run: brew install boost && pip install "esig>=0.9.7,<1.0.0"
```
MatthewMiddlehurst, line 78: Why is this needed?

.github/utilities/run_examples.sh
shinymack: I don’t think this makes it harder to run locally - installing the latest bash via brew is pretty safe. Since the script needs Bash 5.x for `lastpipe`, we could add a symlink from brew’s bash to `/bin/bash` . That way, `#!/bin/bash` works everywhere. should i do this?

.github/workflows/periodic_tests.yml
MatthewMiddlehurst: I would add the esig notebook to the skip list for now.

.github/utilities/run_examples.sh
MatthewMiddlehurst: either one is fine. I would change the name though "Install bash 5.x" will be outdated eventually

shinymack:
@MatthewMiddlehurst I have made the required changes. Regarding the symlink, its not possible due mac-OS's SIP, so `/bin/bash` cannot be modified even by using sudo.

MatthewMiddlehurst:
please avoid force committing if possible it makes reviewing changes harder.

MatthewMiddlehurst (approved):
lgtm

MatthewMiddlehurst:
@aeon-actions-bot assign @MatthewMiddlehurst

MatthewMiddlehurst:
sorry testing something :)

PR #2679, opened 2025-03-23
[ENH] Implementation of Extended Isolation Forest (EIF) anomaly detector

### LINKED ISSUE #2113: [ENH] Implement Extended Isolation Forest

The Extended Isolation Forest improves upon the Isolation Forest method.

Implement directly in aeon:

- Paper: https://doi.org/10.1109/TKDE.2019.2947676
- Code: https://github.com/sahandha/eif
- Code (in TimeEval): https://github.com/TimeEval/TimeEval-algorithms/tree/main/eif

However, support semi-supervised as well as unsupervised usage similar to the [`PyODAdapter`](https://github.com/aeon-toolkit/aeon/blob/main/aeon/anomaly_detection/_pyodadapter.py) or [`KMeansAD`](https://github.com/aeon-toolkit/aeon/blob/main/aeon/anomaly_detection/_kmeans.py)

_No response_

### Additional context

_No response_

Akhil-Jasson:
This PR implements the Extended Isolation Forest (EIF) algorithm.

#### Reference Issues/PRs
Fixes #2113

Yes, it introduces H20.ai as a new dependency

##### For all contributions


MatthewMiddlehurst:
Please fill out the template and use the correct title format.

Akhil-Jasson:
I've fixed the title format. Shall I proceed with adding EIF to the online API documentation?

Ramana-Raja:
Hi @Akhil-Jasson, I checked out your PR, and the code looks great! That said, I don’t think using H2O is the best approach since Aeon doesn’t rely on it. It might be better to have our own implementation instead.

A few updates to consider:
   1. Could you update the section "Does your contribution introduce a new dependency?" and mention H2O there?

    2.The test cases seem to be missing—could you add them?

    3.Instead of importing the entire H2O module, it’s better to import only what’s needed to keep things lightweight.

MatthewMiddlehurst:
New dependencies should be put in pyproject.toml otherwise this won't be tested, Still bits missing from the template

Akhil-Jasson:
I've added the h2o dependency to pyproject.toml, but I'm encountering errors when running the test files. The test attempts to import aeon.anomaly_detection._eif but fails, indicating that the module doesn’t exist yet.  

Is there a step I'm missing for adding new modules to aeon? What could be the possible issue?

MatthewMiddlehurst:
your import is incorrect.

SebastianSchmidl (changes requested):
AI2O looks like a massive package. Do we actually want to include it as a dependency? The issue clearly states that we are looking for an implementation in aeon directly.

I know this is not mentioned in the corresponding issue, but I think it makes sense to work with sliding windows in EIF as well. We can always get the original behavior back by setting the window-size to 1.

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+    window_size : int, default=1
+        Size of the sliding window. If 1, the original point-wise behavior is used.
+    stride : int, default=1
+        Stride of the sliding window.
+    axis : int, default=1
+        The time point axis of the input series if it is 2D. If ``axis==0``, it is
+        assumed each column is a time series and each row is a time point. i.e. the
+        shape of the data is ``(n_timepoints, n_channels)``. ``axis==1`` indicates
+        the time series are in rows, i.e. the shape of the data is
+        ``(n_channels, n_timepoints)``.
```
SebastianSchmidl, line 51: Different data layouts are handled in the base class. Please do not expose an additional `axis`-parameter here.

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+    def _fit(self, X, y=None):
+        """Fit the model using X as training data.
+
+        Parameters
+        ----------
+        X : np.ndarray
+            Training data of shape (n_timepoints,) or (n_timepoints, n_channels)
+        y : np.ndarray, optional
+            Labels for semi-supervised learning. 0 for normal, 1 for anomalous.
+            If None, unsupervised learning is used.
```
SebastianSchmidl, line 93: Semi-supervised means training on normal data only, so just `0` is allowed here. Otherwise, this would be supervised training with normal data and known anomalies. Details: https://www.aeon-toolkit.org/en/stable/examples/anomaly_detection/anomaly_detection.html#Anomaly-Detection-in-aeon

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Returns
+        -------
+        self : object
+            Fitted estimator.
+        """
+        # Ensure X is 2D
+        if X.ndim == 1:
+            X = X.reshape(-1, 1)
```
SebastianSchmidl, line 102: This should not be necessary anymore if you remove the axis-parameter from the class and use a single data layout within the estimator.

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+
+            # Build tree
+            tree = self._build_tree(X_sample, rng)
+            self.forest_.append(tree)
+
+        # Calculate anomaly scores
+        scores = self._predict(X)
+
+        # Set threshold
+        self.threshold_ = float(np.percentile(scores, 100 * (1 - self.contamination)))
```
SebastianSchmidl, line 156: This looks suspicious. Why do you predict in `fit`?

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+        # Convert to anomaly scores
+        scores = 2 ** (-scores / self._c(self.sample_size_))
+
+        # If windowing was applied, we need to map scores back to original samples
+        if self.window_size > 1:
+            scores = reverse_windowing(
+                scores,
+                window_size=self.window_size,
+                stride=self.stride,
+                n_timepoints=self._n_samples_orig,
```
SebastianSchmidl, line 300: This just works if either EIF is used in an unsupervised way or both training and prediction time series have the same length!

aeon/anomaly_detection/_eif.py
```diff
@@ -0,0 +1,331 @@
...
+        Returns
+        -------
+        float
+            Average path length
+        """
+        if n <= 1:
+            return 0
+        elif n == 2:
+            return 1
+        return 2 * (np.log(n - 1) + 0.5772156649) - 2 * (n - 1) / n
```
SebastianSchmidl, line 259: Where is the magic number from? Please at least add a comment.

SebastianSchmidl (changes requested):
The code does not look too complicated, so I think, it was a good decision to not include H2O.ai as a dependency!

Thanks for the implementation effort. Please address the issues below and then compare your implementation to the original implementation by Hariri et al. to demonstrate that your/our implementation produces the same results as theirs.

I think this estimator is a suitable candidate for improving the performance with Numba / JITs. But let us first focus on making the implementation correct, and then tackle the performance optimization in another PR.

Akhil-Jasson:
Previously, the tests were triggered automatically on commit based on the old repo structure. Since then, the structure has changed significantly. I ran the tests locally using the earlier setup, and they passed.

To accommodate the requirements of the test files—such as the need for y in the fit function—I’ve made it an optional parameter. I’ll take a look into the other comments as well.

Given the extent of changes in the repo structure, is it easier to just create a new branch from the updated main branch or is there an easier alternative? Please let me know!

MatthewMiddlehurst:
I would merge main into your branch, the bulk of your code should not conflict. It is mainly that we have refactored the module so it has all moved around a bit. If you do not know which module to put your implementation in just put it anywhere for now and we can discuss after where is best.

Akhil-Jasson:
Changes have been made to this branch to align with the newly refactored module structure. I'll soon address the feedback to remove the axis hyperparameter and add the necessary comments as requested.

MatthewMiddlehurst:
Hello, is this still active?

Akhil-Jasson:
Hello @MatthewMiddlehurst,
Thank you for checking in. I apologize for the long pause. I've had some time away, but I’m still eager to continue work on the Extended Isolation Forest implementation.

Akhil-Jasson:
Hi @SebastianSchmidl  and @MatthewMiddlehurst,

I've addressed the feedback from the previous review:

Key Changes Made:
- Removed threshold calculation from _fit()
- Fixed sliding window length handling - Now works with different-sized prediction data (addresses the length dependency issue)
- Added comments for magic numbers - Explained the Euler-Mascheroni constant in _c() method
- Maintained unsupervised/semi-supervised support - Works with both `y=None` and `y=0` modes

Please let me know if there are any remaining concerns.

Thanks for the detailed feedback, it really helped improve the implementation!

MatthewMiddlehurst:
Hello, based on the current state of the PR and the changes made, I think this is going to take a lot more effort to get anywhere. If you want to continue, please:

1. Look at the series anomaly detection base class. There is a misunderstanding here of what we mean by unsupervised and semi-supervised and how that is configured in the framework. The notebook we linked appears to contradict this, so that is our bad.
2. Look at other anomaly detections to see the format and tags used. The init requires you to provide an axis to the base class which determines the shape of the data
3. Merge to the most recent main branch so testing can be run and the code is up to date with recent changes. There is also no testing for this currently.
4. Run your implementation and compare against the original results

This is a good base, but we cant accept code which breaks the framework or we are unsure works as intended.

aeon/anomaly_detection/outlier_detection/_eif.py
```diff
@@ -0,0 +1,347 @@
...
+                scores = scores[:original_n_timepoints]
+            elif len(scores) < original_n_timepoints:
+                # Pad with edge values if needed
+                pad_length = original_n_timepoints - len(scores)
+                scores = np.pad(scores, (0, pad_length), mode="edge")
+
+        return scores
+
+    def predict_labels(self, X) -> np.ndarray:
+        """Predict if points are anomalies or not.
```
MatthewMiddlehurst, line 322: this method shouldnt exist IMO.

aeon/anomaly_detection/outlier_detection/_eif.py
```diff
@@ -0,0 +1,347 @@
...
+class EIF(BaseAnomalyDetector):
+    """Extended Isolation Forest (EIF) for anomaly detection.
+
+    Implementation of the Extended Isolation Forest algorithm that uses
+    random hyperplanes for splitting instead of axis-parallel splits.
+    This allows for better handling of high-dimensional data and complex distributions.
+
+    The implementation supports both unsupervised and semi-supervised learning:
+    - Unsupervised: No labels provided (y=None)
+    - Semi-supervised: Labels provided (y=0 for normal data)
```
MatthewMiddlehurst, line 21: by unsupervised we mean it makes predictions on the same data with no labels, for semi-supervised it fits on data then this can be used for other data. I do not think this is unsupervised in our framework.

aeon/anomaly_detection/outlier_detection/_eif.py
```diff
@@ -0,0 +1,347 @@
...
+        -------
+        self : object
+            Fitted estimator.
+        """
+        if y is not None:
+            y = np.asarray(y)
+            if not np.all(y == 0):
+                raise ValueError(
+                    "Semi-supervised mode only supports training on normal data (y==0)."
+                )
```
MatthewMiddlehurst, line 98: I do not think this is needed as y is never used. This may be worth putting in the base class, however.

aeon/anomaly_detection/outlier_detection/_eif.py
```diff
@@ -0,0 +1,347 @@
...
+            X_windowed = X
+
+        # Set random state
+        rng = np.random.RandomState(self.random_state)
+
+        # Determine sample size
+        n_samples = X_windowed.shape[0]
+        if isinstance(self.sample_size, str):
+            if self.sample_size != "auto":
+                raise ValueError("sample_size must be 'auto', int or float")
```
MatthewMiddlehurst, line 115: this error catching could be better. it will only run if it is an incorrect str

aeon/anomaly_detection/outlier_detection/_eif.py
```diff
@@ -0,0 +1,347 @@
...
+        # Apply sliding window if window_size > 1
+        if self.window_size > 1:
+            X_windowed, _ = sliding_windows(
+                X, window_size=self.window_size, stride=self.stride
+            )
+        else:
+            X_windowed = X
+
+        # Set random state
+        rng = np.random.RandomState(self.random_state)
```
MatthewMiddlehurst, line 109: can you use https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html for rng, see the docstring of other algorithms which use rng for a better one

aeon/anomaly_detection/outlier_detection/_eif.py
MatthewMiddlehurst: We made a decision previously that different types i.e. unsupervised and semi-supervised should be different classes. I would just go for semi-supervised for now.

TonyBagnall:
hi @Akhil-Jasson in the spirit of housekeeping I am converting this to draft, feel free to reopen if you want to continue with it

TonyBagnall:
made redundant by #3626

Issue #2660, opened 2025-03-20
[ENH] Implement U-Shapelets clusterer


MatthewMiddlehurst:
### Describe the feature or idea you want to propose

It would be nice to have an implementation of the Unsupervised Shapelets (U-Shapelets)

https://ieeexplore.ieee.org/document/6413851
https://epubs.siam.org/doi/pdf/10.1137/1.9781611974010.101
https://link.springer.com/article/10.1007/s10618-015-0411-4

### Describe your proposed solution

Implement the U-Shapelets algorithms and possibly scalability improvements. It will be on the contributor to evaluate the accuracy of their implementation to published code and results.

aryanpola:
@aeon-actions-bot assign @aryanpola

PR #2680, opened 2025-03-23
[ENH] Implementing U-shapelet algorithm

aryanpola:
#### Reference Issues/PRs
Implements #2660 

Implements Unsupervised Shapelet Clustering Algorithm. Clusters the time series based on binary labels assigned.

#### Any other comments?
The paper currently suggests to split the time series into 2 based on a single best shapelet.
 We can incorporate `k` (number of clusters) parameter to include iterative splitting to form more clusters.

### DIFF
--- a/aeon/clustering/_u_shapelet.py
+++ b/aeon/clustering/_u_shapelet.py
@@ -20,7 +20,7 @@ class UShapeletClusterer(BaseClusterer):
     ----------
     subseq_length : int, default=10
         Length of the subsequences to consider as candidate shapelets.
-    sax_length : int, default=8
+    sax_length : int, default=16
         Length of the PAA transformation used in the SAX representation.
     projections : int, default=10
         Number of random projections used in collision counting.
@@ -43,7 +43,7 @@ class UShapeletClusterer(BaseClusterer):
     def __init__(
         self,
         subseq_length: int = 10,
-        sax_length: int = 8,
+        sax_length: int = 16,
         projections: int = 10,
         lb: int = 2,
         ub: int = None,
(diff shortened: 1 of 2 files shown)


aeon/clustering/_u_shapelet.py
```diff
@@ -0,0 +1,679 @@
...
+        self.sax_length = sax_length
+        self.projections = projections
+        self.lb = lb
+        self.ub = ub
+        self.n_clusters = n_clusters
+
+        if isinstance(random_state, np.random.RandomState):
+            self.random_state = random_state
+        else:
+            self.random_state = np.random.RandomState(random_state)
```
MatthewMiddlehurst, line 84: dont do this, parameters must have an attribute i.e. `self.random_state` which is the same as the input.

aryanpola: Thanks for the help!

TonyBagnall:
hi @aryanpola thanks for this, could you describe in the PR description a bit more about your implementation, and its correctness in terms of the original paper? Is it based on other source code or done directly from the paper description? Have you tested it against any other published results? thanks

aryanpola:
Hi @TonyBagnall,
Apologies for the delayed reply. I've implemented the code using the papers as a reference. There are still some enhancements (early abandoning from [this ](https://link.springer.com/article/10.1007/s10618-015-0411-4), r filter ratio from [this](https://epubs.siam.org/doi/pdf/10.1137/1.9781611974010.101)) to be done. I also spent some time testing on the Trace dataset but haven’t been able to reproduce the expected split yet, still looking into why the gap scores aren’t lining up.
I'm tied up with coursework right now, will try to get this done asap.

TonyBagnall:
hi @aryanpola can we close this PR or do you still want to work on it? Its failing some CI. thanks, Tony

aryanpola:
We’ll close this for now and plan to revisit it at a later stage.

PR #2578, opened 2025-02-28
[ENH] KNN n_jobs and updated kneighbours method

chrisholder:
#### Reference Issues/PRs
This PR depends on #2545 

#### What does this implement/fix? Explain your changes.
This PR passes n_jobs to the pairwise distance function so that is now can run on multiple threads. In addition I changed the k_neighbors method to be compatible with sklearns which can be used in various ways. See: https://github.com/scikit-learn/scikit-learn/blob/98ed9dc73/sklearn/neighbors/_base.py#L763.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you after the PR has been merged.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #2687, opened 2025-03-24
[ENH]Use n_jobs parameter in KNeighborsTimeSeriesClassifier.

### LINKED ISSUE #2478: [ENH] functional n_jobs parameter for knn classifier

Current [n_jobs params](https://github.com/aeon-toolkit/aeon/blob/66128decb6f1f17e1e61ada598e3e8d5f1c8909c/aeon/classification/distance_based/_time_series_neighbors.py#L49) is not doing anything in [knn classifier.](https://github.com/aeon-toolkit/aeon/blob/66128decb6f1f17e1e61ada598e3e8d5f1c8909c/aeon/classification/distance_based/_time_series_neighbors.py#L23)

Make use of it ! TBD how exactly. If not possible a warning should at least be raised.

_No response_

### Additional context
Was curious after looking at [sequentia](https://github.com/eonu/sequentia) benchmark on why we were so slow...

steenrotsman:
#### Reference Issues/PRs
Fixes #2478. See also #2545 and #2578.

Uses n_jobs parameter in `_predict` and `predict_proba` of `KNeighborsTimeSeriesClassifier`. Parallelization is done in these methods instead of `_kneighbors` to potentially allow speedup through upper bounding the distance. 

##### For all contributions

### DIFF
--- a/aeon/classification/distance_based/_time_series_neighbors.py
+++ b/aeon/classification/distance_based/_time_series_neighbors.py
@@ -51,6 +51,11 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
         The number of parallel jobs to run for neighbors search.
         ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
         ``-1`` means using all processors.
+    parallel_backend : str, ParallelBackendBase instance or None, default=None
+        Specify the parallelisation backend implementation in joblib, if None
+        a ‘prefer’ value of “threads” is used by default. Valid options are
+        “loky”, “multiprocessing”, “threading” or a custom backend.
+        See the joblib Parallel documentation for more details.
 
     Examples
     --------
@@ -79,11 +84,13 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
         n_neighbors: int = 1,
         weights: Union[str, Callable] = "uniform",
         n_jobs: int = 1,
+        parallel_backend: str = None,
     ) -> None:
         self.distance = distance
         self.distance_params = distance_params
         self.n_neighbors = n_neighbors
         self.n_jobs = n_jobs
+        self.parallel_backend = parallel_backend
 
         self._distance_params = distance_params
         if self._distance_params is None:
@@ -134,7 +141,9 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
             The class probabilities of the input samples. Classes are ordered
             by lexicographic order.
         """
-        preds = Parallel(n_jobs=self.n_jobs)(delayed(self._proba_row)(x) for x in X)
+        preds = Parallel(n_jobs=self.n_jobs, parallel_backend=self.parallel_backend)(
+            delayed(self._proba_row)(x) for x in X
+        )
         return np.array(preds)
 
     def _predict(self, X):
@@ -153,7 +162,9 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
         y : array of shape (n_cases)
             Class labels for each data sample.
         """
-        preds = Parallel(n_jobs=self.n_jobs)(delayed(self._predict_row)(x) for x in X)
+        preds = Parallel(n_jobs=self.n_jobs, parallel_backend=self.parallel_backend)(
+            delayed(self._predict_row)(x) for x in X
+        )
         return np.array(preds, dtype=self.classes_.dtype)
 
     def _proba_row(self, x):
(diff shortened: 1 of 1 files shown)


aeon/classification/distance_based/_time_series_neighbors.py
```diff
@@ -134,16 +135,9 @@ def _predict_proba(self, X):
...
-        for i in range(len(X)):
-            idx, weights = self._kneighbors(X[i])
-            for id, w in zip(idx, weights):
-                predicted_class = self.y_[id]
-                preds[i, predicted_class] += w
-
-            preds[i] = preds[i] / np.sum(preds[i])
-
-        return preds
+        self._check_is_fitted()
```
MatthewMiddlehurst, line 138: not needed

aeon/classification/distance_based/_time_series_neighbors.py
```diff
@@ -162,18 +156,24 @@ def _predict(self, X):
...
+
+    def _proba_row(self, x):
+        scores = self._predict_scores(x)
+        return scores / np.sum(scores)
+
+    def _predict_row(self, x):
+        scores = self._predict_scores(x)
+        return self.classes_[np.argmax(scores)]
+
+    def _predict_scores(self, x):
```
MatthewMiddlehurst, line 170: not used?

MatthewMiddlehurst (commented):
Hi, could you verify output is the same from before and show that the mutlithreading does make the classifier faster.

aeon/classification/distance_based/_time_series_neighbors.py
steenrotsman: Should I also remove it from 158?

aeon/classification/distance_based/_time_series_neighbors.py
steenrotsman: _predict_scores is used in both _proba_row and _predict_row, which are used in _predict_proba and _predict, respectively. If you want, I could delete these methods and have _predict(_proba) use _predict_scores directly.

steenrotsman:
In #2478, I've put some benchmarks. These are the entries on three data sets with 1 or 8 cores with only Aeon's builtin DTW:
|name | avg | std | min |
|:--|--:|--:|--:|
|ACSF1 (1) | 110.2201 | 2.4291 | 107.7615|
|ACSF1 (8) | 49.2053 | 1.0859 | 48.3538|
|ArrowHead (1) | 1.8508 | 0.0506 | 1.7939|
|ArrowHead (8) | 0.5634 | 0.0114 | 0.5545|
|GunPoint (1) | 0.8112 | 0.0238 | 0.7918|
|GunPoint (8) | 1.0206 | 1.6650 | 0.2700|

For the small GunPoint data set, overhead is larger than gains from parallelization. I expect that for ED, data sets would need to be larger to get a performance increase (if that's even possible) compared to DTW.

Equivalence is tested in `aeon/distances/tests/test_sklearn_compatibility.py`: previous version tested equivalence with `sklearn`. The test passes with the new version, so the new output is equal to the old output.

aeon/classification/distance_based/_time_series_neighbors.py
MatthewMiddlehurst: ah did not notice it was already there! Yes it should be redundant, this is handled in the base class.

MatthewMiddlehurst (changes requested):
Thanks. Could you edit the `n_jobs` docstring and consider implementing the backend parameter found in other threaded classifiers. Don't have to do the second but would be a nice addition I think.

aeon/classification/distance_based/_time_series_neighbors.py
```diff
@@ -134,16 +141,10 @@ def _predict_proba(self, X):
...
-        for i in range(len(X)):
-            idx, weights = self._kneighbors(X[i])
-            for id, w in zip(idx, weights):
-                predicted_class = self.y_[id]
-                preds[i, predicted_class] += w
-
-            preds[i] = preds[i] / np.sum(preds[i])
-
-        return preds
+        preds = Parallel(n_jobs=self.n_jobs, backend=self.parallel_backend)(
```
MatthewMiddlehurst, line 144: use check_n_jobs for both functions

MatthewMiddlehurst (approved):

MatthewMiddlehurst:
LGTM. Question is still open on whether this is the best way to do this for KNN, but it is much better than nothing.

TonyBagnall (approved):
yes thanks for this

PR #2715, opened 2025-03-31
[BUG] Updated sbd_distance() to handle multivariate series (#2674)

### LINKED ISSUE #2674: [BUG] Inconsistent Sbd distance with `tslearn` and other implementations
### Describe the bug

While implementing **K-shape clustering**(#2661) I noticed that our implementation for the `sbd_distance` differs slightly from how `tslearn` handles it.

This doesn't matter for univariate series but for multivariate series we do see differences.

**sbd_distance**: finds the distance for each channel independently and then takes its average
**normalized_cc**: finds the correlations for each of the channels and then sums the max of each channels, and then normalizes using the norm of the the entire multivariate series.

I found another implementation of **kshape** online, one of its contributors is the original author of the kshapes paper (https://dl.acm.org/doi/pdf/10.1145/2723372.2737793). They handle this the same way as `tslearn`. https://github.com/TheDatumOrg/kshape-python

I am not sure if this is an intentional choice but this changes the final clustering output. 
If we use aeon's sbd distances for k-shape clustering we will have to change the test cases to reflect the same, but then we wouldn't have something to compare it against.

```python

import numpy as np
from aeon.distances import sbd_distance
from tslearn.metrics.cycc import normalized_cc

# Multivariate time series
np.random.seed(1)
x = np.random.rand(6, 100)
y = np.random.rand(6, 100)
a = sbd_distance(x, y, standardize=False)
print(x.shape, y.shape)

z = np.transpose(x, (1, 0))
w = np.transpose(y, (1, 0))
b = 1.0 - normalized_cc(z, w).max()
print(a, b)


# Univariate time series
np.random.seed(1)
x = np.random.rand(1, 200)
y = np.random.rand(1, 200)
print(x.shape, y.shape)

a = sbd_distance(x, y, standardize=False)

x = np.transpose(x, (1, 0))
y = np.transpose(y, (1, 0))

b = 1.0 - normalized_cc(x, y).max()
print(a, b)
```

### Expected results

Expected result is that sbd distance computed for both univariate and multivariate time series are consistent with `tslearn` and https://github.com/TheDatumOrg/kshape-python

### Actual results

1st example is multivariate time series which is inconsistent with `tslearn` and https://github.com/TheDatumOrg/kshape-python
Image

### Versions

System:
    python: 3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:06:27) [MSC v.1942 64 bit (AMD64)]
executable: D:\Open Source\aeon\aeon-venv\Scripts\python.exe
Python dependencies:
         aeon: 1.0.0
          pip: 24.3.1
   setuptools: 75.8.0
 scikit-learn: 1.5.2
        numpy: 1.26.4
        numba: 0.60.0
        scipy: 1.14.1
       pandas: 2.2.3

pvprajwal:
#### Reference Issues/PRs
Fixes #2674

* Updated sbd_distance() to handle multivariate data consistently with tslearn and other implementations
* added _multivariate_sbd_distance() which finds the correlations for each of the channels and then normalizes using the norm of the multivariate series.

The earlier implementation found the average of the distance after calculating normalized_cc for each channel independenly.
```py
if x.shape[0] == 1 and y.shape[0] == 1:
    _x = x.ravel()
    _y = y.ravel()
    return _univariate_sbd_distance(_x, _y, standardize)
else:
    # independent (time series should have the same number of channels!)
    nchannels = min(x.shape[0], y.shape[0])
    distance = 0.0
    for i in range(nchannels):
        distance += _univariate_sbd_distance(x[i], y[i], standardize)
    return distance / nchannels
```

The is replaced by a new _multivariate_sbd_distance() method which normalizes using the norm of the multivariate series, as is the case with tslearn and kshape-python. 

```py
def _multivariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) -> float:
    x = x.astype(np.float64)
    y = y.astype(np.float64)

    x = np.transpose(x, (1, 0))
    y = np.transpose(y, (1, 0))

    if standardize:
        if x.size == 1 or y.size == 1:
            return 0.0

        x = (x - np.mean(x)) / np.std(x)
        y = (y - np.mean(y)) / np.std(y)

    norm1 = np.linalg.norm(x)
    norm2 = np.linalg.norm(y)
    
    denom = norm1 * norm2
    if denom < 1e-9:  # Avoid NaNs
        denom = np.inf

    with objmode(cc="float64[:, :]"):
        cc = np.array([correlate(x[:, i], y[:, i], mode="full", method="fft") for i in range(x.shape[1])]).T

    sz = x.shape[0]
    cc = np.vstack((cc[-(sz - 1):], cc[:sz]))  # Reorganize correlation values
    norm_cc = np.real(cc).sum(axis=-1) / denom

    return np.abs(1.0 - np.max(norm_cc))

```

Nil

#### Any other comments?
Tests need to be modified. Now the values are consistent for tslearn and kshape-python 

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you after the PR has been merged.

### DIFF
--- a/aeon/distances/_sbd.py
+++ b/aeon/distances/_sbd.py
@@ -98,17 +98,7 @@ def sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool = True) -> floa
     if x.ndim == 1 and y.ndim == 1:
         return _univariate_sbd_distance(x, y, standardize)
     if x.ndim == 2 and y.ndim == 2:
-        if x.shape[0] == 1 and y.shape[0] == 1:
-            _x = x.ravel()
-            _y = y.ravel()
-            return _univariate_sbd_distance(_x, _y, standardize)
-        else:
-            # independent (time series should have the same number of channels!)
-            nchannels = min(x.shape[0], y.shape[0])
-            distance = 0.0
-            for i in range(nchannels):
-                distance += _univariate_sbd_distance(x[i], y[i], standardize)
-            return distance / nchannels
+        return _multivariate_sbd_distance(x, y, standardize)
 
     raise ValueError("x and y must be 1D or 2D")
 
@@ -245,3 +235,34 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
 
     b = np.sqrt(np.dot(x, x) * np.dot(y, y))
     return np.abs(1.0 - np.max(a / b))
+
+@njit(cache=True, fastmath=True)
+def _multivariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) -> float:
+    x = x.astype(np.float64)
+    y = y.astype(np.float64)
+
+    x = np.transpose(x, (1, 0))
+    y = np.transpose(y, (1, 0))
+
+    if standardize:
+        if x.size == 1 or y.size == 1:
+            return 0.0
+
+        x = (x - np.mean(x)) / np.std(x)
+        y = (y - np.mean(y)) / np.std(y)
+
+    norm1 = np.linalg.norm(x)
+    norm2 = np.linalg.norm(y)
+    
+    denom = norm1 * norm2
+    if denom < 1e-9:  # Avoid NaNs
+        denom = np.inf
+
+    with objmode(cc="float64[:, :]"):
+        cc = np.array([correlate(x[:, i], y[:, i], mode="full", method="fft") for i in range(x.shape[1])]).T
+
+    sz = x.shape[0]
+    cc = np.vstack((cc[-(sz - 1):], cc[:sz]))
+    norm_cc = np.real(cc).sum(axis=-1) / denom
+
+    return np.abs(1.0 - np.max(norm_cc))
--- a/aeon/distances/_sbd.py
+++ b/aeon/distances/_sbd.py
@@ -236,8 +236,11 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
     b = np.sqrt(np.dot(x, x) * np.dot(y, y))
     return np.abs(1.0 - np.max(a / b))
 
+
 @njit(cache=True, fastmath=True)
-def _multivariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) -> float:
+def _multivariate_sbd_distance(
+    x: np.ndarray, y: np.ndarray, standardize: bool
+) -> float:
     x = x.astype(np.float64)
     y = y.astype(np.float64)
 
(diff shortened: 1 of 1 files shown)


pvprajwal:
I think the tests have to be modified to be consistent with the new sbd_distance calculation. The same approach is used in the original k-shapes paper at https://dl.acm.org/doi/pdf/10.1145/2723372.2737793. 

image

aeon/distances/_sbd.py
```diff
@@ -245,3 +235,42 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
...
+
+@njit(cache=True, fastmath=True)
+def _multivariate_sbd_distance(
+    x: np.ndarray, y: np.ndarray, standardize: bool
+) -> float:
+    x = x.astype(np.float64)
+    y = y.astype(np.float64)
+
+    x = np.transpose(x, (1, 0))
+    y = np.transpose(y, (1, 0))
```
SebastianSchmidl, line 248: You transpose the output in L270 again. Is this really necessary?

aeon/distances/_sbd.py
```diff
@@ -98,17 +98,7 @@ def sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool = True) -> floa
     if x.ndim == 1 and y.ndim == 1:
         return _univariate_sbd_distance(x, y, standardize)
     if x.ndim == 2 and y.ndim == 2:
-        if x.shape[0] == 1 and y.shape[0] == 1:
-            _x = x.ravel()
-            _y = y.ravel()
-            return _univariate_sbd_distance(_x, _y, standardize)
```
SebastianSchmidl, line 104: Please keep this branch for now.

aeon/distances/_sbd.py
```diff
@@ -245,3 +235,42 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
...
+
+    x = np.transpose(x, (1, 0))
+    y = np.transpose(y, (1, 0))
+
+    if standardize:
+        if x.size == 1 or y.size == 1:
+            return 0.0
+
+        x = (x - np.mean(x)) / np.std(x)
+        y = (y - np.mean(y)) / np.std(y)
```
SebastianSchmidl, line 255: After transposing, this is standardizing along the wrong axis, isn't it?

aeon/distances/_sbd.py
```diff
@@ -245,3 +235,42 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
...
+    if denom < 1e-9:  # Avoid NaNs
+        denom = np.inf
+
+    with objmode(cc="float64[:, :]"):
+        cc = np.array(
+            [
+                correlate(x[:, i], y[:, i], mode="full", method="fft")
+                for i in range(x.shape[1])
+            ]
+        ).T
```
SebastianSchmidl, line 270: Python loops are expensive. This is the reason, we use Numba a lot in the distance module. Is this loop really necessary? Maybe the correlate function can be used in a vectorized way?

aeon/distances/_sbd.py
```diff
@@ -245,3 +235,42 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
...
+
+    sz = x.shape[0]
+    cc = np.vstack((cc[-(sz - 1) :], cc[:sz]))
```
SebastianSchmidl, line 273: Slicing with `[:n]` should be equal to slicing with `[:-1]`, which is simpler to understand, or are you working around a limitation in Numba? Then, please add a comment.

SebastianSchmidl (changes requested):

SebastianSchmidl:
We should benchmark the new implementation against the existing one for univariate series. If it is faster, we could get rid of the existing code.

Please also verify that the `pairwise_distance`-function supports variable-length inputs.

MatthewMiddlehurst:
Seems abandoned. Feel free to re-open if not.

Issue #2661, opened 2025-03-20
[ENH] Implement K-Shape clusterer


MatthewMiddlehurst:
### Describe the feature or idea you want to propose

We have a wrapper for K-Shape, but no internal implementation. It may be good to have one and possibly deprecate the wrapper if results match.

### Describe your proposed solution

Implement the K-Shape algorithm. It will be on the contributor to evaluate the accuracy of their implementation to published code and results. Efficiency is also a concern here, ideally we would be faster than the wrapper.

tanishy7777:
I was trying to implementing the **kshape clustering** algorithm for aeon.
and I noticed that our implementation for the **sbd_distance** in aeon differs slightly from how tslearn handles it. This dosen't matter for univariate series but for multivariate series we do see differences.

sbd_distance: finds the distance for each channel independently and then takes its average
normalized_cc: finds the correlations for each of the channels and them sums them, and then normalizes using the norm of the the entire multivariate series.

This changes the final clustering output? How should we handle this?
If we use aeon's sbd distances we will have to change the test cases to reflect the same, but then we wouldnt have something to compare it against.

I found another implementation of kshape online, one of its contributors is the original author of the kshapes paper. They handle this the same way as tslearn. https://github.com/TheDatumOrg/kshape-python

SebastianSchmidl:
> I found another implementation of kshape online, one of its contributors is the original author of the kshapes paper. They handle this the same way as tslearn. https://github.com/TheDatumOrg/kshape-python

This is the original implementation. In this case, we should adapt the implementation of SBD in aeon to align with the original proposal.

First fix SBD in one PR, then we continue implementing k-shape.

tanishy7777:
> > I found another implementation of kshape online, one of its contributors is the original author of the kshapes paper. They handle this the same way as tslearn. https://github.com/TheDatumOrg/kshape-python
> 
> This is the original implementation. In this case, we should adapt the implementation of SBD in aeon to align with the original proposal.
> 
> First fix SBD in one PR, then we continue implementing k-shape.

Got it, will start working on this.

tanishy7777:
> > I found another implementation of kshape online, one of its contributors is the original author of the kshapes paper. They handle this the same way as tslearn. https://github.com/TheDatumOrg/kshape-python
> 
> This is the original implementation. In this case, we should adapt the implementation of SBD in aeon to align with the original proposal.
> 
> First fix SBD in one PR, then we continue implementing k-shape.

[...]

PR #2742, opened 2025-04-13
[BUG] Fixes sbd distances for multivariate case

### LINKED ISSUE #2674: [BUG] Inconsistent Sbd distance with `tslearn` and other implementations

tanishy7777:
#### Reference Issues/PRs
Fixes #2674
Closes #2715
As mentioned in  https://github.com/aeon-toolkit/aeon/issues/2661#issuecomment-2745266428 and https://github.com/aeon-toolkit/aeon/issues/2661#issuecomment-2779108840

fixes sbd distance for multivariate case

Currently the implementation of **sbd distance** in aeon handles the multivariate case slightly differently than the official version(https://github.com/TheDatumOrg/kshape-python) and tslearn's implementation as well.

Also, I set the `unequal_length=False` (need input from reviewers regarding this as SBD distance(official & tslearn version) supports  unequal length time series but doesn't support unequal channels, but the tests in aeon in `test_distances` check for unequal channels also, more details in  https://github.com/aeon-toolkit/aeon/issues/2661#issuecomment-2799941937 ).

Differences:
sbd_distance: finds the distance for each channel independently and then takes its average
normalized_cc(from tslearn): finds the correlations for each of the channels and then sums them and finds the max from this sum and then normalizes them using the norm of the the entire multivariate series.

##### For all contributions


tanishy7777:
Equivalence between [original implementation](https://github.com/TheDatumOrg/kshape-python), tslearn's implementation and fixed implementation as per (https://github.com/aeon-toolkit/aeon/pull/2742)
Image

tanishy7777:
Reply to https://github.com/aeon-toolkit/aeon/issues/2661#issuecomment-2799962773


Distance measures dependent on the `_dtw_distance` function like `ddtw_distance`, `dtw_gi_distance` and also  `adtw`, `erp`, `edr` distances ignore the other channels **without warning**.

For example in the `dtw_distance`, for a given time step all the channels are passed to the `_univariate_squared_distance`  which takes min of the channels and computes the distance between them.

Image
Reply to https://github.com/aeon-toolkit/aeon/issues/2661#issuecomment-2799962773

Image
Image

Image
where 
```python
def _univariate_euclidean_distance(x: np.ndarray, y: np.ndarray) -> float:
    return np.sqrt(_univariate_squared_distance(x, y))
```

tanishy7777:
Even though currently we are simply ignoring the other channels in all the distance measures(this is the case for atleast the ones mentioned above). Would it be better to add warnings for this?
@SebastianSchmidl

SebastianSchmidl:
As the implementations will be used with the Numba JIT turned on in most cases, I would be more interested in the runtimes and scaling behavior of the aeon implementations with Numba. You do not need to include `sbd_original` and `sbd_tslearn`.

Could you also include the implementation idea in #2715 in the benchmark?

Especially the different usage of `with objmode(cc="float64[:, :]"):` with Numba could make a big difference in certain configurations.

tanishy7777:
> As the implementations will be used with the Numba JIT turned on in most cases, I would be more interested in the runtimes and scaling behavior of the aeon implementations with Numba. You do not need to include `sbd_original` and `sbd_tslearn`.
> 
> Could you also include the implementation idea in #2715 in the benchmark?
> 
> Especially the different usage of `with objmode(cc="float64[:, :]"):` with Numba could make a big difference in certain configurations.

Hmm makes sense, will enable numba and compare with #2715. Couple of assignments and project presentations lined up this weekend in college, but will add it soon.

tanishy7777:
@SebastianSchmidl Sorry for the delay, I have my semester exams and project submissions currently, I will be available from 1st May.

tanishy7777:
> Sorry to ask again for more, but could you scale the experiment further (larger time series)? According to the plots, all runs finish in under 1s.
> 
> Interesting that #2715 is faster than your proposal for two channels, but scales worse with the number of channels 🤔

Yea sure will add them. Actually for 2 channel my fix works better than #2715 
For univariate case only my fix is worse than #2715 but it is still better than the current implementation in aeon.

tanishy7777:
> Yes, I meant the single channel experiment.

Btw edited my message (added a para) in case you missed it

SebastianSchmidl:
It is not obvious how to best benchmark this because we have two scaling dimensions (length of time series and number of channels), which are not completely independent. You could create a full matrix of tests (all combinations of lengths and channels in the search range), but this could take a long time and is harder to plot. Maybe, we just need to increase the number of experiments, e.g. 5 length-scaling experiments with fixed #channels = [1,5,10,100,1000] and 5 channel-scaling experiments with fixed lengths = [1000,5000,10000,50000,100000] or similar

tanishy7777:
> It is not obvious how to best benchmark this because we have two scaling dimensions (length of time series and number of channels), which are not completely independent. You could create a full matrix of tests (all combinations of lengths and channels in the search range), but this could take a long time and is harder to plot. Maybe, we just need to increase the number of experiments, e.g. 5 length-scaling experiments with fixed #channels = [1,5,10,100,1000] and 5 channel-scaling experiments with fixed lengths = [1000,5000,10000,50000,100000] or similar ... still a lot to test

Yep that makes sense, will do this

aeon/distances/_sbd.py
```diff
@@ -104,11 +104,16 @@ def sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool = True) -> floa
...
+            nchannels = x.shape[0]  # both x and y have the same number of channels
+            norm = np.linalg.norm(x.astype(np.float64)) * np.linalg.norm(
+                y.astype(np.float64)
+            )
+            distance = np.zeros((2 * x.shape[1] - 1,))
             for i in range(nchannels):
-                distance += _univariate_sbd_distance(x[i], y[i], standardize)
-            return distance / nchannels
+                distance += _helper_sbd(x[i], y[i], standardize)
+            return np.abs(1 - np.max(distance) / norm)
```
MatthewMiddlehurst, line 116: Do other distances have this in a separate function like the univariate one?

aeon/distances/_sbd.py
```diff
@@ -240,8 +245,16 @@ def _univariate_sbd_distance(x: np.ndarray, y: np.ndarray, standardize: bool) ->
...
 
     b = np.sqrt(np.dot(x, x) * np.dot(y, y))
     return np.abs(1.0 - np.max(a / b))
+
+
+@njit(cache=True, fastmath=True)
+def _helper_sbd(x, y, standardize):
+
+    with objmode(a="float64[:]"):
+        a = correlate(x, y, method="fft")
```
MatthewMiddlehurst, line 258: not really a fan of this but it is the current functionality. Maybe worth an issue.

aeon/testing/expected_results/expected_distance_results.py
```diff
@@ -115,7 +115,7 @@
         0.6617308353925114,
         0.6617308353925114,
         0.5750093257763462,
-        0.5263609881742105,
+        None,
```
MatthewMiddlehurst, line 118: Why is this None? The comment in the file does not seem to line up with the values seen.

MatthewMiddlehurst (commented):
Just catching up really. Is the main issue here scalability or? @TonyBagnall @chrisholder possibly some issues re: the unequal stuff and results comparison

> Even though currently we are simply ignoring the other channels in all the distance measures(this is the case for atleast the ones mentioned above). Would it be better to add warnings for this?

I would open a separate issue on this.

SebastianSchmidl:
> Just catching up really. Is the main issue here scalability or? @TonyBagnall @chrisholder possibly some issues re: the unequal stuff and results comparison

The main issue was that the definition of the SBD distance included a multivariate version that was not implemented in aeon. aeon just computed the SBD for each channel independently and then combined the results (we do this for other distances as well).

However, multiple PRs (this one and #2715) were opened to address this issue. I suggested comparing their performance and integrating the better performing implementation into aeon.

SebastianSchmidl:
This version also seems to perform better than #2715, but let's wait for the new numbers from @tanishy7777. Once, your mentioned issues are addressed, I would also merge this one in favor of #2715 (I already added a dependency to the issue description for this).

TonyBagnall:
catching up a bit, I agree its important to have consistency with the other implementations and this solution seems good to me

MatthewMiddlehurst:
I have closed the other PR, is this still active @tanishy7777?

tanishy7777:
> I have closed the other PR, is this still active @tanishy7777?

Yea will work on this. I think only the benchmarks are remaining

tanishy7777:
> I still have a few comments above. More concerned with correctness that speed currently but if you can benchmark it thats good.

Got it, will address the comments first.

SebastianSchmidl:
I agree. The other implementation is stale, and the preliminary performance looked good. So, you can go ahead with resolving the issues and checking the correctness of your implementation.

aeon/distances/_sbd.py
SebastianSchmidl: Yeah, this was existing before the changes in this PR. I couldn't find a Numba-way to do this. We could dig into the math and re-implement the functionality in Numba with some effort.

TonyBagnall:
came to this through #3314 and noticed the same thing when implementing. I will benchmark with the existing implementation on the UCR, then look at these changes.

TonyBagnall:
this has been around for a while, got claude to look at it and has two blockers. @tanishy7777 do you still want to work on this? I'll convert to draft but feel free to bring back

The switch to summed per-channel cross-correlation normalised by the Frobenius norms is the right multivariate SBD (Paparrizos & Gravano), but two bugs need fixing before merge. Checked against a brute-force oracle (`scipy.signal.correlate(method="direct")`, per-channel z-norm, summed CC / (‖x‖_F‖y‖_F)).

**1. `standardize` has no effect for multivariate input.** `_helper_sbd` ignores it and `norm` uses the raw arrays. For `x, y ~ N(0,1)` of shape `(3, 20)`:

| | PR | oracle |
|---|---|---|
| `standardize=True` | 0.818756 | 0.825624 |
| `standardize=False` | 0.818756 | 0.818756 |

Scaling x's channels by (1, 100, 0.01) and adding 50 gives `sbd_distance(x_scaled, x) = 0.656`, when it should be 0.

**2. Unequal-length multivariate reads out of bounds.** The accumulator is `np.zeros(2 * x.shape[1] - 1)` but the correlation has length `n + m - 1`. numba doesn't enforce broadcasting on `+=`, so this either truncates lags or reads past the buffer, with no error. In 2000 random 2-channel pairs (lengths 3-29, `standardize=False`), 502 disagreed with the oracle, the worst by 3.4e275, e.g. lengths (12, 16): `d(a, b) = 0.4817`, `d(b, a) = 2.77e305`.

**Tests.** The equal-length "multivariate" case in `test_distances` uses one channel, so the new code path isn't exercised. Setting `unequal_support=False` works around the 5-vs-10-*channel* unequal test, but it also drops univariate unequal coverage and leaves bug 2 reachable. Please:
- keep `unequal_support=True` and size the accumulator `x.shape[1] + y.shape[1] - 1`
- z-normalise each channel (for both CC and norm) when `standardize=True`
- add a genuine multi-channel equal-length expected value, and make the unequal test use matching channel counts

**Minor:** update both docstrings (they still describe per-channel averaging and min-channel truncation), drop the trailing space in the ValueError message, remove the unused `standardize` arg from `_helper_sbd` (or inline it), and add a changelog note since results change for existing users.

tanishy7777:
Sure I will have a look at this over the weekend
@TonyBagnall

PR #2747, opened 2025-04-14
[MNT] Use GPU installs for periodic tests

MatthewMiddlehurst:
In PRs we use the CPU version of `torch`/`tensorflow` for faster installs. Time spent is not as important for the periodic tests, so it would be better to test the regular install use case here. It is possible that there could be subtle differences between these installs i.e. versioning for other dependencies.

May end up reverting if it turns out to be more pain than its worth.


TonyBagnall (approved):
lgtm

PR #2540, opened 2025-02-16
[ENH] Test coverage for AEAttentionBiGRUNetwork Improved

lucifer4073:
#### Reference Issues/PRs

Contributes to #2497. Reference PRs #2518 and #2534.

#### What does this implement/fix? Explain your changes.

The given issue improvises the test coverage of AEAttentionBiGRUNetwork class.

#### Does your contribution introduce a new dependency? If yes, which one?

No.

#### Any other comments?

No.

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you after the PR has been merged.
- [X] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #2771, opened 2025-05-02
[BUG] Fix bug of shape variable assignment in AEAttentionBiGRUNetwork

### LINKED ISSUE #2560: [BUG] UnboundLocalError in AEAttentionBiGRUNetwork When temporal_latent_space=True
### Describe the bug

When initializing `AEAttentionBiGRUNetwork` with `temporal_latent_space=True`, calling `build_network` results in an `UnboundLocalError`. The issue occurs because the variable `shape_before_flatten` is referenced but never assigned a value when `temporal_latent_space=True`. The issue does not occur when `temporal_latent_space=False`.  

```python
from aeon.networks import AEAttentionBiGRUNetwork  

# Initialize with temporal_latent_space=True
aeattentionbigru = AEAttentionBiGRUNetwork(latent_space_dim=64, temporal_latent_space=True)  
aeattentionbigru.build_network((1000, 5))  # This causes the error
```

### Expected results

The `build_network` function should execute without errors when `temporal_latent_space=True`, correctly constructing the network.

### Actual results

```python 
  File "<stdin>", line 1, in <module>
line 177, in build_network
    shape=shape_before_flatten, name="decoder_input"
          ^^^^^^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'shape_before_flatten' where it is not associated with a value
```
### Versions

_No response_

hadifawaz1999:
Fix #2560

### DIFF
--- a/aeon/networks/_ae_abgru.py
+++ b/aeon/networks/_ae_abgru.py
@@ -161,6 +161,7 @@ class AEAttentionBiGRUNetwork(BaseDeepLearningNetwork):
             x = tf.keras.layers.Flatten()(x)
             x = tf.keras.layers.Dense(self.latent_space_dim)(x)
         elif self.temporal_latent_space:
+            shape_before_flatten = x.shape[1:]
             x = tf.keras.layers.Conv1D(filters=self.latent_space_dim, kernel_size=1)(x)
 
         encoder = tf.keras.models.Model(inputs=input_layer, outputs=x, name="encoder")


MatthewMiddlehurst:
Would be nice to have a test if this is still not covered.

hadifawaz1999:
> Would be nice to have a test if this is still not covered.

@lucifer4073 made a PR for covering this network, so i mentioned this on #2540 that when we get this bug fix in we update that PR and see if it passes

hadifawaz1999:
> > Would be nice to have a test if this is still not covered.
> 
> yes @MatthewMiddlehurst , @lucifer4073 made a PR for covering this network, so i mentioned this on #2540 that when we get this bug fix in we update that PR and see if it passes

TonyBagnall (approved):

PR #2783, opened 2025-05-08
[REF] Refactor smoothing transformers into a sub-package

MatthewMiddlehurst:

#1920 implemented a set of smoothing filters. This PR does some tidying, refactors them into their own sub-package with more intuitive names and deprecates the old versions.

A lot of the old acronyms were not commonly used I believe, the new names and locations should make them more recognisable.  Big offender was `DFTSeriesTransformer` which does an IFFT at the end. Bit confusing for someone who just wants to do a regular DFT.

Feel free to suggest alternatives. FYI @Cyril-Meyer @Datadote


Cyril-Meyer:
@MatthewMiddlehurst good idea, the proposed new names looked good to me.
The only thing I do not find perfect is having 'smoothing' in the name when already in a 'smoothing' package (`ExponentialSmoothing`) but I do not have better alternative.

Datadote:
@MatthewMiddlehurst  Looks good to me.

TonyBagnall (approved):
LGTM

Issue #2713, opened 2025-03-31
[ENH] Replace argsort by argpartition in similarity search


baraline:
### Describe the feature or idea you want to propose

After review with Patrick, to do after #2473 is merged.

Replace the use of argsort by argpartition in files such as `series/_commons.py` (e.g. `_extract_top_k_from_dist_profile`) to speed up the functions using those.

### Describe your proposed solution

Replace argsort by argpartition and adjust the algorithms, notably for the successive call to argpartition in cases where trivial matches should be avoided.

Kaustbh:
Hi @baraline, I would like to work on this issue. Can I start working from that branch and rebase it once the PR gets merged?

baraline:
Hi, if by that branch you mean the branch from 2473 yes you can.

Kaustbh:
Yes, from 2473. I will start working.

Kaustbh:
@aeon-actions-bot assign @Kaustbh

Kaustbh:
Hi, I have made the changes just waiting for the PR #2473 to get merged.

baraline:
Hey @Kaustbh, PR is in, you can raise yours, thanks !

baraline:
Hi, great question. This is should happen when the number of admissibles matches is less than k. Which can happen with small series and use of trival matches to false. 
Did you have a specific example where the behaviour was suspicious?

Kaustbh:
Thank you for the clarification! Yes, that makes sense. It can happen because of small series, can it also happen if we gave large value for `length` parameter?

Kaustbh:
@baraline, I have made the PR Please review it and let me know if any changes or improvements are needed.

PR #2805, opened 2025-05-14
Use `np.argpartition` for efficient top-k selection instead of `np.argsort`

Kaustbh:
#### Reference Issues/PRs

Fixes: #2713

This PR replaces `np.argsort` with `np.argpartition` in cases where only the top-k smallest elements are needed, improving performance on large arrays.

##### For all contributions

### DIFF
--- a/aeon/similarity_search/series/_commons.py
+++ b/aeon/similarity_search/series/_commons.py
@@ -137,8 +137,10 @@ def _extract_top_k_from_dist_profile(
     top_k_distances = np.full(k, np.inf, dtype=np.float64)
     ub = np.full(k, np.inf)
     lb = np.full(k, -1.0)
-    # Could be optimized by using argpartition
-    sorted_indexes = np.argsort(dist_profile)
+
+    k = min(k, len(dist_profile))
+    partitioned = np.argpartition(dist_profile, k)[:k]
+    sorted_indexes = partitioned[np.argsort(dist_profile[partitioned])]
     _current_k = 0
     if not allow_trivial_matches:
         _current_j = 0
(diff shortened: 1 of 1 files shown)


baraline (changes requested):
Test are failing due to an issue with the implementation, currently, `sorted_index` is only keeping the top-`k` indexes of `dist_profile`. When not allowing trivial matches, it is possible that the top-`k` matches are not included in the first `k` matches.

The original issue mention "adjust the algorithms, notably for the successive call to argpartition in cases where trivial matches should be avoided."

The idea is to try to find these `k` best match in the first argpartition, and look into another `k` argpartition, until you find `k` matches or checked all values of `dist profile`. 

Am I clear or is this a bit confusion ?

Kaustbh:
Sorry for the earlier misunderstanding. Our main goal is to find `k` valid matches. If we find fewer than `k`, we continue searching in the next batch of `k` elements using `argpartition`, and repeat this until we either find all `k` matches or reach the end of `dist_profile`. Am I right?

baraline:
No problem, this is quite a challenging topic. Yes I think you got it. We want to output `k` matches, but if for any reason (length of the distance profile or usage of trivial matches causing a shortage of admissible matches), we cannot find `k`, we will return less.

baraline:
Thanks for the changes, I'll run a benchmark to compare performance before making the decision, but it looks OK at first glance.

baraline:
Bit busy until 15th of june, i'll take a look at it after, thx 🙏

Kaustbh:
Okay, no problem

baraline (approved):
Finally had a change to benchmark it, I confirm that this reduces runtime for large inputs, and the difference is negligible for small ones. Great job, thanks again !

PR #2808, opened 2025-05-15
[MNT] Lower bound keras to discard old bugged versions of early keras3

hadifawaz1999:
The beginning of keras3 had lots of bugs, logical because of the huge change, but i prefer adding a lower bound, as the more i use older version than 3.6.0 the more i notice bugs that can lead to false research results

Example:

with keras 3.0 it discards the "trainable" boolean of a tensorflow operation, making for instance LITE work differently (unreported issue).

This is to archive for future references

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,12 +70,14 @@ all_extras = [
     "statsmodels>=0.12.1",
     "stumpy>=1.5.1",
     "tensorflow>=2.14; python_version < '3.13'",
+    "keras>=3.6.0; python_version < '3.13'",
     "torch>=1.13.1",
     "tsfresh>=0.20.0",
     "tslearn>=0.5.2",
 ]
 dl = [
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst (commented):
is 3.6 the latest version? Don't mind the change but if there are some previous non-buggy versions would be nice to allow those

hadifawaz1999:
> is 3.6 the latest version? Don't mind the change but if there are some previous non-buggy versions would be nice to allow those

the choice for 3.6 is because 3.4 and 3.5 had bugs with saving and loading (if you recall about this, we added an upper bound back then) so 3.6 seems the reasonable choice for me, its also the one i use locally so i dont seem to have issues with it

MatthewMiddlehurst (approved):
sounds fine then. latest is 3.9 so its not like we are limited to a single version.

PR #2812, opened 2025-05-17
[ENH] Series transformer stuff

MatthewMiddlehurst:
todo


aeon/transformations/series/_collection_wrapper.py
```diff
@@ -0,0 +1,103 @@
...
+        tags_to_keep = CollectionToSeriesWrapper._tags
+        tags_to_add = transformer.get_tags()
+        for key in tags_to_keep:
+            tags_to_add.pop(key, None)
+        for key in ["capability:unequal_length", "removes_unequal_length"]:
+            tags_to_add.pop(key, None)
+        self.set_tags(**tags_to_add)
+
+    def _fit(self, X, y=None):
+        X = X.reshape(1, X.shape[0], X.shape[1])
```
baraline, line 56: I'm a bit paranoid on reshape behaviours (prefer to use `X[np.newaxis, :]` in those case), But I think it should work fine here and in the other places

aeon/transformations/base.py
baraline, line 1: Do we still need the fit_transform abstract function here as we inherit from `TransformerMixin` ? I'm unsure how it works when the child class define an abstract method from a class it inherit from which has it defined.

aeon/transformations/series/_collection_wrapper.py
TonyBagnall: I think its fine, because X must be a 2D array (n_channels, n_timepoints)

TonyBagnall (approved):

aeon/transformations/series/_collection_wrapper.py
MatthewMiddlehurst: yeah can understand the concern generally but I think in this case it is pretty safe. Will all be converted or caught earlier than this.

aeon/transformations/base.py
MatthewMiddlehurst: To be honest I am unsure as well, but I would assume that it makes the mixin version unusable which is what I would want. It does some routing stuff that we dont support and I dont want to deal with 🙂.

https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7/sklearn/base.py#L839

The main thing we want is the tags and support for `isinstance` when using sklearn utilities.

MatthewMiddlehurst:
Should be ready @TonyBagnall @baraline

Issue #2753, opened 2025-04-17
[ENH] TS2Vec implementation


gasperpetelin:
### Describe the feature or idea you want to propose

It might be beneficial to include an implementation of **TS2Vec** (Towards Universal Representation of Time Series) in the library. TS2Vec is a popular and efficient contrastive learning approach for time series representation. It could be implemented as a transformer similar to Rocket or TSFresh. The original author's code is available [here](https://github.com/zhihanyue/ts2vec) (MIT licence). I would be happy to add the implementation if this is something the maintainers/developers of aeon agree should be included and if it aligns with future development.

### Describe your proposed solution

I believe the implementation could be similar to Rocket. At first, it could support only `encoding_window='full_series'`, so the returned array would always have the shape `n_instances x output_dim`, primarily useful for classification.

Example: 
```
trf = TS2Vec(output_dim=320)
trf.fit(X_train)
X_train = trf.transform(X_train)
X_test = trf.transform(X_test)
```

### Additional context

Classification accuracy of **TS2Vec** compared to 2 other randomly selected TSC models 
|   TS2Vec |   Catch22 |   MultiRocket | dataset                        |
|---------:|----------:|--------------:|:-------------------------------|
| 0.84     |  0.85     |      0.9      | ACSF1                          |
| 0.565217 |  0.734015 |      0.828645 | Adiac                          |
| 0.788571 |  0.748571 |      0.874286 | ArrowHead                      |
| 0.986667 |  0.913333 |      1        | BME                            |
| 0.766667 |  0.6      |      0.766667 | Beef                           |
| 0.8      |  0.75     |      0.85     | BeetleFly                      |
| 0.8      |  0.9      |      0.9      | BirdChicken                    |
| 1        |  0.943333 |      0.995556 | CBF                            |
| 0.833333 |  0.766667 |      0.933333 | Car                            |
| 0.982507 |  0.900875 |      0.976676 | Chinatown                      |
| 0.590104 |  0.603906 |      0.7875   | ChlorineConcentration          |
| 0.83913  |  0.792754 |      0.95     | CinCECGTorso                   |
| 1        |  1        |      1        | Coffee                         |
| 0.636    |  0.732    |      0.804    | Computers                      |
| 0.766667 |  0.582051 |      0.797436 | CricketX                       |
| 0.75641  |  0.55641  |      0.851282 | CricketY                       |
| 0.776923 |  0.635897 |      0.828205 | CricketZ                       |
[...]

PR #2825, opened 2025-05-19
[ENH] Implementation of TS2Vec

gasperpetelin:
The first approach for implementing TS2Vec is still a work in progress.

Usage:
```python
from aeon.transformations.collection.contrastive_based import TS2Vec
X, _ = load_unit_test(split="train")

# Select the device
transformer = TS2Vec(device="cpu") # or
transformer = TS2Vec(device="cuda") # or
transformer = TS2Vec(device="cuda:0")

transformer.fit_transform(X)
```

#### Reference Issues/PRs

Implements TS2Vec as discussed in #2753 

Implements TS2Vec

No new dependencies. The code is mostly based on Pytorch [from original author](https://github.com/zhihanyue/ts2vec )

#### Any other comments?

The original code has MIT licence

##### For all contributions

### DIFF
--- a/aeon/networks/_ts2vec.py
+++ b/aeon/networks/_ts2vec.py
@@ -1,0 +1,173 @@
+"""Fully Convolutional Network (FCNNetwork)."""
+
+__maintainer__ = ["GasperPetelin"]
+
+
+class TS2VecNetwork:
+    def __init__(
+        self,
+        input_dims,
+        output_dims,
+        hidden_dims,
+        depth,
+    ):
+        self.input_dims = input_dims
+        self.output_dims = output_dims
+        self.hidden_dims = hidden_dims
+        self.depth = depth
+        super().__init__()
+
+    def build_network(self, input_shape, **kwargs):
+        import numpy as np
+        import torch
+        import torch.nn.functional as F
+        from torch import nn
+
+        class ConvBlock(nn.Module):
+            def __init__(
+                self, in_channels, out_channels, kernel_size, dilation, final=False
+            ):
+                super().__init__()
+                self.conv1 = SamePadConv(
+                    in_channels, out_channels, kernel_size, dilation=dilation
+                )
+                self.conv2 = SamePadConv(
+                    out_channels, out_channels, kernel_size, dilation=dilation
+                )
+                self.projector = (
+                    nn.Conv1d(in_channels, out_channels, 1)
+                    if in_channels != out_channels or final
+                    else None
+                )
+
+            def forward(self, x):
+                residual = x if self.projector is None else self.projector(x)
+                x = F.gelu(x)
+                x = self.conv1(x)
+                x = F.gelu(x)
+                x = self.conv2(x)
+                return x + residual
+
+        class DilatedConvEncoder(nn.Module):
+            def __init__(self, in_channels, channels, kernel_size):
+                super().__init__()
+                self.net = nn.Sequential(
+                    *[
+                        ConvBlock(
+                            channels[i - 1] if i > 0 else in_channels,
+                            channels[i],
+                            kernel_size=kernel_size,
+                            dilation=2**i,
+                            final=(i == len(channels) - 1),
+                        )
+                        for i in range(len(channels))
+                    ]
+                )
+
+            def forward(self, x):
+                return self.net(x)
+
+        class SamePadConv(nn.Module):
+            def __init__(
+                self, in_channels, out_channels, kernel_size, dilation=1, groups=1
+            ):
+                super().__init__()
+                self.receptive_field = (kernel_size - 1) * dilation + 1
+                padding = self.receptive_field // 2
+                self.conv = nn.Conv1d(
+                    in_channels,
+                    out_channels,
+                    kernel_size,
+                    padding=padding,
+                    dilation=dilation,
+                    groups=groups,
+                )
+                self.remove = 1 if self.receptive_field % 2 == 0 else 0
+
+            def forward(self, x):
+                out = self.conv(x)
+                if self.remove > 0:
+                    out = out[:, :, : -self.remove]
+                return out
+
+        class TSEncoder(nn.Module):
+            def __init__(
+                self,
+                input_dims,
+                output_dims,
+                hidden_dims=64,
+                depth=10,
+                mask_mode="binomial",
+            ):
+                super().__init__()
+                self.input_dims = input_dims
+                self.output_dims = output_dims
+                self.hidden_dims = hidden_dims
+                self.mask_mode = mask_mode
+                self.input_fc = nn.Linear(input_dims, hidden_dims)
+                self.feature_extractor = DilatedConvEncoder(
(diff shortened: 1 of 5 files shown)


MatthewMiddlehurst:
Testing with soft dependencies also needs to be isolated i.e. https://github.com/aeon-toolkit/aeon/blob/main/aeon/classification/feature_based/tests/test_tsfresh.py

TonyBagnall:
from my initial look, it seems that this transformer outputs a multivariate time series of embeddings, so input a collection of shape (n_cases, n_channels, n_timepoints) to outputs (n_cases, n_embeddings, n_timepoints), is that correct? If so, the output type is not tabular, it is a collection of transformed time series. Have I understood correctly? 

Also, our estimators generally dont work with tensors. Could you format the output to by a 3D numpy array (n_cases, n_embeddings, n_timepoints),? 

I'd like to try this for classification, could be interesting.

hadifawaz1999:
Waiting on the self supervised PR to get in then will review this, @TonyBagnall check the self supervised PR if you want to know more about how the structure of the module is gonna be for now (#2385 )

TonyBagnall:
hi @gasperpetelin https://github.com/aeon-toolkit/aeon/pull/2385 is merged, it would be great if you could get this in

gasperpetelin:
Hi @TonyBagnall. I’ve now moved my code to the self_supervised module. However, I’m not sure why some tests are failing. It looks like they are trying to connect to `timeseriesclassification.com` but are failing because the website is currently unreachable (not 100% sure if there is not another reason).

hadifawaz1999 (changes requested):
Gonna make some general comments before starting to get into details with this:

- The setup we're following in self supervised is that the model should take a network as input parameter (see what TRILITE does) from the networks module of aeon (aeon/networks)
- The only thing the self supervised model should do is the training part, nothing related to the neural networks should be mentioned here.
- The same fit mechanism of TRILITE should be followed, meaning callbacks etc. (see fit function of TRILITE)
- Saving and loading of best model should be followed same as TRILITE
- The implementation for now is supporting tensorfow keras backend, when removing the network related stuff in the model implementaiton the fit function should look like the one in TRILITE, if the model uses a custom loss then it can be implemented as a private function of the class
- Most important thing is: keeping consistency with TRILITE

gasperpetelin:
@hadifawaz1999 I just have 3 questions regarding the implementation:

- I’ve moved the TS2Vec neural network definition to `aeon/networks`. However, TS2Vec internally uses PyTorch, while all other networks in that directory are implemented using TensorFlow. This creates a potential issue: if a user passes a TS2Vec network to TRILITE, it would fail. The same would happen if someone passes a TensorFlow-based network like FCNNetwork to TS2Vec. So, if we're placing networks in the same module, should compatibility checks be implemented to prevent users from mixing incompatible components?

- Networks in `aeon/networks` currently inherit from `BaseDeepLearningNetwork`, which is designed for TensorFlow and expects specific methods, like returning the first and last layers. Since TS2Vec uses PyTorch, it can't inherit from this base class without making the PyTorch code somewhat hacky to conform to the expected interface. Should PyTorch code therefore be in `aeon/networks`?

- TensorFlow/Keras uses callbacks. To my knowledge, PyTorch does not have an equivalent standard (except PyTorch Lightning). Is it okay if the callback parameter simply accepts a regular Python function aka Callable (or a list of functions)?
For example:
```
after_iter_callbacks: Callable | list[Callable] | None = None
after_epoch_callbacks: Callable | list[Callable] | None = None
```

hadifawaz1999:
> @hadifawaz1999 I just have 3 questions regarding the implementation:
> 
>     * I’ve moved the TS2Vec neural network definition to `aeon/networks`. However, TS2Vec internally uses PyTorch, while all other networks in that directory are implemented using TensorFlow. This creates a potential issue: if a user passes a TS2Vec network to TRILITE, it would fail. The same would happen if someone passes a TensorFlow-based network like FCNNetwork to TS2Vec. So, if we're placing networks in the same module, should compatibility checks be implemented to prevent users from mixing incompatible components?
> 
>     * Networks in `aeon/networks` currently inherit from `BaseDeepLearningNetwork`, which is designed for TensorFlow and expects specific methods, like returning the first and last layers. Since TS2Vec uses PyTorch, it can't inherit from this base class without making the PyTorch code somewhat hacky to conform to the expected interface. Should PyTorch code therefore be in `aeon/networks`?
> 
>     * TensorFlow/Keras uses callbacks. To my knowledge, PyTorch does not have an equivalent standard (except PyTorch Lightning). Is it okay if the callback parameter simply accepts a regular Python function aka Callable (or a list of functions)?
>       For example:
> 
> 
> ```
> after_iter_callbacks: Callable | list[Callable] | None = None
> after_epoch_callbacks: Callable | list[Callable] | None = None
> ```

The model should be re-implemented in tensorflow keras for the network and training strategy

gasperpetelin:
Regarding the TensorFlow reimplementation: I’ll give it a try, but to be honest, I have significantly more experience with PyTorch than with TensorFlow/Keras. Because of that, implementing TS2Vec in TensorFlow is not something I can do easily or quickly.

TS2Vec has a non-trivial architecture and training procedure. Additionally, the two frameworks differ in several important aspects (for example, PyTorch's `AveragedModel`, handling of padding or NaNs, and weight initialization), which can lead to degraded performance even when the implementations appear almost identical. So while I’ll do my best, I can’t promise a fully equivalent TensorFlow version in the near future.

Just to confirm, TS2Vec would only be accepted as a TensorFlow/Keras implementation and not as a PyTorch version (similar to how HydraTransformer is handled)?

hadifawaz1999:
> Regarding the TensorFlow reimplementation: I’ll give it a try, but to be honest, I have significantly more experience with PyTorch than with TensorFlow/Keras. Because of that, implementing TS2Vec in TensorFlow is not something I can do easily or quickly.
> 
> TS2Vec has a non-trivial architecture and training procedure. Additionally, the two frameworks differ in several important aspects (for example, PyTorch's `AveragedModel`, handling of padding or NaNs, and weight initialization), which can lead to degraded performance even when the implementations appear almost identical. So while I’ll do my best, I can’t promise a fully equivalent TensorFlow version in the near future.
> 
> Just to confirm, TS2Vec would only be accepted as a TensorFlow/Keras implementation and not as a PyTorch version (similar to how HydraTransformer is handled)?

Yes the implementaiton should be in tensorflow-keras. The network itself should be easilly adapted, i dont really believe a model will be impacted by the initialization of the weights if you had to use another method, because if a model is impacted its really a bad sign on the model itself.

Padding exists in keras as well so there is no issue. Anything can be adapted easilly, there is no impossible way.

aeon supports the use of torch in other modules, but the deep learning framework no it follows a tensorflow backend and keras for the networks module

TonyBagnall:
hi @gasperpetelin in the spirit of housekeeping I am converting this to draft, feel free to reopen if you want to continue with it

TonyBagnall:
closing now, conficted  feel free to reopen if you want to continue with it

PR #2827, opened 2025-05-20
[ENH]: Add kdtw distance implementation

SebastianSchmidl:

Adds KDTW distance (univariate, multivariate, and variable-length support)

Open todos:

- [ ] current implementation is a similarity. How to convert to distance?
- [ ] generate expected values using original Octave implementation and fix tests
- [ ] replace `kdtw` kernel in KernelKMeans implementation

no


aeon/distances/kernel/_kdtw.py
```diff
@@ -0,0 +1,482 @@
...
+    if normalize_dist:
+        self_x = _kdtw_cost_matrix(x, x, gamma, epsilon, normalize_input)[n, n]
+        self_y = _kdtw_cost_matrix(y, y, gamma, epsilon, normalize_input)[m, m]
+        norm_factor = np.sqrt(self_x * self_y)
+    else:
+        norm_factor = 1.0
+    return (
+        1.0
+        - _kdtw_cost_matrix(x, y, gamma, epsilon, normalize_input)[n, m] / norm_factor
+    )
```
SebastianSchmidl, line 196: This similarity-to-distance conversion part is problematic (if no input normalization is used) because the computed similarities are very, very small for most inputs and, thus, vanish.

E.g., all the following similarities will disappear in floating-point arithmetic (first two basic motions time series with different gammas):
- `1.0 - 2.7435e-123 == 1.0`
- `1.0 - 4.9180e-72 == 1.0`
- `1.0 - 6.5392e-53 == 1.0`

SebastianSchmidl:
Do we want to include this distance in aeon? It does not perform super well and has numerical stability issues. However, it was used in a couple of publications, and would be nice for comparisons.

MatthewMiddlehurst:
If we can get it functioning with CI and stable I do not see why not. May be best to ask @chrisholder or @TonyBagnall when it comes to formatting in the module im not so familiar with it!

aeon/distances/kernel/_kdtw.py
MatthewMiddlehurst: Thats a bit troublesome. Is it possible to just scale it up or is that going to cause a bunch of issues. If thats fundamentally changing the algorithm it may be wise to just avoid it 🙂

PR #2830, opened 2025-05-22
[ENH] Series transformer pipeline and datatype list tidy

MatthewMiddlehurst:

Implements a pipeline for series transforms, including an Id transform and testing. Also:

- Tidies up and structures the datatypes lists. These should now all be located in `utils`. Most of the file changes are refactoring this
- For some reason `LogTransformer` was in the boxcox file and unavailable locally, I have separated this and included it in the init as i use it for examples


TonyBagnall (dismissed):
its a big un, looked through it and it all looks good

PR #2828, opened 2025-05-20
[ENH] AutoETS, ARIMA, NaiveForecaster, Dataset loading updates, Differencing/Windowing/Train test transforms.

alexbanwell1:
#### What does this implement/fix? Explain your changes.
AutoETS, ARIMA, NaiveForecaster, Dataset loading updates, Differencing/Windowing/Train test transforms.

#### Does your contribution introduce a new dependency? If yes, which one?
No

#### Any other comments?
Likely not tidy enough to go into main aeon branch, but needs to be reviewed/cleaned up to merge in.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you **after** the PR has been merged.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

PR #2838, opened 2025-05-24
[ENH] Introduce AutoSARIMA forecaster

TonyBagnall:
refactoring of code from #2828 

@alexbanwell1 currently ARIMA seems to automatically include seasonal components (SARIMA). I would like to separate these out.


aeon/forecasting/_arima.py
```diff
@@ -0,0 +1,492 @@
...
+    lags = np.concatenate((np.array([1.0]), lags))
+    peaks = []
+    mean_lags = np.mean(lags)
+    for i in range(1, len(lags) - 1):  # Skip the first (lag 0) and last elements
+        if lags[i] >= lags[i - 1] and lags[i] >= lags[i + 1] and lags[i] > mean_lags:
+            peaks.append(i)
+    if not peaks:
+        return 1
+    else:
+        return peaks[0]
```
alexbanwell1, line 492: I feel like this should be somewhere else, it is also used in the AutoETS implementation, and so, to reduce code duplication, should probably be in a file somewhere else, it's also more generally applicable as well. I guess the same could be said about the nelder-mead optimisation algorithm too, but that's slightly less easy to split out

alexbanwell1 (commented):
Otherwise looks like the code I wrote, so should be fine!

alexbanwell1:
Now implements AutoSARIMA Forecaster.
Based on #2860 #2861 and #2862
Final section to implement #2848

TonyBagnall:
superseded

TonyBagnall:
will need to revisit this once ARIMA is in

Issue #2179, opened 2024-10-11
[BUG] freeze on import of aeon.classification.convolution_based


user706:
### Describe the bug

Hi, 
when doing 

```python
from aeon.classification.convolution_based import Arsenal
```

it takes forever.

- It was working a few days back. In fact in the old venv it still works. (**Edit:** might be that the 1st loading always takes long. [Not quite sure how long it should take...])
- But when trying a **new venv** (see below -> reproduce) it **takes forever**. Why?

### my system

* `Ubuntu 24.04.1 LTS` (according to `cat /etc/os-release`)
* `Python 3.12.3` (according to `python --version` after starting the venv (see below))

### reproduce (or does it work on your side?)

```bash
# in the shell (I'm using bash on linux)
python3 -m venv pyvenv
. pyvenv/bin/activate
pip install --upgrade pip
pip install -U aeon
python -c "print('loading'); from aeon.classification.convolution_based import Arsenal; print('done')" # takes forever
```

Any ideas?

### details

stopping the freeze by hitting Ctrl-C, results in 

<details>
<summary>click for details</summary>

```
KeyboardInterrupt: 
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/dude/proj/test_aeon/pyvenv/lib/python3.12/site-packages/aeon/classification/convolution_based/__init__.py", line 10, in <module>
    from aeon.classification.convolution_based._arsenal import Arsenal
  File "/home/dude/proj/test_aeon/pyvenv/lib/python3.12/site-packages/aeon/classification/convolution_based/_arsenal.py", line 20, in <module>
    from aeon.transformations.collection.convolution_based import (
  File "/home/dude/proj/test_aeon/pyvenv/lib/python3.12/site-packages/aeon/transformations/collection/__init__.py", line 33, in <module>
    from aeon.transformations.collection.channel_selection import (
  File "/home/dude/proj/test_aeon/pyvenv/lib/python3.12/site-packages/aeon/transformations/collection/channel_selection/__init__.py", line 16, in <module>
    from aeon.transformations.collection.channel_selection._channel_scorer import (
  File "/home/dude/proj/test_aeon/pyvenv/lib/python3.12/site-packages/aeon/transformations/collection/channel_selection/_channel_scorer.py", line 7, in <module>
    from aeon.classification.convolution_based._rocket_classifier import RocketClassifier
[...]

PR #2858, opened 2025-05-28
[BUG] Remove typing from numba functions

TonyBagnall:
Fixes https://github.com/aeon-toolkit/aeon/issues/2179

removes the typing from numba, which was slowing down the cache building considerably, see issue for discussion

seems I have corrupted my local main branch again, ignore the arima first commit


TonyBagnall:
> Did you have any benchmarks on before vs after ? Would have been nice to understand why it caused it, but who knows... Otherwise LGTM.

I didnt formally time it, just turned it off an on, difference was noticeable.

Issue #2224, opened 2024-10-18
[BUG] PLAID Dataset is being loaded as unequal always


hadifawaz1999:
### Describe the bug

When loading the PLAID dataset (originally unequal length) with `load_classification` it is always loading the list version  never the equal length version

### Steps/Code to reproduce the bug

```python
x, y = load_classification(name="PLAID")
print(x.shape)
```

### Expected results

shape output

### Actual results

Attribute error, list does not have attribute shape

lucifer4073:
Hi @hadifawaz1999, I would like to work on this issue. Can you please assign it to me?

MatthewMiddlehurst:
@lucifer4073 Please have a look through the contributing guide. This issue involved loading external data so theres a chance you wont be able to fix this.

lucifer4073:
Hi @MatthewMiddlehurst, thank you for your response. I attempted to debug the issue and found that x has an inhomogeneous shape, which prevents it from being converted into a NumPy array. This seems to stem from the external data being loaded.

```
Type of x -> <class 'list'>
Type of y -> <class 'numpy.ndarray'>
Type of x[0] -> <class 'numpy.ndarray'>
Type of x[0][0] -><class 'numpy.ndarray'>
```
I’ve attached this for reference in case someone else wants to review or provide further insights. 

Attempt: `x=np.asarray(x)`

Result:  `ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (1074, 1) + inhomogeneous part.`

MatthewMiddlehurst:
The function has a `load_equal_length` parameter which should handle this, but I don't think its an issue with the function itself.

hadifawaz1999:
@TonyBagnall what do you think

saadaltohamy:
Hi, I found that the Plaid dataset doesn't have _disc, _eq, or _nmv versions. So when calling _load_saved_dataset(), it returns a list of 2D arrays due to variable-length multivariate series.

> Shape of the nested list:
> (1074,)
> Shapes of the first 5 elements in X:
> (1, 500)
> (1, 500)
> (1, 500)
> (1, 200)
> (1, 544)

Trying to convert this list to a 3D NumPy array with _from_np_list_to_numpy3d() raises:
ValueError: Cannot convert unequal length to numpy3D.

[...]

PR #2859, opened 2025-05-28
[ENH] Refactor example dataset for unequal length univariate

TonyBagnall:
this PR deprecates the PLAID baked in dataset and adds PickupGestureWiimoteZ to represent unequal length univariate series. 

reason for this is 
1. PLAID is too big, and the package size is creeping up agains
2. The current files do not have the equal length version. To put them in (as with JapaneseVowels) would increase size further. Not having them causes a bug see #2224 

once deprecated, #2224 can be closed, but will remain a bug until then.

### DIFF
--- a/aeon/datasets/_single_problem_loaders.py
+++ b/aeon/datasets/_single_problem_loaders.py
@@ -381,10 +381,10 @@ def load_pickup_gesture_wiimoteZ(split=None):
 
 # TODO: remove in v1.3.0
 @deprecated(
-    version="1.1.0",
+    version="1.2.0",
     reason="load_plaid and the PLAID dataset will be removed in version 1.3. It has "
     "been replaced by a smaller unequal length univariate problem "
-    "PickupGestureWiimote. ",
+    "PickupGestureWiimoteZ.",
     category=FutureWarning,
 )
 def load_plaid(split=None):


hadifawaz1999:
would this not raise the same issue for the new dataset as what was happening with Plaid ? If yes we should add a warning in load_classificaiton for this one saying that it will always load the list version not equal length

TonyBagnall:
> would this not raise the same issue for the new dataset as what was happening with Plaid ? If yes we should add a warning in load_classificaiton for this one saying that it will always load the list version not equal length

no it wont, because the problem is small enough to include the equal length version in the datasets directory. We decided avoid doubling the PLAID directory size by incluidng PLAID_eq_TEST/TRAIN, hence the change

aeon/datasets/_single_problem_loaders.py
```diff
@@ -344,6 +345,48 @@ def load_basic_motions(split=None, return_type="numpy3d"):
...
+    """
+    return _load_tsc_dataset("PickupGestureWiimoteZ", split, return_type="np-list")
+
+
+# TODO: remove in v1.3.0
+@deprecated(
+    version="1.1.0",
+    reason="load_plaid and the PLAID dataset will be removed in version 1.3. It has "
+    "been replaced by a smaller unequal length univariate problem "
+    "PickupGestureWiimote. ",
```
MatthewMiddlehurst: 1.2.0 and should be "PickupGestureWiimoteZ" rather than PickupGestureWiimote probably?

MatthewMiddlehurst (approved):

PR #2876, opened 2025-06-04
[ENH] Refactor forecast predict to return a single value

TonyBagnall:
I think we should change the model for a Forecaster, make it as simple as possible to start with then consider more complex scenarios. Points made on slack, but basis is forecasters train on a 
```python
model = MyForecaster(horizon=1)
model.fit(y)
y_hat = model.predict()
```
predict aways returns the same value, and `y_hat` is the one ahead prediction.


TonyBagnall:
closing for now, just experimental, will wait until the recursive and direct are in

PR #2877, opened 2025-06-04
[MNT] Update Slack link

MatthewMiddlehurst:
Refreshing this since I got a message regarding it. Could someone else check the old and new link to see if they work?


TonyBagnall (approved):
invite links time out I think, not sure if you can make a persistent one

PR #2880, opened 2025-06-05
[ENH] Add a recursive forecasting function

TonyBagnall:
dependent on #2410 adds a function for recursive forecasting

### DIFF
--- a/aeon/forecasting/_recursive.py
+++ b/aeon/forecasting/_recursive.py
@@ -5,7 +5,9 @@ import numpy as np
 from aeon.forecasting.base import BaseForecaster
 
 
-def recursive_forecasting(forecaster_class, y: np.ndarray, steps_ahead: int, exog=None):
+def recursive_forecasting(
+    forecaster: BaseForecaster, y: np.ndarray, steps_ahead: int, exog=None
+):
     """
     Forecast ``steps_ahead`` using a BaseForecaster trained on y then recursively .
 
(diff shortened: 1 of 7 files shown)


aeon/forecasting/_recursive.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    if isinstance(forecaster, type):
+        raise TypeError(
+            "Passed a class in the forecaster parameter rather than a "
+            "object that is an instance of class that inherits from BaseForecaster"
+        )
+    if not isinstance(forecaster, BaseForecaster):
+        raise TypeError("Passed an object that does not inherit from BaseForecaster.")
+    preds = np.zeros(steps_ahead)
+    forecaster.fit(y, exog=exog)
```
alexbanwell1, line 64: This would need to be inside the loop - you need to fit the forecaster again each time

aeon/forecasting/_recursive.py
alexbanwell1, line 1: I feel this would be better as a method on the base_forecaster class? This has some major drawbacks doing it this way - can't currently pass parameters to the object (although there are workaround for this), it also doesn't allow any optimisation of any individual algorithms  if we were to go down the route of just updating the parameters rather than refitting the entire model!

TonyBagnall: I think thats definitely an option and we can discuss it. Meanwhile, I have stopped it passing the class so user can set the parameters.

aeon/forecasting/_recursive.py
TonyBagnall: recursive does not get refit each time for regressors, you fit once then just change the input to predict. How we resolve this for stats is open to debate.

TonyBagnall:
replaced by #2899

PR #2881, opened 2025-06-05
[ENH] Add an evaluate function

TonyBagnall:
third in a series of simple functions, I will probably put them all in one file in the end, and can resolve passing object/class, just want to get the concept first. This is dependent on the previous two

All this does is take a y_train and y_test split, assumed to be contiguous in time, then iteratively makes test predictions using all the data prior to the point in question. There is probably a name for this model, can change it later.


TonyBagnall:
this is superceded by the base class model for iterative and direct

PR #2887, opened 2025-06-08
[ENH] Adding a simple SFA code

hadifawaz1999:
still a draft, to be incrementally enhanced. For now simply an SFA with basic binning, no windowing no other parameters.


aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+        for c in range(self.n_channels):
+
+            X_c = X[:, c, :]
+
+            dft_features = np.array([self._compute_dft(x_c) for x_c in X_c])
+
+            bins = self._uniform_binning(dft_features)
+            self.bins_.append(bins)
+
+    def _compute_dft(self, x):
```
patrickzib, line 96: There is a `_binning_dft` in `_sfa_fast.py` which has a branch for no windowing. Why not use that one? Also there is `_fast_fourier_transform`in `_sfa_fast.py` 

One problem is that SFA does divide by stds by default, which this code does not. Also, SFA normalizes by the inverse window size, to allow for lower bounding, which is missing here.

aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+            dft[1::2] = imags[
+                : (self.word_length - 1 - np.uint32(self.word_length / 2))
+            ]
+        else:
+            dft[0::2] = reals[: np.uint32(self.word_length / 2)]
+            dft[1::2] = imags[: (self.word_length - np.uint32(self.word_length / 2))]
+
+        return dft
+
+    def _uniform_binning(self, values):
```
patrickzib, line 117: This is different from what SFA does. SFA creates its own bins for each Fourier value.. I.e. your breakpoints are 2d and not 1d with first dimension being the word length. This process is called MCB and fundamentally different from the one implemented here. 

I would prefer to have this MCB applied here, too.

aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+                edges = [-np.inf] + [np.inf] * (self.alphabet_size - 1)
+            else:
+                edges = list(
+                    np.linspace(min_val_t, max_val_t, self.alphabet_size + 1)[1:-1]
+                )
+
+            bins.append(edges)
+        return bins
+
+    def _discretize(self, values, bins):
```
patrickzib, line 135: Accordingly, the correct bins for each symbol must be used. See `_transform_words_case`

aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+from sklearn.utils import check_random_state
+
+from aeon.transformations.collection import BaseCollectionTransformer
+
+
+class SFA_Simple(BaseCollectionTransformer):
+    """Symbolic Fourier Approximation (SFA) [1]_ Transformer.
+
+    For each series, shorten the series with DFT
+    discretise the shortened series into bins set by MFC
```
patrickzib, line 19: what is MFC?

aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+        self.random_state = random_state
+
+        self.n_cases = 0
+        self.n_channels = 0
+        self.n_timepoints = 0
+
+        super().__init__()
+
+    def _fit(self, X, y=None):
+        """Calculate word breakpoints using MCB or IGB.
```
patrickzib, line 67: where is MCB or IGB applied?

aeon/transformations/collection/dictionary_based/_sfa_simple.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def _discretize(self, values, bins):
+        word = []
+        for i, val in enumerate(values):
+            edges, symbols = bins[i], self.breakpoints
+            bin_idx = np.searchsorted(edges, val, side="right")
+            word.append(symbols[bin_idx])
+        return word
+
+    def _transform(self, X, y=None):
```
patrickzib, line 143: please check `_transform_words_case` in _sfa_fast.py - might be similar to this one

patrickzib (changes requested):
Thank you very much. The code differs from the SFA paper in at least two ways:
1) Way how DFT is computed and normalized
2) Way how MCB works

PR #2893, opened 2025-06-09
[ENH] remove fit_is_empty tag from forecasting base

TonyBagnall:
the way its structured currently, it makes no sense to have a tag fit_is_empty, since fit is the only place the forecaster sees the series. I dont see any use for a completely random forecaster.


TonyBagnall:
seems fit_is_empty is an estimator tag, so will leave it there

PR #2895, opened 2025-06-09
[ENH] `fit_is_empty` ETS

MatthewMiddlehurst:
Fits better this way IMO, but up for discussion.


TonyBagnall:
Won't work with next version which has fitted parameters

MatthewMiddlehurst:
We can wait for that to go in then

TonyBagnall:
it has gone in

TonyBagnall:
so happy to revisit the issue but I think this PR is too far behind main to keep? @MatthewMiddlehurst

TonyBagnall:
this should be an issue not a PR, I will set it up

PR #2897, opened 2025-06-11
[BUG] Fixes a bug with SFAFast throwing an error when calling transform after fit

patrickzib:
This PR fixes a bug with SFAFast throwing an error when calling `transform()` after `fit()`

As reported by @baraline SFAFast had the following error:

```python
from aeon.datasets import load_classification
from aeon.transformations.collection.dictionary_based import SFAFast
X_train, y_train = load_classification("SmoothSubspace", split="train")
SFAFast().fit(X_train, y_train).transform(X_train, y_train)
```

results in

```
File ~\Documents\aeon\aeon\transformations\collection\dictionary_based\_sfa_fast.py:1245, in create_bag_transform(X_index, feature_count, feature_selection, relevant_features, sfa_words, remove_repeat_words)
   1241         all_win_words[j, :] = np.bincount(
   1242             sfa_words[j][masked], minlength=feature_count
   1243         )
   1244     else:
-> 1245         all_win_words[j, :] = np.bincount(sfa_words[j], minlength=feature_count)
   1246 else:
   1247     if remove_repeat_words:

ValueError: could not broadcast input array from shape (20324,) into shape (0,)
```

This PR fixes the above problem and also adds a test case.

### DIFF
--- a/aeon/transformations/collection/dictionary_based/tests/test_sfa.py
+++ b/aeon/transformations/collection/dictionary_based/tests/test_sfa.py
@@ -5,7 +5,8 @@ import sys
 import numpy as np
 import pytest
 
-from aeon.transformations.collection.dictionary_based._sfa import SFA
+from aeon.datasets import load_classification
+from aeon.transformations.collection.dictionary_based import SFA, SFAFast
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


aeon/transformations/collection/dictionary_based/tests/test_sfa.py
```diff
@@ -224,3 +225,27 @@ def test_typed_dict():
     word_list2 = p2.bag_to_string(p2.transform(X, y)[0][0])
 
     assert word_list == word_list2
+
+
+def test_sfa_fast_transform_after_fit():
+    """Test transform called after fit returns the same result as fit_transform()."""
+    X_train, y_train = load_classification("SmoothSubspace", split="train")
```
baraline: I think it would be better to use the specific test datasets in the testing module, it avoid using load classif which can fail due to unavailability.

baraline (changes requested):
Small comment on the testing data, otherwise looks good. 
I think @MatthewMiddlehurst wanted to add a generic test for all estimators, but I think it would fall into another PR?

aeon/transformations/collection/dictionary_based/tests/test_sfa.py
patrickzib: Sure.

MatthewMiddlehurst (commented):
We test .fit.transform against .fit_transform in the tests. does the newly added test fail without the fix? Only thing we dont test would be .fit_transform into another .transform i think

No issue with the new test since it uses default parameters.

baraline (approved):
Thanks, lgtm !

PR #2899, opened 2025-06-11
[ENH] Add recursive and direct methods to forecasting base class 

TonyBagnall:
see #2410 and #2880

this option is to have base class methods recursive_forecast and direct_forecast. 

1. Direct is the same for all, but some forecasters cannot do it, so we can just check the tag here
2. The default recursive is how you would do it with regression with repeated calls to predict, but allows stats models to override for their ways

I think this is the easiest way to generalise. It also covers the generative model use case: direct_forecast can cover the situation where the forecaster outputs a series rather than a single forecast

closes #2410 and #2880 and adjusts notebooks

note we may later move to a mixin model


TonyBagnall:
So if this is agreed I will remove the two alternatives and work notebooks into one

alexbanwell1 (commented):
Looks good

aeon/forecasting/_regression.py
```diff
@@ -74,10 +74,10 @@ def _fit(self, y, exog=None):
         # Ignore the final horizon values: need to store these for pred with empty y
         X = X[: -self.horizon]
         # Extract y
```
hadifawaz1999: extract y_train ? to avoid confusion

hadifawaz1999 (approved):
LGTM, just made a small comment on a comment in regression forecaster

MatthewMiddlehurst (approved):
discussed this on slack quite a bit. Fine with having this in while we think about framework further.

PR #2911, opened 2025-06-18
[DOC,MNT] Documentation configuration clean-up

MatthewMiddlehurst:
- Fixes changelog and docstring typos
- Adds GSoC mentees to the funding section
- Tidies up the conf.py into sections and removes some useless attributes (from removed/deprecated extensions). Adds more comments such as where the config option is from and what they do
- Removes the manual and Latex configuration. I dont think anyone uses this and it can be easily added back. It seems like clutter in an already difficult to understand file for newcomers
- Add option to show which classes an estimator inherits in the API

Old:
image
New:
image

- Adds config to remove some useless functions we inherit from the API page i.e. `get_metadata_routing` used by `scikit-learn`

Old:
image
New:
image

-  Moves the binder badge from the top of examples to the bottom

Old:
image
image
New:
image
image

### DIFF
--- a/docs/changelogs/v1.2.md
+++ b/docs/changelogs/v1.2.md
@@ -9,7 +9,7 @@ June 2025
 into `anomaly_detection.collection` and `anomaly_detection.series` submodules.
 - `ProximityForest` and `ProximityTree` classifiers have been improved to support
 unequal length and multivariate series. The classifiers are also significantly faster.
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst:
This may come with some unintended side effects. Keep an eye out, but from what I checked there is nothing off.

hadifawaz1999 (approved):
lgtm

PR #2913, opened 2025-06-20
[MNT] Add testing workflow for multithreading and make threaded estimators use `check_n_jobs`

### LINKED ISSUE #1886: [ENH] n_jobs/_n_jobs, parameter in classifiers

MatthewMiddlehurst:
Fixes #1886

I use the keyword to close the issue, but it's more like "fixes". I am not entirely happy with multithreading, but I think this is better than having it in the base class. Adds testing to ensure that estimators use `check_n_jobs` by enforcing a `self._n_jobs` attribute if you have the multithreading capability tag. 

Adds a separate workflow to run estimators which can thread, which doesn't use `pytest-xdist`. This tests attributes, tags and whether output is the same when threading and not. Attribute tests do not apply when the `fit_is_empty` tag is true.

The vast majority of the PR is adding `self._n_jobs` and `check_n_jobs` to current estimators. There is likely still some issues with documentation. No guarantee that passing the attribute to external libraries will work the same way (though not much we can do about that).


baraline (approved):
This at least makes the use of the parameter and how its set standard across all estimators, which was at the core of the original issue. I'm in favour of this change.

TonyBagnall (approved):
you have bitten that bullet! nice one

PR #2920, opened 2025-06-27
[ENH] Implement Time-Varying Parameter (TVP) regression forecaster using Kalman filter 

TonyBagnall:
Time-Varying Parameter is a whole sub topic in forecasting, with a strong Bayesian flavour. 

This is a simple implementation using a Kalman filter that also models the parameter variance. It mirrors the RegressionForecaster approach.

Basically you update the parameters of the model each time. My interpretation is that it is a recency weighted regression fitting algorithm, along the lines of discount functions in reinforcement learning or stochastic gradient descent, with a data driven learning rate.


TonyBagnall:
```python
from statsmodels.tsa.statespace.mlemodel import MLEModel

class TVPModel(MLEModel):
    def __init__(self, endog, exog):
        k_states = exog.shape[1]
        super().__init__(endog, k_states=k_states, k_posdef=k_states)
        self.exog = exog
        self['design'] = exog[:, None, :]
        self['transition'] = np.eye(k_states)
        self['selection'] = np.eye(k_states)
        self.k_states = k_states

    def start_params(self):
        return [0.1] + [0.01] * self.k_states

    def update(self, params, **kwargs):
        self['obs_cov'] = np.array([[params[0]]])
        self['state_cov'] = np.diag(np.square(params[1:]))

X = np.column_stack([np.ones(n), x])  # intercept + x
model = TVPModel(y, X)
res = model.fit(disp=False)

filtered_betas = res.filtered_state.T
```

TonyBagnall:
> From what I can see it looks good, I have no idea if it matches the model described in the paper, but it looks like it works if the tests pass! It looks like it does what it's supposed to do. I guess the only thing is to check it against any other implementations if they exist

yeah tried that, but the statsmodel one is weird. I can say it forecasts [1,2,3,4,5,6,7,8,9,10] pretty well :)

PR #2926, opened 2025-07-02
[ENH] Auto naive forecaster

TonyBagnall:
Simple forecaster that chooses either naive, mean or seasonal based on train error (or "in sample" as they call it).

Seems like a sensible benchmark to me, although Ive not seen it used. Main issues are

1) It uses the train sample mean to measure performance
2) It evaluates seasonality for a max of 1/4 series length

could easily cross validate, but seems OTT for this algorithm.


aeon/forecasting/_auto_naive.py
```diff
@@ -0,0 +1,86 @@
...
+            self.best_mse_ = best_seasonal
+
+        return self
+
+    def _predict(self, y, exog=None):
+        if self.strategy_ == "last" or self.strategy_ == "mean":
+                    return self._fitted_scalar_value
+            # For "seasonal_last" strategy
+        prediction_index = (self.horizon - 1) % self.season_
+        return self.self._y [prediction_index]
```
alexbanwell1, line 84: I believe this should be self._y[-self.season_ - 1 + prediction_index)] that way it is predicting the last seasonal value in the series, rather than the corresponding seasonal value at the start of the series

aeon/forecasting/_auto_naive.py
```diff
@@ -0,0 +1,86 @@
...
+from enum import Enum
+from aeon.forecasting.base import BaseForecaster
+
+
+class AutoNaiveForecaster(BaseForecaster):
+    """
+    Naive forecaster with strategy set based on minimising error.
+
+    Searches options, "last", "mean", and "seasonal_last", with season in range [2,max_season].
+    If max_season is not passed to the constructor, it will be set to series length/2.
```
alexbanwell1, line 19: Should be length/4

aeon/forecasting/_auto_naive.py
```diff
@@ -0,0 +1,86 @@
...
+        # series mean strategy, in sample
+        mse_mean = np.mean((y - np.mean(y)) ** 2)
+
+        # seasonal strategy, in sample
+        max_season = self.max_season
+        if self.max_season is None:
+            max_season = len(y)/4
+        best_s = None
+        best_seasonal = np.inf
+        for s in range(1, max_season + 1):
```
alexbanwell1, line 55: Should be 2 surely? - as the interval is [2, max_season]

aeon/forecasting/_auto_naive.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class AutoNaiveForecaster(BaseForecaster):
+    """
+    Naive forecaster with strategy set based on minimising error.
+
+    Searches options, "last", "mean", and "seasonal_last", with season in range [2,max_season].
+    If max_season is not passed to the constructor, it will be set to series length/2.
+
+    Simple first implementation, splits the train series into 70% train and 30% validation split
```
alexbanwell1, line 21: I don't believe it does this, it simply minimises RMSE over the whole series passed in _fit

alexbanwell1 (changes requested):

TonyBagnall:
nice to do, but too far behind main

PR #2927, opened 2025-07-02
[ENH] Refactor to set forecast_ in fit and to require y in predict

### LINKED ISSUE #2909: [ENH] Forecasting predict function: require y

after discussion on slack, an opinion was put forward to remove the option of calling `predict` without y. This issue  puts that in context and opens up discussion.

Currently the simplest forecasting use case is

```python
f = MyForecaster(horizon=1)
f.fit(y)
pred = f.predict()
```
`predict` signature is `predict(y=none)`. If `y` is not passed, the prediction is the next value from the series passed in fit, if it is passed the prediction is the next value of the series passed in `y`. 

```python
f = MyForecaster(horizon=1)
f.fit(y)
pred = f.predict(other_y) # predict the value horizon after `other_y`
```

Predict applies the model from fit to a previously unseen series, with no time index. This is a natural model for regression, but a bit confusing for stats forecasters. The equivalent operation to `predict(y)` in stats models is `filter(y)`, in stats parlance I think this is called making out of sample predictions, although we only make a single prediction from `y` which is assumed to be prior to our `horizon`. We do not assume `y` in predict directly follows `y` in train. So for MA component in ARMA we can use the out of sample residuals from `y` in predict, but it doesnt work with seasonal component models such as ETS.

So one way forward is to not make fit/predict the main access, instead we retain the temporal link between train/test and use

 predict one ahead  
```
f = MyForecaster(horizon=1)
pred = f.forecast(y)
```
predict `prediction_horizon` ahead using iterative or direct
```
f = MyForecaster(horizon=1)
preds1 = f.direct_forecast(y, prediction_horizon=10)
preds1 = f.iterative_forecast(y, prediction_horizon=10)
```

`predict` is retained primarily for regression models. We then change `predict` to require `y` and make naive have tag `fit_is_empty` and forecast then becomes
```python
def _forecast(self, y):
   return self.predict(y)
```
and put the Naive functionality in `_predict`. We then work out whether ETS can also have fit is empty as true. 

_No response_

### Additional context

_No response_

TonyBagnall:
after discussion, we decided to remove the y=None option in predict for forecasters, and this PR refactors our current forecasters to do this.

the model is now that predict stores the single prediction from the train set in self.forecast_, so that now 

```python
forecast(y):
   self.fit(y)
   return self.forecast_
```
predict is now reserved for algorithms than can fit independent of time, so nothing seasonal etc. These should use `forecast()` or `iterative_forecast()`

closes #2909


aeon/forecasting/_ets.py
```diff
@@ -214,18 +224,7 @@ def _predict(self, y=None, exog=None):
...
-            self.level_,
-            self.trend_,
-            self.seasonality_,
-            self.phi,
-            self.horizon,
-            self.n_timepoints_,
-            self._seasonal_period,
-        )
-        return fitted_value
+        return self.forecast_
```
MatthewMiddlehurst, line 227: Currently there is nothing to tell people using ETS that inputting `y` does nothing, similar to the previous issue with Naive.

aeon/forecasting/_regression.py
```diff
@@ -122,26 +125,52 @@ def _predict(self, y=None, exog=None):
...
-        last_window_flat = last_window.reshape(1, -1)
+        last_window = combined_data.reshape(1, -1)
+
+        return self.regressor_.predict(last_window)[0]
 
-        return self.regressor_.predict(last_window_flat)[0]
+    def _forecast(self, y, exog=None):
+        """Forecast values for time series X."""
+        self.fit(y, exog)
+        return self.forecast_
```
MatthewMiddlehurst, line 173: Same as the base class i think?

aeon/forecasting/tests/test_naive.py
```diff
@@ -31,36 +29,36 @@ def test_naive_forecaster_seasonal_last_strategy():
...
     np.testing.assert_array_equal(pred, expected)
     np.testing.assert_array_equal(pred2, expected)
 
     # Test horizon wrapping around to a new season
     forecaster = NaiveForecaster(strategy="seasonal_last", seasonal_period=3, horizon=7)
+    pred = forecaster.forecast(data)
     forecaster.fit(data)
-    pred = forecaster.predict()
-    pred2 = forecaster.predict(y=data)
+    pred2 = forecaster.forecast_
```
MatthewMiddlehurst, line 52: predict actually uses y useful to test it returns the same as forecast/self.forecast_

MatthewMiddlehurst (commented):
in favour of making `y` required. What are some of the pros and cons of the `self.forecast_` model vs `fit_is_empty`?

aeon/forecasting/tests/test_naive.py
TonyBagnall: added a test for predict and removed the testing of array equality

aeon/forecasting/_regression.py
TonyBagnall: base class calls fit and predict, so would result in calculating forecast_ twice. Can remove this if we go to fit_is_empty and ditch forecast_

aeon/forecasting/_ets.py
TonyBagnall: this is the current behaviour, so its not introducing this. We can change when we revise ETS

MatthewMiddlehurst (approved):

PR #2939, opened 2025-07-10
[ENH] Rework ETS Forecaster

TonyBagnall:
some adaptations of ETS to make it compliant. The basic structure of ETS is this

```python
_fit
    _numba_fit
    self.forecast_ = _numba_predict


_numba_fit
    for t in range(n_timepoints):
            _update_states()
                _predict_value

_numba_predict
    _predict_value
```
predict_value makes a step ahead prediction (and also returns two other variables, so maybe poorly named)

1. Override iterative_forecasting
             the default method would result in refitting ETS each time. It  now calls fit once then iterates numba_predict

```python
        self.fit(y)
        preds = np.zeros(prediction_horizon)
        preds[0] = self.forecast_
        for i in range(1, prediction_horizon):
            preds[i] = _numba_predict(
                self._trend_type,
                self._seasonality_type,
                self.level_,
                self.trend_,
                self.seasonality_,
                self.phi,
                i + 1,
                self.n_timepoints_,
                self._seasonal_period,
            )
``` 

2. With no damping, now sets phi to the horizon

3. Not sure if I will make it "fit_is_empty" here, may be better to wait until we see how auto ets works.


alexbanwell1 (approved):
Looks good to me, I've fixed the one bug that seemed to exist, so otherwise set up with the right parameters, produces similar forecasts to statsmodels

TonyBagnall:
our next move is to auto set the parameters 
```python
    alpha : float, default=0.1
        Level smoothing parameter.
    beta : float, default=0.01
        Trend smoothing parameter.
    gamma : float, default=0.01
        Seasonal smoothing parameter.
```
in fit, it doesnt really make sense for the user to set them, other packages dont and the forecasts are very sensitive to them. So ETS will fit the parameters in train, AutoETS will set the hyper parameters (additive etc) in fit using ETS

TonyBagnall:
test expected values changed because of the seasonality bug

PR #2948, opened 2025-07-13
[ENH] forecasting coverage

TonyBagnall:
coverage for forecasting, will aim to keep it at 100%, leaving ETS until after the refactor


aeon/forecasting/tests/test_naive.py
```diff
@@ -62,3 +63,19 @@ def test_naive_forecaster_seasonal_last_strategy():
...
     expected = 6  # predicts the 2nd element of the new last season.
     np.testing.assert_allclose(pred, pred2, pred3, expected)
+
+
+def test_predict():
+    """Test different input for private predict."""
+    forecaster = NaiveForecaster(strategy="mean")
+    y = np.array([1, 2, 3, 4, 5, 6, 7, 8])
+    x = forecaster._predict(y)
+    assert isinstance(x, float)
```
alexbanwell1: Is it worth doing assert x == 8.0 here for an extra check

alexbanwell1 (dismissed):
Looks like good additional coverage, looks good

aeon/forecasting/tests/test_naive.py
TonyBagnall: 4.5 :)

PR #2953, opened 2025-07-19
[MNT] add n_jobs parameter for sax invertion

### LINKED ISSUE #2873: [MNT] SAX parallel numba without n_jobs parameter
### Describe the issue

SAX uses numba parallel, albeit only for symbol inversion, but does not expose a n_jobs parameter to control the number of threads through numba set_num_threads.

_No response_

### Additional context

_No response_

baraline:
#### Reference Issues/PRs

Fixes #2873

add n_jobs parameter and use check n_jobs with set num threads to set it for the parallel numba function for sax invertion.

### DIFF
--- a/aeon/transformations/collection/dictionary_based/_sax.py
+++ b/aeon/transformations/collection/dictionary_based/_sax.py
@@ -64,6 +64,7 @@ class SAX(BaseCollectionTransformer):
 
     _tags = {
         "capability:multivariate": True,
+        "capability:multithreading": True,
         "fit_is_empty": True,
         "algorithm_type": "dictionary",
     }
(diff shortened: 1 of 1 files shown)


hadifawaz1999:
Nice ! can we benchmark the time gain you think ?

baraline:
Hey, I don't think there will be any, the default is to use all threads for numba if you don't specify anything. Or were you talking about making the other part of SAX use numba parallel too ? This only impact the inverse transform.

hadifawaz1999:
> Hey, I don't think there will be any, the default is to use all threads for numba if you don't specify anything. Or were you talking about making the other part of SAX use numba parallel too ? This only impact the inverse transform.

ah yes you're right never mind !

hadifawaz1999 (approved):

PR #2962, opened 2025-07-28
[ENH] Forecasting refactor

TonyBagnall:
Reorganise the forecasting module as follows

forecasting
--- stats
           ETS, ARIMA
--- deep_learning (when we have some)
--- machine_learning (when we have some)
--- utils (including nelder mead)
--- tests

Also rename ETSForecaster as ETS

edit: when sorting out the API docs I noticed that there were still incorrect references to registry, so also changed them here


aeon/forecasting/__init__.py
MatthewMiddlehurst, line 1: Shouldn't import from both here and the submodules IMO

TonyBagnall: wondered about that, ok

TonyBagnall:
> GitHub is showing some of these as new files rather than moved i.e. ARIMA and the utils. Can you check that these arent duplicates or contents from another PR

that will be because I branched this of the ARMA branch I assume

aeon/forecasting/utils/_hypo_tests.py
alexbanwell1, line 1: Was this in the base branch? This is used for AutoARIMA but I don't think anything else, so not sure why it's in here already? Presumably it should go in with AutoARIMA? Not a problem but surprised it's made it in already

aeon/forecasting/utils/_seasonality.py
alexbanwell1, line 1: Likewise with this one

docs/api_reference/forecasting.rst
```diff
@@ -0,0 +1,36 @@
+.. _forecasting_ref:
+
+Forecasting
+===========
+
+The :mod:`aeon.forecasting` module contains algorithms for forecasting.
+
+All clusterers in `aeon` can be listed using the `aeon.registry.all_estimators`
```
alexbanwell1: forecasters

docs/api_reference/forecasting.rst
```diff
@@ -0,0 +1,36 @@
...
+.. currentmodule:: aeon.forecasting
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseForecaster
+    NaiveForecaster
+    RegressionForecaster
+    TVPForecaster
```
alexbanwell1, line 26: Are we leaving all of these just at the top level? I would have thought TVP would fall into stats, and Regression into machine learning? And then that only leaves Naive at the top, which arguable could be put into stats, but could also be left at the top level. Doesn't matter too much, but just seems like it would be sensible to split them all down while we're doing this refactoring?

aeon/forecasting/utils/_hypo_tests.py
TonyBagnall: I wondered, it was in the general utils folder so I guess came in with ARIMA, I'll take it out for now, makes sense to include with Auto

aeon/forecasting/utils/_seasonality.py
TonyBagnall: I thought they might be legacy :)

docs/api_reference/forecasting.rst
TonyBagnall: actually I should check that actually works, just cut and paste :)

docs/api_reference/forecasting.rst
TonyBagnall: Well I left TVP open as its not exactly obvious, TVP is a regressor after all. I think once we have a machine learning module we can discuss. I think RegressionForecaster should be in base, it can be used with any regressor and I dont classify Linear Regression as machine learning .... its all debatable

docs/api_reference/forecasting.rst
TonyBagnall: found it was incorrect for other types of estimator too, so thats useful

alexbanwell1 (approved):
Looks fine now

PR #2963, opened 2025-07-28
[DEP] Remove deprecated stats

MatthewMiddlehurst:
Deprecation for 1.3.0


baraline (dismissed):
lgtm

MatthewMiddlehurst:
@SebastianSchmidl could you double-check the anomaly detection bits if you have time, should be quick and dont think I missed anything. `range_precision` never used its alpha parameter so fixed that. I have a slight memory of a discussion that it is not really a useful parameter, but should either be used or deprecated IMO.

aeon/benchmarking/metrics/anomaly_detection/_range_metrics.py
```diff
@@ -207,3 +204,352 @@ def range_f_score(
...
+
+    if start is not None:
+        ranges.append((start, len(binary_sequence) - 1))
+
+    return ranges
+
+
+def _ts_precision(
+    y_pred_ranges, y_real_ranges, gamma="one", bias_type="flat", alpha=0.0
+):
```
SebastianSchmidl, line 425: `alpha` is missing in doc

aeon/benchmarking/metrics/anomaly_detection/_range_metrics.py
```diff
@@ -67,7 +62,7 @@ def range_precision(
     y_pred_ranges = _binary_to_ranges(y_pred)
     y_true_ranges = _binary_to_ranges(y_true)
     return _ts_precision(
-        y_pred_ranges, y_true_ranges, gamma=cardinality, bias_type=bias
+        y_pred_ranges, y_true_ranges, gamma=cardinality, bias_type=bias, alpha=alpha
```
SebastianSchmidl, line 67: Maybe include a section about the `alpha`-discussion in the documentation of `range_precision`?

> Because precision emphasizes prediction quality, there is no need for an existence reward and this value should always be set to 0.

You can add `[1]` because the paper backs this statement.

SebastianSchmidl (changes requested):

MatthewMiddlehurst:
Included the suggestion in both docstrings

SebastianSchmidl (approved):

PR #2968, opened 2025-07-29
[ENH] Improvements to ST transformer and classifier

MatthewMiddlehurst:
- Enables unequal length for STC. Changes ST unequal length transform to flip the series and shapelet in distance calculation if the series is smaller. This replaces the current code extracting up to the smallest length seen in the train set only. This means it can now handle lengths smaller than that seen in the train set as well as extracting a larger range of shapelets.
- Removes early abandon IG calculation from ST. Running some experiments on the UCR archive showed that it makes it a lot slower overall. Alternative methods for keeping the early abandon showed that speed also improved, but with no significant difference from just removing it. Removing it allows easier implementation of alternative quality measures and makes the code simpler.
- Saves the distances calculated during shapelet extraction for `fit_transform`. Small optimisation which adds a bunch of lines. Tested for correctness but not speed. It skips a full transform call though so I doubt it is going to be worse than the current runtime.
- Adds a verbose parameter for STC and ST just printing out progress at certain points.


MatthewMiddlehurst:
testFitTime | stc-no-ea | stc-reduced-ea | stc-bisect-ea | stc
-- | -- | -- | -- | --
testFitTimeMean | 535886.3 | 535459 | 537589 | 1694658
testFitTimeAvgRank | 1.618182 | 1.945455 | 2.754545 | 3.681818

testAccuracy | stc-no-ea | stc | stc-bisect-ea | stc-reduced-ea
-- | -- | -- | -- | --
testAccuracyMean | 0.861645 | 0.861668 | 0.861668 | 0.861668
testAccuracyAvgRank | 2.486364 | 2.504545 | 2.504545 | 2.504545

Early abandon testing on UCR datasets

TonyBagnall:
also introduces a batch mechanism for threading?

MatthewMiddlehurst:
That was just me moving the parameter, the batch is how many shapelets we extract before merging it into the full collection, removing similar etc.

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -257,19 +258,23 @@ def _fit(self, X, y):
...
                         n_shapelets_extracted + i,
-                        shapelets,
-                        max_shapelets_per_class,
                         check_random_state(rng.randint(np.iinfo(np.int32).max)),
                     )
                     for i in range(self._batch_size)
                 )
+                candidate_shapelets, candidate_distances = zip(*p)
+
+                if save_distances:
```
TonyBagnall, line 274: not sure whats going on here, seems to overwrite s[6] each time?

MatthewMiddlehurst: GitHub is not showing the correct bit of code I think

MatthewMiddlehurst:
I think the expected results are due to the early abandon on datasets with a small amount of cases. I found no difference on a run with the full BasicMotions but did on the subsample we use in testing.

TonyBagnall (approved):
ignore this

PR #2975, opened 2025-07-30
[ENH] Update testing data dictionary with series testing data

MatthewMiddlehurst:
The current testing data for series estimators is rather shallow as around the time of the rework most series modules were much more experimental. Now that more work is going in to these modules i.e. anomaly detection and forecasting this PR expands the testing data available in the general testing and formats it more like the collection sets. For now the options are no labels and anomaly labels, but it is easy to expand this.

### DIFF
--- a/aeon/testing/data_generation/_labels.py
+++ b/aeon/testing/data_generation/_labels.py
@@ -10,7 +10,8 @@ def make_anomaly_detection_labels(
 ) -> np.ndarray:
     """Generate anomaly detection labels.
 
-    Generates a simple 0/1 anomaly labels for testing purposes.
+    Creates a boolean array of length ``n_labels`` with ``ceil(anomaly_rate*n_labels)`` 
+    anomalies at random positions.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


aeon/testing/data_generation/_labels.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def make_anomaly_detection_labels(
+    n_labels: int,
+    anomaly_rate: float = 0.01,
+    random_state: Optional[int] = None,
+) -> np.ndarray:
+    """Generate anomaly detection labels.
+
+    Generates a simple 0/1 anomaly labels for testing purposes.
```
TonyBagnall: generates a boolean array of length ``n_labels`` with ``ceil(anomaly_rate*n_labels)`` anomalies at random positions.

TonyBagnall (approved):
just a comment on the docstring, but looks good otherwise

aeon/testing/data_generation/_labels.py
MatthewMiddlehurst: Done

baraline (approved):
LGTM, as you said the labels can be extend later so good for now.

PR #2976, opened 2025-07-30
[DOC] Simple imbalance notebook

TonyBagnall:
Simple notebook for class imbalance


LinGinQiu (approved):
I have add ohit and references in #2986, and expression about ADASYN seems a bit vague. I have replaced it #2986

PR #2979, opened 2025-07-30
[ENH] Series data validation improvements and data validation deprecations

MatthewMiddlehurst:
Updates series validation functions

- Removes/deprecated the current hierarchical datatypes and utilities. We can reimplement if we deal with hierarchical in the future, for currently these are not used anywhere. There are currently no plans to use them either.
- Implements a bunch of old utilities from `validation/__init__.py` which were missed in the major release. I dislike putting functions in the init and none of these have lots of usage if any. Replaces any code which use them with different functions.
- Minor updates to the collection validation code and docstring. Main change is that importing from `validation/__init__.py` is deprecated, import from `validation.collection` instead.
- Reworks series validation in the style of  `validation.collection` and deprecates old functions. Most old functionality is still available in new functions. This finally gets rid of `check_series` which is an old function we didnt want to touch as it was embedded in so much old code we have removed since.


aeon/utils/validation/tests/test_series.py
```diff
@@ -17,161 +9,62 @@
...
+
+
+def test_has_missing():
+    """Test has_missing validation."""
+    assert not has_missing(make_example_pandas_series())
+    assert not has_missing(make_example_1d_numpy())
+    assert not has_missing(make_example_dataframe_series(n_channels=1))
+    assert not has_missing(make_example_dataframe_series(n_channels=2))
+    assert not has_missing(make_example_2d_numpy_series(n_channels=1))
+    assert not has_missing(make_example_2d_numpy_series(n_channels=2))
```
baraline: A test with an actual assert to True would be nice to have here, testing a series with numpy nan and a DF with pd NA.

baraline (changes requested):
Deprecation and other stuff LGTM, just a comment on a test that doesn't actually test the capability

aeon/utils/validation/tests/test_series.py
MatthewMiddlehurst: Agreed, but I would like https://github.com/aeon-toolkit/aeon/pull/2975 in first to make testing these easier.

MatthewMiddlehurst: Done

baraline:
I'm seeing changes that seem a bit unrelated to the PR (changes to collections etc.), Is this normal ? It looks like the change since last viewed is showing me stuff from merging from main too ... Wondering if GitHub is playing a trick on me.

Issue #2972, opened 2025-07-29
[ENH] Parallel PAA & SAX transform


baraline:
### Describe the feature or idea you want to propose

Sax transform currently have n_jobs but use it only for inverse transform. We should also use it for the transform step.

### Describe your proposed solution

We should introduce numba parallel replacements for `_get_sax_symbols(self, X_paa)` and also introduce n_jobs to `PAA`. 

These functions are fairly simple to parallelize. For exemple the sax symbols (replace uint8 with appropriate dtype): 
```python
@njit(fastmath=True, cache=True, parallel=True)
def _discretize_input(X, breakpoints):
    n_cases, n_channels, n_timepoints = X.shape
    X_new = np.zeros((n_cases, n_channels, n_timepoints), dtype=np.uint8)
    n_break = breakpoints.shape[0] - 1
    for i_x in prange(n_cases):
        for i_c in prange(n_channels):
            for i_b in prange(n_break):
                mask = np.where(
                    (X[i_x,i_c] >= breakpoints[i_b]) & (X[i_x,i_c] < breakpoints[i_b + 1])
                )[0]
                X_new[i_x, i_c, mask] += np.uint8(i_b)

    return X_new
```

PR #2980, opened 2025-07-31
[ENH] Parallelize SAX and PAA transformers

aadya940:
Related to #2972


hadifawaz1999:
Nice addition @aadya940 !

Could you run examples comparing the old and new code in terms of output accuracy ? to make sure the algorithm didnt change
Also can we benchmark the time gain ? like see how faster it becomes as plot in function of n samples length of the series etc.

aadya940:
@hadifawaz1999 Yes sure

aeon/transformations/collection/dictionary_based/_paa.py
```diff
@@ -143,3 +146,45 @@ def _get_test_params(cls, parameter_set="default"):
...
+            for j in range(n_channels):
+                acc = 0.0
+                for k in range(seg_len):
+                    acc += X[i, j, segment[k]]
+                X_paa[i, j, _s] = acc / seg_len
+
+    return X_paa
+
+
+@njit(parallel=True, fastmath=True)
```
patrickzib: please add `cache=True`

aeon/transformations/collection/dictionary_based/_paa.py
```diff
@@ -143,3 +146,45 @@ def _get_test_params(cls, parameter_set="default"):
         """
         params = {"n_segments": 10}
         return params
+
+
+@njit(parallel=True, fastmath=True)
```
patrickzib: please add `cache=True`

aeon/transformations/collection/dictionary_based/_paa.py
```diff
@@ -143,3 +146,45 @@ def _get_test_params(cls, parameter_set="default"):
...
+        segment = split_segments[_s]
+        seg_len = segment.shape[0]
+
+        if seg_len == 0:
+            continue  # skip empty segment
+
+        for i in range(n_samples):
+            for j in range(n_channels):
+                acc = 0.0
+                for k in range(seg_len):
```
patrickzib: Why not use the `.mean()` here as in the original version??

aeon/transformations/collection/dictionary_based/_paa.py
```diff
@@ -143,3 +146,45 @@ def _get_test_params(cls, parameter_set="default"):
...
+
+    for _s in prange(n_segments):  # Parallel over segments
+        segment = split_segments[_s]
+        seg_len = segment.shape[0]
+
+        if seg_len == 0:
+            continue  # skip empty segment
+
+        for i in range(n_samples):
+            for j in range(n_channels):
```
patrickzib, line 165: `prange`?

aeon/transformations/collection/dictionary_based/_paa.py
```diff
@@ -143,3 +146,45 @@ def _get_test_params(cls, parameter_set="default"):
...
+    X_paa = np.zeros((n_samples, n_channels, n_segments), dtype=X.dtype)
+
+    for _s in prange(n_segments):  # Parallel over segments
+        segment = split_segments[_s]
+        seg_len = segment.shape[0]
+
+        if seg_len == 0:
+            continue  # skip empty segment
+
+        for i in range(n_samples):
```
patrickzib, line 164: `prange`?

patrickzib (commented):
Thank you for your additions. Some comments of mine.

aeon/transformations/collection/dictionary_based/_paa.py
aadya940: numba doesn't have an implementation for `mean`

aeon/transformations/collection/dictionary_based/_paa.py
aadya940: Wouldn't that spawn too many threads, given that the outer loop is with a `prange`?

aeon/transformations/collection/dictionary_based/_paa.py
patrickzib: Not sure, I understand. Both seem to work fine (I might be missing something)

```python
        for i in range(n_samples):
            for j in range(n_channels):
                acc = X[i, j, segment].mean()
```

```python
        for i in range(n_samples):
            for j in range(n_channels):
                acc = np.mean(X[i, j, segment])
```

aeon/transformations/collection/dictionary_based/_paa.py
patrickzib: You have used three nested prange loops in `_parallel_inverse_paa_transform` ? 

But, in fact numba seems to ignore nested loops. I did not know this before:

> Loop serialization
> Loop serialization occurs when any number of prange driven loops are present inside another prange driven loop. In this case the outermost of all the prange loops executes in parallel and any inner prange loops (nested or otherwise) are treated as standard range based loops. Essentially, nested parallelism does not occur.

https://numba.pydata.org/numba-doc/dev/user/parallel.html

aeon/transformations/collection/dictionary_based/_paa.py
aadya940: I meant we used `axis=-1` in the original implementation, however, numba doesn't support optional arguments. Hence, implemented it this way.
:))

aeon/transformations/collection/dictionary_based/_paa.py
aadya940: Yep, missed it. However, I'd also like to point out, the previously implemented numba functions in SAX use multiple nested `pranges` as well which is essentially dead code. I can remove it in this PR?

aeon/transformations/collection/dictionary_based/_paa.py
patrickzib: Thanks. I understand.. 

Yet, why not use one of the two alternatives presented above that do not require `axis=-1`?

aeon/transformations/collection/dictionary_based/_paa.py
patrickzib: Not sure, who wrote these. @hadifawaz1999 ?

hadifawaz1999: yes i wrote these, i dont the issue though from the discussion, SAX was working fine why is it dead code ? @patrickzib

hadifawaz1999:
nothing is dead code @aadya940  :) just probably inconsistent usage of prange, but code works fine

aadya940:
@hadifawaz1999 By dead code I meant it has no effect on the output and gives identical results in terms of speed and accuracy to if there was a single `prange` since numba ignores other nested `prange` :)

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -292,3 +296,31 @@ def _invert_sax_symbols(sax_symbols, n_timepoints, breakpoints_mid):
                 ]
 
     return sax_inverse
+
+
+@njit(fastmath=True, cache=True, parallel=True)
```
hadifawaz1999, line 301: Not 100% sure what the gain would be of doing all this, if its not significant am not for doing this, it would mean instead of simply doing ```sax_symbols = np.digitize(x=X_paa, bins=self.breakpoints)``` we have 30+ new lines of code wiith nested loops and prange.

hadifawaz1999 (changes requested):
Should be compared in terms of accuracy before runtime, and should see how worth the gain is compared to the new amount of code

hadifawaz1999:
> @hadifawaz1999 By dead code I meant it has no effect on the output and gives identical results in terms of speed and accuracy to if there was a single `prange` since numba ignores other nested `prange` :)

if am not mistaken numba considers a prange as a normal range when its nested, so its simply useless usage of prange nested yes

aeon/transformations/collection/dictionary_based/_sax.py
aadya940: I have done a benchmark in the past, I think it performs little better after `n_jobs > 2` and significantly better post 4 threads as compared to `np.digitize`

baraline: Having dealt with MONSTER datasets lately, not being able to parallelize stuff for such huge size datasets is a big downside.
It might not be significant for UCR/UEA, but can help a lot when dataset size grow.

I'm not a fan of the flatten and reshape, though, but I guess you did it to avoid the nested loop parallelism problem ?

You could add that breakspoints need to be sorted for the function to work for clarity.

patrickzib: You might want to check `_sfa_fast.py`

I use np.digitize within a parallel for-loop. This approach strikes a balance between using only explicit loops and relying solely on vectorized digitize operations.

```python
        for a in prange(dfts.shape[0]):
            for i in range(word_length):  # range(dfts.shape[2]):
                words[a, : dfts.shape[1]] = (
                    words[a, : dfts.shape[1]] << letter_bits
                ) | np.digitize(dfts[a, :, i], breakpoints[i], right=True)
```

aadya940:
LGTM! The changes suggested have been made.

aadya940:
@baraline Any updates here?

baraline:
Will take a look tomorrow

baraline (dismissed):
All the issues raised by previous members seem fixed, and digitize is supported in numba so no issue on that, using mean make it more readable.

aeon/transformations/collection/dictionary_based/_paa.py
MatthewMiddlehurst: I would keep it consistent and use `prange` for all of them or `range` for the inner ones on both functions.

aeon/transformations/collection/dictionary_based/_sax.py
MatthewMiddlehurst: Was the above implemented? the version I see now looks more like that.

MatthewMiddlehurst (commented):
PAA can take n_jobs now but the SAX implementation does not set n_jobs for PAA (line 128)

MatthewMiddlehurst:
The inverse is parallelised for PAA but not SAX, any reason for that?

aadya940:
> The inverse is parallelised for PAA but not SAX, any reason for that?

Hello, sorry for the late reply! I'll take a look this week.

aadya940:
> PAA can take n_jobs now but the SAX implementation does not set n_jobs for PAA (line 128)

Good catch, it should in-fact be set.

aadya940:
> The inverse is parallelised for PAA but not SAX, any reason for that?

I think it is parallel, maybe the name is misleading? 
[Link](https://github.com/aeon-toolkit/aeon/blob/e193b1e249e97f9e91bd19aa3c6a14d9a9c1a5c8/aeon/transformations/collection/dictionary_based/_sax.py#L288)

hadifawaz1999 (approved):
LGTM ! can you just show a plot before merging to compare runtime with respect to length, channels, samples, n segments, alphabet size etc. ? @aadya940

aadya940:
image

hadifawaz1999:
@aadya940 but do you have the same one of the code on main branch ? to see how much the PR adds in speed

aadya940:
@hadifawaz1999 I assume n_jobs=1 would give the same effect? It'd be the same as the sequential version.

hadifawaz1999:
> @hadifawaz1999 I assume n_jobs=1 would give the same effect? It'd be the same as the sequential version.

Yes, because those plots are with n_jobs=-1 right?

aadya940:
Okay, so ya these are with `n_jobs=4` except the last one (threads vs runtime).

hadifawaz1999:
> Okay, so ya these are with `n_jobs=4` except the last one (threads vs runtime).

is it normal the behavior of sax with lots of threads ? @baraline @MatthewMiddlehurst @patrickzib what do you think

patrickzib:
> > Okay, so ya these are with `n_jobs=4` except the last one (threads vs runtime).
> 
> is it normal the behavior of sax with lots of threads ? @baraline @MatthewMiddlehurst @patrickzib what do you think

The bottom right plot looks kind of odd. But I have no instant explanation for this. One would expect, that adding threads, which obviously speeds up PAA, would also benefit SAX.

aadya940:
Honestly the speed ups weren't the most impressive. The average speedup was around 1.3x

MatthewMiddlehurst:
There are a few comments I have above in the conversations still. For your evaluation use bigger data and average it over 100 runs or something. I agree it would be good to compare it to the old version as well. Good to post the code you used here also. Agreed the last plot is a bit weird, but the actual deviations are rather small.

As Antoine says there's is little use case for multithreading on the UCR, but with bigger datasets having decent speed ups can make a difference. Doing multithreading correctly can be a bit difficult, its possible that there is a better way to go about it than numba.

MatthewMiddlehurst:
> > The inverse is parallelised for PAA but not SAX, any reason for that?
> 
> I think it is parallel, maybe the name is misleading? [Link](https://github.com/aeon-toolkit/aeon/blob/e193b1e249e97f9e91bd19aa3c6a14d9a9c1a5c8/aeon/transformations/collection/dictionary_based/_sax.py#L288)

weird if this was already parallel. Yeah could you make sure the code and naming is consistent with what you have done.

MatthewMiddlehurst:
If you do the timing experiment please make sure the output remains the same as the old one also.

aadya940:
>  Good to post the code you used here also. Agreed the last plot is a bit weird, but the actual deviations are rather small.
@MatthewMiddlehurst Here's the code.

https://gist.github.com/aadya940/4c909ef0f8cca794e93316c526d4e7b9

PR #2987, opened 2025-07-31
[BUG] Fix incorrect undifferencing in ARIMA

alexbanwell1:
Fixes incorrect undifferencing in ARIMA, adds inverse transform to DifferenceTransformer

##### For all contributions

### DIFF
--- a/aeon/transformations/series/tests/test_diff.py
+++ b/aeon/transformations/series/tests/test_diff.py
@@ -12,41 +12,26 @@ def test_diff():
     dt1 = DifferenceTransformer(order=1)
     Xt1 = dt1.fit_transform(X)
     expected1 = np.array([[3.0, 5.0, 7.0, 9.0, 11.0]])
-    # X_hat1 = dt1.inverse_transform(Xt1, X)
 
     np.testing.assert_allclose(
         Xt1, expected1, equal_nan=True, err_msg="Value mismatch for order 1"
     )
-    # np.testing.assert_allclose(
-    #     X_hat1, X, equal_nan=True, err_msg="Inverse transform failed for order 1"
-    # )
 
     dt2 = DifferenceTransformer(order=2)
     Xt2 = dt2.fit_transform(X)
     expected2 = np.array([[2.0, 2.0, 2.0, 2.0]])
-    # X_hat2 = dt2.inverse_transform(Xt2, X)
 
     np.testing.assert_allclose(
         Xt2, expected2, equal_nan=True, err_msg="Value mismatch for order 2"
     )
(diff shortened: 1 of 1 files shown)


aeon/transformations/series/_diff.py
```diff
@@ -1,8 +1,11 @@
+"""Difference Transformer."""
+
 import numpy as np
+from numba import njit
 
 from aeon.transformations.series.base import BaseSeriesTransformer
 
-__maintainer__ = ["TinaJin0228"]
+__maintainer__ = ["TinaJin0228", "Alex Banwell"]
```
TonyBagnall: this should be your github name

aeon/transformations/series/_diff.py
```diff
@@ -90,3 +94,96 @@ def _transform(self, X, y=None):
...
         return Xt
+
+    def _inverse_transform(self, X, y=None):
+        """
+        Inverse transform to reconstruct the original time series.
+
+        Parameters
+        ----------
+        X : Time series to inverse transform. With shape (n_channels, n_timepoints).
+        y : ignored argument for interface compatibility
```
TonyBagnall: this comment on y is not now true

aeon/transformations/series/_diff.py
```diff
@@ -90,3 +94,96 @@ def _transform(self, X, y=None):
...
+
+        Returns
+        -------
+        Xt : np.ndarray
+            Reconstructed original time series.
+        """
+        if y is None or y.shape[1] < self.order:
+            raise ValueError(
```
TonyBagnall: so you could store the difference values in fit, then if y is none you could use them, i.e. assume the series being inverted is the same as the one in fit?

TonyBagnall (changes requested):
could you take out the inverse of difference from this PR and make a separate one? I'm not sure atm how to do this to be consistent with the general interface, so it wpould be better as a separate PR imo

TonyBagnall (approved):
actually want this in, we can discuss the differencer at a later date

aeon/transformations/series/_diff.py
alexbanwell1: Yeah that sounds like a good idea! I'll make those changes in a bit!

aeon/transformations/series/_diff.py
alexbanwell1: Deferred to #3001

aeon/transformations/series/_diff.py
alexbanwell1: Deferred to #3001

PR #3002, opened 2025-08-05
[ENH] Deprecate MatrixProfile collection transformer and MPDist

### LINKED ISSUE #833: [ENH] Rationalise matrix profile transformers

TonyBagnall:
resolves #833 

after discussion, we have decided to remove our bespoke implementations of matrix profile to rely instead on those in the stumpy package, which are better maintained and more comprehensive than ours. This involves

deprecate MatrixProfile in collections
deprecate MPDist in distances
rename MatrixProfileSeriesTransformer -> MatrixProfileTransformer 

we could call it just MatrixProfile, but most others have the suffix.

### DIFF
--- a/aeon/distances/_mpdist.py
+++ b/aeon/distances/_mpdist.py
@@ -3,6 +3,7 @@
 from typing import Optional, Union
 
 import numpy as np
+from deprecated.sphinx import deprecated
 from numba import njit
 from numba.typed import List as NumbaList
 
@@ -10,6 +11,14 @@ from aeon.utils.conversion._convert_collection import _convert_collection_to_num
 from aeon.utils.validation.collection import _is_numpy_list_multivariate
 
 
+# TODO: remove in v1.5.0
+@deprecated(
+    version="1.3.0",
+    reason="mp_distance will be removed in v1.5.0. We recommend using the stumpy "
+    "version instead, see "
+    "https://stumpy.readthedocs.io/en/latest/api.html#stumpy.mpdist.",
+    category=FutureWarning,
+)
 def mp_distance(x: np.ndarray, y: np.ndarray, m: int = 0) -> float:
     r"""Matrix Profile Distance.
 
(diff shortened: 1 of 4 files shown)


aeon/distances/_mpdist.py
```diff
@@ -3,13 +3,22 @@
...
 
 
+# TODO: remove in v1.5.0
+@deprecated(
+    version="1.3.0",
+    reason="mp_distance will be removed in v1.5.0. We recommend using the stumpy "
+    "version instead, see "
+    "https://stumpy.readthedocs.io/en/latest/api.html#stumpy.mpdist.",
+    category=FutureWarning,
+)
```
MatthewMiddlehurst, line 17: Can we actually use this in estimators etc with our functionality? May be worth creating an example at some point

aeon/transformations/collection/_matrix_profile.py
```diff
@@ -192,6 +193,13 @@ def _stomp_self(ts, m):
     return mp
 
 
+# TODO: remove in v1.5.0
+@deprecated(
+    version="1.3.0",
+    reason="MatrixProfile will be removed in v1.5.0. Instead, use the "
+    "MatrixProfileSeriesTransformer and the SeriesToCollectionBroadcaster.",
+    category=FutureWarning,
+)
```
MatthewMiddlehurst: This points to another deprecated class

aeon/transformations/series/_matrix_profile.py
```diff
@@ -1,11 +1,20 @@
...
 from aeon.transformations.series.base import BaseSeriesTransformer
 
 
+# TODO: remove in v1.5.0
+@deprecated(
+    version="1.3.0",
+    reason="MatrixProfileSeriesTransformer will be renamed MatrixProfileTransformer in "
+    "v1.5.0.",
+    category=FutureWarning,
+)
```
MatthewMiddlehurst: I would implement `MatrixProfileTransformer` now, you can just have `MatrixProfileSeriesTransformer` inherit from it or duplicate the code I dont really care.

aeon/distances/_mpdist.py
TonyBagnall: dont know, probably not.

aeon/transformations/series/_matrix_profile.py
TonyBagnall: done

PR #3005, opened 2025-08-06
[ENH] forecast broadcaster for applying univariate forecasters to multivariate time series

### LINKED ISSUE #3004: [ENH] Forecaster broadcaster for multivariate forecasting

wrapper class to broadcast a univariate forecaster to each channel independently. 

simple wrapper, but we will need to resolve that predict and forecast return an array rather than a single value in testing

_No response_

### Additional context

_No response_

TonyBagnall:
resolves #3004 

This branch adds BroadcastForecaster, a lightweight wrapper that applies a univariate BaseForecaster independently to each channel of a multivariate series. The wrapper clones
  the supplied forecaster once per channel, fits each clone on the corresponding channel, and returns channel-wise forecasts. For univariate input it preserves the existing
  scalar output convention.

  One design issue to flag: for multivariate input, predict necessarily returns a vector of shape (n_channels,), one forecast per channel. That is new behaviour relative to the
  current generic forecasting checks, which assume predict always returns a scalar. I have updated the checker locally so multivariate forecasters may return one value per
  channel, but this is a broader base-class contract question and may need a separate discussion before merge.

  Related base-class/API points that may need resolving outside this PR:

  - BaseForecaster.predict docstring currently says it returns a float, which is too narrow for multivariate forecasting.
  - check_forecaster_output currently encodes the scalar-only assumption.
  - The forecasting API may need an explicit convention for multivariate output shape, likely (n_channels,) for single-horizon forecasts.
  - We may need to decide whether all multivariate-capable forecasters should follow that convention, or whether broadcasters/wrappers are a special case.


MatthewMiddlehurst:
Needs some tests IMO as we dont really have anything that does multivariate currently in forecasting. Possibly a more descriptive name.

TonyBagnall:
> Needs some tests IMO as we dont really have anything that does multivariate currently in forecasting. Possibly a more descriptive name.

any ideas on the name? still draft, not written tests yet, need to decide on whether output polymorphism is an issue or not

TonyBagnall:
oh, not draft!

MatthewMiddlehurst:
Maybe just `MultivariateBroadcastForecaster`

PR #3007, opened 2025-08-06
[BUG] Issue with axis in `TCNNetwork`

hadifawaz1999:
The network had issues in docstrings flipping channels and time axes. Also the last step was taking the last channel not the last time stamp

### DIFF
--- a/aeon/networks/_tcn.py
+++ b/aeon/networks/_tcn.py
@@ -133,7 +133,7 @@ class TCNNetwork(BaseDeepLearningNetwork):
         Parameters
         ----------
         input_tensor : tf.Tensor
-            Input tensor of shape (batch_size, channels, sequence_length).
+            Input tensor of shape (batch_size, sequence_length, channels).
         chomp_size : int
             Number of time steps to remove from the end.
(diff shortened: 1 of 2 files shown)


hadifawaz1999:
Made sure original code before bug fix is failing on the added test

```FAILED aeon/networks/tests/test_tcn.py::test_tcn_final_dense_weights_shape - AssertionError: Expected input dim 64, got 40```

File: aeon/networks/_tcn.py [hunk 1]
```diff
@@ -322,6 +322,6 @@ def build_network(self, input_shape: tuple, **kwargs) -> tuple:
             kernel_size=self.kernel_size,
             dropout=self.dropout,
         )
-        output = tf.keras.layers.Dense(input_shape[1])(x[:, :, -1])
+        output = tf.keras.layers.Dense(input_shape[1])(x[:, -1, :])
```
lucifer4073, line 325: Yes, this seems to be correct. Instead of choosing the last feature and all timepoints we are choosing the last timepoint and all feature dimensions.

PR #3011, opened 2025-08-07
[ENH] Convert ETS to automatically optimise for the alpha, beta, gamma and phi parameters

alexbanwell1:

Removes alpha, beta, gamma and phi parameters from the ETS constructor and uses nelder-mead optimisation algorithm to fit them.

##### For all contributions

### DIFF
--- a/aeon/forecasting/stats/_ets.py
+++ b/aeon/forecasting/stats/_ets.py
@@ -14,6 +14,11 @@ import numpy as np
 from numba import njit
 
 from aeon.forecasting.base import BaseForecaster, IterativeForecastingMixin
+from aeon.forecasting.utils._loss_functions import (
+    _ets_fit,
+    _ets_initialise,
+    _ets_predict_value,
+)
 
 ADDITIVE = "additive"
 MULTIPLICATIVE = "multiplicative"
@@ -189,6 +194,7 @@ class ETS(BaseForecaster, IterativeForecastingMixin):
         )
         data = y.squeeze()
         (
+            self.aic_,
             self.level_,
             self.trend_,
             self.seasonality_,
@@ -198,8 +204,7 @@ class ETS(BaseForecaster, IterativeForecastingMixin):
             self.avg_mean_sq_err_,
             self.liklihood_,
             self.k_,
-            self.aic_,
-        ) = _numba_fit(self.parameters_, data, self.model_)
+        ) = _ets_fit(self.parameters_, data, self.model_)
         self.forecast_ = _numba_predict(
             self._trend_type,
             self._seasonality_type,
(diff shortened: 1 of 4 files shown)


TonyBagnall:
perhaps we should also allow the string inputs for multiplicative and additive that statsmodels and statsforecast use? 
options = ("add", "mul", "additive", "multiplicative")
is statsmodels

TonyBagnall:
I've not isolated the case, but when running experiments I get this

                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Code\aeon\local\ets_evaluate.py", line 235, in run_ets_experiment
    aeon_model.fit(y_train)
  File "C:\Code\aeon\aeon\forecasting\base.py", line 97, in fit
    self._fit(y, exog)
  File "C:\Code\aeon\aeon\forecasting\stats\_ets.py", line 185, in _fit
    (self.parameters_, self.aic_) = nelder_mead(
                                    ^^^^^^^^^^^^
ZeroDivisionError: division by zero

TonyBagnall:
inside nelder mead 


        (self.parameters_, self.aic_) = nelder_mead(
            1,
            1 + 2 * (self._trend_type != 0) + (self._seasonality_type != 0),
            data,
            self._model,
            max_iter=self.iterations,
        )

its in the loss function, happens if the trend goes negative in multiplicative models with phi < 1.0

        damped_trend = trend**phi

or generally if the forecast overflows 1xe300. In both cases it become inf or negative inf. Several possible solutions. The one I like is to just return infinite cost for the model if that happens. The optimiser wont choose it, and if it can find an alternative, it would be ok.

needs more explanation though, no need to hold this up for it

TonyBagnall:
so the divide by zero is due to overflow or negative forecasts with multiplicative models. I suspect ETS is not really suitable for longer series, since the trend tends to blow up. But the Auto version should mitigate this somewhat. In terms of run time its very good

image

TonyBagnall:
statsmodels seems consistently better, which is weird. Might be the strange data. We could squeeze some more out of this implementation, maybe splitting the loss function into nelder_mead and final fit version as with arima, but its tiny, just skip some lines like this

        avg_mean_sq_err_ += (time_point - fitted_value) ** 2

aeon/forecasting/stats/_ets.py
```diff
@@ -99,33 +105,35 @@ def __init__(
...
-        self.gamma = gamma
-        self.phi = phi
         self.forecast_val_ = 0.0
         self.level_ = 0.0
         self.trend_ = 0.0
         self.seasonality_ = None
-        self._beta = beta
-        self._gamma = gamma
         self.error_type = error_type
+        self._error_type = error_type
```
TonyBagnall: you dont need to declare these in the constructor, I think its a bit confusing, they are private and set in fit

TonyBagnall (approved):
one tiny change

PR #3012, opened 2025-08-07
[ENH] TAR forecaster

### LINKED ISSUE #2817: [ENH] Implement Threshold Autoregressive model(s)

We want an implementary of SETAR-Tree (#2816), but it would probably be useful to have implementations of other TAR variants. A paragraph from the SETAR-Tree paper lists a few:

> Piecewise linear models such as Threshold Autoregressive (TAR) models (Tong, [1978](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR80); [1993](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR81); Terasvirta, [1994](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR78)) have greatly contributed to the econometric and forecasting fields during the past few decades. These models partition the instance space of a particular single time series to be forecast into several subspaces where each subspace is modeled by a separate Autoregressive (AR) model. There are many variations of TAR models such as Self Exciting Threshold Autoregressive (SETAR, Tong, [1993](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR81)), Smooth Transition Autoregressive (STAR, Terasvirta, [1994](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR78)), Exponential STAR (ESTAR), Logistic STAR (LSTAR) and Neuro-Coefficient STAR (Medeiros & Veiga, [2000](https://link.springer.com/article/10.1007/s10994-023-06316-x#ref-CR60)) models.

Implement TAR based forecasters.

_No response_

### Additional context

If anyone wants to take this on, suggest individual algorithms and we can create a list in the issue.

TonyBagnall:
Resolves #2817

Implements a numba TAR and AutoTAR forecaster. TAR is simple, if not a bit odd to me.

Given or find a threshold value. All points where previous delay point was below the threshold go in one series ("regime"), all values above in another. Fit separate AR models to each.

Also refactors TVPForecaster to just TVP for consistency

### DIFF
--- a/aeon/forecasting/stats/_tar.py
+++ b/aeon/forecasting/stats/_tar.py
@@ -12,12 +12,13 @@ from aeon.forecasting.base import BaseForecaster, IterativeForecastingMixin
 
 
 class TAR(BaseForecaster, IterativeForecastingMixin):
-    r"""Threshold Autoregressive (TAR) forecaster with fixed parameters.
+    r"""Threshold Autoregressive (TAR) [1] forecaster with fixed parameters.
 
-    Two regimes split by a threshold :math:`r` on the variable :math:`z_t=y_{t-d}`:
-    observations with :math:`z_t \le r` follow the **below/left** AR model, and
-    those with :math:`z_t > r` follow the **above/right** AR model. Each regime is
-    fit by OLS. **No parameter optimisation/search** is performed.
+    A series is split into two series (regimes) by a threshold :math:`r` on the
+    variable :math:`z_t=y_{t-d}`: observations with :math:`z_t \le r` follow the
+    **below/left** AR model, and those with :math:`z_t > r` follow the
+    **above/right** AR model. Each regime is fit by Ordinary Least Squares. No
+    parameter optimisation/search is performed.
 
     Defaults:
     - ``delay=1``
@@ -53,11 +54,15 @@ class TAR(BaseForecaster, IterativeForecastingMixin):
     params_ : dict
         Snapshot of configuration and a simple AIC diagnostic.
 
+    See Also
+    --------
+    AutoTAR
+
     References
     ----------
-    Tong, H., & Lim, K. S. (1980).
+    .. [1] Tong, H., & Lim, K. S. (1980).
     Threshold autoregression, limit cycles and cyclical data.
-    *JRSS-B*, 42(3), 245–292.
+    RSS-B, 42(3), 245–292.
     """
 
     def __init__(


TonyBagnall:
comparing to our current optimised for speed ARIMA and current SETAR tree. MSE on simulated low order ARMA. Cheating a bit as giving TAR the correct AR order (doesnt fit MA though). 

image

runtime
image

super fast TAR, but also super simple. Not trying to make the definitive version, if someone is a fan of the forecaster they might do so, Im only interested in potential components. Also shows we can optimise SETAR tree.

pyproject.toml
alexbanwell1, line 1: Presumably this is a hangover from something? Don't think this should be commented out!

aeon/forecasting/stats/tests/test_tar.py
alexbanwell1, line 1: Might be good to add some numerical tests to test the forecasted value is equal to an expected value as that usually catches anything that breaks

aeon/forecasting/stats/_tar.py
alexbanwell1, line 1: Is the numba OLS function something that would be useful elsewhere? If so would it be better to put it in its own file in utils? Just thinking that seems like quite a common thing to want to do and if we now have a very fast numba version it would be good to make use of it!

TonyBagnall: yes good point, it is used in Theta too

TonyBagnall: I will refactor when Theta and that is in, too messy atm

TonyBagnall:
so this has got more complex. It now includes TAR: fixed parameters with optional defaults and AutoTAR, which finds the parameters. ChatGPT gave me a much more efficient way of auto fitting based on storing sum of squares (reminded me of shapelets). Its complex, so I have tested this against a grid search using the basic one, output identical (tested for each value of n), run times much faster (30x-70x),
image

For correctness, there are no proper TAR implementations in python to compare against and I will not go into R. It should be as good as AR/ARMA though, which gives us a core experiment 

AutoTAR comparable in ARMA data, which you would hope. 
image
Interestingly faster than ARMA, which has fixed p and q, which is the fastest ARMA I could find. much faster than our current SETAR, which I will optimise. Setting AR parameters with OLS is very fast

image

aeon/forecasting/stats/tests/test_tar.py
TonyBagnall: added a bunch of tests

alexbanwell1 (approved):
Looks good, as with previous comment can't comment on the correctness of the model, but it certainly appears to be a good forecaster either way!

PR #3032, opened 2025-08-14
[ENH] Add AutoETS Algorithm

alexbanwell1:
Adds AutoETS

No

##### For all contributions


alexbanwell1:
Branch based off AutoARIMA branch, to avoid file conflicts when that is merged. Need to work out how to deal with ZeroDivisionErrors in the loss function, as currently number is not used in the main AutoETS function as it needs to catch any ZDEs

alexbanwell1:
Also Added at the top of the AutoETS forecaster class:    
Attempts to make this forecaster stable:
    - Issues with Zero division Errors:
        - If any data points are non-positive, multiplicative options are excluded.
        - With numba fastmath=true, the compiler moves the operations around. This means
          zero division guards are ineffective. As such I have tried putting the guards
          in a separate fastmath=false function, with inline=true. This slows it down a
          lot though.
        - Need to make sure initialisation function never assigns slices of the data
        array
        - fixed bug where the first few values were being changed in the seasonality
          calculation array
    - Issues with the nelder mead not finding good parameters,
      usually ending up with alpha approx 1:
        - Tested updating the initial conditions (initial level, trend,
          seasonality array) to be heuristically calculated across the whole
          array at the start.
          This seemed to fix some issues, but cause others.
        - Tested optimising over the initial conditions in the nelder-mead array.
          This didn't really help, although is how statsmodels does it.
        - Added guards to the nelder-mead algorithm to reflect points back in when they
          go above or below (0,1) to ensure parameters stay valid.
        - Added sigmoid function to output of nelder-mead to ensure parameters stay
          in (0,1).
        - Initialised the simplex array with reasonable starting values
          (a=0.4, b=0.25, phi=0.95, g=0.35)
    - The algorithms sometimes produce really extreme forecasts of the order of 1e100
      larger than the data. I assume this is due to the guards on the zero division,
      but haven't been able to work out how to fix it.

alexbanwell1:
Superseded by #3253

PR #3033, opened 2025-08-15
[ENH] AR forecaster

TonyBagnall:
if we have TAR and SETAR, seem natural to have AR.

I'll do this with OLS parameter setting. We already have a statsmodels burg version in transformers to benchmark against. A long time ago I implemented the Durban-Levinson method for this, might dig it out to compare. This is not ready, still draft and will wait until TAR is in so I can consolidate common methods


TonyBagnall:
out of date now

PR #3042, opened 2025-08-18
[ENH,DEP] Update `.ts` writer and deprecate old writers

MatthewMiddlehurst:
Updates and renames .ts file writer. I have been using this function recently and it is pretty inconvenient when writing different train and test files for the same problem or folds. It is also missing some of the metadata options. 

Uses all the new(?) validation and conversion functions rather than doing it itself, so now it accepts all the different collections. It all just gets converted to np-list anyway.

Deprecates the old version and arff writer. The arff writer only works for univariate equal length classification and I dont really know if anyone uses the format for time series anymore. There are much better ways for that now and .ts has much more functionality..


aeon/datasets/_data_writers.py
```diff
@@ -1,11 +1,164 @@
+"""Functions to write aeon datasets to file."""
+
+__maintainers__ = ["MatthewMiddlehurst"]
+__all__ = ["write_to_ts_file"]
```
chrisholder: Not sure if you want all the functions included in the __all__

__all__ = ["save_to_ts_file", "write_to_ts_file", "write_to_arff_file"]

aeon/datasets/_data_writers.py
```diff
@@ -1,11 +1,164 @@
...
 
-__all__ = ["write_to_ts_file", "write_to_arff_file"]
+        # start writing data
+        file.write("@data\n")
+        for i in range(n_cases):
+            for j in range(n_channels):
+                series = ",".join(
+                    [str(num) if not np.isnan(num) else "NaN" for num in X[i][j]]
+                )
+                file.write(str(series))
```
chrisholder, line 148: Just to preface this comment I ran the PRs code through an AI after I had looked at it to check for spelling mistakes but it flagged this to me:

"I found a critical issue! The data writing format in the new save_to_ts_file function is incorrect. In the old function, each channel is separated by : but in the new function, the channels are concatenated without separation. Let me fix this:"

It suggested:

```
                series = ",".join(
                    [str(num) if not np.isnan(num) else "NaN" for num in X[i][j]]
                )
                file.write(str(series))
                if j < n_channels - 1:  # Add separator between channels
                    file.write(":")
```

This could be complete nonsense but thought i'd mention it just in case

chrisholder:
Looks good. I have suggested two things but both can be ignored if you think they arent needed.

aeon/datasets/_data_writers.py
MatthewMiddlehurst: just the new one I think. others are deprecated. Not sure it really does anything practically, just for infromation.

aeon/datasets/_data_writers.py
MatthewMiddlehurst: Thanks, insufficient testing. Added a fix and new tests.

MatthewMiddlehurst:
Was broken yes,good catch. Try not to approve with stuff like that though.

aeon/datasets/_data_writers.py
```diff
@@ -1,11 +1,164 @@
+"""Functions to write aeon datasets to file."""
+
+__maintainers__ = ["MatthewMiddlehurst"]
+__all__ = ["save_to_ts_file"]
```
TonyBagnall, line 4: the file name should perhaps change to data_savers

TonyBagnall (approved):

PR #3047, opened 2025-08-20
[DEP] Remove `update` function from base transformer and small tidying

MatthewMiddlehurst:
Nothing uses `update` and I do not think we want to support it currently.

Nothing uses the return from `_fit`, it is always called in `fit` where the output is ignored.


TonyBagnall:
The comment
        Writes to self:
        _is_fitted : flag is set to True.

needs changing

TonyBagnall:
personally, I would rather reserve ellipsis for abstract methods and use pass for empty functions such as _fit, but thats just my preference

TonyBagnall (dismissed):

hadifawaz1999 (dismissed):

MatthewMiddlehurst:
Made the changes. Also editing files through GitHub now auto generates commit descriptions using AI, weird

PR #3074, opened 2025-09-17
[BUG] FIX DRNN load bug

PipaFlores:
##### For all contributions

### DIFF
--- a/aeon/clustering/deep_learning/_ae_dcnn.py
+++ b/aeon/clustering/deep_learning/_ae_dcnn.py
@@ -319,6 +319,37 @@ class AEDCNNClusterer(BaseDeepClusterer):
 
         return self
 
+    def load_model(self, model_path, estimator):
+        """Load a pre-trained keras model instead of fitting.
+
+        When calling this function, all functionalities can be used
+        such as predict, predict_proba etc. with the loaded model.
+
+        Parameters
+        ----------
+        model_path : str (path including model name and extension)
+            The directory where the model will be saved including the model
+            name with a ".keras" extension.
+            Example: model_path="path/to/file/best_model.keras"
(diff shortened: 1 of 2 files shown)


MatthewMiddlehurst:
Is there an issue with the inherited function? If so, wonder why it is not captured by testing. No description to go by.

PipaFlores:
I don't know why my message did not appear in the original post. I did not notice

Yes, the load function is inherited from BaseDeepClusterer, but it uses the basic functionality to "simply" load the model. 

The Dilated Recurrent NN autoencoder uses a custom module, which needs to be provided to the load function to work properly. That is the change made. This is not necessary for the other deepclusterers because they do not have any particularly special modules.

To replicate the bug you can just try training an AEDRNN clusterer on any number of epochs, saving, and loading.

hadifawaz1999 (dismissed):
Good catch, we do this in the end of the fit to load the best model, but didnt think about this other base function, but it also shows we dont test the function... approving this if @MatthewMiddlehurst is okay too cz it solves the issue, but if you're also motivated to PRing a test for the functions of save load in deep clustering would be great too :)

PipaFlores:
Glad to know it might help. 

In a similar sense than the previous PR on adding the validation_split argument to the deepclusterers, I am only modifying the code that is useful to my research. If I eventually have time, I can get to the testing units, for know I keep to the practical. But if any other contributor find this, feel free to implement, I am sure is pretty straightforward.

PipaFlores:
Added the same fix to DCNN. I did not notice the custom class until I stepped into the bug, as it was not in the network script, as in the DRNN, but in another module utils.networks.weight_norm.

Both should work fine now, as they are for me.

hadifawaz1999 (approved):
Added the issue here #3080

PR #3099, opened 2025-10-28
[ENH] MCM plot function tidy.

MatthewMiddlehurst:
This is already an 'MCM Lite' function so removing some extra bits which may confuse users and make sense in the original package but perhaps not aeon. Also does a bit of tidying and fixes typos.

### DIFF
--- a/aeon/visualisation/results/_mcm.py
+++ b/aeon/visualisation/results/_mcm.py
@@ -50,7 +50,7 @@ def create_multi_comparison_matrix(
 
     Note: this implementation uses different pvalue parameters from the original
     by default. To use the original parameters, set ``pvalue_test_params`` to
-    ``{"zero_method": "pratt", "alternative": "two-sided"}`` and 
+    ``{"zero_method": "pratt", "alternative": "two-sided"}`` and
     ``pvalue_correction`` to ``None``.
 
     Parameters


MatthewMiddlehurst:
Currently the docs for `pvalue_test_params` does not match the actual default I believe. Would like advice on which one to go with.

aeon/visualisation/results/_mcm.py
```diff
@@ -43,20 +40,24 @@ def create_multi_comparison_matrix(
...
+    show_symmetry=True,
 ):
     """Generate the Multi-Comparison Matrix (MCM) [1]_.
 
     MCM summarises a set of results for multiple estimators evaluated on multiple
     datasets. The MCM is a heatmap that shows absolute performance and tests for
     significant difference. It is configurable inmany ways.
 
+    Note: this implementation uses different pvalue parameters from the original
+    by default. To use the original parameters, set ``pvalue_test_params`` to
```
hadifawaz1999, line 52: I would also mention that the Holm is not the default of the paper not only wilcoxon params

MatthewMiddlehurst: good catch, forgot to update the note after changing the default.

PR #3104, opened 2025-11-14
[ENH] Change default dataset download path to user home directory

### LINKED ISSUE #3088: [ENH] Storage of TSC datasets when loading from a package using import

writing labs on machine learning using aeon for data loading. 

```python
from aeon.datasets import load_gunpoint, load_from_ts_file
X, y = load_gunpoint(split="train", return_type="numpy2D")
print(X.shape, y.shape)
```
I have only ever done it within aeon, and wondered where the actual ts file gets placed. Turns out it goes in the environment. 

Image

I dont think it should go there!

all to do with

Image
caused by this, which looks for the default path for aeon, which is in the env. Not sure whats best here

Image

_No response_

### Additional context

_No response_

satwiksps:
# [ENH] Change default dataset download path to user home directory

## What does this implement/fix?
- This PR resolves #3088.

Currently, `aeon` downloads datasets into the installed package directory (`site-packages` / `venv`) by default. This causes issues with virtual environment persistence, permissions, and makes it difficult for users to locate the downloaded data.

**This change updates the default download behavior to use the user's home directory (`~/.aeon/data`) while preserving backward compatibility for bundled datasets used in CI/tests.**

## Updated files:
* `aeon/datasets/_data_loaders.py`: Updated logic in `_load_tsc_dataset`, `load_classification`, `load_forecasting`, `load_regression`, and `download_all_regression` to prioritize bundled data, then default to the user home directory. Updated `_download_and_extract` to support this default.
* `aeon/datasets/tests/test_data_loaders.py`: Updated `shutil.rmtree` cleanup calls to `ignore_errors=True` to prevent test failures when the deprecated `local_data` folder is not created.

## Summary
The logic for determining `local_module` (where data is loaded/saved) has been updated with a hybrid check:
1.  **Check Bundled:** It first checks if the dataset exists inside the package (`aeon/datasets/data`). If found (e.g., `UnitTest`), it uses that path. This ensures CI and unit tests remain green.
2.  **Check/Download User Home:** If not found internally, it defaults to `Path.home() / ".aeon" / "data"`.
3.  **User Override:** If the user provides a specific `extract_path`, that path is respected and used.

## Does your contribution introduce a new dependency?
No

## Any other comments?
I have verified this locally by running `load_gunpoint()` (which now correctly downloads to `~/.aeon`) and by running the unit tests `pytest aeon/datasets/tests/test_data_loaders.py` (which pass).

## PR checklist

- [x] PR title begins with [ENH]
- [ ] I've added myself to the list of contributors. Alternatively, I will use the @all-contributors bot after the PR is merged.


satwiksps:
### CI Failures Summary

Hi! @MatthewMiddlehurst  @TonyBagnall 

The CI reports several failing jobs, but the errors seem unrelated to the changes in this PR.

### **Details**
- The failures occur in tests such as `test_kshapes` and a small numerical mismatch in a `ROCKAD` doctest.
- These belong to clustering and anomaly-detection modules, which this PR does **not** touch.
- All dataset-related tests pass successfully.
- No path-handling issues or import errors appeared, and local tests for the modified files succeeded.

### **Context**
This PR only updates dataset directory handling to use `pathlib.Path`, so it should not affect the modules where the failures occurred.  
The failing tests look like numerical-sensitivity issues on newer Python/NumPy versions rather than something caused by this change.

### **Request**
Could a maintainer please confirm whether these CI test failures are expected on the current Python versions?  
I’m happy to adjust anything if required, but the issues seem independent of this PR.

Thank you! 🙏

Nithurshen:
The CI failures is common to all the PRs. It would be fixed.

MatthewMiddlehurst:
Was there an issue with this PR?

satwiksps:
> Was there an issue with this PR?

Hi, thanks for following up @MatthewMiddlehurst 

I closed the previous PR because, after reviewing it again, I realised that the change I made (replacing `os.path` with `Path`) did not actually fix the underlying issue raised in #3088 
It only modernised the path code but did not change the dataset storage location.

The real problem is that downloaded TSC datasets are currently saved inside the package directory (`site-packages/aeon/datasets/data/`), which is not ideal because package directories should be treated as read-only and can cause permission issues across environments.

For the proper fix, I’m planning to move the default dataset storage location to a user-writable directory using `platformdirs`.
On Windows this resolves to something like - `C:\Users\satwiksps\AppData\Local\aeon\aeon-toolkit\datasets\` which avoids polluting the installation directory.

Before I start the new PR, could you please confirm that `user_data_dir("aeon", "aeon-toolkit") / "datasets"` is the preferred default location for TSC dataset downloads ?

I’ll proceed as soon as I have your confirmation.

MatthewMiddlehurst:
It would be better to edit your current PR rather than create a new one for changes like this. Rather than AppData it should probably be wherever the code is run from right? i.e. if you run `code/my_code.py` it should create `code/aeon_data/my_data.ts` or something. Up for discussion.

If you don't mind answering, why did you initially only change the method to find the path? Just curious whether it was an problem with the issue.

satwiksps:
> It would be better to edit your current PR rather than create a new one for changes like this. Rather than AppData it should probably be wherever the code is run from right? i.e. if you run `code/my_code.py` it should create `code/aeon_data/my_data.ts` or something. Up for discussion.

I’m happy to continue with the existing PR instead of opening a new one. I closed it only because I thought the fix was incomplete and possibly misleading.

About the default location:
I’m open to storing the data relative to the working directory (e.g. `./aeon_data/...`) as you suggested. My initial thought was to use a user-writable platform directory (e.g. `AppData/Local/aeon/... on Windows, ~/.local/share/aeon/... on Linux`) since many ML libraries (e.g., HuggingFace, torchtext) follow that pattern, but using a project-local folder like `./aeon_data` also has advantages for reproducibility and portability.

I’m happy to implement this and I’ll update the PR accordingly.

satwiksps:
> If you don't mind answering, why did you initially only change the method to find the path? Just curious whether it was an problem with the issue.

Initially, I interpreted the issue as primarily a problem with how the dataset directory was being resolved, not where it should live. The use of `os.path.dirname(aeon.__file__)` sometimes resolves to unexpected paths under editable instals, venvs, or when the package is symlinked.

So my first instinct was to fix the path resolution method to reduce those inconsistencies. Only after digging deeper into the datasets module did I realize the real problem is not about resolution, but about location.

So my first PR only improved the mechanics of path construction, not the logic of choosing the correct base directory. Once I connected that the issue is actually about selecting a new default storage root, I realized the initial fix wasn’t addressing the underlying problem so I stepped back to rethink the design.

MatthewMiddlehurst:
Interesting to hear that other packages put it in AppData. May be worth asking on Slack if there is a preference? I think for now if you want to edit the PR just do it locally as its an improvement over the current method, and we can change it later if necessary.

aeon/datasets/tests/test_data_loaders.py
```diff
@@ -57,7 +57,7 @@ def test_load_forecasting_from_repo():
     assert not meta["contain_missing_values"]
     assert not meta["contain_equal_length"]
 
-    shutil.rmtree(os.path.dirname(__file__) + "/../local_data")
+    shutil.rmtree(os.path.dirname(__file__) + "/../local_data", ignore_errors=True)
```
MatthewMiddlehurst, line 60: Why do we need to ignore errors here now?

satwiksps:
Closing this as it has been superseded by PR #3304

PR #3109, opened 2025-11-16
[ENH] Optimize ElasticEnsemble _fit to avoid redundant cross-validation

### LINKED ISSUE #2854: [ENH] Implement optimisation to `ElasticEnsemble` grid search

From the code comments:

```
            # once the best parameter option has been estimated on the
            # training data, perform a final pass with this parameter option
            # to get the individual predictions with cross_cal_predict (
            # Note: optimisation potentially possible here if a GridSearchCV
            # was used previously. TO-DO: determine how to extract
            # predictions for the best param option from GridSearchCV)
            else:
                best_model = KNeighborsTimeSeriesClassifier(
                    n_neighbors=1,
                    distance=this_measure,
                    distance_params=grid.best_params_["distance_params"],
                    n_jobs=self._n_jobs,
                )
                preds = cross_val_predict(
                    best_model, full_train_to_use, y, cv=LeaveOneOut()
                )
                acc = accuracy_score(y, preds)
```

Explore if the grid search and/or random parameter search can be optimised to avoid doing an extra cross-validation when weighting.

_No response_

### Additional context

This can be done in a few ways possible, dont know what is correct:
- Redo CV if `self.proportion_train_in_param_finding` is less than 1
- Use the subsamples train set to generate accuracy estimate and always avoid an extra CV

Nithurshen:
#### Reference Issues/PRs

Fixes #2854

This PR optimizes the `_fit` method in the `ElasticEnsemble` classifier to eliminate a redundant cross-validation step, speeding up the fitting process under specific conditions.

### Problem

The original implementation first used `GridSearchCV` or `RandomizedSearchCV` to find the best parameters for a given distance measure. Then, to get the accuracy score for weighting the ensemble, it performed a second, separate `cross_val_predict` using those best parameters.

This second `cross_val_predict` step was redundant when the initial parameter search was already performed on the entire training set (`proportion_train_in_param_finding == 1.0`), as it was essentially re-running the same validation.

### Solution

I've introduced a new logic path that runs only when **`self.proportion_train_in_param_finding == 1.0`** and **`not self.majority_vote`**.

The modification:

1.  **Bypasses `GridSearchCV`/`RandomizedSearchCV`** entirely.
2.  Manually iterates through the parameter grid (respecting `self.proportion_of_param_options` to mimic the randomized search behavior).
3.  Runs `cross_val_predict` once for each parameter set inside a single loop, calculating the accuracy.
4.  Keeps track of the best accuracy score and the corresponding best parameters as it iterates.

This change combines the parameter search and the accuracy-for-weighting calculation into a single loop, completely removing the redundant N-fold cross-validation pass.

The original logic (using `GridSearchCV` + the second `cross_val_predict`) is fully preserved for all other cases (i.e., when subsampling for parameter finding or when `majority_vote` is `True`), ensuring no existing behavior is broken.

No.

##### For all contributions
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance. ##### For new estimators and functions


Nithurshen (commented):
.

docs/changelogs/v1.3.md
MatthewMiddlehurst, line 1: Don't change this

Nithurshen:
> Do we need both options? Can we not just what is there already?
> 
> Please ensure that the output is still the same, or if it is different why.

Regarding the two options, The manual loop is necessary for the optimization because `GridSearchCV` only exposes the final `best_score_`. For `ElasticEnsemble` weighting, we need the actual individual predictions from the best parameter set.

By capturing these predictions manually during the search phase (for the `prop_train == 1.0` case), we can calculate the accuracy weighting immediately. This is what allows us to skip the second, redundant N-fold `cross_val_predict` pass that the current implementation performs.

I have verified that the output is identical.

docs/changelogs/v1.3.md
Nithurshen: Done

MatthewMiddlehurst:
> we need the actual individual predictions from the best parameter set

do we though? cant we just store the best score then add an extra elif to the majority vote check for if we already have it.

> I have verified that the output is identical.

please post the code you used to check this, or the test that would cover this scenario

PR #3111, opened 2025-11-16
[MNT] Add missing load_model test for deep clusterers (Fixes #3080)

### LINKED ISSUE #3080: [MNT] Add testing for loading/saving for deep clusterers
### Describe the issue

In the deep clustering module there is the possibility of loading a pre-trained model, however the function is not tested, which lead to the bug that was fixed in #3074 

Add a test for this functionality

### Additional context

_No response_

satwiksps:
# [MNT] Add missing load_model test for deep clusterers

## Reference Issues/PRs
- Fixes #3080

This PR adds a missing unit test for the deep clustering autoencoder models (`AEDRNNClusterer` and `AEDCNNClusterer`) to ensure that the `load_model` functionality works as expected.

Issue #3080 identified that while deep clusterers support loading a pre-trained Keras model, this behavior was not covered by any tests. This lack of test coverage allowed a bug fixed in PR #3074 to remain undetected.

This PR introduces a new test file, `test_deep_clusterer_io.py`, which:

- Trains each deep autoencoder clusterer on a tiny synthetic dataset.
- Ensures that a `.keras` checkpoint is saved via `save_best_model=True`.
- Calls `load_model` to load the saved model, including custom layers
  (`_TensorDilation` for AEDRNN and `_WeightNormalization` for AEDCNN).
- Verifies that the loaded model is valid and assigned to `model_`.
- Confirms that the load path added in PR #3074 works without errors.

The test does **not** call `predict()` on the loaded estimator, because estimator state
is intentionally not restored by `load_model` and would correctly raise `NotFittedError`.
This matches the design of the deep clusterer classes.

This adds the required IO coverage to ensure regressions do not reoccur.

**No** new dependencies are introduced.
The test is guarded by a TensorFlow soft dependency check.
It will be skipped if TensorFlow is not installed, consistent with other deep learning tests.

## Any other comments?

- The test is placed in a dedicated file (`test_deep_clusterer_io.py`) to keep IO-specific behavior separate from existing feature and base tests.
- The test follows the same pattern as other deep learning tests within aeon.

### For all contributions
- [x] The PR title starts with `[MNT]`, matching aeon’s conventions.

- Not applicable.

- Not applicable.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
         args: [ "--create", "--python-folders", "aeon" ]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.4
+    rev: v0.14.5
     hooks:
       - id: ruff
         args: [ "--fix"]
@@ -53,7 +53,7 @@ repos:
         args: [ "--max-line-length=88", "--extend-ignore=E203" ]
 
   - repo: https://github.com/psf/black
-    rev: 25.9.0
+    rev: 25.11.0
     hooks:
       - id: black
         language_version: python3


hadifawaz1999 (changes requested):
Thanks for taking care of this issue. Two comments for now:

- The test should be for all models under clustering/deep_learning
- The test should use tempfile (see [this](https://github.com/aeon-toolkit/aeon/blob/eaa51b40dc185187bf4ba6bb5109b7212d079da8/aeon/classification/deep_learning/tests/test_deep_classifier_base.py#L62) for example)

satwiksps:
> You should be able to call predict because when "fit" is called on a deep clusterer, it creates an instance self._estimator for the clustering estimator that you can fetch after calling fit and give it to the load_model

I've updated the test as suggested:
- Now calling model.fit(X) first and retrieving model._estimator (so the test uses the actual trained estimator created during fit)
- Passing the trained estimator to loaded.load_model(....)
- Added a predict(X) check to ensure the restored estimator works end-to-end

All tests pass locally with TensorFlow enabled.

hadifawaz1999 (approved):

hadifawaz1999:
LGTM ! thanks

PR #3106, opened 2025-11-15
[ENH] Add prediction intervals to ARIMA iterative_forecast (#3100)

satwiksps:
# [ENH] Add prediction intervals to ARIMA iterative_forecast (#3100)

#### Reference Issues/PRs
- Fixes #3100.

---

#### What does this implement/fix?

This PR adds **optional prediction interval support** to the `ARIMA.iterative_forecast` method in `aeon.forecasting.stats._arima`.

**Key features:**

- Adds a new optional parameter:  
  **`alpha` (default: None)**  
- **Backward compatible**:
  - If `alpha=None` → returns the same `np.ndarray` as before.
  - If `alpha` is provided → returns a `pandas.DataFrame` with columns:
    - `"mean"` - point forecasts  
    - `"lower"` - lower prediction interval  
    - `"upper"` - upper prediction interval  
- Prediction intervals are computed using:
  - Empirical residual standard deviation  
  - Normal quantiles via `scipy.stats.norm.ppf`  
  - Standard error scaling with `sqrt(horizon)`  

Additional updates:
- `AutoARIMA.iterative_forecast` now forwards the `alpha` argument to the wrapped ARIMA model.
- Added new unit tests covering prediction intervals.
- All existing tests pass with no modification.

This enhancement provides users with basic forecast uncertainty estimates while keeping the ARIMA implementation lightweight and backward compatible.

---

#### Does this contribution introduce a new dependency?

No.  
Uses existing NumPy, SciPy, and pandas packages already in the project.

---

#### Any other comments?

- All pre-commit hooks pass (`ruff`, `flake8`, `isort`, `black`, etc.).
- Tests pass locally, including new interval tests.
- Changes are intentionally minimal to avoid regressions.

---

##### For all contributions
- [ ] I've added myself to the list of contributors (after merge).
- [x] The PR title starts with `[ENH]`.

##### For new estimators/functions
- [ ] Not applicable - this PR modifies an existing estimator.

##### For developers with write access
- [ ] Not applicable.

PR #3116, opened 2025-11-17
[ENH] Add exogenous variable support to ARIMA

### LINKED ISSUE #3077: [ENH] ARIMA support for exogenous variables

Hi,

First, thank you for the great work on implementing a Python-native and efficient version of ARIMA.

It would be very helpful if ARIMA could support the use of exogenous variables (exog). This would enable users to incorporate additional predictors into their time series models, making the implementation more flexible and powerful.

Thanks again for your efforts on this project!

I don’t have a concrete implementation proposal at the moment. My request is mainly for adding support for exogenous variables (exog) within the ARIMA implementation, similar to what is available in libraries like statsmodels. I’d be happy to test or provide feedback once a design or prototype is available.

_No response_

### Additional context

_No response_

satwiksps:
## Reference Issues
- Closes #3077.

## What does this implement ? Explain your changes.
This PR adds full support for **exogenous variables (exog)** to `ARIMA` and `AutoARIMA`.

#### Summary of enhancements:
- **Base Interface**: Updated `IterativeForecastingMixin.iterative_forecast` in `aeon/forecasting/base.py` to accept `exog`.
- **ARIMA Logic**: Added `"capability:exogenous": True` and implemented OLS regression in `ARIMA` (`_fit`, `_predict`, `iterative_forecast`) to regress `y` on `exog` and fit ARIMA on residuals.
- **AutoARIMA**: Updated `AutoARIMA` to correctly forward `exog` to the wrapped `ARIMA` model.
- **Compatibility**: Updated `ETS` and `Theta` method signatures to match the new `IterativeForecastingMixin` interface.

#### Tests added
- **`test_arima_with_exog_basic_fit_predict`**: Verifies ARIMA fits, predicts correctly when exogenous variables are supplied.
- **`test_arima_exog_shape_mismatch_raises`**: Ensures shape mismatch errors are raised for incorrect exog dimensions.
- **`test_arima_iterative_forecast_with_exog`**: Tests multi-step forecasting with future exogenous values.
- **`test_autoarima_passes_exog_correctly`**: Ensures AutoARIMA correctly propagates exog to the inner model.
- **`test_arima_no_exog_backward_compatibility`**: Ensures ARIMA without exog behaves exactly as before.

NO

## Any other comments?
Logic includes an optimization in `iterative_forecast` to skip re-fitting if the input series matches the training data, allowing for smooth handling of future-only exogenous data.

### For all contributions


JoaquinAmatRodrigo:
Hi @satwiksps,

Thanks for working on this issue. I'm not involved in the Aeon development, so I'm not very familiar with the codebase. After looking at your modifications, I have a question regarding input alignment: the input data (y and exog) are assumed to be aligned. If the ARIMA model includes differencing, it’s necessary to drop from exog the same number of initial observations as the differencing applied to y; otherwise, the alignment is lost.

Is this already handled in your implementation?

satwiksps:
> Is this already handled in your implementation?

Hello @JoaquinAmatRodrigo 
I believe the alignment is handled correctly in the current implementation.

- Exogenous variables are used **only once**, during the initial OLS regression step `y - (X @ beta)`. This produces a residual series that is **the same length as y**, so alignment is preserved at this stage.

- Differencing is applied **after** residualization `np.diff(residual_series, n=d)`. At this point, the model works entirely on the differenced residuals and **does not use exog anymore**, so no trimming or shifting of `exog` is required after differencing.

- For forecasting (one-step or multi-step), future `exog` rows are used only to compute the regression contribution `[1, future_exog_row] @ beta`. This does not depend on differencing and does not require dropping the first `d` rows.

So I think this implementation preserves correct alignment without needing to manually drop the first `d` rows of `exog`. I might be wrong. I would appreciate if you could suggest me if any changes I need to do in my implementation if I am missing something. Thank you.

MatthewMiddlehurst:
Thanks for the input Joaquin. We should hopefully be back to forecasting soon if things calm down this semester 🙂

aeon/forecasting/stats/_arima.py
```diff
@@ -336,7 +423,11 @@ def _fit(self, y, exog=None):
         ) = model
         self.constant_term_ = constant_term_int == 1
         self.final_model_ = ARIMA(self.p_, self.d_, self.q_, self.constant_term_)
-        self.final_model_.fit(y)
+
+        if exog is not None:
+            self.final_model_.fit(y, exog=exog)
+        else:
+            self.final_model_.fit(y)
```
alexbanwell1: Imo this should just pass it through, so it should just call self.final_model_.fit(y, exog) there isn't any need for the extra check

aeon/forecasting/stats/_arima.py
```diff
@@ -207,42 +268,63 @@ def _forecast(self, y, exog=None):
         self._fit(y, exog)
         return float(self.forecast_)
 
-    def iterative_forecast(self, y, prediction_horizon):
+    def iterative_forecast(self, y, prediction_horizon, exog=None):
```
alexbanwell1, line 271: For this you should update the IterativeForecatingMixin to include exog in the method (and also update any other cases where IterativeForecastingMixin is used) otherwise this class ceases to implement IterativeForecastingMixin and just has another custom method!

alexbanwell1 (changes requested):
Generally fairly happy with how this has been implemented and seems to be a sensible addition, but definitely needs a couple of fixes

satwiksps:
Hi @alexbanwell1, thank you for the review!

I have implemented your suggestions, specifically:
1. Updated the `IterativeForecastingMixin` signature to accept `exog` (and updated `ETS` and `Theta` to match).
2. Simplified the `AutoARIMA` logic to pass `exog` through directly without the extra check.

Once this PR is merged, I will update PR #3106 to resolve any resulting conflicts.

satwiksps:
Thank you @alexbanwell1 for the review and the optimization for moving this pr forward! Removing the copy in fit makes sense as copying large arrays is expensive

TonyBagnall:
@alexbanwell1 could you re-review?

PR #3133, opened 2025-11-24
[ENH] Implement LOESS Smoother (Locally Estimated Scatterplot Smoothing)

### LINKED ISSUE #2974: [ENH] LOESS Smoothing

thanks to @Cyril-Meyer we have a nice collection of smoothing algorithms. It would be nice to have a simple version of LOESS: 

"LOESS (Locally Estimated Scatterplot Smoothing) is a non-parametric regression technique used to produce a smooth curve through noisy data. It was proposed by Cleveland (1979, 1988), and is often employed in exploratory data analysis, time series smoothing, and data visualization because of its flexibility and simplicity."

The general definition is for x and exogenous variables, but I think we only need the univariate version such as used in STL.

_No response_

### Additional context

_No response_

Nithurshen:
#### Reference Issues/PRs

Fixes #2974 

**Implementation Details:**
- **Class:** `LoessSmoother` extending `BaseSeriesTransformer`.
- **Parameters:**
    - `span`: Float (0, 1] defining the proportion of data used in the local window.
    - `degree`: Int (1 or 2) defining the local polynomial degree (Linear or Quadratic).
- **Multivariate Support:** Iterates over channels while reusing the Design Matrix ($H$) and Weight Matrix ($W$) for the time axis to improve efficiency.
- **Safety:** Includes input validation for parameters and a fallback check for very short time series ($n \le degree$).
- **Tests:** Added `test_loess.py` covering:
    - Linear and Quadratic function recovery.
    - Constant value preservation.
    - Noise reduction effectiveness (MSE decrease on noisy sine waves).
    - Edge cases (short series, extreme span values).

No

#### Any other comments?

**Reference:**
[1] Cleveland, W. S. (1979). Robust locally weighted regression and smoothing scatterplots. Journal of the American statistical association, 74(368), 829-836.

- **Performance:** The current implementation uses NumPy vectorization. While it is efficient for typical time series lengths, the core loop is Python-bound. I am happy to convert the solver loop to a Numba-optimized function if performance gain is needed for this iteration.

##### For all contributions


Nithurshen (commented):
.

Nithurshen (commented):
.

TonyBagnall:
@Nithurshen thanks for this, sorry it has taken me a while to get back to it.

For correctness, I plan to run comparisons against statsmodels.nonparametric.smoothers_lowess.lowess, and we can use that as a benchmark for the degree=1 case (with it=0, delta=0, and the same frac/span). I'll post comparisons here along with timing.

The degree parameter is a nice addition, but I suggest defaulting it to 1 to align with statsmodels (LOWESS is local linear only).

Statsmodels also supports “robust iterations” (it, default 3), which is essentially an iteratively reweighted fit based on residuals. We do not need that functionality yet, but it is worth noting, and any parity comparisons should be done with it=0 since we are not doing robust reweighting.

TonyBagnall:
I've run this vs statsmodels and its much slower, assuming chatgpt set up the comparison correctly. Issues with differences in output but would need to dig deeper into this, the slowness is a blocker Im afraid

image





```python
import time
from dataclasses import dataclass
from typing import Callable, List, Tuple

import numpy as np

# aeon
from loess_smoother import LoessSmoother

from statsmodels.nonparametric.smoothers_lowess import lowess as sm_lowess


@dataclass
class Row:
    n: int
    aeon_ms: float
    sm_ms: float
    max_abs_diff: float
    rmse: float


def _to_1d(a) -> np.ndarray:
    """Convert aeon output to 1D float array."""
    arr = np.asarray(a)
    return arr.squeeze().astype(np.float64, copy=False)


def _aeon_loess(y: np.ndarray, span: float) -> np.ndarray:
    # degree=1 to match statsmodels LOWESS family
    out = LoessSmoother(span=span, degree=1).fit_transform(y)
    return _to_1d(out)


def _statsmodels_lowess(y: np.ndarray, span: float) -> np.ndarray:
    if sm_lowess is None:
        raise ImportError(
            "statsmodels is not installed. Install it locally to run this benchmark: pip install statsmodels"
        )
    x = np.arange(len(y), dtype=np.float64)
    out = sm_lowess(y, x, frac=span, it=0, delta=0.0, return_sorted=False)
    return np.asarray(out, dtype=np.float64)


def benchmark_loess_vs_statsmodels(
    max_n: int = 2000,
    step: int = 100,
    span: float = 0.3,
    repeats: int = 5,
    warmup: int = 1,
    seed: int = 0,
) -> List[Row]:
    """
    Deterministic benchmark of aeon LoessSmoother vs statsmodels LOWESS (it=0, delta=0),
    measuring runtime and output differences for increasing n, and plotting both.

    Produces two plots:
      1) Runtime vs n (ms per call)
      2) Differences vs n (max abs diff and RMSE)
    """
    rng = np.random.default_rng(seed)
    ns = list(range(step, max_n + 1, step))
    rows: List[Row] = []

    for n in ns:

        # Deterministic test signal, smooth-ish with some noise
        t = np.arange(n, dtype=np.float64)
        y = (0.05 * t) + np.sin(t / 15.0) + 0.1 * rng.standard_normal(n)
        y = y.astype(np.float64, copy=False)
        print(f"n = {n}\n")

        # Warmup (not timed)
        for _ in range(warmup):
            _aeon_loess(y, span)
            _statsmodels_lowess(y, span)
        # Time aeon
        t0 = time.perf_counter()
        for _ in range(repeats):
            aeon_out = _aeon_loess(y, span)
        t1 = time.perf_counter()
        aeon_ms = (t1 - t0) * 1000.0 / repeats
        print(f" Time PR = {aeon_ms}")

        # Time statsmodels
        t0 = time.perf_counter()
        for _ in range(repeats):
            sm_out = _statsmodels_lowess(y, span)
        t1 = time.perf_counter()
        sm_ms = (t1 - t0) * 1000.0 / repeats
        print(f" Time statsmodels = {sm_ms}")

        # Differences (use the last outputs computed above)
        diff = aeon_out - sm_out
        max_abs = float(np.max(np.abs(diff)))
        rmse = float(np.sqrt(np.mean(diff * diff)))

        rows.append(Row(n=n, aeon_ms=aeon_ms, sm_ms=sm_ms, max_abs_diff=max_abs, rmse=rmse))

    # ---- Plot 1: runtimes
    plt.figure()
    plt.plot([r.n for r in rows], [r.aeon_ms for r in rows], marker="o", label="aeon LoessSmoother")
    plt.plot([r.n for r in rows], [r.sm_ms for r in rows], marker="o", label="statsmodels lowess (it=0)")
    plt.xlabel("n (series length)")
    plt.ylabel("Runtime (ms per call)")
    plt.title("LOESS runtime vs series length")
    plt.legend()
    plt.tight_layout()
    plt.show()

    # ---- Plot 2: differences
    plt.figure()
    plt.plot([r.n for r in rows], [r.max_abs_diff for r in rows], marker="o", label="max abs diff")
    plt.plot([r.n for r in rows], [r.rmse for r in rows], marker="o", label="RMSE")
    plt.xlabel("n (series length)")
    plt.ylabel("Difference")
    plt.title("aeon vs statsmodels output differences vs series length")
    plt.legend()
    plt.tight_layout()
    plt.show()

    # Small summary
    last = rows[-1]
    print(
        f"n={last.n}, aeon={last.aeon_ms:.3f} ms, statsmodels={last.sm_ms:.3f} ms, "
        f"max|diff|={last.max_abs_diff:.3e}, rmse={last.rmse:.3e}"
    )

    return rows
```

TonyBagnall:
hi, closing this as #2974 has been resolved by #3321

PR #3136, opened 2025-11-24
[ENH] Standardise check_y usage via separate _convert_y method

### LINKED ISSUE #2970: [ENH] Standardise check_y use case

Further to #2958 

several base classes have `check_y` functionality. We cannot really generalise this, since there are different constraints on different tasks. However, we can standardise the use case. Currently there are two models

`self._check_y(y)` or `self._check_y(y, n_cases=self.metadata_["n_cases"])`

this simply checks that y is a compliant type and shape for the task in question. This is  used in `BaseClassifier`, both `BaseTransformer` classes, `BaseSegmenter` (after #2958). Other case is

`y = self._check_y(y)` or y = `self._check_y(y, self.metadata_["n_cases"])`

this is used in `BaseRegressor` as follows

```python
        if isinstance(y, pd.Series):
            y = pd.Series.to_numpy(y)
        return y.astype(float)
```
and the two `BaseAnomalyDetector` classes, where it gets converted to numpy booleans. 

I think conversion should be by a separate function `convert_y` and standard usage would be
```python
self._check_y(y)
y = self._convert_y(y)
```

_No response_

### Additional context

_No response_

satwiksps:
# [ENH] Standardise `check_y `usage via separate `_convert_y `

## Reference Issues/PRs
- Fixes #2970  

This PR standardises the usage of `check_y` across base classes to follow the pattern proposed in #2970.

Previously, two inconsistent patterns existed:
- Validation and conversion combined via `y = self._check_y(y)`
- Validation only via `self._check_y(y)`

This caused ambiguity and inconsistent behaviour across different estimator types.

This PR introduces a clear separation of responsibilities by:
- Ensuring `_check_y()` is used strictly for validation
- Moving all type conversions into a new dedicated method `_convert_y()`

The standard usage is now:
`self._check_y(y)`  
`y = self._convert_y(y)`

This pattern is applied consistently to:
- BaseRegressor  
- BaseCollectionAnomalyDetector  
- BaseSeriesAnomalyDetector  

This improves clarity, consistency, and prevents unintended side-effects such as overwriting `y` or implicit type coercion during validation.

No

## Any other comments?
No

### For all contributions
- [ ] I've added myself to the list of contributors.
- [x] The PR title starts with [ENH].

- [ ] I've added the estimator/function to the online API documentation.
- [ ] I've added myself as a `__maintainer__`

- [ ] I've updated CODEOWNERS.

### DIFF
--- a/.github/workflows/fast_release.yml
+++ b/.github/workflows/fast_release.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Setup Python 3.12
         uses: actions/setup-python@v6
--- a/.github/workflows/issue_assigned.yml
+++ b/.github/workflows/issue_assigned.yml
@@ -21,7 +21,7 @@ jobs:
           private-key: ${{ secrets.PR_APP_KEY }}
 
       - name: Checkout main
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           sparse-checkout: .github/utilities
 
--- a/.github/workflows/issue_comment_edited.yml
+++ b/.github/workflows/issue_comment_edited.yml
@@ -22,7 +22,7 @@ jobs:
           private-key: ${{ secrets.PR_APP_KEY }}
 
       - name: Checkout main
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           sparse-checkout: .github/utilities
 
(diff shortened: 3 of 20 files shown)


SebastianSchmidl:
I think this needs a more careful look through the code base. A quick sanity check for the classification base class shows that its `_check_y` also transforms `y` to numpy (L564f):

https://github.com/aeon-toolkit/aeon/blob/adb7e455d26e20ff72cdaf490c438e0a2a2deeeb/aeon/classification/base.py#L553-L575

@TonyBagnall Should the check function `_check_y()` return anything at all? Shouldn't we annotate it with `None` and remove the `return`-statements?

MatthewMiddlehurst:
I would check what the equivalent function for X is doing but yeah I would probably return nothing in these functions. Could you comment here what the other base classes mentioned in the issue (i,e, classification) are doing and if they meet the new format?

MatthewMiddlehurst:
No response. Beginning to tackle base class y stuff in #3310 but does not resolve this issue completely. Think separating these is a good idea though.

PR #3137, opened 2025-11-24
[ENH] Add general estimator test for common input dtypes

### LINKED ISSUE #2803: [ENH] Add general testing for common data `dtypes`
### Describe the issue

Currently we do not test specific data dtypes. We should add a general estimator test converting testing data to common types i.e. float32, float64, int32, int64 and make sure it still runs as expected.

Add new check to the estimator testing in the `testing` module.

### Additional context

_No response_

satwiksps:
# [ENH] Add general estimator test for common input dtypes
## Reference Issues/PRs
- Closes #2803

## What does this implement/fix?

This PR adds a general estimator test to ensure estimators correctly handle common numeric input dtypes.  
The test verifies that estimators can run successfully when inputs are provided as:

- `float32`
- `float64`
- `int32`
- `int64`

The check reuses existing estimator testing utilities rather than manually converting arrays, ensuring consistent behavior with the rest of the testing framework.

NO

## Any other comments?
NO

### For all contributions
- [ ] I've added myself to the list of contributors. Alternatively, I will use the @all-contributors bot after the PR is merged.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV].

- [x] Not applicable (no new estimators or public functions added).

- [x] Not applicable.


aeon/testing/estimator_checking/_yield_estimator_checks.py
```diff
@@ -690,3 +695,34 @@ def check_fit_deterministic(estimator, datatype):
...
+    y_train = deepcopy(FULL_TEST_DATA_DICT[datatype]["train"][1])
+    X_test = deepcopy(FULL_TEST_DATA_DICT[datatype]["test"][0])
+
+    dtypes = [np.float32, np.float64, np.int32, np.int64]
+
+    X_train_np = np.asarray(X_train)
+    X_test_np = np.asarray(X_test)
+
+    if X_train_np.dtype == object:
+        return
```
MatthewMiddlehurst: when is this going to be the case?

aeon/testing/estimator_checking/_yield_estimator_checks.py
```diff
@@ -690,3 +695,34 @@ def check_fit_deterministic(estimator, datatype):
...
+    estimator = _clone_estimator(estimator)
+
+    X_train = deepcopy(FULL_TEST_DATA_DICT[datatype]["train"][0])
+    y_train = deepcopy(FULL_TEST_DATA_DICT[datatype]["train"][1])
+    X_test = deepcopy(FULL_TEST_DATA_DICT[datatype]["test"][0])
+
+    dtypes = [np.float32, np.float64, np.int32, np.int64]
+
+    X_train_np = np.asarray(X_train)
+    X_test_np = np.asarray(X_test)
```
MatthewMiddlehurst: use our data conversion utilities if you need a numpy array format or create your own. I am not so convinced `np.asarray` will always work.

aeon/testing/estimator_checking/_yield_estimator_checks.py
```diff
@@ -690,3 +695,34 @@ def check_fit_deterministic(estimator, datatype):
...
+    for dtype_cast in dtypes:
+        try:
+            X_train_cast = X_train_np.astype(dtype_cast)
+            X_test_cast = X_test_np.astype(dtype_cast)
+
+            est = estimator.clone()
+            est.fit(X_train_cast, y_train)
+
+            if hasattr(est, "predict"):
+                est.predict(X_test_cast)
```
MatthewMiddlehurst: what about estimators which dont have predict? There are some testing utilities for using predict/transform type methods

MatthewMiddlehurst (changes requested):
It would be better to use the current conversion and testing utilities.

is it possible to merge this with another test to reduce the amount of fit calls? I would make suggestions here in comments rather than alter the code for now.

aeon/testing/estimator_checking/_yield_estimator_checks.py
satwiksps: We do not need this. We do not need to skip object dtypes

aeon/testing/estimator_checking/_yield_estimator_checks.py
satwiksps: Okay. I will switch to using the existing estimator test utilities instead of manually calling np.asarray, so dtype handling is consistent with the rest of the test suite.

aeon/testing/estimator_checking/_yield_estimator_checks.py
satwiksps: I will update the execution logic to rely on the existing `_run_estimator_method` helper, which correctly handles estimators without predict, rather than caling predict directly

MatthewMiddlehurst:
Sorry I do not think this is going to make progress. The revised test does not achieve anything. If the suggested utilities cant be used you can say that but the changes made give the impression you do not understand the task.

Issue #2985, opened 2025-07-31
[ENH] loaders for MONSTER datasets


baraline:
### Describe the feature or idea you want to propose

Include loaders to the MONSTER datasets in the datasets module. 

The only downside is that we would have to put huggingface as an optional dependency. I'm not sure if there is other channels to load the datasets from to avoid another dependency?

### Describe your proposed solution

Code to load the datasets from huggingface

```python
import numpy as np
from aeon.utils.numba.general import z_normalise_series_3d
from huggingface_hub import hf_hub_download

univariate_monster_datasets = [
    "CornellWhaleChallenge",
    "AudioMNIST",
    "WhaleSounds",
    "Pedestrian",
    "FruitFlies",
    "AudioMNIST-DS",
    "Traffic",
    "LakeIce",
    "MosquitoSound",
    "InsectSound",
]


def load_monster(dataset_name, fold, normalize=True):
    repo_id = f"monster-monash/{dataset_name}"

    # Download data
    data_path = hf_hub_download(
        repo_id=repo_id, filename=f"{dataset_name}_X.npy", repo_type="dataset"
    )
    X = np.load(data_path, mmap_mode="r")  # (#Samples, #Channel, #Length)
    if normalize:
        X = z_normalise_series_3d(X)
    # Download labels
    label_filename = f"{dataset_name}_Y.npy"
    try:
        label_path = hf_hub_download(
            repo_id=repo_id, filename=label_filename, repo_type="dataset"
        )
    except:
        label_filename = f"{dataset_name}_y.npy"
        label_path = hf_hub_download(
            repo_id=repo_id, filename=label_filename, repo_type="dataset"
        )
    y = np.load(label_path)
    # Load test indices
    try:
        test_index_path = hf_hub_download(
            repo_id=repo_id,
            filename=f"test_indices_fold_{fold}.txt",
            repo_type="dataset",
        )
        test_index = np.loadtxt(test_index_path, dtype=int)
    except Exception as e:
        logger.error(f"Failed to load test indices: {e}")
        raise

    test_bool_index = np.zeros(len(y), dtype=bool)
    test_bool_index[test_index] = True
    return (
        X[~test_bool_index],
        y[~test_bool_index],
        X[test_bool_index],
        y[test_bool_index],
    )
```

rwtarpit:
@hadifawaz1999 can i work this one out?

baraline:
Sure go ahead

rwtarpit:
[...]

PR #3141, opened 2025-11-27
[ENH] Add loaders  for Monster datasets with hugging_face dependency

rwtarpit:
#### Reference Issues/PRs
Fixes  [#2985 ](url)

Implements a new file for loading Monster datasets from `hugginngface_hub`.
Creates methods for loading list of available monster datasets in [huggingface_hub](https://huggingface.co/monster-monash/datasets) and downloading the datasets.

`hugging_face`

#### Any other comments?
I am yet to implement the `test_monster_loaders` file.

##### For all contributions


rwtarpit:
@baraline pls have a look.
and also can you guide me on testing file for this.
thank you.

baraline:
At first glance the file looks good. I think you can remove the univariate dataset name list, as you pull all the dataset names already.

Concerning the tests, you can look at existing tests file from other loaders, for exemple https://github.com/aeon-toolkit/aeon/blob/main/aeon/datasets/tests/test_data_loaders.py
More specifically, the xxxx_from_repo tests. You can perform a similar test with similar function tags but trying to fecth from Monster, and validate that what you fetched is in the expected format and not empty.

Also, you have some pre-commit failure, you can refer to the developer documentation it explains how to install it, it will correct most of the issues by itself when you try to commit and signal you the others.

rwtarpit:
@baraline how do i resolve `FAILED aeon/datasets/tests/test_monster_loader.py::test_load_monster_dataset - ModuleNotFoundError: No module named 'huggingface_hub'`. 
i tried removing `_check_soft_dependencies` but still same error.

baraline:
The check soft dep should stay, it is used to check if the dependency is installed (as it is optional) and raide appropriate messages to the users. Don't have access to my computer rn, so add the check soft dep back  and i'll check the errors throwed when I can later

baraline (changes requested):
Test are failing because the test skip is wrongly parameterized. The wrong check is applied to validate if the test should be run or not. You should do 
```
    @pytest.mark.skipif(
        not _check_soft_dependencies("hugging-face", severity="none"),
        reason="required soft dependency hugging-face not available",
    )
```
Instead of
```
@pytest.mark.skipif(
    PR_TESTING,
    reason="Only run on overnights because of read from internet.",
)
```

baraline:
You should test with huggingface-hub (the dependency you added in the project.toml), not with hugging-face. You correctly did it when you created the loader like this :

_check_soft_dependencies("huggingface-hub", severity="none")
from huggingface_hub import hf_hub_download

My example didn't take the correct name.

rwtarpit:
> You should test with huggingface-hub (the dependency you added in the project.toml), not with hugging-face. You correctly did it when you created the loader like this :
> 
> _check_soft_dependencies("huggingface-hub", severity="none") from huggingface_hub import hf_hub_download
> 
> My example didn't take the correct name.

my bad!!
did it in a hurry

baraline:
The remaining failure is not linked to what you did, so we should be good to go. We'll just wait for it to be resolved. Thanks !

baraline (changes requested):
To keep thing simple and similar to other dataset loaders, could you remove the lazy loading for the dataset names ? 

I'd rather expose a function that fetch it everytime and let user cache it if they want. 

We'll be good to go after that.

MatthewMiddlehurst:
Thanks for the reviews Antoine. Prefer the soft dependency approach and this was first so closing the other PR. Possibly go over the documentation again and reconsider whether we want to normalize in this function. Some of these datasets are massive so I think we can to reduce as much data duplication as we can in the function itself.

rwtarpit:
please let me know if any changes are to be made

baraline:
If think what would be missing is :
-  add the new public functions to the API reference in the docs/api_reference/datasets.md file
- I agree with Matthew, remove the normalisation option and let the user deal with any preprocessing himself, as the datasets can be massive
- I would remove the _fetch_dataset_name function and put its code the in load_dataset_name function, the public function has no additional code, so I don't think we need the private function here.

baraline:
From a quick look at the last changes, everything seem to be in place, so LGTM

PR #3144, opened 2025-11-29
[ENH] Add load_classification_train_test wrapper for label encoding

### LINKED ISSUE #2807: [ENH] Add a label encoder option when loading dataset

When loading datasets, we often get class labels as strings, which is not the best case when wanting to do plotting or other operation (e.g. plotting samples coloured by their class label).

It would be nice to have an option as a boolean parameter to apply (and return) a label encoder directly when loading, instead of having to do it outside this function every time we want too.

Add a boolean parameter (default to false to not change default behaviour) that applies and returns a label encoder to the class labels when loading classification datasets.

Some thinking might be needed on how to handle properly the case where we load the training data, and then need to re-use the label encoder for the testing data.

_No response_

### Additional context

I often need to do this when loading classification datasets, so I might not be the only one (hopefuly!)

satwiksps:
## Reference Issue
- Fixes #2807.

This PR implements a new wrapper function, `load_classification_train_test`, in `aeon/datasets/_data_loaders.py`.

As discussed in issue #2807, this function addresses the need to easily load classification datasets with target labels automatically encoded as integers (using `sklearn.LabelEncoder`) without modifying the signature of the base `load_classification` function. It also simplifies the common workflow of loading both train and test splits simultaneously.

**Key changes:**
1.  **New Function:** Added `load_classification_train_test`. 
    * Loads both train and test splits in one call.
    * Includes an `encode_labels` parameter (default `True`) to automatically fit a `LabelEncoder` on the training set and transform both splits.
    * Optionally returns the fitted encoder via `return_encoder=True`.
    * Docstrings match the style and content of the original `load_classification` function for consistency.
2.  **Export:** Exposed the function in `aeon/datasets/__init__.py`.
3.  **Testing:** Added a unit test in `aeon/datasets/tests/test_data_loaders.py` to verify integer conversion, encoder return behavior, and the option to disable encoding.

No

## Any other comments?

No

### For all contributions


baraline (changes requested):
Hi, the code itself is good, but (minor point) I'm unsure about naming. 

We already have a function named load_classification, which expects a split paramater. 

We don't want that paramater here as it would mean we need to carry around the label encoder, so we want to return both train and test, as you did. 

To avoid contusion on the expected paramater and output of the function, I'd rather rename it to something like ` load_classification_train_test`.
The encoded part being optional, including it in the naming looks weird to me.

MatthewMiddlehurst (changes requested):
Yes agreed could use a better name, Antoine's suggestion seems fine if the optional encoding bit is included.

Documentation should be improved and match the original function where possible.

MatthewMiddlehurst:
Could you explain what has changed? by force-pushing you damage the commit history so we cannot easily compare what you had before to what you have now.

satwiksps:
> by force-pushing you damage the commit history

@MatthewMiddlehurst I am sorry for this. I will keep this in mind from now on.

> Could you explain what has changed

I have updated the PR based on the feedback to renamethe function and make the encoding behavior optional.

### Summary of Changes

1.  **Renaming**
    - Renamed the function from `load_classification_encoded` to `load_classification_train_test`.

2.  **Optional Encoding Logic**
    - Added a new parameter `encode_labels=True`.
    - **Default Behavior (`True`):** The function fits a `LabelEncoder` on the training set and transforms both train and test labels to integers.
    - **Optional Behavior (`False`):** Users can pass `encode_labels=False` to get the raw string labels while still benefiting from the convenience of loading both splits simultaneously.

3.  **Improve Documentation**
    - Updated the docstring parameter descriptions to match the original `load_classification` function exactly, ensuring documentation consistency across the API.
    - Added an `Examples` section demonstrating the default usage, the return of the encoder, and the no-encoding case.
    - **Note:** I am still not fully confident about the documentation part. If there are further improvements or conventions I missed, I am more than willing to make changes

4.  **Test Coverage**
    - Updated `test_data_loaders.py` to test the new function name.
    - Added a specific assertion to verify that `encode_labels=False` correctly returns string labels.


Apologies again for the force push. To help you trace the changes, you can refer to:
- **Previous state:** [6fa2bad](https://github.com/aeon-toolkit/aeon/commit/6fa2badd205e68f3511152659fdf32e9a29171cf) (The initial implementation of `load_classification_encoded`)
- **Major changes:** [f25bcab](https://github.com/aeon-toolkit/aeon/commit/f25bcab31bf5ada40a0fe7fd5bc287dd79a398d5) (This has the function renaming, addition of optional encoding logic and most of the doc changes)
- **Minor docstring changes:** [08614f8](https://github.com/aeon-toolkit/aeon/commit/08614f88259011c51b313f142cf529bea79fb4cf) (This has only some minor changes to the docstring of thr returns section)

aeon/datasets/_data_loaders.py
```diff
@@ -1427,6 +1429,114 @@ def load_classification(
...
+
+    Examples
+    --------
+    >>> from aeon.datasets import load_classification_train_test
+    >>> X_train, y_train, X_test, y_test = load_classification_train_test(
+    ...     "ArrowHead"
+    ... ) # doctest: +SKIP
+    """
+    if return_encoder:
+        encode_labels = True
```
MatthewMiddlehurst, line 1354: would raise an error instead IMO.

aeon/datasets/_data_loaders.py
```diff
@@ -1427,6 +1429,114 @@ def load_classification(
...
+        split="test",
+        extract_path=extract_path,
+        return_metadata=False,
+        load_equal_length=load_equal_length,
+        load_no_missing=load_no_missing,
+    )
+
+    le = None
+    if encode_labels:
+        le = LabelEncoder()
```
MatthewMiddlehurst, line 1379: just create the class in 1 line, no real overhead here.

MatthewMiddlehurst (commented):
Please add this to the documentation. 

@baraline any thoughts when you get some free time? I'm fine with it if this solves your issue.

aeon/datasets/_data_loaders.py
```diff
@@ -1427,6 +1429,114 @@ def load_classification(
...
+        this function will look in the extract_path first, and if it is not present,
+        attempt to download the data from www.timeseriesclassification.com, saving it to
+        the extract_path.
+    extract_path : str, default=None
+        the path to look for the data. If no path is provided, the function
+        looks in `aeon/datasets/local_data/`. If a path is given, it can be absolute,
+        e.g. C:/Temp/ or relative, e.g. Temp/ or ./Temp/.
+    return_metadata : boolean, default = False
+        If True, returns a dictionary of metadata found in the header.
+    return_encoder : bool, optional (default=False)
```
TonyBagnall, line 1310: we decided the word "optional" was redundant, just

return_encoder : bool, default=False

aeon/datasets/_data_loaders.py
TonyBagnall: agreed

TonyBagnall (changes requested):
looks fine but best wait for the new load_classification.

TonyBagnall:
hi, this needs resolving with data loaders, I will convert to draft for now, feel free to reopen if you want to continue @satwiksps

TonyBagnall:
closing this now, feel free to reopen if you want to continue

PR #3146, opened 2025-12-01
[ENH] Implement series_to_series forecaster mixin

### LINKED ISSUE #2910: [ENH] Forecasting series to series models

If we want to make a single forecast 1 ahead we do this
```python
f = MyForecaster()
p=f.forecast(y)
```
so if y is length `t` this forecasts step `t+1`.  We currently have two ways of making more than one prediction. So if we have a prediction horizon of `ph` and we want to predict `t+1` to `t+ph`, we do use either direct or iterative (aka recursive) to get an array size `ph`. Direct refits the model with increasing horizons (so fits `ph` models), iterative feeds predictions back into a single model.
```python
f = MyForecaster()
preds1=f.direct_forecast(y)
preds2= f.interative_forecast(y)
```
some are not capable of direct (e.g. ARMA) and this is a separate issue (mixin thread to follow). The issue here is some forecasters do not fit either model. ETS, for example, can only normally be fit on a horizon of 1 (unless we downsampled), but is not iterative, it is sometimes called state iterative, in that it simply progresses its internal model through time. Elso generative models may fit a series to series model, so they might predict a whole series in one go without iteration. The issue is how to deal with these?

we could either 

1. try shoe-horn them into the existing model or define a new function. So, for example, we could override `direct_forecast` and `recursive_forecast` for ETS to do the same thing, 
2. define a new behaviour, say, `series_forecast` that fits a single model and predicts a whole series in one go rather than point by point

rwtarpit:
#### Reference Issues/PRs
Fixes #2910 

This PR implements the Mixin class for `series_to_series ` forecasters by introducing a abstract method `_series_to_series_forecast`. 
A dummy series forecaster is also implemented in `_dummy_series_forecaster.py` for testing purpose.

##### For all contributions

### DIFF
--- a/aeon/forecasting/deep_learning/tests/test_dummy_series_forecaster.py
+++ b/aeon/forecasting/deep_learning/tests/test_dummy_series_forecaster.py
@@ -8,10 +8,15 @@ from aeon.forecasting.deep_learning._dummy_series_forecaster import (
     DummySeriesForecaster,
 )
 from aeon.forecasting.deep_learning.base import BaseDeepForecaster
+from aeon.utils.validation._dependencies import _check_soft_dependencies
 
 Y_TEST = np.array([[1.0, 2.0, 3.0, 4.0, 5.0]])
 
 
+@pytest.mark.skipif(
+    not _check_soft_dependencies("tensorflow", severity="none"),
+    reason="skip test if required soft dependency not available",
+)
 def test_abstract_method_raises_error():
     """Checks that a forecaster must implement the abstract method."""
 
@@ -26,6 +31,10 @@ def test_abstract_method_raises_error():
         BrokenForecaster(horizon=5, axis=1)
 
 
+@pytest.mark.skipif(
+    not _check_soft_dependencies("tensorflow", severity="none"),
+    reason="skip test if required soft dependency not available",
+)
 def test_series_to_series_forecast_output():
     """
     Test the functionality of the series_to_series_forecast method directly.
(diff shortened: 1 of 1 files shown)


rwtarpit:
@hadifawaz1999 this PR is regarding the unimplemented `SeriesToSeriesForecastingMixin` we talked about on slack.
pls suggest any changes

aeon/forecasting/_dummy_series_forecaster.py
```diff
@@ -0,0 +1,36 @@
...
+    is functional.
+    """
+
+    def __init__(self, horizon: int = 1, axis: int = 1, value_to_return: float = 1.0):
+        self.value_to_return = value_to_return
+        super().__init__(horizon=horizon, axis=axis)
+
+    def _fit(self, y, exog=None):
+        self.n_timepoints_ = y.shape[1]
+        self.last_observed_value_ = y[0, -1]
```
hadifawaz1999: this line seems unnecessary, you can just say the forecaster return the same value chosen in parameter default to 1

rwtarpit:
@hadifawaz1999 the last few tests that are failing are in python 3.13 enviroment specifically, due to tensorflow compatibility issues. Tensorflow is being imported in `BaseDeepForecaster`.
should i skip tests with tensorflow dependency?

hadifawaz1999:
> @hadifawaz1999 the last few tests that are failing are in python 3.13 enviroment specifically, due to tensorflow compatibility issues. Tensorflow is being imported in `BaseDeepForecaster`. should i skip tests with tensorflow dependency?

Hello, for all tests you should add this decorator:
```python
@pytest.mark.skipif(
    not _check_soft_dependencies("tensorflow", severity="none"),
    reason="skip test if required soft dependency not available",
)
```

alexbanwell1:
Just as another pair of eyes, looks good to me @hadifawaz1999

PR #3153, opened 2025-12-06
[ENH] Made SAX transformations and PAA parallel #2972

Lucky-Lodhi2004:
#### Reference Issues/PRs

Fixes issue #2972

This PR introduces parallelism in _get_sax_symbols() and PAA class. It introduces n_jobs to PAA class and its functions. Although instead of using nested prange() inside loops like it is in inverse function, i have used prange() only in outer loop. For example : 

in _invert_sax_symbols()
```python
for i in prange(n_samples):
        for c in prange(n_channels):
            for _current_sax_index in prange(sax_length):
                start_index = _current_sax_index * segment_length
                stop_index = start_index + segment_length
```

and in new paa transform 

```python
for i_x in prange(n_cases):
            for i_c in range(n_channels):  # keep inner loops non-parallel
                for i_t in range(n_segments):
                    value = X_paa[i_x, i_c, i_t]
```

This is because numba only makes outer loops parellel and ignores inner loops (documentation). inner prange also introduces extra overhead.

image

No optional dependency is introduced

##### For all contributions

### DIFF
--- a/.github/workflows/fast_release.yml
+++ b/.github/workflows/fast_release.yml
@@ -22,7 +22,7 @@ jobs:
           python -m build
 
       - name: Store build files
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: dist
           path: dist/*
@@ -38,7 +38,7 @@ jobs:
       id-token: write
 
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: dist
           path: dist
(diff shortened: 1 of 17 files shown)


hadifawaz1999:
Thank you but it seems we already have an active PR open for this that was mentioned on the issue, [here](https://github.com/aeon-toolkit/aeon/pull/2980)

Lucky-Lodhi2004:
> Thank you but it seems we already have an active PR open for this that was mentioned on the issue, [here](https://github.com/aeon-toolkit/aeon/pull/2980)

Requested changes were not made there, I think

MatthewMiddlehurst:
Leaning towards closing this due to the failing tests and duplicate PR with more discussion. If there is any valid reason these changes are more suitable please make a case.

PR #3173, opened 2025-12-14
[BUG] Fix handling series with extremely small scale

### LINKED ISSUE #2799: [BUG] Handling series with extremely small scale
### Describe the issue

Series with extremely small values (i.e. e-23, see #2787) can have unintended consequences. I do not think we should support series of this scale, as any floating point imprecision or mitigation we use for this will have an impact on these series. It is not worth accounting for this throughout the package, and silently processing (or crashing) does not seem like a good idea.

In the series/collection base class, check the scale of the input series. If this is too small, raise an error.

### Additional context

This does not make the users data unusable, as they should be able to just increase the scale of the data prior to input if necessary.

satwiksps:
## Reference Issues/PRs
- Fixes #2799.

This PR implements a validation check to reject time series data with an extremely small scale (max absolute value < `1e-19`) before it reaches estimators. This prevents floating point underflows and "divide by zero" errors during standard deviation calculations in algorithms.

**Changes:**
1. **`aeon/base/_base.py`**: Added `_check_scale(self, X)` to `BaseAeonEstimator`. This method checks if the maximum absolute value of the input `X` is below a safe threshold (`1e-19`).
2. **`aeon/base/_base_collection.py`**: Integrated `_check_scale` into `BaseCollectionEstimator._preprocess_collection`. This ensures all classifiers, clusterers, and regressors fail fast with a clear error message when given micro-scale data.
3. **`aeon/base/_base_series.py`**: Integrated `_check_scale` into `BaseSeriesEstimator._preprocess_series`. This extends the safety check to forecasters and transformers.
4. **`aeon/base/tests/test_base_collection.py`**: Added `test_check_small_scale` to verify that inputs with extremely small scales correctly raise a `ValueError`.

**Performance Benchmark:**
I verified that adding this check to the base class does not introduce significant overhead. A benchmark (10,000 runs) shows the check takes approximately **6–45 microseconds** per call, adding effectively zero cost to the training process.

| Data Type | Total (s) | Per call (μs) |
| :--- | :--- | :--- |
| Numpy Array | 0.06744 | **6.74 μs** |
| Pandas DataFrame | 0.07735 | **7.73 μs** |
| List of Arrays | 0.45155 | **45.15 μs** |

The threshold `1e-19` was chosen as a safety buffer (the reported crash was at `1e-23`) to catch instability before it reaches the floating point precision limit.

No.

## Any other comments?

No.

### For all contributions


SebastianSchmidl (commented):
How expensive is this check?

If we want to include such functionality, we should include some tests. Maybe, @TonyBagnall still has access to the data that caused the original bug.

satwiksps:
Thank you for the review, @SebastianSchmidl 

> we should include some tests

I have updated the PR to include unit tests for `_check_scale` to ensure inputs with extremely small scales (e.g., `1e-20`) correctly raise a ValueError.

> How expensive is this check?

I ran a benchmark to verify that adding this check to the base class does not introduce any significant overhead to the `fit` method.The check takes approximately **6–45 microseconds** per call depending on the data structure. This effectively adds zero cost to the model training process.

<details>
<summary>Please click to view code used for benchmark test</summary>

```python

import timeit
import numpy as np
import pandas as pd
from aeon.base._base import BaseAeonEstimator

class Benchmarker(BaseAeonEstimator):
    def __init__(self):
        super().__init__()

    def check(self, X):
        return self._check_scale(X)

X_np = np.random.rand(100, 10)
X_list = [np.random.rand(10, 10) for _ in range(10)]
X_df = pd.DataFrame(X_np)

checker = Benchmarker()

print("--- Performance Benchmark (10,000 runs) ---")
N = 10000
results = {
    "Numpy Array": timeit.timeit(lambda: checker.check(X_np), number=N),
    "List of Arrays": timeit.timeit(lambda: checker.check(X_list), number=N),
    "Pandas DataFrame": timeit.timeit(lambda: checker.check(X_df), number=N)
}

print(f"\n{'Data Type':<20} | {'Total (s)':<10} | {'Per call (us)':<15}")
print("-" * 50)
for name, t in results.items():
    print(f"{name:<20} | {t:.5f}    | {t/N*1e6:.2f} us")


```

</details>

<details>
<summary>Please click to view output of benchmark test</summary>

```
PS C:\Users\satwi\OneDrive\Desktop\dex\aeon> & C:/Users/satwi/AppData/Local/Programs/Python/Python311/python.exe c:/Users/satwi/OneDrive/Desktop/dex/aeon/benchmark.py
--- Performance Benchmark (10,000 runs) ---

Data Type            | Total (s)  | Per call (us)  
--------------------------------------------------
Numpy Array          | 0.06521    | 6.52 us
List of Arrays       | 0.41140    | 41.14 us
Pandas DataFrame     | 0.07866    | 7.87 us
```

</details>

I have also updated the PR description.

MatthewMiddlehurst:
Yes I don't think cost should be an issue for any estimators.

This does not cover the necessary data types available in aeon currently. Ideally we should have a function in utils which does this check for all possible collection and series datatypes.

This should be called in `_check_X` in both datatype base classes. I wouldnt touch `BaseEstimator`.

MatthewMiddlehurst:
Thanks for having a go at this.

I have ended up creating a PR for this as the issue has been lingering for a while, and we keep coming back to it. Dealing with all the datatypes can be complex and this does not do it adequately. Closing in favour of #3315.

Issue #1248, opened 2024-02-25
[MNT] Test consistency between CPU & GPU version of ROCKET


baraline:
### Describe the issue

After the introduction of the GPU version of ROCKET in #1199, we need new test in order to make sure the CPU and GPU version of rocket are consistent when given the same random seed.

### Suggest a potential alternative/fix

The goal of this issue is to implement a test [here](https://github.com/aeon-toolkit/aeon/tree/main/aeon/transformations/collection/convolution_based/rocketGPU/tests) that will transform a small test dataset using both CPU and GPU versions of ROCKET initialized with the same random seed, and test that the transformation result are similar using the `assert_array_almost_equal` function from `numpy.testing`

hadifawaz1999:
given both versions uses numpy to randomly generate the convolution parameters, it should be simple, if it was one numpy one tensorflow than it would have been problematic

we should make sure both versions are using the new api generator in numpy, use the same dataset, and make sure we both normalize the input

NOTE: the rocket-gpu only works on equal length series, limitations of accelerating with gpu is batching, and batching only works with equal length series
So we should test only with equal length series, univariate and multivariate

TonyBagnall:
replaced by #1699

Adityakushwaha2006:
Hey, I would like to work on this issue, I've reviewed the previous changes and understood the issue.
I believe I can prop up the test cases and a possible fix for the same.

Adityakushwaha2006:
@aeon-actions-bot assign @Adityakushwaha2006

Adityakushwaha2006:
Hey moderators, I’ve been digging into the Rocket (CPU) vs ROCKETGPU implementations to figure out why exactly they diverge.

It turns out it’s a bit deeper than just the Random Number Generator. There are two structural mismatches we need to handle:

**Sparse vs Dense: The CPU version randomly selects a subset of channels for each kernel (Sparse behavior). The current GPU version convolves over all channels (Dense behavior).

Padding differences: CPU uses explicit integer padding logic, whereas the GPU relies on TensorFlow's "SAME"/"VALID" strings, which handles edge cases differently.**

The Fix I'm Proposing: To get them consistent without rewriting the entire GPU kernel, I think we should just bridge them:

Unified Gen: Inside ROCKETGPU.fit, we call the standard CPU _generate_kernels to get the "source of truth" weights.

[...]

PR #3175, opened 2025-12-14
[ENH] Add unified ROCKET interface with device parameter

### LINKED ISSUE #3174: [ENH] Proposal: Unified ROCKET Interface with 'device' parameter

while I was deep-diving into the codebase to fix the ROCKET GPU consistency issue (working on that parallelly), I spent a lot of time tracing the execution flow between Rocket and ROCKETGPU.

One thing that really stood out to me was the friction in switching between hardware. Right now, if a user wants to move from CPU to GPU,they have to change their imports, rename the class, and rewrite their script.It seemed very redundant and like we were exposing the internal architecture to the user rather than giving them a clean API.

### The Proposal
I’ve been researching this extensively, and I propose we unify these two into a single interface. The user should just be able to express their intent.

What it would look like:

Python
```

# Default (Legacy behavior)
Rocket(n_kernels=10000, device="cpu")

# Explicit GPU usage
Rocket(n_kernels=10000, device="gpu")

# Automatic selection (Best available hardware)
Rocket(n_kernels=10000, device="auto")
```

_No response_

### Additional context

**The Logic & "Nitty-Gritties"**
I didn't want to suggest this until I was sure it wouldn't break anything, so I’ve researched the potential edge cases. Here is the implementation plan I’ve drafted to handle the complexities:

**The "Switchboard" Architecture:**

I plan to rename the existing Rocket to _RocketCPU (internal) so the original Numba logic remains strictly untouched.

The new Rocket class will act as a Facade, dispatching the work to either the CPU or GPU backend based on the device param.

**Handling Dependencies (The tricky part):**

Lazy Loading: I’ve ensured that tensorflow is only imported if the user explicitly asks for device='gpu'. CPU users won't carry that overhead.

Pickling: Since GPU objects (TensorFlow) can't be pickled, I’ve designed the wrapper to throw clear, helpful errors for the GPU variant while keeping the CPU variant fully pickle-safe.

Parameter Mismatches: I noticed CPU and GPU params differ (e.g., n_jobs vs batch_size). I’ve mapped these out so the wrapper warns the user if they pass an incompatible parameter, rather than silently ignoring it.

**_Current Status_**
I have already prototyped this locally and verified that it maintains 100% backward compatibility with existing imports. I'll create a PR for this as soon as i can confirm that it shall work consistently and wont require further work on this.

Adityakushwaha2006:
#### Reference Issues/PRs

Fixes #3174

Relates to the ongoing GPU implementation tasks and channel consistency issues discussed in the issues.

This PR implements a unified interface for the `Rocket` transformer, enabling seamless switching between CPU and GPU backends via a `device` parameter.

**Key Changes:**
1.  **Facade Pattern:** Renamed the existing implementation to `_RocketCPU` (internal) and created a new `Rocket` class that acts as a dispatcher.
2.  **Device Parameter:** Users can now pass `device="cpu"` (default) or `device="gpu"`.
3.  **Backward Compatibility:** Full backward compatibility is preserved for all existing users (Numba backend remains the default).
4.  **Soft Dependency:** GPU logic uses lazy imports, so users without TensorFlow installed can still use the CPU version without errors.

This is **Step 1** of the roadmap to fix the random state/channel inconsistencies in the GPU implementation.

No new mandatory dependencies.
* **CPU:** No changes (Numba).
* **GPU:** Requires `tensorflow`, but it is implemented as a **soft dependency**. The code only attempts to import TensorFlow if `device="gpu"` is explicitly selected.

#### Any other comments?

@hadifawaz1999 As discussed in the issue, here is the Unified Interface (`device` parameter).

I’ve opened this now so the CI can run overnight. This PR is the structural prerequisite for the Random State/Channel Fix we discussed.

**Status:**
*  Verified backward compatibility (CPU/Numba remains untouched).
*  Added `device='gpu'` support via a Facade pattern.
*  Added comprehensive tests in `test_rocket_unified.py`.

Once this is reviewed/merged, I have the environment ready to implement the `tf.random` synchronization fix immediately.

##### For all contributions

### DIFF
--- a/aeon/forecasting/base.py
+++ b/aeon/forecasting/base.py
@@ -5,9 +5,14 @@ A simplified first base class for forecasting models.
 """
 
 __maintainer__ = ["TonyBagnall"]
-__all__ = ["BaseForecaster", "DirectForecastingMixin", "IterativeForecastingMixin"]
+__all__ = [
+    "BaseForecaster",
+    "DirectForecastingMixin",
+    "IterativeForecastingMixin",
+    "SeriesToSeriesForecastingMixin",
+]
 
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from typing import final
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


Adityakushwaha2006:
I will be working to identify the cause of the soft dependencies failure.
Please hold off a deep review until i get the Cl green .
Thanks

hadifawaz1999:
i am not sure this is a good way to go, i think its best if first thing to do is keep both implementation separate and make them produce the same thing and check with a test. Then we can see if merging them under the "device" parameter would be better

Adityakushwaha2006:
@hadifawaz1999 , Makes complete sense, thank you for the guidance, i was worried a bit about the same aswell.
I'll change my implementation strategy , Ill hold off #3175 for later and convert this to draft, ill work on the Gpu consistency issue  #1248, first ensuring that we're done with that.
Thanks for the direction , really appreciate it : )

I am away from my workstation until Wednesday night (IST), but I will start on the numerical fix immediately when I return.

Adityakushwaha2006:
Closing this PR as Tensorflows convolution cannot achieve parity for multivariate datasets, and since we cant get them to produce the same results, this wont fit well as a feature addition.

PR #3185, opened 2025-12-21
[BUG] Fix broken fit_predict in anomaly detection & add tests

### LINKED ISSUE #3208: [BUG] BaseCollectionAnomalyDetector: fit_predict fails with AttributeError and TypeError
### Describe the bug

The `fit_predict` method in BaseCollectionAnomalyDetector (and subclasses like ClassificationAdapter) is currently non-functional. It crashes immediately upon execution with 100% reproducibility due to logic errors likely inherited from the Series module.

```
import numpy as np
from aeon.anomaly_detection.collection import ClassificationAdapter
from sklearn.ensemble import RandomForestClassifier

# Minimal reproduction case
X = np.random.rand(10, 1, 20)
y = np.random.randint(0, 2, 10)

detector = ClassificationAdapter(classifier=RandomForestClassifier(n_estimators=2))
# Crashes here
detector.fit_predict(X, y)
```

### Expected results

The method should execute the pipeline (fit then predict) and return the anomaly scores/labels without error.

### Actual results

Crash 1 (Immediate): The code fails with an AttributeError because it attempts to call a method that does not exist on Collection estimators:

Image

Crash 2 (Secondary): If the first bug is patched locally, a second crash occurs due to a missing argument in the y-check:

`TypeError: BaseCollectionAnomalyDetector._check_y() missing 1 required positional argument: 'n_cases'`

### Additional Context:
These regressions appear to have persisted because fit_predict was previously excluded from the centralized aeon.testing checks for Collection Anomaly Detectors. Unlike Classification or Regression modules, the CI was not validating this method for this specific module type.

### Versions

_No response_

Adityakushwaha2006:
### Reference Issues/PRs
Fixes #3208 

### What does this implement/fix? Explain your changes.
This PR fixes two critical bugs in `BaseCollectionAnomalyDetector.fit_predict()` and updates the centralized test suite to prevent future regressions.

1. Critical Bug Fixes in BaseCollectionAnomalyDetector: Prior to this PR, the fit_predict() method was completely non-functional.

**Fix 1:** Changed `self._preprocess_series(X)` to` self._preprocess_collection(X)`. The adapter previously crashed with AttributeError because it tried to call a series-specific method.

**Fix 2:** Updated `self._check_y(y)` to `self._check_y(y, self.metadata_["n_cases"])`. The previous call was missing the required n_cases argument, causing crashes when labels were provided.

**Cleanup:** Removed the unused axis parameter from` fit_predict` (collection data does not require axis specification).

2. Testing Infrastructure Update:

Updated `aeon/testing/estimator_checking/_yield_collection_anomaly_detection_checks.py` to explicitly test `fit_predict` for all Collection Anomaly Detectors.

Note: Previously, fit_predict was missed in the centralized checks for this module, which allowed these bugs to persist. This change ensures strictly enforced coverage for all current and future estimators.

### Does your contribution introduce a new dependency? If yes, which one? 
No.

### Any other comments?
Refactored the PR to align with architectural guidance of removing the initial custom test suite in favor of patching aeon.testing directly.

### PR checklist
**For all contributions** 
[x] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you after the PR has been merged.

[x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

**For new estimators and functions** 
[ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).

[ ] (OPTIONAL) I've added myself as a maintainer at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

**For developers with write access** 
[ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

### DIFF
--- a/aeon/anomaly_detection/collection/base.py
+++ b/aeon/anomaly_detection/collection/base.py
@@ -201,14 +201,14 @@ class BaseCollectionAnomalyDetector(BaseCollectionEstimator, BaseAnomalyDetector
         # reset estimator at the start of fit
         self.reset()
 
-        X = self._preprocess_series(X, axis, store_metadata=True)
+        X = self._preprocess_collection(X)
 
         if self.get_tag("fit_is_empty"):
             self.is_fitted = True
             return self._predict(X)
 
         if y is not None:
-            y = self._check_y(y)
+            y = self._check_y(y, self.metadata_["n_cases"])
 
         pred = self._fit_predict(X, y)
 
--- a/aeon/testing/estimator_checking/_yield_collection_anomaly_detection_checks.py
+++ b/aeon/testing/estimator_checking/_yield_collection_anomaly_detection_checks.py
@@ -79,3 +79,12 @@ def check_collection_anomaly_detector_output(estimator, datatype):
         ), "y_pred must contain only 0s, 1s, True, or False"
     else:
         raise ValueError(f"Unknown anomaly output type: {out_type}")
+
+    # Test fit_predict method
+    estimator_fp = _clone_estimator(estimator)
+    y_pred_fp = estimator_fp.fit_predict(
+        FULL_TEST_DATA_DICT[datatype]["train"][0],
+        FULL_TEST_DATA_DICT[datatype]["train"][1],
+    )
+    assert isinstance(y_pred_fp, np.ndarray)
+    assert len(y_pred_fp) == get_n_cases(FULL_TEST_DATA_DICT[datatype]["train"][0])


aeon/anomaly_detection/collection/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests for collection anomaly detection adapters."""
```
SebastianSchmidl:
```suggestion
"""Tests for collection anomaly detectors."""
```

SebastianSchmidl (approved):
I think, we need to address the discrepancy between testing `fit` and then `predict` and `fit_predict` both in `aeon.testing` holistically. However, to fix the existing bugs in the AD module, let's merge this already and refactor later 🚀 .

Please update the PR description to reflect the new changes.

Issue #3039, opened 2025-08-18
[ENH] Validation loss for deep learning clustering, classification and regression


PipaFlores:
### Describe the feature or idea you want to propose

I am testing the different AutoEncoders implemented in aeon. I have noticed that the models are only storing the training loss, accessible for plotting/inspection in `model.summary()`. However, there are no validation losses. 

### Describe your proposed solution

I made a small addition in the `self.training_model_.fit()` in 

https://github.com/aeon-toolkit/aeon/blob/0412d5b50272f58efbcd989ffe18bb46d6378965/aeon/clustering/deep_learning/_ae_abgru.py#L284-L292

where only a new argument of `validation_split = int [0,1] ` is needed according to keras api https://keras.io/api/models/model_training_apis/

I did this on my experiments and it seems to work fine. I would consider this enhancement very useful, as naive users might go for the 2000 n_epochs default run and end up overfitting all their models, and then the manifold learning and subsequent clustering makes no much sense 

I would be happy to create a PR for most, if not all, the models in the deep learning clustering module. 
Basically, it would be a new argument for the initialization of the clusterer classes (validation_split).

I am new to Keras, so please let me know if I overlooked any easier way to do this.

PipaFlores:
Another open question on my side is on how the `save_best_model` works. Its creating some `callbacks.modelCheckpoint(..., monitor = "loss", save_best_only = True)`
So the best model is defined as the one with min(loss)? I would switch this to a validation loss criteria instead, as it is overlooking the issue of overfitting.

hadifawaz1999:
Hello, thank you for raising the enhancement proposal. We've been asked a lot before if all deep learning models do support validation split, in classification regression and clustering, its always been on our to do list but still havent done it, am happy to have this functionality, although to make it clear because you also mentioned the save_best_model thing:

We follow this rule: a deep learning model can be trained to infinite amount of epochs, making it random if we use the "last model" for evaluation in predict. In practice we use the "best model" for evaluation. However the "best model" can be chosen either on the validation loss if a validation split is provided and on the training loss if not. By default in aeon the ModelCheckpoint saves to file the best model using the training loss. Now if ```save_best_model``` is set to true, the keras file will be kept in file after fit is done, if its false then the file will be deleted (see docs of any deep model in aeon). The saving directory is the ```file_path``` parameter and the name of the file will be ```best_file_name``` parameter with a default added ```.keras``` extension.

[...]

PR #3199, opened 2025-12-25
[ENH] extend dl classifiers/regressors to use full keras compile/fit API

jsquaredosquared:
feat: add ability to use full compile and fit APIs

#### Reference Issues/PRs

Partially addresses [this issue](https://github.com/aeon-toolkit/aeon/issues/3039).

For purposes like using validation, it is helpful to be able to use the full Keras compile and fit APIs from the deep learning estimators in Aeon. 

This pull requests modifies the constructors and invocations to the compile and fit methods so that all the arguments from the Keras API can be used.

No  new dependency.

#### Any other comments?

Extending this to the clusterers and forecasters would require some additional work.

##### For all contributions


satwiksps:
Hi @jsquaredosquared, great work on this!

I checked the CI logs, and the failures are due to two main issues:
1.  **Mutation Error:** `scikit-learn` forbids modifying parameters passed to `__init__` (like `self.compile_args`) inside the `fit` method. The line `self.compile_args = {} ...` changes the object's state, which causes the `check_fit_updates_state_and_cloning` test to fail.
2.  **TypeError:** Since `compile_args` defaults to `None`, unpacking it via `**self.compile_args` raises a `TypeError`.

And I think instead of fixing this in every single subclass (which creates a lot of code duplication), would it be a good idea in  moving the logic into a helper method in `BaseDeepClassifier` and `BaseDeepRegressor`. This fixes the mutation bug by using local variables and keeps the codebase clean. 

### Proposed changes for `aeon/classification/deep_learning/base.py` (and similarly for Regressor):

**1. Update `__init__`:**

```python
@abstractmethod
def __init__(self, ..., compile_args=None, fit_args=None):
   # ...
   self.compile_args = compile_args
   self.fit_args = fit_args
   # ...
```

**2. Add the safe helper method:**
```python
def _fit_keras_model(self,model,X,y,batch_size,epochs,verbose,callbacks):
        c_args=self.compile_args if self.compile_args is not None else {}
        # Validation for compile_args...
        # ...
        model.compile(...,**c_args)
        # Use local variable for fit_args
        f_args=self.fit_args if self.fit_args is not None else {}
        # Validation for fit_args...
        # ...
        self.history=model.fit(
             X, y, 
             batch_size=batch_size, 
             epochs=epochs, 
             verbose=verbose, 
             callbacks=callbacks, 
             **f_args
        )
        return self.history
```

Then, you can simply call `self._fit_keras_model(...`) in the `_fit `method of the subclasses. This should clear up the CI errors!

jsquaredosquared:
Thanks for taking a look.

There is a lot of duplication in the model building and fitting. Perhaps a refactor is in order. For now I will just copy and paste because a full-on refactor seems daunting and might need a PR of its own :smiling_face_with_tear: .

MatthewMiddlehurst:
Is this still active? We do not usually review drafts. Feel free to post on discord if you want feedback on changes.

TonyBagnall (changes requested):
thanks @jsquaredosquared, this is a useful addition and the mechanics seem correct. Thre blockers

1. validation_data has to be in Keras's internal format. The classifiers transpose X to channels-last and one-hot encode y before calling fit, but fit_args goes straight to Keras. Passing validation_data=(X_val, y_val) in aeon's own format fails:
   ValueError: ... expected shape=(None, 16, 2), found shape=(None, 2, 16)

To make it work, a user has to transpose X_val themselves and, for classifiers, one-hot encode y_val using the same class order as aeon's internal label encoder. Few users will get that right. validation_data should be converted the same way as X and y.
.
2. ResNetRegressor is missing. It's the only DL classifier or regressor not changed: ResNetRegressor(fit_args={}) raises TypeError: unexpected keyword argument 'fit_args'.
. 
3. No tests. Nothing checks that the arguments reach Keras, that forbidden keys raise, or that validation_data works.

other points
Please move the duplicated blocks into a helper on `BaseDeepClassifier`/`BaseDeepRegressor`, as @satwiksps suggested; that is also the natural place for the conversion in (1).
- Add a docstring note that Keras's `validation_split` takes the *last* fraction before shuffling, so it can validate on a single class when data is sorted by class.
- Also note that the default `ModelCheckpoint`/`ReduceLROnPlateau` callbacks monitor `loss`, so validation data does not affect best-model selection or the LR schedule unless custom callbacks are passed.
- Minor: trailing space in the `compile_args` error message; use double backticks for `compile`/`fit` in the docstrings.

PR #3206, opened 2025-12-26
[ENH] Fix binary IG calculation to respect identical split values

### LINKED ISSUE #1322: [BUG] ShapeletTransform: binary ig calculation problem
### Describe the bug

The current _calc_binary_ig( ) evaluates split points between data points with the same feature values but different labels, which might not be suitable for datasets that contain a lot of such data points.

from aeon.transformations.collection.shapelet_based._shapelet_transform import _calc_binary_ig
orderline = [(2,-1),(2,-1),(2,1),(3,1),(3,1)]
c1, c2 = 3, 2
_calc_binary_ig(orderline,c1,c2)

### Expected results

0.42

### Actual results

0.97

### Versions

System:
    python: 3.9.7 (tags/v3.9.7:1016ef3, Aug 30 2021, 20:19:38) [MSC v.1929 64 bit (AMD64)]
executable: c:\xxx\python.exe
   machine: Windows-10-10.0.19041-SP0

Python dependencies:
          pip: 22.3.1
   setuptools: 57.4.0
 scikit-learn: 1.4.0
         aeon: 0.7.1
  statsmodels: None
        numpy: 1.24.0
        scipy: 1.10.1
       pandas: 2.0.3
   matplotlib: 3.5.0
       joblib: 1.3.2
        numba: 0.58.1
     pmdarima: None
      tsfresh: None

musaqlain:
#### Reference Issues/PRs
Fixes #1322

This PR fixes a logic error in `ShapeletTransform` where the binary Information Gain (IG) calculation was incorrectly evaluating split points between identical distance values.

**Problem:**
The `_calc_binary_ig` function iterated through every index in the sorted orderline to test split points. It failed to check if the current distance was equal to the next distance (`orderline[i].distance == orderline[i+1].distance`). This caused the algorithm to calculate splits *inside* groups of identical values, which is mathematically invalid and resulted in inflated IG scores (e.g., returning ~0.97 instead of ~0.42 for the reproduction case).

**Solution:**
Added a check to strictly skip invalid split points where the distance values are identical.
```python
if orderline[split][0] == orderline[split+1][0]:
    continue

### DIFF
--- a/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
+++ b/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
@@ -696,12 +696,15 @@ def _calc_binary_ig(orderline, c1, c2):
     c2_count = 0
 
     # evaluate each split point
-    for split in range(len(orderline)):
+    for split in range(len(orderline) - 1):
         next_class = orderline[split][1]  # +1 if this class, -1 if other
         if next_class > 0:
             c1_count += 1
         else:
             c2_count += 1
+
+        if orderline[split][0] == orderline[split + 1][0]:
+            continue
 
         left_prop = (split + 1) / total_all
         ent_left = _binary_entropy(c1_count, c2_count)
--- a/aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
+++ b/aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
@@ -1,0 +1,25 @@
+"""Tests for shapelet transform functions."""
+
+__maintainer__ = []
+
+from aeon.transformations.collection.shapelet_based._shapelet_transform import (
+    _calc_binary_ig,
+)
+
+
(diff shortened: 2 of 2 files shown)


musaqlain:
PTAL @MatthewMiddlehurst @TonyBagnall

MatthewMiddlehurst:
This is less of a fix and more of a change to the algorithm itself. I would want to see it evaluated on benchmark datasets to see how it changes results before going forward. Either way skipping the last IG check does not seem correct.

TonyBagnall:
hi, I looked at this and I dont really think the change is necessary. Its a design decision, and it as no impact on performance when I tested it, so I think atm I go by if it aint broke, dont fix it. I'll convert to draft, but if you can show it adds value, happy to switch back

TonyBagnall:
hi, closing this now, its still an issue but given the whole module will be revamped, I think this is stale, reopen if you want to come back to it

Issue #313, opened 2023-04-21
[ENH] ROCKET variants in GPU implementation


hadifawaz1999:
As done in [TSAI](https://github.com/timeseriesAI/tsai/blob/main/tsai/models/ROCKET_Pytorch.py), ROCKET and its variants can be implemented in GPU (tensorflow then pytorch at some point after closing this issue). Here is the to do list

- [x] ROCKET
- [ ] MiniROCKET
- [ ] MiniROCKET multivariate
- [ ] MiniROCKET multivariate variable
- [ ] MultiROCKET
- [ ] MultiROCKET multivariate

TonyBagnall:
replaced by #1699

PR #3211, opened 2025-12-29
[ENH] GPU Rocket 2.0: CuPy Backend ,Exact CPU Parity (<1e-7), 100x Speedup, Full Multivariate Support

Adityakushwaha2006:
#### Reference Issues/PRs

Supersedes #3177
Relates to #313
Relates to #1248

This pull request implements a high-performance, GPU-accelerated backend for the `Rocket` transformer utilizing **CuPy**.

This implementation achieves **strict numerical parity** with the CPU baseline and introduces full support for **multivariate** time series, resolving the architectural limitations and accuracy issues identified in the previous TensorFlow-based approach (PR #3177).

### Key Improvements

* **Performance:** Delivers significant acceleration over both the standard CPU implementation (up to 14x) and the Legacy TensorFlow prototype (up to 100x).
* **Numerical Accuracy:** Guarantees numerical equivalence to the CPU implementation. Benchmarks confirm that while the Legacy TF version suffered from massive deviations (MAE > 600), the CuPy implementation maintains strict parity (MAE < 1e-7).
* **Multivariate Support:** Natively handles multivariate input data, addressing a critical functional gap in the previous TensorFlow prototype.
* **Minimal Footprint:** Integrates `cupy` as a soft, optional dependency, avoiding the substantial package overhead associated with TensorFlow, reducing excessive bloat for the users.

### Benchmarks & Validation

The following benchmarks demonstrate the performance gains and correctness verification.

(GPU: 1650 ti, CPU: i7)
(Speed and breakeven points will be better on better GPUs like RTX 4090)
(Note: legacy TF refers to the original TF implementation from #1199)
**1. Performance: CuPy vs. CPU**
*Detailed speedup analysis across varying dataset sizes.*
image

**Analysis:** The CuPy implementation demonstrates a non-linear performance gain as dataset dimensions increase. For standard time-series lengths (5,000 - 10,000), the GPU backend achieves a speedup factor of approximately **14.7x** compared to the CPU implementation. The break-even point occurs at approximately 1,000 timepoints, making this backend highly efficient for medium-to-large scale datasets.

**2. Performance: CuPy vs. TensorFlow (Legacy)**
*Comparison against the previous GPU prototype.*
image

**Analysis:** The architectural shift to CuPy yields massive performance gains over the TensorFlow legacy code. At a time series length of 10,000, the CuPy implementation is **101.5x faster** than the TensorFlow version. This also confirms that the overhead of the TensorFlow graph execution was the primary bottleneck in the previous iteration.

**3. Deviation Analysis & Numerical Parity**
*Verification of numerical stability and equivalence with CPU kernels.*

image

**Analysis:** I conducted an extensive audit across 15 datasets (6 Univariate, 9 Multivariate) to compare the Mean Absolute Error (MAE) against the CPU baseline. The results below demonstrate that the CuPy implementation is statistically indistinguishable from the CPU version, whereas the Legacy TF version failed to maintain scientific precision.

| Metric | Legacy TF (GPU) | CuPy (New GPU) |
| :--- | :--- | :--- |
| **Best MAE** | 1.33 | **6.28e-08** |
| **Worst MAE** | 648.18 | **3.02e-05** |
| **Datasets Passing Parity (< 1e-5)** | 0 / 15 (0%) | **15 / 15 (100%)** |

#### Detailed Dataset Breakdown

**Univariate Datasets**
| Dataset | Legacy TF MAE | CuPy MAE |
| :--- | :--- | :--- |
| GunPoint | 1.40 | **6.30e-08** |
| ItalyPowerDemand | 1.34 | **6.32e-08** |
| Chinatown | 648.18 | **3.02e-05** |
| TwoLeadECG | 1.53 | **8.54e-08** |
| ECG200 | 1.42 | **8.31e-08** |
| Coffee | 1.33 | **6.28e-08** |

**Multivariate Datasets**
| Dataset | Legacy TF MAE | CuPy MAE |
| :--- | :--- | :--- |
| BasicMotions | 2.29 | **1.41e-07** |
| Epilepsy | 1.89 | **1.12e-07** |
| NATOPS | 3.61 | **1.50e-07** |
| ERing | 2.11 | **1.44e-07** |
| Handwriting | 3.30 | **2.67e-07** |
| ArticularyWord | 2.42 | **1.67e-07** |
| RacketSports | 3.20 | **1.97e-06** |
| StandWalkJump | 3.64 | **1.77e-07** |
| UWaveGesture | 1.61 | **1.12e-07** |

### Architectural Justification: Transitioning from TensorFlow to CuPy

Following the development of the TensorFlow-based implementation (PR #3177), extensive testing identified two critical blockers that necessitated a pivot to CuPy:

1.  **Multivariate Complexity:** TensorFlow lacked the flexibility required for raw kernel manipulations in multivariate scenarios, leading to inefficient workarounds and the high error rates observed in the table above and the general functioning.
2.  **Dependency Overhead:** The inclusion of TensorFlow as a dependency introduced disproportionate bloat relative to the performance gains for smaller datasets.

**The CuPy Advantage**
Switching to **CuPy** allowed for direct, CUDA-like kernel definitions within Python. This enabled a direct replication of the `aeon` CPU logic, successfully solving the multivariate issue while maintaining a lightweight dependency profile.

### Future Development

This architecture establishes a robust foundation for a "GPU-First" convolution module. This framework will facilitate the rapid porting of:
* MiniRocket
* MultiRocket
* Hydra
and subsequently other extremely important algorithms like HIVE-COTE and Shapelets.

Yes, `cupy`.

It is implemented strictly as an **optional/soft dependency**. The GPU backend is only initialized if explicitly requested by the user, ensuring no additional overhead for standard CPU-only workflows.

#### Any other comments?

**Hardware version**
NVIDIA Gtx 1650 ti / i7 / 16GB RAM

**Reviewers**
cc: @hadifawaz1999 @TonyBagnall @MatthewMiddlehurst

##### For all contributions


Adityakushwaha2006:
Thanks for flagging issue #3212 @MatthewMiddlehurst . I've analyzed this against my CuPy implementation, and it won't negatively impact the GPU variant. Since ROCKETGPU delegates kernel generation to the CPU Rocket class to ensure parity, fixing the RNG handling here will automatically propagate the correct behavior to the GPU version without requiring kernel changes. 

Just wanted to bring to your notice.

Adityakushwaha2006:
Hey @MatthewMiddlehurst, I understand the concerns youve raised about the implementation.
I want to make sure Im understanding the direction correctly before doing a major rework. I'm willing to put in the effort to bring this in line with aeons standards.

This seems like significant progress...so if you can provide guidance on what the 'right' approach looks like, we can definitely have a better ,more standard implementation.
If the fundamental direction is incompatible with what aeon needs, please let me know that too so I don't waste your time with extensive revisions that arent needed

Thanks for considering this.

Adityakushwaha2006:
In the meanwhile i was trying to optimise the other rocket family , with their GPU implementations, and see how CuPY fairs out there, i have some benchmarking results from the **minirocket GPU implementation** aswell, which i just finished with about 10 minutes back.

Attaching them here:
image
image
image


I do still think that CuPy is the way to go forward if we want a GPU implementation in aeon...the effort in adding the entire structure is expected and im not seeing this as a one off PR either, i do understand that this will need major work, but that too seems essential for our progress.
Let me know : )

MatthewMiddlehurst:
Ignoring the testing and unrelated changes to the issue which I'm not sure why it has been implemented like this, I do not want to start maintaining CUDA code and add in dependencies for 1 second of speed up. Your images seem to plot the same things multiple times?

Adityakushwaha2006:
Each graph set plots the speedup between the CPU version and the GPU version on different channels (1,2...6) (so ranging from univariate to multivariate)
Just to clarify the speedup, its not about saving 1 second on a single run. Its about 24x faster execution on multivariate data.

As per my understanding why i thought this was of immense importance was because in a typical research workflow where someone would run rocket on 100 different parameter configurations or datasets, that would make the difference between 2 hours vs 5 minutes of run time.

I understand the concern about CUDA maintenance and dependencies though. If the maintenance burden outweighs the performance benefit for aeons use cases, I completely get that too.

baraline:
I agree with Matthew on maintain CUDA kernels, I don't think we want to get into that level of complexity.

One alternative if we want to use GPU implementation (which I think is a good thing) would be to use CuPy native operation (i.e. cupy.convolve etc...) so that the code is close to a numpy implementation, which is the whole point of CuPy, to be a drop in replacement for scipy/numpy.

hadifawaz1999:
putting my comment here as well

the only change am happy to get is to use the same kermel generation function of the cpu implementation, but keep using tensorflow for the convolution

MatthewMiddlehurst:
Sorry the labels for the latest plot all say number of cases up to 10,000 so it is a little confusing if it is altering other metrics. Generally yes there will be multiple runs, and speed ups are good, but this as it stands is a bit of a maintainability liability to be honest. In real terms ROCKET is a quick algorithm already, and the speed-up is not worth the potential issues.

Adityakushwaha2006:
Understandable hiccups , i'll try and list all the options available to us as of our current standing.

**1) CuPy implementation with convolve (without custom kernel):**
Pros: easy to maintain, code similar to CPU, removes TF bloat,easy to understand
Cons: Adds soft dependency
For this approach , i cant say a lot about the speed and parity without testing it first, my understanding would place the speedup to be slightly lesser than the custom kernel method.

**2) Previous TF implementation with Kernel Parity changes:**
Pros: Less codebase change, already exists in PR #3177
Cons: Incredibly slow (~2-3x speedup only), hard to map CPU to GPU, dependency bloat, equivalence point at larger sizes (compared to CuPy)
Speed: ~2-3x vs CPU 
Parity:  supports Univariate parity (MAE < 1e-5) ,  does not support Multivariate parity (MAE 1.61-648.18 due to different convolution algorithms at transform stage)

**3) CuPy custom kernel :**
Pros: extremely fast (14.7x peak speedup), code similar to CPU except kernel, easy to understand,  removes TF bloat , equivalence around ~1k datapoints
Cons: Hard to maintain kernel, adds soft dependency, understandable concerns about mixing with codebase ...all in all does not seem favourable here
Speed: 14.7x peak vs CPU, average 8-12x depending on dataset size
Parity: Supports Univariate parity (MAE < 1e-5, 6/6 datasets pass), supports Multivariate parity (MAE < 1e-5, 9/9 datasets pass)


With a final weigh in of opinions and discussion we can decide if we wish to abandon the CuPy approach in its entirety , or try out the 1st option as suggested by @baraline , or if we wanna revert to the TF implementation opened in #3177 (opening it again for review) as seen favourable by @hadifawaz1999

baraline:
As much as I love CuPy, I agree with Ali too, if we can do it with tensorflow we should, it keeps the GPU framework we use consistent across aeon. 
Even a 2x or 3x speedup is already really nice to have, especially if the maintenance cost is low. But yeah we might want to work on the other issues it has in terms of parity

MatthewMiddlehurst:
"Parity: supports Univariate parity (MAE < 1e-5) , does not support Multivariate parity (MAE 1.61-648.18 due to different convolution algorithms at transform stage)"

Not really sure how this works, why is it different?

Either way this does not seem to be the approach we want, so closing.

Adityakushwaha2006:
TF conv1d uses matrix ops that accumulate in parallel. For univariate , theres only one multiplication so order 
doesn't matter. For multivariate, the different accumulation order causes float rounding differences to compound.
Also , TFs convolution algorithm isnt fixed either, so its completely unpredictable

Totally able to understand the concerns tho, ill get on with the TF implementation betterment : )

PR #3241, opened 2026-01-15
[ENH] adding type hints for files in aeon/classification/deep_learning

### LINKED ISSUE #1454: [ENH] Type hints for primitives and string arguments

stephanielees:
#### Reference Issues/PRs

Fixes #1454 some part

Adding type hints to files in `aeon/classification/deep_learning`

None

#### Any other comments?

Just a little question... I notice that in the documentation of the deep learning models, the parameter like `n_filters` is said to be default to a list of integers. However, in the `def init`, these parameters are always defaulted to `None`. Is that intentional?

Don't worry, I didn't change anything about the default values of any parameters; I only added the type hints. I'm just curious about the design of the code :)

##### For all contributions


aeon/classification/deep_learning/_cnn.py
```diff
@@ -27,7 +27,7 @@ class TimeCNNClassifier(BaseDeepClassifier):
     kernel_size : int or list of int, default = 7
         Kernel size of convolution layers, if not a list, the same kernel size
         is used for all layer, len(list) should be n_layers.
-    n_filters : int or list of int, default = [6, 12]
+    n_filters : int or list of int, default = None
```
MatthewMiddlehurst, line 30: Does None map to [6, 12] somewhere in fit? If so it should be mentioned in the docstring.

Are there similar cases in other files?

stephanielees: Thanks for your help, now I get it. `None` is mapped to [6, 12] when initializing the classifier object. More specifically, when we initialize the `TimeCNNClassifier`, it will call `TimeCNNNetwork` from `/aeon/networks` to do the mapping.

I'll modify the docstring then, mentioning that the mapping would happen automatically if the `n_filters` and/or `kernel_size` is left to `None`.

MatthewMiddlehurst: The wording used is a bit stiff. "If set to None, defaults to [6, 12]" would be better imo. same for others.

PR #3246, opened 2026-01-18
[MNT] Update `numba` versioning

MatthewMiddlehurst:
Allows the latest version of `numba`. Skips some expected results tests.


MatthewMiddlehurst:
Agreed. I will do this in a separate PR though and figure out if I can get them removed from the exclude list at the same time.

PR #3250, opened 2026-01-19
[MNT] Update `scikit-learn` upper bound

MatthewMiddlehurst:
problem estimators:

1. rotation_forest_regressor: I've regenerated results as the changes are genuine (see below)
2. weasel. seems to be a typing issue with the words used


TonyBagnall:
just looking at this

Catch22Regressor, 
[
](https://github.com/aeon-toolkit/aeon/actions/runs/23682122267/job/68995718383?pr=3250)

this seems a matter of precision
Expected:
    array([0.638218, 1.090666, 0.583235, 1.575507, 0.484134,
           0.709761, 1.332061, 1.099275, 1.516734, 0.316833])
Got:
    array([0.63821896, 1.0906666 , 0.58323551, 1.57550709, 0.48413489,
           0.70976176, 1.33206165, 1.09927538, 1.51673405, 0.31683308])

TonyBagnall:
FAILED aeon/classification/dictionary_based/tests/test_weasel.py::test_weasel_v2_score - AssertionError: 
Arrays are not almost equal to 4 decimals
 ACTUAL: 0.5454545454545454
 DESIRED: 0.90909
FAILED aeon/classification/dictionary_based/tests/test_weasel.py::test_weasel_score - AssertionError: 
Arrays are not almost equal to 4 decimals
 ACTUAL: 0.5454545454545454
 DESIRED: 0.727272

TonyBagnall:
FAILED aeon/regression/sklearn/tests/test_rotation_forest_regressor.py::test_rotf_output - AssertionError: 
Arrays are not almost equal to 4 decimals

Mismatched elements: 15 / 15 (100%)
Max absolute difference among violations: 0.0045145
Max relative difference among violations: 0.1914099
 ACTUAL: array([0.026 , 0.0245, 0.0224, 0.0453, 0.0892, 0.0314, 0.026 , 0.0451,
       0.0287, 0.04  , 0.026 , 0.0378, 0.0265, 0.0356, 0.0281])
 DESIRED: array([0.0269, 0.0269, 0.02  , 0.0428, 0.0903, 0.0271, 0.0255, 0.0408,
       0.029 , 0.0425, 0.0269, 0.0367, 0.0236, 0.0344, 0.0236])

TonyBagnall:
I looked at the rotation forest regressor, and it looks like a fix in scikit that our testing case hits


claude this time:


_"From the [1.8 changelog](https://scikit-learn.org/stable/whats_new/v1.8.html):

"Fixed a regression in decision trees where almost constant features were not handled properly" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html) (#32259)
"Fixed splitting logic during training in tree.DecisionTree* (and consequently in ensemble.RandomForest*) for nodes containing near-constant feature values and missing values" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html)
"Fix decision tree splitting with missing values present in some features. In some cases the last non-missing sample would not be partitioned correctly" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html) (#32351)

Rotation Forest is a near-perfect trigger for #1 and #2. Each tree is trained on PCA-rotated features built from a bootstrap sample of only 3 columns at a time. It's extremely common for one of the three PCA components on a bootstrap subsample to capture almost no variance — i.e. a "near-constant feature." sklearn 1.7 split those columns one way; 1.8 splits them another (correctly). That cascades through every tree in the ensemble, so every prediction shifts a little — exactly the pattern you're seeing (all 15 elements differ, magnitude ~1–20% relative, no element wildly off).From the [1.8 changelog](https://scikit-learn.org/stable/whats_new/v1.8.html):

"Fixed a regression in decision trees where almost constant features were not handled properly" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html) (#32259)
"Fixed splitting logic during training in tree.DecisionTree* (and consequently in ensemble.RandomForest*) for nodes containing near-constant feature values and missing values" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html)
"Fix decision tree splitting with missing values present in some features. In some cases the last non-missing sample would not be partitioned correctly" [scikit-learn](https://scikit-learn.org/stable/whats_new/v1.8.html) (#32351)

Rotation Forest is a near-perfect trigger for #1 and #2. Each tree is trained on PCA-rotated features built from a bootstrap sample of only 3 columns at a time. It's extremely common for one of the three PCA components on a bootstrap subsample to capture almost no variance — i.e. a "near-constant feature." sklearn 1.7 split those columns one way; 1.8 splits them another (correctly). That cascades through every tree in the ensemble, so every prediction shifts a little — exactly the pattern you're seeing (all 15 elements differ, magnitude ~1–20% relative, no element wildly off)."_


and chatgpt

_The most likely upstream cause of the 1.7 -> 1.8 drift is scikit-learn PR #32259, which landed in 1.8. The 1.8 release notes describe it as fixing a regression where almost constant features were not handled properly in decision trees. The PR discussion is more explicit: after the tree refactor in PR #29458, FEATURE_THRESHOLD was accidentally initialised to 0.0 instead of 1e-7, and 1.8 fixes that. The constant is commented as being there to mitigate precision differences between 32-bit and 64-bit.

That fits RotationForestRegressor unusually well, because aeon casts the rotated PCA features to float32 before fitting the tree. So a tree-side fix specifically about how near-constant features are treated under 32-bit vs 64-bit precision is exactly the sort of change that can move RotF predictions while leaving the high-level algorithm unchanged_

TonyBagnall:
chatgpt thinks this 

es. These two failures are not the logistic-regression issue.

WEASEL uses RidgeClassifierCV on its default path because support_probabilities=False by default, and WEASEL_V2 also fits a RidgeClassifierCV. So test_weasel_score and test_weasel_v2_score are both going through the same sklearn ridge-CV path, not the LogisticRegression(liblinear) path.

The common upstream input is also important: both classifiers feed bag-of-words count features into that ridge classifier. In aeon’s SFA code, sparse bags are explicitly built as csr_matrix(..., dtype=np.uint32), and the dense bag constructors also allocate np.uint32 count arrays. WEASEL stacks those bags straight into RidgeClassifierCV, and WEASEL_V2 does the same after building dense SFA outputs with return_sparse=False.

The likely source of the regression is the sklearn 1.8 ridge refactor. The release notes say RidgeCV, RidgeClassifier, and RidgeClassifierCV gained array-API support in 1.8. In the 1.8 _ridge.py source, _RidgeGCV.fit now records original_dtype = X.dtype, fits in a high-precision float dtype, and then casts intercept_, dual_coef_, and coef_ back to original_dtype at the end. That original_dtype capture and end-of-fit cast-back are not present in the 1.7.2 source.

For WEASEL and WEASEL_V2, that means the ridge model is very likely being fit from uint32 count matrices and then having its learned parameters cast back to uint32. For a linear classifier, that is disastrous because the coefficients and intercept are supposed to be signed floating-point values. That neatly explains why both tests drop together under 1.8, despite using different transforms: the shared failure point is RidgeClassifierCV receiving integer-count bags and then storing integer-cast parameters.

The fix in aeon should be simple: cast the bag matrix to float before calling RidgeClassifierCV.fit in both places. In practice, just before fitting:

all_words = all_words.astype(np.float64, copy=False)

for WEASEL, and

words = words.astype(np.float64, copy=False)

TonyBagnall (approved):
lgtm, thanks

PR #3280, opened 2026-01-26
[MNT] Change stale branch check to run monthly, wait longer and stop tagging people

MatthewMiddlehurst:
Should only create issues for really old branches. Stops tagging people, as it sometimes tags randoms and bots. Gives longer to react to compensate.


.github/workflows/monthly_github_maintenance.yml
```diff
@@ -0,0 +1,30 @@
+name: Weekly GitHub maintenance
```
SebastianSchmidl: "Monthly"

SebastianSchmidl (approved):

PR #3296, opened 2026-02-09
[ENH] Make the ACF function more efficient

TonyBagnall:
speeds up the ACF used in ETS to find seasonality by making it more numba friendly

Current version
```python

def acf(X: np.ndarray, max_lag: int) -> np.ndarray:
    """
    Baseline sample autocorrelation function for lags 1..max_lag.

    For each lag k:
    - Let x1 = X[:-k], x2 = X[k:]
    - Returns Pearson correlation(x1, x2)
    - If both variances are ~0, returns 1 for that lag
    - If exactly one variance is ~0, returns 0 for that lag
    """
    n = X.shape[0]
    if n <= 1 or max_lag <= 0:
        return np.empty(0, dtype=np.float64)

    if max_lag > n - 1:
        max_lag = n - 1

    out = np.empty(max_lag, dtype=np.float64)
    eps = 1e-9

    for lag in range(1, max_lag + 1):
        m = n - lag

        x1 = X[:-lag]
        x2 = X[lag:]

        s1 = np.sum(x1)
        s2 = np.sum(x2)
        m1 = s1 / m
        m2 = s2 / m

        ss1 = np.sum(x1 * x1)
        ss2 = np.sum(x2 * x2)

        v1 = ss1 - s1 * m1
        v2 = ss2 - s2 * m2

        v1_is_zero = v1 <= eps
        v2_is_zero = v2 <= eps

        if v1_is_zero and v2_is_zero:
            out[lag - 1] = 1.0
        elif v1_is_zero or v2_is_zero:
            out[lag - 1] = 0.0
        else:
            out[lag - 1] = np.sum((x1 - m1) * (x2 - m2)) / np.sqrt(v1 * v2)

    return out
```

new version avoids slicing, numpy calls, reduces the iterations and stores and reuses some recalculations

```python

def acf_fused(X: np.ndarray, max_lag: int) -> np.ndarray:
    """
    Faster ACF variant with fused accumulation loops.

    It computes s1, s2, ss1, ss2, s12 in one loop per lag, which reduces
    temporary allocations and repeated passes.
    """
    n = X.shape[0]
    if n <= 1 or max_lag <= 0:
        return np.empty(0, dtype=np.float64)

    if max_lag > n - 1:
        max_lag = n - 1

    out = np.empty(max_lag, dtype=np.float64)
    eps = 1e-12

    for lag in range(1, max_lag + 1):
        m = n - lag

        s1 = 0.0
        s2 = 0.0
        ss1 = 0.0
        ss2 = 0.0
        s12 = 0.0

        for i in range(m):
            a = X[i]
            b = X[i + lag]
            s1 += a
            s2 += b
            ss1 += a * a
            ss2 += b * b
            s12 += a * b

        v1 = ss1 - (s1 * s1) / m
        v2 = ss2 - (s2 * s2) / m

        if v1 <= eps and v2 <= eps:
            out[lag - 1] = 1.0
        elif v1 <= eps or v2 <= eps:
            out[lag - 1] = 0.0
        else:
            cov = s12 - (s1 * s2) / m
            out[lag - 1] = cov / np.sqrt(v1 * v2)

    return out
```

### DIFF
--- a/aeon/forecasting/utils/_seasonality.py
+++ b/aeon/forecasting/utils/_seasonality.py
@@ -77,31 +77,6 @@ def acf(X, max_lag):
 
     return acf_function
 
-    # length = len(X)
-    # X_t = np.zeros(max_lag, dtype=float)
-    # for lag in range(1, max_lag + 1):
-    #     lag_length = length - lag
-    #     x1 = X[:-lag]
-    #     x2 = X[lag:]
-    #     s1 = np.sum(x1)
-    #     s2 = np.sum(x2)
-    #     m1 = s1 / lag_length
-    #     m2 = s2 / lag_length
-    #     ss1 = np.sum(x1 * x1)
-    #     ss2 = np.sum(x2 * x2)
-    #     v1 = ss1 - s1 * m1
-    #     v2 = ss2 - s2 * m2
-    #     v1_is_zero, v2_is_zero = v1 <= 1e-9, v2 <= 1e-9
-    #     if v1_is_zero and v2_is_zero:  # Both zero variance,
-    #         # so must be 100% correlated
-    #         X_t[lag - 1] = 1
-    #     elif v1_is_zero or v2_is_zero:  # One zero variance
-    #         # the other not
-    #         X_t[lag - 1] = 0
-    #     else:
-    #         X_t[lag - 1] = np.sum((x1 - m1) * (x2 - m2)) / np.sqrt(v1 * v2)
-    # return X_t
-
 
 @njit(cache=True, fastmath=True)
 def calc_seasonal_period(data):


TonyBagnall:
10x-50x speed up, its very fast anyway and only called once but every little helps! Seems to be a big caching boost
image

TonyBagnall:
mean and max differences are small, range is >10e-17 when used standard normal series but I want another test to be sure

image

TonyBagnall:
ran another correctness test with range of values, 
image
no difference, although I do think we should probably use a relative tolerance for tiny values

TonyBagnall:
now to see if it changes the test values again!

MatthewMiddlehurst:
Adding a change to this, as this doctest failure is impacting other places as well. I think this should ignore any minor differences

aeon/forecasting/utils/_seasonality.py
```diff
@@ -40,30 +42,65 @@ def acf(X, max_lag):
...
+    #     v1_is_zero, v2_is_zero = v1 <= 1e-9, v2 <= 1e-9
+    #     if v1_is_zero and v2_is_zero:  # Both zero variance,
+    #         # so must be 100% correlated
+    #         X_t[lag - 1] = 1
+    #     elif v1_is_zero or v2_is_zero:  # One zero variance
+    #         # the other not
+    #         X_t[lag - 1] = 0
+    #     else:
+    #         X_t[lag - 1] = np.sum((x1 - m1) * (x2 - m2)) / np.sqrt(v1 * v2)
+    # return X_t
```
MatthewMiddlehurst: should probably remove this. will be in the version history

MatthewMiddlehurst:
nevermind, just seems a bit unstable. Going between two values at least it seems 🙂

aeon/forecasting/utils/_seasonality.py
TonyBagnall: yes sorry, thought it was draft still, changed at haste :)

PR #3304, opened 2026-02-17
[ENH] Multiverse: change all loaders to zenodo, include new problems

TonyBagnall:
add the zenodo keys for the upcoming Multiverse, its coming!

Fixes #3104 

main changes: 
1. Now downloads exclusively from zenodo
2. Generalised into ``_load_collection``, which is then called by ``load_classification`` and ``load_regression``
3. Enforce you cant load a regression with ``load_classification`` and vice versa
4. Split baked in Covid3Months so the discrete version is in its own directory
5. Fixes the download issue #3104 to model scikit-learn. Now if you dont give it a path it downloads to your home directory unless an environment variable is set

This adds all the zenodo keys which will need maintaining if the data is altered. Ultimately tsc.com will also just point to zenodo. There is more tidying up to do and coverage needed, but want this in for next release. 

The complication is with unequal length and missing value data. In e.g. unequal length case, there will be two files: JapaneseVowels.ts (unequal) and JapaneseVowels_eq.ts. Currently these are stored in a single zip on tsc.com. The loader downloads the zip, then inspects the content to decide which to load. 

Zenodo stores the files separately. The new function downloads any present to model current behaviour. 

It currently defaults to always loading the equal length series since that is how it is now. There is a design choice here, and I think possibly change to load the original by default, but will I think require deprecation.

For regression
```python
        str = "ParkingBirmingham" # Unequal length problem
        #Both equal length
        X,y,meta=load_regression(str,extract_path="C:\\Temp\\Regression1\\")
        X,y,meta=load_regression(str+"_eq",extract_path="C:\\Temp\\Regression2\\") 
        #unequal length
        X,y,meta=load_regression(str, extract_path="C:\\Temp\\Regression3\\",load_equal_length=False) 
```

The downloadable files are in dictionaries in the  tser_datasets and tsc_datasets files tsr_zenodo and tsc_zenodo. Choice to keep them separate, because python cant really tell one from the other in memory, and you dont want to build classifiers on regression problems without processing, but they are essentially the same operation.  

simplified ``load_regression`` to this 
```python

    X, y, meta = load_regression(name, split, extract_path,return_metadata=True,
                                 load_equal_length=load_equal_length,
                                 load_no_missing=load_no_missing, problem_dict=tsc_zenodo)
    # Check this is a classification problem
    if "classlabel" not in meta or not meta["classlabel"]:
        raise ValueError(
            f"You have tried to load a regression problem called {name} with "
            f"load_classifier. This will cause unintended consequences for any "
            f"classifier you build. If you want to load a regression problem, "
            f"use load_regression "
        )
    if return_metadata:
        return X, y, meta
    return X, y
```

I have run this with all the zenodo keys and it works as expected.

### DIFF
--- a/aeon/datasets/_data_loaders.py
+++ b/aeon/datasets/_data_loaders.py
@@ -7,7 +7,6 @@ __all__ = [
     "load_from_tsf_file",
     "load_from_arff_file",
     "load_from_tsv_file",
-    "_load_collection",
     "load_classification",
     "load_forecasting",
     "load_regression",
@@ -1608,7 +1607,7 @@ def _load_collection(
 
     Examples
     --------
-    >>> from aeon.datasets import _load_collection
+    >>> from aeon.datasets._data_loaders import _load_collection
     >>> X, y = _load_collection("FloodModeling1")  # doctest: +SKIP
     """
     # If the request is for _eq or _nmv versions, strip that out and set flags.


TonyBagnall:
I think I will need to add some coverage tests, but there are also some legacy functions kicking around, the load forecasting etc need a tidy. What I have done to test this is download the entire zenodo repos with checks the flags are correctly set, reload it all again from disk. Code has changed a bit since then so will rerun it all again

aeon/datasets/dataset_collections.py
```diff
@@ -160,3 +187,32 @@ def get_downloaded_tsf_datasets(extract_path=None):
             if name + ".tsf" in all_files:
                 datasets.append(name)
     return datasets
+
+
+def get_data_home(data_home: str | os.PathLike[str] | None = None) -> str:
```
MatthewMiddlehurst: would be better private IMO

examples/datasets/data_unequal.ipynb
MatthewMiddlehurst, line 1: not looked in-depth but this contains an error

TonyBagnall: its the scikit learn issue
ImportError: cannot import name 'validate_data' from 'sklearn.utils.validation'
they have moved it around, had this somewhere else too, annoying

aeon/datasets/__init__.py
```diff
@@ -9,8 +9,10 @@
     "load_from_timeeval_csv_file",
     "load_anomaly_detection",
     "load_classification",
+    "_load_collection",
```
MatthewMiddlehurst: would remove this and import directly from the file, only public bits should be in the init if possible IMO

aeon/datasets/tsc_datasets.py
```diff
@@ -1,41 +1,65 @@
-"""Datasets in the UCR/tsml data archives from https://timeseriesclassification.com.
+"""Time Series Classification (TSR) datasets in the UCR/tsml data archives.
```
MatthewMiddlehurst: TSC not TSR

examples/datasets/data_unequal.ipynb
MatthewMiddlehurst, line 1: Needs to be ran again now the fix PR is in

aeon/datasets/tsc_datasets.py
```diff
@@ -421,15 +445,364 @@
...
+# 26 EEG TSC datasets in the multiverse [7]
+# See aeon-neuro for more info.
+eeg = [
+    "Alzheimers",
+    "Blink",
+    "ButtonPress",
+    # "EpilepticSeizures", #Univariate
+    "EyesOpenShut",
+    "FaceDetection",
+    # "FeedbackButton", #Univariate
```
MatthewMiddlehurst: We have the 2 univariate ones now i believe.

PR #3310, opened 2026-02-21
[ENH] Improve and add testing to label validation

MatthewMiddlehurst:
Overhaul label validation for classification, regression and anomaly detection to use utility functions

- Unify anomaly detection label checking (@SebastianSchmidl) - main difference is pandas datatypes can accept int instead of just bool and it uses a util function
- disallow single class input for classification (this can still be overridden for cases such as SMOTE, easier to do so now IMO)
- disallow binary inputs for regression (labels with only 0 and 1)

### DIFF
--- a/examples/datasets/datasets.ipynb
+++ b/examples/datasets/datasets.ipynb
@@ -247,7 +247,7 @@
     "\n",
     "X = np.random.random((5, 1, 10))  # Collection of five series\n",
     "y1 = np.array([\"run\", \"walk\", \"run\", \"run\", \"walk\"])\n",
-    "y2 = np.array([1, 0, 1, 1, 0])\n",
+    "y2 = np.array([1, 0, 1, 2, 0])\n",
     "y3 = np.array([0.5, 1.11, 0.545, -0.3675, 2.662])\n",
     "cls = DummyClassifier()\n",
     "reg = DummyRegressor()\n",


aeon/utils/validation/labels.py
```diff
@@ -50,21 +60,30 @@ def check_regression_y(y):
...
+    if len(y) == 0:
+        raise ValueError("y must not be empty.")
 
     y_type = type_of_target(y, input_name="y")
-    if y_type != "continuous":
+    if y_type != "continuous" and y_type != "multiclass":
         raise ValueError(
             f"y type is {y_type} which is not valid for regression. "
             f"Should be continuous according to sklearn.utils.multiclass.type_of_target"
         )
```
TonyBagnall, line 79: this is a bit confusing, if its multiclass its ordinal?

MatthewMiddlehurst: multiclass covers integer values, its a best guess really. impossibly to determie from array input alone. only doing continuous would make [1.0,1.0,1.0,2.0,2.0,2.0,3.0,3.0,3.0] invalid input.

using the method to filter out binary and then manually combing for strings seems sensible to me.

MatthewMiddlehurst: ordinal is a whole other beast we dont really cover yet I think 🙂

SebastianSchmidl (approved):
Anomaly detection parts look good 🚢

PR #3314, opened 2026-02-23
[ENH] Port in kshape

### LINKED ISSUE #3306: [ENH] Port in KShape clustering algorithm

there were some issues with out tslearn wrapper for this, and I would like a bespoke version.

easiest solution is to port in the version from the datum lab, obviously with attribution, and with any numba optimisations we can think of.  

_No response_

### Additional context

_No response_

TonyBagnall:
closes #3306, #2661

This ports in K-Shape, based on TheDatum version managed by the algorithm author
https://github.com/TheDatumOrg/kshape-python
but using the existing SBD distance function and numba normalisation. We currently have a wrapper of the tslearn version, but had some issues with it recently, so given we already have SBD, seems logical to port it in. Some minor differences in implementation 

[1] Datum version computes NCC on the full multivariate array at once using a Frobenius norm in the denominator and summing CC over channels before normalisation. aeon sbd_pairwise_distance computes SBD independently per channel and then averages across channels. See #2674 
[2] We do not guard against zero std dev in SBD.
[3] TheDatum terminate early if labels don't change, we also terminate for very low inertia change.

I'm running on the 112 to compare to tslearn and TheDatum versions for quality and speed, will post results below before making it non draft

### DIFF
--- a/aeon/clustering/_k_shape.py
+++ b/aeon/clustering/_k_shape.py
@@ -149,7 +149,7 @@ class KShape(BaseClusterer):
         if self.n_clusters > X.shape[0]:
             raise ValueError(
                 f"n_clusters ({self.n_clusters}) cannot be larger than"
-                f"n_cases ({X.shape[0]})."
+                f" n_cases ({X.shape[0]})."
             )
         if not isinstance(self.max_iter, int) or self.max_iter < 1:
             raise ValueError("max_iter must be a positive integer.")
@@ -491,8 +491,6 @@ def _shape_extraction(aligned: np.ndarray, P: np.ndarray) -> np.ndarray:
     if aligned.ndim != 2:
         raise ValueError("aligned must be 2D, shape (n_members, n_timepoints)")
     n_members, L = aligned.shape
-    if n_members < 1:
-        raise ValueError("aligned must have at least one member series")
 
     # Normalise each member (rows) across time
     Y = z_normalise_series_2d(np.ascontiguousarray(aligned, dtype=np.float64))  # (n, L)


TonyBagnall:
ran tests on the UCR on the aeon, tslearn and datum implementations. Finished 78  and there is no significant difference between aeon and datum in AMI and ARI on unseen data
image
image
as expected, tslearn is worse.  This is in my mind acceptable so making this ready for review

image
image

TonyBagnall:
in terms of runtim, tslearn was the fastest, completing 78 problems in 51 minutes, then aeon in 73 minutes then datum in 85 minutes. 

In terms of averages for ARI tslearn 0.193, datum 0.266 and aeon 0.271

aeon/clustering/_k_shape.py
```diff
@@ -1,12 +1,264 @@
...
+        dists = self._pairwise_sbd_distance(X, self.cluster_centres_)
+        return dists.argmin(axis=1)
+
+    def _check_params(self, X: np.ndarray) -> None:
+        if not isinstance(self.n_clusters, int) or self.n_clusters < 1:
+            raise ValueError("n_clusters must be a positive integer.")
+        if self.n_clusters > X.shape[0]:
+            raise ValueError(
+                f"n_clusters ({self.n_clusters}) cannot be larger than"
+                f"n_cases ({X.shape[0]})."
```
chrisholder: Need a space:

f" n_cases ({X.shape[0]})."

or the output will be cannot be larger thann_cases

aeon/clustering/_k_shape.py
```diff
@@ -181,3 +433,82 @@ def _get_test_params(cls, parameter_set="default"):
...
+    Note: we re-normalise each aligned member series because zero-padding alignment
+    changes mean and variance, even if the originals were z-normalised.
+    """
+    if aligned.size == 0:
+        raise ValueError("aligned cannot be empty")
+    if aligned.ndim != 2:
+        raise ValueError("aligned must be 2D, shape (n_members, n_timepoints)")
+    n_members, L = aligned.shape
+    if n_members < 1:
+        raise ValueError("aligned must have at least one member series")
```
chrisholder: This last check can never be reached as the two if statements above ensure that n_members >= 1. Remove this check

PR #3315, opened 2026-02-23
[BUG] Handling series with extremely small scale 

### LINKED ISSUE #2799: [BUG] Handling series with extremely small scale

MatthewMiddlehurst:
Closes #2799

Adds two utility functions which raise an error when series scale is very small. This currently happens when any series standard deviation is less than 1e-7.  Our current numba stdev check treats anything less than 1e-8 as 0. This does not raise an error for constant series, so [0, 0, 0, 0, 0] is fine, because that would legitimately have a stdev of 0.

Calls these functions in the _check_X in both series and collection base classes.


TonyBagnall:
this might pick up some errors with EEG, which are sometimes on this scale with valid reasons, but thats a good thing

aeon/clustering/deep_learning/tests/test_deep_clusterer_io.py
```diff
@@ -31,7 +31,7 @@
 @pytest.mark.parametrize("cls", ALL_DEEP_CLUSTERERS)
 def test_deep_clusterer_load_model(cls):
     """Test that all deep autoencoder clusterers load saved Keras models correctly."""
-    X = np.random.randn(4, 10, 1).astype(np.float32)
+    X = np.random.randn(4, 1, 10).astype(np.float32)
```
TonyBagnall, line 34: just a logical mistake of shape I guess, wont have any effect?

TonyBagnall:
I guess the only thing would be whether we want to disable this tiny test with, for example, massive data or for algorithms where its not an issue. But thats not really for this PR

aeon/clustering/deep_learning/tests/test_deep_clusterer_io.py
MatthewMiddlehurst: What I think happened here, was that it was treated at a series of 1 value, which when going through the new scale test broke as it could calculate either the range or stdev. I am a bit surprised it worked before, assuming nothing weird with shape is going on for deep learning. Maybe @hadifawaz1999 knows if so.

MatthewMiddlehurst:
Should be possible to add a global variable and/or flag to disable this if needed yeah.

aeon/utils/validation/series.py
```diff
@@ -306,3 +307,85 @@ def get_type(X, raise_error=True):
...
+        If any channel has std <= threshold and raise_error=True.
+        threshold is negative.
+        Input_type not in SERIES_DATA_TYPES.
+        X is 2D but axis is not 0 or 1.
+    """
+    if threshold < 0:
+        raise ValueError("threshold must be non-negative.")
+
+    t = get_type(X)
+    if (t == "pd.DataFrame" or (t == "np.ndarray" and X.ndim == 2)) and axis not in [
```
SebastianSchmidl, line 356: Didn't we have a utility class to check for valid input types?

MatthewMiddlehurst: that is `get_type`, but it doesnt say which is 2d and which is 1D (np.ndarray can be both)

MatthewMiddlehurst:
yeah makes less sense for series, but keeps it similar to collection and lets me loop through the test data array. Could loop through a list of strings and create a new one but not much difference really

PR #3321, opened 2026-02-25
LOWESS smoother

### LINKED ISSUE #2974: [ENH] LOESS Smoothing

TonyBagnall:
Resolves #2974
Implements the LOWESS smoother in the transformer.series.smoothing package.

Fairly simple algorithm, fits a local linear model. Details in the docstring and testing, explicit comparison to statsmodels version below

Added to the smoothing notebook, put in tests and put it in the API


TonyBagnall:
comparison with statsmodels with this code

```python
import time
from dataclasses import dataclass
from typing import Callable, List, Tuple

import numpy as np

# aeon
from aeon.transformations.series.smoothing import LOWESS

from statsmodels.nonparametric.smoothers_lowess import lowess as sm_lowess


@dataclass
class Row:
    n: int
    aeon_ms: float
    sm_ms: float
    max_abs_diff: float
    rmse: float


def _to_1d(a) -> np.ndarray:
    """Convert aeon output to 1D float array."""
    arr = np.asarray(a)
    return arr.squeeze().astype(np.float64, copy=False)


def _aeon_loess(y: np.ndarray, span: float) -> np.ndarray:
    # degree=1 to match statsmodels LOWESS family
    out = LOWESS(frac=span, it=0, delta=0.0).fit_transform(y)
    return _to_1d(out)


def _statsmodels_lowess(y: np.ndarray, span: float) -> np.ndarray:
    x = np.arange(len(y), dtype=np.float64)
    out = sm_lowess(y, x, frac=span, it=0, delta=0.0, return_sorted=False)
    return np.asarray(out, dtype=np.float64)


def benchmark_loess_vs_statsmodels(
    max_n: int = 10000,
    step: int = 500,
    span: float = 0.3,
    repeats: int = 5,
    warmup: int = 1,
    seed: int = 0,
) -> List[Row]:
    """
    Deterministic benchmark of aeon LoessSmoother vs statsmodels LOWESS (it=0, delta=0),
    measuring runtime and output differences for increasing n, and plotting both.

    Produces two plots:
      1) Runtime vs n (ms per call)
      2) Differences vs n (max abs diff and RMSE)
    """
    rng = np.random.default_rng(seed)
    ns = list(range(step, max_n + 1, step))
    rows: List[Row] = []

    for n in ns:

        # Deterministic test signal, smooth-ish with some noise
        t = np.arange(n, dtype=np.float64)
        y = (0.05 * t) + np.sin(t / 15.0) + 0.1 * rng.standard_normal(n)
        y = y.astype(np.float64, copy=False)
        print(f"n = {n}\n")

        # Warmup (not timed)
        for _ in range(warmup):
            _aeon_loess(y, span)
            _statsmodels_lowess(y, span)
        # Time aeon
        t0 = time.perf_counter()
        for _ in range(repeats):
            aeon_out = _aeon_loess(y, span)
        t1 = time.perf_counter()
        aeon_ms = (t1 - t0) * 1000.0 / repeats
        print(f" Time PR = {aeon_ms}")

        # Time statsmodels
        t0 = time.perf_counter()
        for _ in range(repeats):
            sm_out = _statsmodels_lowess(y, span)
        t1 = time.perf_counter()
        sm_ms = (t1 - t0) * 1000.0 / repeats
        print(f" Time statsmodels = {sm_ms}")

        # Differences (use the last outputs computed above)
        diff = aeon_out - sm_out
        max_abs = float(np.max(np.abs(diff)))
        rmse = float(np.sqrt(np.mean(diff * diff)))

        rows.append(Row(n=n, aeon_ms=aeon_ms, sm_ms=sm_ms, max_abs_diff=max_abs, rmse=rmse))

    # ---- Plot 1: runtimes
    plt.figure()
    plt.plot([r.n for r in rows], [r.aeon_ms for r in rows], marker="o", label="aeon LoessSmoother")
    plt.plot([r.n for r in rows], [r.sm_ms for r in rows], marker="o", label="statsmodels lowess (it=0)")
    plt.xlabel("n (series length)")
    plt.ylabel("Runtime (ms per call)")
    plt.title("LOESS runtime vs series length")
    plt.legend()
    plt.tight_layout()
    plt.show()

    # ---- Plot 2: differences
    plt.figure()
    plt.plot([r.n for r in rows], [r.max_abs_diff for r in rows], marker="o", label="max abs diff")
    plt.plot([r.n for r in rows], [r.rmse for r in rows], marker="o", label="RMSE")
    plt.xlabel("n (series length)")
    plt.ylabel("Difference")
    plt.title("aeon vs statsmodels output differences vs series length")
    plt.legend()
    plt.tight_layout()
    plt.show()

    # Small summary
    last = rows[-1]
    print(
        f"n={last.n}, aeon={last.aeon_ms:.3f} ms, statsmodels={last.sm_ms:.3f} ms, "
        f"max|diff|={last.max_abs_diff:.3e}, rmse={last.rmse:.3e}"
    )

    return rows


benchmark_loess_vs_statsmodels()

```

TonyBagnall:
Speed up vs statsmodels is about an order of magnitude (x8-x10)
image

output the same within tolerances (e-10)
image

PR #3349, opened 2026-03-16
[ENH] SpokenArabicDigits unequal length

TonyBagnall:
run through of multiverse-mini found the key for SpokenArabicDigits was an old version, this corrects that


MatthewMiddlehurst:
This seems to have some other changes which have come along with it. If it is just changing the id number no issue though.

Issue #3371, opened 2026-03-28
[ENH] Dont use Parallel when n_jobs is 1


TonyBagnall:
### Describe the issue

We can avoid the overhead of Parallel creation when n_jobs == 1 by creating a selection branch. This will reduce overhead for heavily pooled classifiers such as ShapeletTransform that currently creates a Parallel for each case in transform. I'll list them here as I find them

- [ ] ShapeletTransform
- [ ] IntervalBased base class (DrCIF etc)

aadya940:
An interesting experiment here would be to try `backend="threading"` in `joblib.Parallel` but with Python 3.13's free threaded build. Since the interpreter now supports no-gil in pure python, the cost of context switch can be genuinely near zero.

Also, could default to `backend="threading"` instead of `loky` whenever free threaded build is available. Happy to try this out when possible if we are interested :))

One potential benefit (not sure), is we work with large arrays and `Parallel` with ` > 1MB` arrays will pickle it to share across processes but free threaded `threading` backend will share it in memory since threads share address space, so that could be beneficial!

PR #3372, opened 2026-03-28
[ENH] ShapeletTransform block transform for threads

TonyBagnall:
#### Reference Issues/PRs
dependent on #3370 
part of #3371 

ST currently creates a new Parallel for each case when transforming. This is tool large a thread pool. This changes transform function to work in blocks of cases with n_blocks dependent on number of jobs.

1. Added a pass through for n_jobs =1, see #3371 
2. _transform_block should really be in Numba since it calles a numba function repeatedly, but that would require unpacking the shapelets away from lists, and that is a bigger change than I want in this PR


baraline (dismissed):
Block logic and skip for n_jobs=1 LGTM for the scope you described

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -412,36 +413,77 @@ def _fit_shapelets(self, X, y, save_distances):
...
+        """Transform a contiguous block of cases."""
+        out = np.empty((stop - start, len(shapelets)))
+
+        for i in range(stop - start):
+            series = X[start + i]
+            for n, shapelet in enumerate(shapelets):
+                out[i, n] = _online_shapelet_distance(
+                    series[shapelet[3]],
+                    shapelet[6],
+                    sorted_indicies[n],
```
dguijo, line 428: Tony, could you fix this typo?
"indices" instead of "indicies"

dguijo (changes requested):
Two minor details, Tony. Thank you!

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: done, good spot

dguijo (dismissed):
LGTM now. Thank you, Tony!

dguijo (approved):
LGTM, Tony! ty

PR #3374, opened 2026-03-28
[BUG] Fix LeftSTAMPi predict() returning incorrect output shape

### LINKED ISSUE #2819: [BUG] `LeftSTAMPi` output is incorrect shape

Nandinisingh07:
Fixes #2819

predict(X) was returning scores for the entire accumulated series instead
of just the new input X. After fit(20 points), predict(15 points) would
return 35 scores instead of 15.

Root cause: fit_is_empty was True, so the base class skipped fit()
during fit_predict(), leaving self.mp as None when _predict() ran.

Changes:
- Set fit_is_empty to False
- _fit() now initialises stumpy on only the first n_init_train points
- _predict() prepends zeros for the init region and scores only new points
- Replaced reverse_windowing with end-of-window score assignment

No new dependencies introduced.

Only _left_stampi.py and its test file are changed.
All unit tests pass locally.


MatthewMiddlehurst:
closing in favour of #3591

Issue #2815, opened 2025-05-18
[AET] Estimator verbose mode


TonyBagnall:
### Describe the feature or idea you want to propose

this is an aeon enhancement proposal open to discussion

I would like to add the ability to print more useful information to the screen through a verbose parameter (or some other approved mechanism).

The primary utility I would like is to tell the user an estimate of how long training will take. Many of our estimators are relatively slow on big problems, and starting them off with no idea how long it will take is frustrating. We have been running experiments on EEG, and I feel the pain. To me this is a good stepping point to proper checkpointing and all part of the road map for improvements for next year or two of improving usability. 

How we do this train time estimation is algorithm specific. I think start with collection transformers, since this is the backbone of most estimators. Broadcast type transformers should be quite easy to estimate: simply measure how long each takes then extrapolate for rough estimate. 

All of this is complicated by threading mechanism, which I think in itself needs re-examining. 

### Describe your proposed solution

As an AET, I would ask people to not make a PR until we have discussed and decided on how we would like to do this.  PRobably best to start without timing estimates and just generalise "verbose" to print easy to calculate information.

SebastianSchmidl:
Are you suggesting to add some kind of progress indicators, such as [tqdm](https://tqdm.github.io), or "just" having more verbose log messages? We do not use any logging framework in aeon currently, right?

baraline:
Both use tqdm and including logging standard library with appropriate message levels (debug, info, warning, error) could be a welcome addition IMO.

TonyBagnall:
> Are you suggesting to add some kind of progress indicators, such as [tqdm](https://tqdm.github.io), or "just" having more verbose log messages? We do not use any logging framework in aeon currently, right?

Im suggesting progress indicators, but logging would be good first step. I've only used logging in java with Log4j, is there a standard python equivalent?

baraline:
The `logging` package is the standard for python, pretty easy to use too. Not sure if we can put it inside numba functions somehow tho

SebastianSchmidl:
We have a monkey-patch to create a tqdm progress bar for joblib parallel loops. However, it does not modify the ETA calculation of tqdm:

https://github.com/TimeEval/TimeEval/blob/main/timeeval/utils/tqdm_joblib.py

There are probably similar ways for other multithreading/multiprocessing backends.

---

> The `logging` package is the standard for python, pretty easy to use too. Not sure if we can put it inside numba functions somehow tho

[...]

PR #3382, opened 2026-04-10
[ENH] Improve verbose for HC2

TonyBagnall:
#### Reference Issues/PRs
Part of #2815 

This PR is primarily to add better verbose print information to HC2, and its components. Decided to not go fully logging as yet. 

Changes
1. Added a log function, with levels we can use later
2. Added a _fit_component method to tidy it up

To do:
make alpha weight a parameter
put initialisation in a function
store a local copy of parameters if passed in init

level 1 verbose

image

level 2 will print out classifier


TonyBagnall:
want to do this, but not on this conflicted branch, easier to start afresh

Issue #1646, opened 2024-06-11
[BUG] BasicMotions results not reproducible when numba turned off


TonyBagnall:
### Describe the bug

replaces https://github.com/aeon-toolkit/aeon/issues/622 with a more specific issue. Since #1598 the only numba related exclusions I am aware of is ACF #625

and this one. I thought #1598 might also fix HC2, but it seems a different issue. Core of it is some numpy functions work differently if numba is turned on than when it is turned off. For example, #1598 was caused by np.argsort treating ties differently gibing different results on rare cases with Catch22. I would guess this fail is also caused by Catch22 and numba/numpy interaction, but it will require some trial and error to find it. 

### Steps/Code to reproduce the bug

disable numba, run results recreation

### Expected results

results should (probably?) be the same with both

### Actual results

test fails

Mukul477:
Hi @TonyBagnall, I’d like to work on this issue if it’s still open.
I’ll investigate the HC2 behavior when numba is disabled and propose a fix or test adjustment.

vickysharma-prog:
Hi @Mukul477 👋  
Nice to see you picked this up.
If you’d like any help with reproducing, testing with numba disabled, or validating fixes across environments, I’d be happy to assist.  
Good luck with the investigation!

Mukul477:
Hi! Thanks for reaching out @vickysharma-prog 

I took some more time to evaluate the issue and realized it’s more complex than I initially expected. Rather than slow things down, I’ll unassign myself so someone else can continue the investigation.

Thanks again for the guidance and support.

vickysharma-prog:
Hi @Mukul477, thanks for the update and for the transparency.
I’m happy to take this on and start investigating. I’ll begin by reproducing the issue locally with numba disabled and try to isolate where the non-determinism is coming from (likely in the Catch22 / HC2 path as discussed earlier).

I’ll post updates here once I have some findings.

TonyBagnall:
so this bug is still there

Results:

  - Numba enabled: maximum difference 0.0071
  - Numba disabled: maximum difference 0.1638

  Issue #1646 is clearly reproduced. Disabling Numba causes substantially different HC2 probabilities. The enabled result also narrowly differs from the stored baseline, consistent with the newer
  Numba-related exclusion noted in testing_config.py.

```python
"""Reproduce https://github.com/aeon-toolkit/aeon/issues/1646."""

import os
import sys
from copy import deepcopy
from pathlib import Path

[...]

PR #3386, opened 2026-04-12
[ENH] Refactor HIVE-COTE V1 and V2 to use a modular BaseHIVECOTE framework

### LINKED ISSUE #663: [GOV] Interpretability for HIVE-COTE V2 Classifier
### Describe the issue

Hi Team,
Just wanted to check if there are any interpretability frameworks that are currently supported for HIVECOTE V2 Classifier like SHAP, LIME, etc? If not, does the team plan to integrate anything in the near future?

_No response_

### Additional context

 I am working on a time-series classification problem statement using HIVECOTE V2 and would really love to interpret the model predictions.

Also, how do you suggest interpreting the results in the current scenario?

li-f8:
#### Reference Issues/PRs

Resolves #663.
See also #1646.

This PR addresses the HIVE-COTE refactoring project goal by combining V1 and V2 into a single framework, modularising the ensemble stage, and allowing users to easily extract component weights (resolving #663).

Key Changes:
1. Modular Framework (`BaseHIVECOTE`): Created a new base class to centralize the CAWPE weighting, `fit_predict` logic, and probability ensembling. It accepts any list of component estimators dynamically.
2. Refactored V1 & V2: Updated `HIVECOTEV1` and `HIVECOTEV2` to inherit from this base class. They now simply act as configuration classes that pass their specific components to `super()._fit()`. Code duplication has been significantly reduced.
3. Interpretability: Exposed a `get_component_weights()` method in the base class. Backward compatibility is fully maintained using `@property` decorators for old attributes like `stc_weight_`.
4. Testing Passed: Both refactored classifiers pass `check_estimator` successfully locally. `_tags` (`capability:multivariate`, `capability:contractable`, etc.) have been properly overridden for V1.

No.

#### Any other comments?

The expected `SKIPPED` status for `check_classifier_against_expected_results` for HIVECOTEV2 (due to the Numba interaction discussed in #1646) is maintained.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you **after** the PR has been merged. *(Note: I will use the bot after merge).*

- [ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html). *(N/A - just refactoring existing estimators)*


TonyBagnall:
thank you for this, we will have a look as soon as we can

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -0,0 +1,132 @@
...
+    _tags = {
+        "capability:multivariate": True,
+        "capability:contractable": True,
+        "capability:multithreading": True,
+        "algorithm_type": "hybrid",
+    }
+
+    def __init__(
+        self,
+        estimators=None,
```
TonyBagnall: given fit doesnt work with estimators =None, I think this should not have default values. Then in fit you should have a check function to make sure they are valid estimators

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -0,0 +1,132 @@
...
+            # 1. Clone model to prevent mutating the original instance
+            est = clone(estimator)
+
+            # 2. Pass global parameters
+            if hasattr(est, "random_state") and self.random_state is not None:
+                est.random_state = self.random_state
+            if hasattr(est, "n_jobs") and self.n_jobs is not None:
+                est.n_jobs = self._n_jobs
+
+            try:
```
TonyBagnall: I dont think you need it in a try catch block. Tsting for the methods existence is better, but if you have checked that the estimators passed are valid, then the interface contract ensures that fit_predict will be present

aeon/classification/hybrid/_hivecote_v1.py
```diff
@@ -106,6 +101,8 @@ class HIVECOTEV1(BaseClassifier):
     """
 
     _tags = {
+        "capability:multivariate": False,
```
TonyBagnall: It inherits these from BaseClassifier, not really needed

aeon/classification/hybrid/_hivecote_v1.py
```diff
@@ -166,133 +152,53 @@ def _fit(self, X, y):
...
+    def stc_weight_(self):
+        return self.get_component_weights().get("STC", 0.0)
 
-        Returns
-        -------
-        y : array-like, shape = [n_cases, n_classes_]
-            Predicted probabilities using the ordering in classes_.
-        """
-        dists = np.zeros((X.shape[0], self.n_classes_))
+    @property
```
TonyBagnall, line 182: I'm not sure about this structure, I will discuss with Matthew

TonyBagnall (commented):
some initial comments, but looks like its going in the right direction, great stuff. We will need to do some tests of logical equivalence and timing, but we can help with that

aeon/classification/hybrid/_hivecote_v1.py
li-f8: These tags are in the Settings that override BaseHIVECOTE, not BaseClassifier.

TonyBagnall:
hi,

I think `BaseHIVECOTE._fit` needs to reset the fitted component state at the start of fitting.

At the moment `fitted_estimators_`, `weights_`, and `component_names_` are initialised in `__init__`, then appended to in `_fit`. This means that calling `fit` twice on the same object will retain the fitted components from the first call and append the new ones from the second call.

For example, after two calls to `fit`, `HIVECOTEV2` would have 8 fitted estimators rather than 4.

I think the base `_fit` should start with something like:

```python
self.fitted_estimators_ = []
self.weights_ = []
self.component_names_ = []

**Comment by li-f8** [comment]
The failing test (examples/base/base_classes.ipynb) is unrelated to this PR — it's a CI environment issue with CPU feature flags. All HIVE-COTE related tests pass.

File: aeon/classification/hybrid/_base_hive_cote.py [hunk 5]
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+from sklearn.base import clone
+from sklearn.metrics import accuracy_score
+from sklearn.utils import check_random_state
+
+from aeon.classification.base import BaseClassifier
+from aeon.utils.validation import check_n_jobs
+
+
+class BaseHIVECOTE(BaseClassifier):
```
MatthewMiddlehurst: It may be better to make this private (name starting with a `_`) until we decide where it is best to put this. Also relaxes documentation responsibilities with the webpage.

aeon/classification/hybrid/_base_hive_cote.py
MatthewMiddlehurst: agreed this seems like a mandatory parameter if possible.

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -0,0 +1,121 @@
...
+        self.verbose = verbose
+
+        super().__init__()
+
+    def _fit(self, X, y):
+        """Fit the ensemble to training data and calculate CAWPE weights."""
+        self._n_jobs = check_n_jobs(self.n_jobs)
+
+        if self.estimators is None or len(self.estimators) == 0:
+            raise ValueError("No estimators provided to BaseHIVECOTE.")
```
MatthewMiddlehurst: Could also check that they are a valid `BaseClassifier` here

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Dynamically traverse and train all the underlying components
+        for name, estimator in self.estimators:
+            est = clone(estimator)
+
+            # Pass global parameters to the cloned estimator
+            if hasattr(est, "random_state") and self.random_state is not None:
+                est.random_state = self.random_state
+            if hasattr(est, "n_jobs") and self.n_jobs is not None:
+                est.n_jobs = self._n_jobs
```
MatthewMiddlehurst, line 95: I would also do this for verbose

MatthewMiddlehurst: The default for self.n_jobs is 1.

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -0,0 +1,121 @@
...
+        Seed for random number generation.
+    n_jobs : int, default=1
+        The number of jobs to run in parallel.
+    verbose : int, default=0
+        Level of output printed to the console.
+    """
+
+    _tags = {
+        "capability:multivariate": True,
+        "capability:contractable": True,
```
MatthewMiddlehurst: This is dynamic so I would default them to false in the base class.

aeon/classification/hybrid/_hivecote_v1.py
```diff
@@ -7,23 +7,18 @@
...
-from datetime import datetime
-
-import numpy as np
-from sklearn.metrics import accuracy_score
-from sklearn.utils import check_random_state
-
-from aeon.classification.base import BaseClassifier
 from aeon.classification.dictionary_based import ContractableBOSS
+
+# import base
```
MatthewMiddlehurst: remove the comment

TonyBagnall:
> The failing test (examples/base/base_classes.ipynb) is unrelated to this PR — it's a CI environment issue with CPU feature flags. All HIVE-COTE related tests pass.

yes thats not on you :)

aeon/classification/hybrid/_hivecote_v2.py
```diff
@@ -292,40 +217,34 @@ def _predict_proba(self, X, return_component_probas=False) -> np.ndarray:
             Predicted probabilities using the ordering in classes_.
         """
         dists = np.zeros((X.shape[0], self.n_classes_))
+        if self.save_component_probas:
+            self.component_probas = {}
```
MatthewMiddlehurst: this should be raising an error as predict/predict_proba should not be changing the object state. not sure why it is not!

MatthewMiddlehurst (commented):
Thanks. Some comments:

aeon/classification/hybrid/_hivecote_v1.py
TonyBagnall: BaseHIVECOTE should default to False

li-f8:
the save_component_probas comment — this feature mutates state inside _predict_proba (self.component_probas = {}), which I agree violates the predict/predict_proba contract. 
I see two options:
A) Remove save_component_probas entirely — cleanest, but a breaking change for any users relying on it.
B) Replace it with a new method like predict_proba_with_components(X) that returns (final_probas, component_probas_dict) without modifying object state.Which approach do you prefer? 
Happy to implement either.

TonyBagnall:
I agree that `_predict_proba` should not mutate estimator state and we do not detect this currently because there is not a test configuration

Dont remove `save_component_probas` outright since a breaking change is unnecessary for this refactor PR. So I prefer option B. 

A method predict_proba_with_components(X) returning final_probas, component_probas seems like a good idea, then implement the shared logic through a private helper so that `_predict_proba` remains side-effect free.

For example:

```python
def _predict_proba_with_components(self, X):
    dists = np.zeros((X.shape[0], self.n_classes_))
    component_probas = {}

    for i, est in enumerate(self.fitted_estimators_):
        probas = est.predict_proba(X)
        dists += probas * self.weights_[i]
        component_probas[self.component_names_[i]] = probas

    sums = dists.sum(axis=1, keepdims=True)
    sums[sums == 0] = 1.0

    return dists / sums, component_probas
```

then:

```python
def _predict_proba(self, X):
    probas, _ = self._predict_proba_with_components(X)
    return probas
```
We should deprecate `save_component_probas` rather than removing it immediately. Any thoughts @MatthewMiddlehurst ?

li-f8:
ok I will try that method,while my current method is     
```
def _predict_proba(self, X) -> np.ndarray:
        """Predict class probabilities for X using CAWPE."""
        dists = np.zeros((X.shape[0], self.n_classes_))

        for i, est in enumerate(self.fitted_estimators_):
            weight = self.weights_[i]
            probas = est.predict_proba(X)
            dists = np.add(dists, probas * weight)

        sums = dists.sum(axis=1, keepdims=True)
        sums[sums == 0] = 1.0
        return dists / sums
```
I delete the method of 
```
def _predict_proba(self, X):
    probas, _ = self._predict_proba_with_components(X)
    return probas
```
if I get the approval of Matthew, I would add the method I deleted before.

aeon/classification/hybrid/_hivecote_v1.py
TonyBagnall: @li-f8 this comment still needs removing

aeon/classification/hybrid/_base_hive_cote.py
TonyBagnall: resolved

TonyBagnall (changes requested):
thanks, getting closer. Return the public save_component_probas parameter to HIVECOTEV2.__init__, preferably deprecate and direct users to predict_proba_with_components(X) instead. 

I would also like some evidence of equivalence of outcome for old and new (run some small experiments) and a timing experiment (compare times for increasing number of cases) just as a sanity check. 

I need to double check but  does _fit mutate the estimators init parameter? V1/V2 call super().__init__(estimators=None, ...), then their _fit does self.estimators = [("STC", ...), ...]. If so, need to store a version in estimators_ and use that for scikit compatibility

li-f8:
Hi Tony,

Done — all three points addressed:

1. Restored `save_component_probas` parameter on `HIVECOTEV2.__init__` with a `DeprecationWarning` directing users to `predict_proba_with_components(X)`.

2. Fixed the scikit-learn compatibility issue you spotted: `_fit` no longer mutates the `estimators` init parameter. V1/V2 now construct their components  into `self._estimators` (a fit-time attribute), and `_BaseHIVECOTE._fit` reads  from `_estimators` first, falling back to `self.estimators`.

3. Equivalence + timing benchmark on `ItalyPowerDemand` (test params, random_state=42, increasing n_cases):

n_cases | old_fit | new_fit | old_pred | new_pred | preds_eq | proba_diff | old_acc | new_acc
--------|---------|---------|----------|----------|----------|------------|---------|--------
     25 |  13.29* |    0.07 |    1.221 |    1.033 |     True |   0.00e+00 |  0.9048 | 0.9048
     50 |   0.09  |    0.12 |    1.420 |    1.428 |     True |   0.00e+00 |  0.9349 | 0.9349
    100 |   0.14  |    0.13 |    1.591 |    1.613 |     True |   0.00e+00 |  0.9427 | 0.9427
    200 |   0.26  |    0.25 |    2.036 |    2.126 |     True |   0.00e+00 |  0.9310 | 0.9310
    400 |   0.61  |    0.62 |    2.846 |    2.823 |     True |   0.00e+00 |  0.9300 | 0.9300

(*n=25 old_fit is the cold-start measurement — numba JIT compilation overhead. Excluding that, fit/predict times are within measurement noise of each other.)

Predictions are bit-exact (preds_eq=True, proba_diff=0 for every n). No speed regression observed.
[benchmark_hc2.py](https://github.com/user-attachments/files/28886417/benchmark_hc2.py)
The benchmark script is at `benchmark_hc2.py` for reference.

TonyBagnall:
Nearly there — a few minor things to address:

  1. Tests needed for new functionality
  
  No test verifies that get_component_weights() returns correct values after fit, that the base class rejects non-BaseClassifier estimators, or that re-fitting resets state correctly (i.e.
  fitted_estimators_, weights_, component_names_ are not accumulated across calls).

  2. Weight properties raise AttributeError before fit

  The original code initialised weights to 0 in __init__:
  self.stc_weight_ = 0
  The PR replaces these with properties:
  @property
  def stc_weight_(self):
      return self.get_component_weights().get("STC", 0.0)
  get_component_weights() accesses self.weights_ and self.component_names_, which only exist after fit. Any code that reads clf.stc_weight_ before fitting will now get AttributeError instead of
   0. The original behaviour should be preserved, e.g. by guarding with hasattr:
  @property
  def stc_weight_(self):
      if not hasattr(self, "weights_"):
          return 0.0
      return self.get_component_weights().get("STC", 0.0)

  3. Verbose output — I intend to rework this in #3382, so happy to let it go for now.

TonyBagnall (changes requested):
almost there, see comments

MatthewMiddlehurst (commented):
Comments have been addressed. Thank you. Happy for this to go in when Tony is satisfied his new comment has been resolved.

li-f8:
Solve all the things mentioned before and also some new checks in test_hc.py

TonyBagnall (changes requested):
This looks very close, and I think the main structural review points have now been addressed. 

I found one small issue that I think should be fixed before merge.

The legacy component weight properties appear to be broken. For example, `stc_weight_`, `tsf_weight_`, `rise_weight_`, `cboss_weight_`, and the equivalent HC2 properties check for private attributes such as `_stc_weight_`. However, `_BaseHIVECOTE._fit` now stores weights in `weights_` and `component_names_`; it does not set `_stc_weight_`, `_tsf_weight_`, etc. As a result, these backward-compatible properties will return `0.0` after fitting, even though `get_component_weights()` contains the actual learned weights.

I would fix this by having the legacy properties read directly from `get_component_weights()`, and making `get_component_weights()` robust before fit, for example by using `getattr(self, "component_names_", [])` and `getattr(self, "weights_", [])`.

Once that is fixed, I would be happy for this to merge, assuming CI is green.

li-f8:
Thanks for catching this! You're right — the legacy `*_weight_` properties were checking for non-existent `_stc_weight_`-style private attributes, so they always returned 0.0 after fit. 
Fixed by following your suggestion:
 - `get_component_weights()` now uses `getattr(self, "component_names_", [])` and `getattr(self, "weights_", [])`, so it returns an empty dict before fit and the real weights after fit.
 - The `*_weight_` properties simply call `get_component_weights().get("<name>", 0.0)` — no separate `hasattr` guard needed.

TonyBagnall (changes requested):
sorry, found another thing really quick to fix

predict_proba_with_components / save_component_probas crashes

_hivecote_v2.py:263 calls self.check_is_fitted(), which doesn't exist in aeon's BaseClassifier — the method is _check_is_fitted() (with the underscore). It's the only wrong usage of that name in the whole classification module. Change this by adding the underscore and its fixed.

Tests dont find this, but thats a separate coverage issue.

li-f8:
Hi Tony, the only failing test on CI is IndividualLITEClassifier with multivariate input on pytest (windows-2022, 3.12, true) — a Keras shape mismatch in the LITE deep learning module ('hybrid-increasse-0-2' layer expects axis -1 to have value 1 but receives shape (4, 20, 2)). This is unrelated to the HIVE-COTE refactor (which only touches aeon/classification/hybrid/) and only fails on Windows + Python 3.12. Linux and macOS CI jobs all pass. Could this be a pre-existing issue on main?

li-f8:
This test failure is not related to the HIVE-COTE refactoring in this PR. 

The `IndividualLITEClassifier` test is failing because:
1. The test uses multivariate input data `(4, 20, 2)` - 2 channels
2. The default LITE mode (when `use_litemv=False`) uses `Conv1D` layers that expect single channel
3. The test parameters in `_get_test_params()` don't set `use_litemv=True` for multivariate support

**Solution**: This is a pre-existing issue in `aeon/classification/deep_learning/_lite_time.py` that should be fixed separately. The test parameters should either:
- Set `use_litemv=True` by default (since LITEMV is designed for multivariate)
- OR add `"capability:multivariate": False` tag to LITE

This is unrelated to the HIVE-COTE refactoring changes in this PR after some checks

TonyBagnall:
> Hi Tony, the only failing test on CI is IndividualLITEClassifier with multivariate input on pytest (windows-2022, 3.12, true) — a Keras shape mismatch in the LITE deep learning module ('hybrid-increasse-0-2' layer expects axis -1 to have value 1 but receives shape (4, 20, 2)). This is unrelated to the HIVE-COTE refactor (which only touches aeon/classification/hybrid/) and only fails on Windows + Python 3.12. Linux and macOS CI jobs all pass. Could this be a pre-existing issue on main?

yes nothing to do with this PR, its an itermittent fault, we have a few, not seen that one before

TonyBagnall (approved):
thanks for this, great job, we got there in the end

li-f8:
examples/base/base_classes.ipynb error,i never meet this before...,this are no details in the error traceback, do I need to change something on the code?

li-f8:
@all-contributors please add @li-f8 for code, test

Issue #3406, opened 2026-05-20
[BUG] error in soft dependencies check for HF hub in monster loader


Cyril-Meyer:
### Describe the bug

Monster loader error when checking soft dependencies of `huggingface-hub` (which is `huggingface_hub`)
https://github.com/aeon-toolkit/aeon/blob/main/aeon/datasets/monster_loader.py

```python
_check_soft_dependencies("huggingface-hub")
```

Bug reproduced on WSL2 and Ubuntu with @hadifawaz1999

### Steps/Code to reproduce the bug

```python
import aeon
import aeon.datasets
d = aeon.datasets.tsc_datasets.monster[0]
aeon.datasets.load_monster_dataset(dataset_name=d)
```
Error

```python
aeon.datasets.monster_loader._check_soft_dependencies("huggingface-hub")
```
Error

```python
aeon.datasets.monster_loader._check_soft_dependencies("huggingface_hub")
```
True

### Expected results

Code should not make dependencies error

### Actual results

```python-traceback
ModuleNotFoundError: No module named 'huggingface-hub'

The above exception was the direct cause of the following exception:

ModuleNotFoundError                       Traceback (most recent call last)
Cell In[2], line 4
      2 import aeon.datasets
      3 d = aeon.datasets.tsc_datasets.monster[0]
----> 4 aeon.datasets.load_monster_dataset(dataset_name=d)

File [~/venv/lib/python3.12/site-packages/aeon/datasets/monster_loader.py:90](http://bothiel:3434/home/cyril/venv/lib/python3.12/site-packages/aeon/datasets/monster_loader.py#line=89), in load_monster_dataset(dataset_name, fold)
     37 def load_monster_dataset(
     38     dataset_name: str, fold: int = 0
     39 ) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
     40     """Load a Monster dataset from Hugging Face Hub.
     41 
     42      MONSTER— the MONash Scalable Time Series Evaluation Repository,
   (...)     88 
     89     """
---> 90     _check_soft_dependencies("huggingface-hub")
     91     from huggingface_hub import hf_hub_download
     92     from huggingface_hub.utils import HfHubDownloadError

[...]

PR #3409, opened 2026-05-20
[BUG] add raise of softdep not working when we install all extras

baraline:
#### Reference Issues/PRs

Related to #3406 

While running tests in CI on environment where we install all_extras, we didn't raise when softdep checks failed, as they had severity=None in the test files. Added a flag to make these test raise. 

One issue is that it's kind of the point of the softdeps checks, as some of these are OS dependent. But the linked issue showed that it can hide real failures. So with this PR we basically say, we trust one OS (here ubuntu) to be able to take all softdeps.

This is more of a quickfix. Maybe a better future solution would be to label those test under different pytest marks (e.g. @pytest.mark.softdepschecks with additional marks for specific OS) that would only run on these CI environment using all_extras.

This PR CI should fail due to the "hugging-face" softdep check.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,6 +10,20 @@ least once, but not necessarily on each operating system / python version combin
 
 __maintainer__ = ["MatthewMiddlehurst"]
 
+import pytest
+
+
+def _is_soft_dependency_skip(report):
+    if not report.skipped:
+        return False
+
+    skip_reason = (
+        report.longrepr[2]
+        if isinstance(report.longrepr, tuple)
+        else str(report.longrepr)
+    )
+    return "soft dependency" in skip_reason.lower()
+
 
 def pytest_addoption(parser):
     """Pytest command line parser options adder."""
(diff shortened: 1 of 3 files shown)


hadifawaz1999:
we had this issue 2 years ago with tensorflow-addons, where it was actually a _ instead of -, and so the tests were not running without us knowing :)

baraline:
If we are OK with it, we can get this one in to check for that while we discuss a better solution. It is supposing we have all soft deps on ubuntu, but for now it is the case. This way we can move forward with the MONSTER data issue.

conftest.py
```diff
@@ -10,6 +10,20 @@
...
+
+def _is_soft_dependency_skip(report):
+    if not report.skipped:
+        return False
+
+    skip_reason = (
+        report.longrepr[2]
+        if isinstance(report.longrepr, tuple)
+        else str(report.longrepr)
+    )
```
MatthewMiddlehurst, line 29: pretty evident what the function is doing as a whole but coud you leave a comment on why this is needed if possible

MatthewMiddlehurst (commented):
Go with 3.11 to avoid the mrseql and mrsqm timebombs for now.

Is there a way to do this as its own test without actually running anything? e.g. install everything, collect all tests then fail skils without running non-skips.

Fine with this going in if you want to link an issue to revisit later.

baraline:
There is definitly a cleaner way to do this, but we need to plan for it properly, so I think linking an issue and listing our options is the way to go here.

conftest.py
baraline: Added some comments/docs any better ?

MatthewMiddlehurst:
This doesnt seem to interact well the current tests due to some classifiers missing from all_extras (which in itself it a bit of a hack due to annoying packages, but thats another topic). I can try make a PR into this branch with its own test if you don't want to take that on. Alternatives involve removing whole estimators which is probably a good idea but takes too long for this case.

MatthewMiddlehurst:
Could also hack it to ignore specific estimators, hacks within hacks

conftest.py
MatthewMiddlehurst: specifically meant that bit of code, im not really sure what in a test could cause the output type the change! other bits seem reasonable to understand just reading through

MatthewMiddlehurst: either way is minor

baraline:
The other option is that we acknowledge that we have an issue in our pipepline, but don't adress it with hacks for now. Just close this, and get the monster PR in with just the corrections.

Then we raise an issue for a proper update of the testing to try to avoid the current pitfalls we are having.

MatthewMiddlehurst:
Sure. Probably can close this or not but feel free to merge the fixes by @Cyril-Meyer

baraline:
I'd prefer that so we don't stack hacks on hacks, closing and opened #3410

Issue #3410, opened 2026-05-23
[ENH] Handling of soft deps in testing


baraline:
### Describe the feature or idea you want to propose

Related to #3406 

We currently call _check_soft_dependencies(...) inside a @pytest.mark.skipif(not ...) decorator.

#3406 : Name mismatch bugs: _check_soft_dependencies("huggingface-hub") always returns False because Python cannot import a module named "huggingface-hub". The test is permanently skipped — even in all_extras CI runs where huggingface_hub is installed. Real bug in monster_loader.py goes undetected.

Minor : 

use pytest.skip(message) in function like _yield_*_checks.py to mark tests as SKIPPED and not PASSED

### Describe your proposed solution

To discuss, but rather than relying on all the @pytest.skip(not ....) we could define a @pytest.mark.requires("package1", "package2", ...) marker. Hook it to a function in conftest.py that handles all skip/pass/raise logic (name resolution for package_import_alias, add hyphen→underscore fallback). Might be worth centralizing package_import_alias as a registry.

Similarly, we can have a mark as  @pytest.mark.run_on_os("ubuntu"), if we need to ignore some OS for some estimator testings.

This is a first thought, so I'm sure I've not covered every angle here, but we need to discuss it.

MatthewMiddlehurst:
Happy to discuss, put together a first pass here based on your old PR. Had half of it done anyway 🙂 

#3411

baraline:
Thanks Matthew, I'll have a look at it once you're done and comment on it.

MatthewMiddlehurst:
Merged that now, should give us some indication if this happens again.

baraline:
Thanks, closing for now

PR #3411, opened 2026-05-23
[MNT] Add test to check that soft dependency skipping is working as intended

MatthewMiddlehurst:
See also #3406 #3409 #3410

Adds a test run that simply checks no tests are skipped containing the reason "soft dependency" or similar when all dependencies are installed.

Skips all actual tests after the initial pytest check for skipping.

pytest checks generated by out testing system no longer xfail (still runs but expected to fail) but skips (does not run at all). I honestly thought we were already skipping and we have never run into and unexpected pass I believe.

Only wraps tests if the pytest flag is set.

### DIFF
--- a/.github/workflows/periodic_tests.yml
+++ b/.github/workflows/periodic_tests.yml
@@ -177,7 +177,7 @@ jobs:
         run: python -m pip list
 
       - name: Run tests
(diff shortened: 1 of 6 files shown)


baraline (approved):
Good for me, it's a good compromise to check whether these tests are actually run with softdeps installed.

PR #3412, opened 2026-05-24
[MNT] Fix `check_non_state_changing_method`

MatthewMiddlehurst:
The previous check used a shallow copy of `__dict__`, so in-place changes inside lists, dictionaries, fitted estimators, and other nested objects could go undetected. Direct deep comparison is not reliable for all estimator state, particularly objects that cannot be pickled.

The new approach detects nested changes for serializable attributes while retaining a safe identity-based fallback.


aeon/classification/dictionary_based/_redcomets.py
```diff
@@ -588,7 +588,7 @@ def _parallel_sax(self, sax_transforms, X):
         """
 
         def _sax_wrapper(sax):
-            return np.squeeze(sax.fit_transform(X), 1)
+            return np.squeeze(sax.transform(X), 1)
```
TonyBagnall, line 591: is this related to the PR? something the new test picked up? Not mentioned in the description

aeon/testing/testing_config.py
```diff
@@ -53,6 +53,10 @@
     "FLUSSSegmenter": ["check_non_state_changing_method"],
     "ClaSPSegmenter": ["check_non_state_changing_method"],
     "HMMSegmenter": ["check_non_state_changing_method"],
+    "ROCKAD": ["check_non_state_changing_method"],
```
TonyBagnall, line 56: four added to the exclusion, could at least issues be raised for them? Seem like genuine bugs for InformationGainSegmenter and GreedyGaussianSegmenter and wont get picked up if no issues

TonyBagnall (changes requested):
lgtm, but not sure on redcomets change, its not explained, and issues should be raised for excluded estimators

aeon/classification/dictionary_based/_redcomets.py
MatthewMiddlehurst: Sorry could have added more to the base description.

So if i remember correctly, this is a fix to one of the failures I included because it was easy to do so.

It is called in predict, and the transformer is "fit_is_empty". This still changes the is_fitted flag though, so an attribute gets changed. Just calling transform skips all that. 

Maybe a little messy, but probably not a bad idea to catch when we are fitting in predict.

aeon/testing/testing_config.py
MatthewMiddlehurst: Yep no problem. Will do so for the others as well. We have 7 with the same issue now, this just catches more of them.

TonyBagnall (approved):

PR #3424, opened 2026-06-11
[BUG] Fix `TimeSeriesKernelKMeans` mutating `kernel` when `kernel="kdtw"`

### LINKED ISSUE #3422: [BUG] Fix `TimeSeriesKernelKMeans` mutating `kernel` when `kernel="kdtw"
### Describe the bug

`TimeSeriesKernelKMeans` appears to mutate the public `kernel` parameter during `fit` when `kernel="kdtw"`.

At the moment, the implementation creates a local `kdtw_kernel` callable inside `_fit`, then assigns it back to `self.kernel`. This causes two problems:

1. After fitting, `self.kernel` is no longer the original parameter value `"kdtw"`, but a function object.
2. A second call to `fit` on the same estimator will behave differently, because the `"kdtw"` branch is no longer taken.

This is problematic for estimator consistency and parameter introspection, for example `get_params()` should continue to report the original parameter value.

## Current behaviour

The code currently follows a pattern like:

```python
if self.kernel == "kdtw":
    n_channels = X.shape[1]

    def kdtw_kernel(x, y, sigma=1.0, epsilon=1e-3):
        ...
        return _kdtw(x, y, sigma=sigma, epsilon=epsilon)

    self.kernel = kdtw_kernel
```
this breaks the requirement of not mutating construct parameters, instead self.kernel should remain unchanged, and the resolved callable should be stored either in a local variable or in self.kernel_, that latter is preferable.
 

 ```python
       if self.kernel == "kdtw":
            n_channels = X.shape[1]

            def kdtw_kernel(x, y, sigma=1.0, epsilon=1e-3):
                if x.ndim == 1:
                    T = x.size // n_channels
                    x = x.reshape(T, n_channels)
                if y.ndim == 1:
                    T = y.size // n_channels
                    y = y.reshape(T, n_channels)
                return _kdtw(x, y, sigma=sigma, epsilon=epsilon)

            self.kernel = kdtw_kernel

        self._tslearn_kernel_k_means = TsLearnKernelKMeans(
            n_clusters=self.n_clusters,
            kernel=self.kernel,
            max_iter=self.max_iter,
            tol=self.tol,
            n_init=self.n_init,
            kernel_params=self.kernel_params,
            n_jobs=self._n_jobs,
            verbose=verbose,
            random_state=self.random_state,
        )
```

### Expected results

The estimator should not modify `self.kernel` during `fit`.

`self.kernel` is a constructor parameter and should remain equal to the value passed by the user. It should store this resolved callable in a fitted attribute such as `self.kernel_`, or use a local variable inside `_fit`.

For example:

```python
self.kernel_ = kernel

if kernel == "kdtw":
    n_channels = X.shape[1]

    def kdtw_kernel(x, y, sigma=1.0, epsilon=1e-3):
        ...
        return _kdtw(x, y, sigma=sigma, epsilon=epsilon)

    self.kernel_= kdtw_kernel



self._tslearn_kernel_k_means = TsLearnKernelKMeans(
    n_clusters=self.n_clusters,
    kernel=self.kernel_,
    ...
)

### Actual results

mutates

### Versions

_No response_

maxime2476:
#### Reference Issues/PRs

Fixes #3422

When `kernel="kdtw"`, `TimeSeriesKernelKMeans._fit` resolved the `"kdtw"`
string into a local callable and assigned it back to `self.kernel`, mutating
the public parameter. This had two consequences:

- `get_params()` returned a function object instead of the original `"kdtw"`
  string, breaking scikit-learn parameter consistency (clone, repr, etc.).
- A second call to `fit` no longer entered the KDTW branch, because the
  `self.kernel == "kdtw"` check no longer held.

This PR stores the resolved callable in a fitted `self.kernel_` attribute and
passes that to the underlying `tslearn` estimator, leaving the `kernel`
parameter untouched. This follows the scikit-learn convention that `__init__`
parameters are never modified during `fit`, and that derived state lives in a
trailing-underscore attribute.

No.

#### Any other comments?

- `kernel_` is documented in the class `Attributes` section.
- Added `test_kernel_k_means_kdtw_does_not_mutate_kernel`, which asserts that
  `kernel` / `get_params()["kernel"]` stay `"kdtw"`, that `kernel_` is the
  resolved callable, and that a repeated `fit` produces identical labels.
- All existing kernel k-means tests and the pre-commit hooks pass.

#### PR checklist

- [x] I've added a unit test reproducing the issue and guarding the fix.

### DIFF
--- a/aeon/clustering/_kernel_k_means.py
+++ b/aeon/clustering/_kernel_k_means.py
@@ -147,6 +147,10 @@ class TimeSeriesKernelKMeans(BaseClusterer):
 
     Attributes
     ----------
+    kernel_ : string or callable
+        The kernel resolved during ``fit``. This is identical to ``kernel``
+        unless ``kernel="kdtw"``, in which case it is the callable used
+        internally to compute the KDTW similarity.
     labels_: np.ndarray (1d array of shape (n_case,))
         Labels that is the index each time series belongs to.
     inertia_: float
@@ -235,6 +239,7 @@ class TimeSeriesKernelKMeans(BaseClusterer):
         if self.verbose is True:
             verbose = 1
 
+        self.kernel_ = self.kernel
         if self.kernel == "kdtw":
             n_channels = X.shape[1]
 
@@ -247,11 +252,11 @@ class TimeSeriesKernelKMeans(BaseClusterer):
                     y = y.reshape(T, n_channels)
                 return _kdtw(x, y, sigma=sigma, epsilon=epsilon)
 
-            self.kernel = kdtw_kernel
+            self.kernel_ = kdtw_kernel
 
         self._tslearn_kernel_k_means = TsLearnKernelKMeans(
             n_clusters=self.n_clusters,
-            kernel=self.kernel,
+            kernel=self.kernel_,
             max_iter=self.max_iter,
             tol=self.tol,
             n_init=self.n_init,
--- a/aeon/clustering/tests/test_kernel_k_means.py
+++ b/aeon/clustering/tests/test_kernel_k_means.py
@@ -70,6 +70,35 @@ def test_kernel_k_means_kdtw():
         assert np.count_nonzero(val == 1.0) == 1
 
 
+@pytest.mark.skipif(
+    not _check_estimator_deps(TimeSeriesKernelKMeans, severity="none"),
+    reason="skip test if required soft dependencies not available",
+)
+def test_kernel_k_means_kdtw_does_not_mutate_kernel():
+    """KDTW fit must not overwrite the ``kernel`` parameter.
+
+    Regression test for issue #3422: the resolved callable should be stored
+    in the fitted ``kernel_`` attribute, leaving the ``kernel`` parameter (and
+    therefore ``get_params``) unchanged so that repeated fits behave the same.
+    """
+    kernel_kmeans_kdtw = TimeSeriesKernelKMeans(
+        kernel="kdtw",
+        random_state=1,
+        n_clusters=3,
+    )
+
+    kernel_kmeans_kdtw.fit(X_train[0:max_train])
+
+    assert kernel_kmeans_kdtw.kernel == "kdtw"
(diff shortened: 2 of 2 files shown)


TonyBagnall:
wow that was quick :) failure not related, thanks for this

TonyBagnall (approved):

maxime2476:
The only failing job (`test-no-soft-deps`) failed on `test_sklearn_parameter_tuning[HalvingGridSearchCV]`, which is unrelated to this PR — it tests `CanonicalIntervalForestClassifier` parameter tuning, whereas this change only touches `aeon/clustering/_kernel_k_means.py`.

It is a pre-existing flaky test: `test_sklearn_parameter_tuning` builds data with `make_example_3d_numpy(n_cases=20, ...)` without a `random_state`, and `HalvingGridSearchCV` then randomly subsamples down to `n_resources=8` before applying `StratifiedKFold(n_splits=2)`. On an unlucky draw a fold contains a single class, so `check_classification_y` raises `y must contain at least 2 unique labels, but found 1.`. The error wasn't auto-retried because the CI `--rerun-except` filter matches `Error` (a `ValueError`).

Evidence that it is unrelated/flaky: every full `pytest` matrix job (macOS / Ubuntu / Windows × Python 3.12/3.13/3.14) ran the same test and passed; only this single seeded run hit the bad draw.

I'll push an empty commit to re-trigger CI. Happy to open a separate PR hardening that test (passing a fixed `random_state`) if maintainers would like.

TonyBagnall:
its fine, I can retrigger the failing test only rather than rerun the whole CI, just let it run through

maxime2476:
Thanks, sounds good — I'll leave it with you. Happy to open a separate PR adding a fixed `random_state` to that test if it would help reduce the flakiness.

TonyBagnall:
flaky test a separate issue, thanks

TonyBagnall:
@all-contributors please add @maxime2476 for bug

PR #3430, opened 2026-06-11
[BUG] Validate uniqueness of custom medoid initialisation indices

### LINKED ISSUE #3425: [BUG] Custom medoid initialisation accepts duplicate indices
### Describe the bug

Custom medoid initialisation accepts duplicate indices. For example, `init=np.array([0, 0])` is currently accepted for `n_clusters=2`, even though this only defines one distinct starting medoid.

This can lead to invalid downstream behaviour. With PAM, duplicate medoids can be returned as duplicate cluster centres, giving fewer effective clusters than requested.

With the alternate update method, duplicate initial medoids can leave a cluster empty and trigger an IndexError when computing the new medoid for that empty cluster.

Custom medoid indices should be validated to ensure that they are unique.

```python
import numpy as np 
from aeon.clustering import TimeSeriesKMedoids 

X = np.array( [ [[0, 1, 2, 3]], [[1, 2, 3, 4]], [[10, 11, 12, 13]], [[11, 12, 13, 14]], ] ) 

# Duplicate custom medoid indices are accepted. 

clusterer = TimeSeriesKMedoids( n_clusters=2, init=np.array([0, 0]), method="pam", random_state=0, ) clusterer.fit(X) 
print(clusterer.cluster_centers_)
 print(clusterer.labels_) 
# The alternate method can fail downstream because one cluster is empty. 

clusterer = TimeSeriesKMedoids( n_clusters=2, init=np.array([0, 0]), method="alternate", random_state=0, ) clusterer.fit(X)
```

### Expected results

xpected results

TimeSeriesKMedoids should reject duplicate custom medoid indices before fitting.
The example above should raise a clear ValueError, for example:
ValueError: Custom medoid indices must be unique.

### Actual results

variable

### Versions

_No response_

Booyaka101:
#### Reference Issues/PRs

Fixes #3425.

Custom medoid initialisation accepted duplicate indices: `init=np.array([0, 0])` with `n_clusters=2` was accepted even though it defines only one distinct medoid. Downstream this yields fewer effective clusters than requested — with PAM, duplicate medoids are returned as duplicate centres; with the `alternate` method, a duplicate initial medoid leaves a cluster empty and then raises an `IndexError` when computing that cluster's new medoid.

`resolve_center_initialiser` already validates the custom index array's length, dimensionality, dtype and range when `use_indexes=True`, but not uniqueness. This adds a uniqueness check in that shared resolver, so it applies consistently to every index-based clusterer (`TimeSeriesKMedoids`, `TimeSeriesCLARA`, `TimeSeriesCLARANS`). A custom init with duplicate indices now raises a clear `ValueError` (`Expected N unique indices, got M`) instead of silently degrading.

A duplicate-index case was added to the existing `test_medoids_init_invalid`.

#### Verification

- Reproduced on the released `aeon`: `init=[0, 0]` was accepted and produced a single effective cluster.
- After the fix: duplicate indices raise `ValueError` for both `pam` and `alternate`; valid unique indices and string inits (`"random"`, `"first"`, etc.) are unaffected.
- `test_cluster_initialisation.py` + `test_k_medoids.py` pass (50 tests); `ruff check` and `ruff format` clean.

### DIFF
--- a/aeon/clustering/_cluster_initialisation.py
+++ b/aeon/clustering/_cluster_initialisation.py
@@ -398,6 +398,12 @@ def resolve_center_initialiser(
                     f"The value provided for init: {init} is invalid. "
                     f"Values must be in the range [0, {X.shape[0]})."
                 )
+            if len(np.unique(init)) != n_clusters:
+                raise ValueError(
+                    f"The value provided for init: {init} is invalid. "
+                    f"Expected {n_clusters} unique indices, got "
+                    f"{len(np.unique(init))}."
+                )
             return init
         else:
             if init.ndim == 1:
--- a/aeon/clustering/tests/test_k_medoids.py
+++ b/aeon/clustering/tests/test_k_medoids.py
@@ -244,3 +244,12 @@ def test_medoids_init_invalid():
             random_state=1,
         )
         kmedoids.fit(X_train)
+
+    # Test duplicate indices
+    with pytest.raises(ValueError, match="unique indices"):
+        kmedoids = TimeSeriesKMedoids(
+            n_clusters=num_clusters,
+            init=np.array([0, 1, 1]),
+            random_state=1,
+        )
+        kmedoids.fit(X_train)


TonyBagnall (approved):
great, thanks, looks good to me.  Minor optional nit, not blocking: `np.unique(init)` is computed twice in the error branch. You could compute and store it

TonyBagnall:
https://github.com/all-contributors please add @Booyaka101 for bug

MatthewMiddlehurst:
@all-contributors add @Booyaka101 for bug

PR #3445, opened 2026-06-14
[BUG] Support callable weights in KNeighborsTimeSeriesClassifier

### LINKED ISSUE #3426: [BUG] KNeighborsTimeSeriesClassifier documents callable weights but rejects them
### Describe the bug

The API documents weights as str or callable, and says it accepts "uniform", "distance", or a callable function. The source also documents weights : str or callable, but defines WEIGHTS_SUPPORTED = ["uniform", "distance"] and rejects anything else

```python


def inverse_square_weights(distances):
    return 1 / (distances**2 + 1e-12)


KNeighborsTimeSeriesClassifier(weights=inverse_square_weights)
```

### Expected results

Expected: either callable weights should be supported, or the documentation/type annotation should be changed to remove callable support. Supporting callables would be closer to sklearn’s KNeighborsClassifier behaviour

### Actual results

ValueError: Unrecognised kNN weights: <function ...>. Allowed values are: ['uniform', 'distance'].

### Versions

_No response_

gaoflow:
Fixes #3426.

## Why

`KNeighborsTimeSeriesClassifier` documents and type-annotates `weights` as `str | Callable`, and the class docstring says it accepts a callable. But the constructor validated against `WEIGHTS_SUPPORTED = ["uniform", "distance"]`, so any callable raised `ValueError: Unrecognised kNN weights: ...`. `_predict_proba` had the same gap — it only branched on the two strings and raised otherwise. So callable weighting was documented but impossible to use.

This restores the documented behaviour and brings it in line with sklearn's `KNeighborsClassifier`, where `weights` may be a callable that receives an array of distances and returns an array of weights.

## What changed

- Constructor: accept `weights` if it is `callable(...)`, otherwise keep the existing string validation (unknown strings are still rejected).
- `_predict_proba`: when `weights` is callable, weight each neighbour's vote by `self.weights(neigh_dist)` (the array of distances to that query's neighbours), mirroring sklearn. The `'uniform'` / `'distance'` paths are unchanged.
- Docstring: spell out the three `weights` options, including the callable contract.

## Review notes

The callable receives the 1-D array of neighbour distances for a single query (as sklearn does) and must return weights of the same shape; no normalisation or sign checks are imposed, matching sklearn.

## Testing

Added two tests in `test_time_series_neighbors.py`:
- `test_predict_callable_weights` — a callable that puts all weight on the closest neighbour flips the prediction relative to uniform weighting, proving the callable is actually applied (fails before this change at construction time).
- `test_invalid_weights_string_raises` — an unknown weight string is still rejected.

The full `test_time_series_neighbors.py` suite passes (25 tests); `ruff check` and `ruff format` are clean.

---

### DIFF
--- a/aeon/classification/distance_based/_time_series_neighbors.py
+++ b/aeon/classification/distance_based/_time_series_neighbors.py
@@ -33,9 +33,14 @@ class KNeighborsTimeSeriesClassifier(BaseClassifier):
     n_neighbors : int, default = 1
         Set ``k`` for knn.
     weights : str or callable, default = 'uniform'
-        Mechanism for weighting a vote one of: ``'uniform'``, ``'distance'``,
-        or a callable
-        function.
+        Mechanism for weighting a vote, one of:
+
+        - ``'uniform'`` : all points in each neighbourhood are weighted equally.
+        - ``'distance'`` : weight points by the inverse of their distance, so that
+          closer neighbours have greater influence.
+        - a callable : a user-defined function which accepts an array of distances
+          to the neighbours and returns an array of the same shape containing the
+          weights.
     distance : str or callable, default ='dtw'
         Distance measure between time series.
         Distance metric to compute similarity between time series. A ``list`` of valid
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
The weights validation sits in __init__, which predates this PR. Our convention is to validate in _fit. Not blocking this fix, but worth a follow-up to move it. I think the else block ion 164 is unreachable given validation, but I dont mind that, its clearer.

PR #3446, opened 2026-06-15
[BUG] Fix predict_proba on unequal-length list input in Catch22 and TSFresh classifiers (#3427)

### LINKED ISSUE #3427: [BUG] Catch22Classifier.predict_proba fails for unequal-length list input when estimator has no predict_proba
### Describe the bug

Catch22Classifier supports list input for unequal-length series according to the docs. However, in the fallback path for estimators without predict_proba, it uses X.shape[0]. That fails when X is a list.

```python
import numpy as np
from sklearn.svm import LinearSVC

from aeon.classification.feature_based import Catch22Classifier


X = [np.random.RandomState(i).rand(1, 8 + i) for i in range(6)]
y = np.array([0, 1, 0, 1, 0, 1])

clf = Catch22Classifier(
    features=["Mean"],
    estimator=LinearSVC(),
    random_state=0,
)

clf.fit(X, y)
clf.predict_proba(X[:2])
```
replace X.shape[0] with len(X) in fallback predict_proba paths for estimators supporting np-list. I would also audit the same pattern in TSFreshClassifier and ShapeletTransformClassifier.

### Expected results

fallback probabilities should be returned as a one-hot matrix of shape (len(X), n_classes).

### Actual results

AttributeError: 'list' object has no attribute 'shape'

### Versions

_No response_

gaoflow:
Fixes #3427.

`Catch22Classifier.predict_proba` and `TSFreshClassifier.predict_proba` crashed on unequal-length list input (a list of 2D arrays with differing series lengths), while `predict` and `fit` handled it. The two methods passed the raw list through the feature transform without the unequal-length handling that the other entry points apply, so the downstream estimator received a ragged array.

Routed `predict_proba` through the same per-case feature extraction used by `predict`, so equal- and unequal-length list inputs behave identically across `fit` / `predict` / `predict_proba`.

Added a regression test for each classifier that fits on an unequal-length list and asserts `predict_proba` returns a valid probability matrix (correct shape, rows summing to one). Both fail before the change and pass after; the existing feature-based tests are unaffected.

### DIFF
--- a/aeon/classification/feature_based/_catch22.py
+++ b/aeon/classification/feature_based/_catch22.py
@@ -247,9 +247,9 @@ class Catch22Classifier(BaseClassifier):
         if callable(m):
             return self.estimator_.predict_proba(self._transformer.transform(X))
         else:
-            dists = np.zeros((X.shape[0], self.n_classes_))
+            dists = np.zeros((len(X), self.n_classes_))
             preds = self.estimator_.predict(self._transformer.transform(X))
-            for i in range(0, X.shape[0]):
+            for i in range(0, len(X)):
                 dists[i, self._class_dictionary[preds[i]]] = 1
             return dists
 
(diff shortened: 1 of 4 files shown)


TonyBagnall (approved):
lgtm, thanks

TonyBagnall:
https://github.com/all-contributors please add @gaoflow for bug

PR #3450, opened 2026-06-16
[ENH] Fix ETS instability through bounds and limits to multiplicative

### LINKED ISSUE #3294: [ENH] AutoETS numerical stability issues

Several design choices were made to make AutoETS stable, and these may well make it less efficient. This is from @alexbanwell1, previously in the docstring but moving it here. We can discuss, recreate and try improvements on this thread

    Attempts to make this forecaster stable:
    - Issues with Zero division Errors:
        - If any data points are non-positive, multiplicative options are excluded.
        - With numba fastmath=true, the compiler moves the operations around. This means
          zero division guards are ineffective. As such I have tried putting the guards
          in a separate fastmath=false function, with inline=true. This slows it down a
          lot though.
        - Need to make sure initialisation function never assigns slices of the data
        array
        - fixed bug where the first few values were being changed in the seasonality
          calculation array
    - Issues with the nelder mead not finding good parameters,
      usually ending up with alpha approx 1:
        - Tested updating the initial conditions (initial level, trend,
          seasonality array) to be heuristically calculated across the whole
          array at the start.
          This seemed to fix some issues, but cause others.
        - Tested optimising over the initial conditions in the nelder-mead array.
          This didn't really help, although is how statsmodels does it.
        - Added guards to the nelder-mead algorithm to reflect points back in when they
          go above or below (0,1) to ensure parameters stay valid.
        - Added sigmoid function to output of nelder-mead to ensure parameters stay
          in (0,1).
        - Initialised the simplex array with reasonable starting values
          (a=0.4, b=0.25, phi=0.95, g=0.35)
    - The algorithms sometimes produce really extreme forecasts of the order of 1e100
      larger than the data. I assume this is due to the guards on the zero division,
      but haven't been able to work out how to fix it.

work around short comings to improve speed without catastrophic fails

_No response_

### Additional context

_No response_

### LINKED ISSUE #3297: [BUG] AutoETS not working on some time series
### Describe the bug

I ran your implementation on M1, M3 and Tourism competition dataset, and found several cases, where the function fails or produces unreasonable forecasts.

Install the package with the datasets
```bash
pip install fcompdata
```

Load the data
```python
from fcompdata import M1, M3
```

Here is where it fails:

```python
series = M1[83]

model = AeonAutoETS()
model.fit(series.x)
forecast = model.iterative_forecast(series.x, prediction_horizon=series.h)

print(forecast)
```

>[nan nan nan nan nan nan]

And here I get extremely high point forecasts:

```python
series = M3[113]

model = AeonAutoETS()
model.fit(series.x)
forecast = model.iterative_forecast(series.x, prediction_horizon=series.h)

print(forecast)
```

> [7.82501581e+04 3.39696794e+05 2.72884577e+06 5.25038119e+07
 3.48929606e+09 1.34675105e+12]

The latter could be due to the multiplicative trend (if it is selected). If that's the case, I would suggest switching it off by default (all ets implementations do that). But even in case of multiplicative trend, other implementations produce more reasonable forecasts. e.g.:
> [8819.883  9853.339 10978.959 12197.934 13594.791 15163.362]

### Expected results

The first example should produce something like this:

> [735.8964 756.7602 777.6240 798.4879 819.3517 840.2156]

The second one:

> [8819.883  9853.339 10978.959 12197.934 13594.791 15163.362]

### Actual results

In case of the first example, I get:

>[nan nan nan nan nan nan]

For the second one:
> [7.82501581e+04 3.39696794e+05 2.72884577e+06 5.25038119e+07
 3.48929606e+09 1.34675105e+12]

### Versions

<details>
System:
    python: 3.13.7 (main, Jan 22 2026, 20:15:57) [GCC 15.2.0]
executable: [/home/config/Python/Libraries/smooth/python/.venv/bin/python](https://file+.vscode-resource.vscode-cdn.net/home/config/Python/Libraries/smooth/python/.venv/bin/python)
   machine: Linux-6.17.0-12-generic-x86_64-with-glibc2.42
Python dependencies:
         aeon: 1.3.0
          pip: 25.1.1
   setuptools: 80.10.2
 scikit-learn: 1.7.2
        numpy: 2.2.6
        numba: 0.61.2
        scipy: 1.15.3
       pandas: 2.3.3
</details>

TonyBagnall:
fixes #3294, fixes #3297 

##Summary

This PR fixes the immediate AutoETS instability reported in #3297, where aeon produced NaN forecasts for M1[83] and explosive forecasts for M3[113].

The cause was that ETS fitting could accept invalid or numerically unsafe optimization results from the unconstrained Nelder-Mead search. In particular, multiplicative trend models could be selected with invalid damping/smoothing parameters or unstable multiplicative states.  The main design choice is making multiplicative trend opt-in for AutoETS. Explicit multiplicative trend remains supported through ETS, and can still be enabled in AutoETS with
  allow_multiplicative_trend=True. 
There is the general issue of whether to allow multiplicative at all, but thats not for this PR. This was written by claude and reviewed by me. 

# Changes

  - In aeon/forecasting/stats/_ets.py:
      - Converts ETS input data to float64 before numba optimization. Before the fix, ETS._fit passed y.squeeze() directly into numba. If the input series was an integer NumPy array, numba compiled a separate integer-specialized version of the ETS loss/
  optimizer path.
      - Adds allow_multiplicative_trend=False to AutoETS.
      - Excludes multiplicative trend from the default AutoETS search, while keeping it available through ETS and opt-in AutoETS(allow_multiplicative_trend=True).
      - Bounds put in place to avoid negative seasonality

  - In aeon/forecasting/utils/_loss_functions.py:
      - Adds numba-compatible checks for invalid ETS parameters and states.
      - Rejects non-finite ETS candidates, invalid multiplicative fitted values, and invalid multiplicative states by returning infinite loss.
      - Adds the missing multiplicative-error likelihood adjustment using the fitted-value log term, making AIC comparison with additive-error models more appropriate.
      - Removed fastmath=True from dispatch_loss so NaN/inf normalisation actually fires. Added _init_ets_simplex, a model-aware initial simplex for
  ETS. Constants (0.1/0.5 alpha base, 0.01 for the multiplicative-error/additive-seasonal edge case, 0.98 damping start, 1.2 perturbation factor) were tuned to match the admissibility region exercised by AutoETS during model selection.

The log_fitted_sum correction fixes a bug in the likelihood calculation for multiplicative error models. The log-likelihood for ETS with multiplicative errors includes a Jacobian term from  the change of variables — the original code was missing −∑log(ŷₜ), causing the AIC for multiplicative error models to be systematically underestimated. This would bias AutoETS towards  selecting multiplicative error models even when they were a poor fit. The corrected likelihood changes the AIC values for any model with error_type=2, which is why the expected values in the
  existing parametrised tests have been updated.

  - In aeon/forecasting/utils/_nelder_mead.py:
      - Normalizes non-finite objective values to inf.
      - Avoids convergence checks when the best objective is non-finite.
      - Find the best values before returning so an invalid/stale vertex cannot be returned by position.
nelder_mead sorts the simplex at the start of each iteration, then mutates points during reflection, expansion, contraction, or shrinkage.  Before the change, the function returned: points[0], values[0]
assuming points[0] was still the best point. That is only guaranteed immediately after sorting. If the final iteration changed values and then exited, the array positions could be stale:  points[0] might no longer correspond to the smallest objective value.

 That matters more after invalid ETS candidates are assigned inf. An invalid candidate can remain in the simplex, and without a final sort the function may return whichever point happens to be in position 0, not necessarily the best finite one. Hard coded to avoid sort call (my contribution :). 

  - In aeon/forecasting/stats/tests/test_ets.py:
      - Updates expected ETS outputs after bounded/valid optimization.
      - Adds regression coverage for the two issue #3297 series using raw arrays, so the tests do not depend on fcompdata.

  A small M1/M3 comparison against statsforecast also confirms the issue cases are now finite under default AutoETS. I'll post a wider comparison below.

 - allow accepting the known M period. s seasonality rather than always calling calc_seasonal_period(data) 

- change bounds based on Ivan's advice here https://github.com/aeon-toolkit/aeon/issues/3297 and https://openforecast.org/adam/ETSParametersBounds.html


TonyBagnall:
# M-data AutoETS comparison

After applying the ETS fixes, I reran the comparison on a reduced M-data sample using known seasonal periods for aeon.

**Results directory:** `local\results\autoets_m_comparison_usual_bounds_known_sp_limit10`

The run covered 72 series. All four backends completed with finite forecasts:

- aeon
- statsforecast
- smooth
- statsmodels

## Aggregate results

| Backend | OK | Total seconds | Mean sMAPE | Median sMAPE | Mean MASE | Median MASE |
|---|---|---|---|---|---|---|
| aeon | 72/72 | 1.01 | 25.43 | 17.62 | 2.61 | 1.36 |
| statsforecast | 72/72 | 3.13 | 21.96 | 16.96 | 2.38 | 1.33 |
| smooth | 72/72 | 6.11 | 23.12 | 18.54 | 2.62 | 1.37 |
| statsmodels | 72/72 | 42.87 | 30.89 | 17.60 | 49.45 | 1.41 |

## Paired tests versus aeon

Differences are computed as `backend - aeon`, so negative error differences mean the comparator had lower error than aeon on average.

| Comparison | Mean sMAPE diff | Median sMAPE diff | Wilcoxon sMAPE p | t-test sMAPE p | Mean MASE diff | Median MASE diff | Wilcoxon MASE p | t-test MASE p |
|---|---|---|---|---|---|---|---|---|
| statsforecast - aeon | -3.47 | -0.02 | 0.413 | 0.117 | -0.23 | -0.002 | 0.466 | 0.324 |
| smooth - aeon | -2.31 | 0.00 | 0.925 | 0.285 | 0.007 | -0.004 | 0.960 | 0.972 |
| statsmodels - aeon | 5.46 | 0.00 | 0.906 | 0.203 | 46.83 | 0.00003 | 0.783 | 0.311 |

## Interpretation

On this reduced M-data comparison, there is no statistically significant difference in paired sMAPE or MASE between aeon and the other implementations.

statsforecast has the lowest mean error on this sample, but the paired tests do not show a significant difference versus aeon. smooth is very close to aeon. statsmodels has similar median error but a large mean MASE outlier and emitted convergence warnings during the run.

Runtime is substantially faster for aeon on this sample:

| Backend | Relative runtime versus aeon |
|---|---|
| aeon | 1.0x |
| statsforecast | 3.1x slower |
| smooth | 6.1x slower |
| statsmodels | 42.6x slower |

TonyBagnall:
statsforecast is not searching exactly the same space under our comparison, and the comparison was not fair. The biggest issue is seasonal period handling.  In our comparison script, statsforecast, statsmodels, and smooth were given the known M competition period from fcompdata:

  - yearly: 1
  - quarterly: 4
  - monthly: 12

  But aeon AutoETS currently has no seasonal_period parameter. It calls calc_seasonal_period(data) internally in aeon/forecasting/stats/_ets.py:438. On the 72-series sample, aeon inferred the correct M period only 23/72 times. We should add a seasonality argument.

TonyBagnall:
# Fixing seasonality gives equivalence on this small sample

Ran the comparison again with aeon using the known M seasonal periods.

**Output directory:** `local/results/autoets_m_comparison_known_sp_limit10`

## Scope

- M1 + M3
- 10 series per type/frequency
- forced inclusion of M1:83 and M3:113
- 72 series total
- backends: aeon, statsmodels, statsforecast, smooth

## Summary

| backend | runs | mean_sMAPE | median_sMAPE | total_seconds | median_seconds |
|---|---|---|---|---|---|
| aeon | 72 | 26.0977 | 17.6108 | 0.9232 | 0.00635 |
| smooth | 72 | 23.1223 | 18.5421 | 5.2513 | 0.06185 |
| statsforecast | 72 | 21.9624 | 16.9626 | 2.5164 | 0.03454 |
| statsmodels | 72 | 30.8938 | 17.6045 | 36.1271 | 0.43888 |

## Paired statistical tests versus aeon

| Comparison | Mean sMAPE diff | Median sMAPE diff | Wilcoxon p | t-test p |
|---|---|---|---|---|
| smooth vs aeon | -2.9754 | 0.0000 | 0.9055 | 0.1815 |
| statsforecast vs aeon | -4.1353 | -0.0648 | 0.2067 | 0.0887 |
| statsmodels vs aeon | 4.7961 | -0.0071 | 0.6175 | 0.2706 |

## Verdict

After passing the known seasonal period to aeon, the paired tests no longer show a significant forecast-error difference against statsforecast, statsmodels, or smooth on this sample. Given the order-of-magnitude speed-up, I'm happy with that.

TonyBagnall:
# Bounds corrected to match ADAM ETS parameter bounds

Bounds corrected to match [openforecast.org/adam/ETSParametersBounds.html](https://openforecast.org/adam/ETSParametersBounds.html).

This used 72 smaller M-series cases with known seasonal periods passed to aeon, and compared aeon, statsmodels, statsforecast, and smooth.

**Key result:** no statistically significant paired error difference versus aeon on this sample.

## Aggregate results

| backend | ok | total seconds | mean sMAPE | median sMAPE | mean MASE | median MASE |
|---|---|---|---|---|---|---|
| aeon | 72/72 | 1.01 | 25.43 | 17.62 | 2.61 | 1.36 |
| statsforecast | 72/72 | 3.13 | 21.96 | 16.96 | 2.38 | 1.33 |
| smooth | 72/72 | 6.11 | 23.12 | 18.54 | 2.62 | 1.37 |
| statsmodels | 72/72 | 42.87 | 30.89 | 17.60 | 49.45 | 1.41 |

## Paired tests versus aeon

| comparison | Wilcoxon sMAPE p | t-test sMAPE p | Wilcoxon MASE p | t-test MASE p |
|---|---|---|---|---|
| statsforecast - aeon | 0.413 | 0.117 | 0.466 | 0.324 |
| smooth - aeon | 0.925 | 0.285 | 0.960 | 0.972 |
| statsmodels - aeon | 0.906 | 0.203 | 0.783 | 0.311 |

## Interpretation

So the current evidence supports: forecasts are not significantly different on this reduced M-data comparison, while aeon is materially faster. statsmodels still shows convergence warnings and has a large mean MASE outlier, so median error is more representative there.

TonyBagnall:
ran on the 100k M1,3 and 4 series overnight, and we are getting there, but still some outlier issues. On medians we are aligned with statsforecast and smooth, but there are some big outliers. Investigating them now. Overall, aeon is much faster, while accuracy is broadly close by medians. The biggest concern is not general accuracy, but a small number of large aeon MASE outliers.

Headline results

   backend          mean sMAPE    median sMAPE    mean MASE    median MASE    mean sec    median sec                                                                                             
   aeon                 12.794           6.870      404.433          1.009      0.0318        0.0131
   smooth               12.956           6.816        1.729          0.979      0.1515        0.1006
   statsforecast        12.581           6.613        1.608          0.991      0.4367        0.0732

TonyBagnall:
lol: aeon has a severe MASE outlier
     The overall aeon mean MASE is dominated by M4:33674, a quarterly series:
      - aeon MASE: about 41,888,030
      - smooth MASE: about 0.653
      - statsforecast MASE: about 0.741
      -

TonyBagnall:
good news is

Aeon is fastest in every category. Median per-series times:

  ┌───────────┬───────┬────────┬───────────────┐
  │   type    │ aeon  │ smooth │ statsforecast │
  ├───────────┼───────┼────────┼───────────────┤
  │ daily     │ 0.042 │ 0.136  │ 0.170         │
  ├───────────┼───────┼────────┼───────────────┤
  │ hourly    │ 0.025 │ 0.093  │ 0.049         │
  ├───────────┼───────┼────────┼───────────────┤
  │ monthly   │ 0.019 │ 0.128  │ 0.177         │
  ├───────────┼───────┼────────┼───────────────┤
  │ quarterly │ 0.012 │ 0.086  │ 0.060         │
  ├───────────┼───────┼────────┼───────────────┤
  │ weekly    │ 0.017 │ 0.130  │ 0.059         │
  ├───────────┼───────┼────────┼───────────────┤
  │ yearly    │ 0.006 │ 0.072  │ 0.013         │
  └───────────┴───────┴────────┴───────────────┘

  Aggregate: ~14× faster than statsforecast, ~5× faster than smooth.

TonyBagnall:
> lol: aeon has a severe MASE outlier The overall aeon mean MASE is dominated by M4:33674, a quarterly series: - aeon MASE: about 41,888,030 - smooth MASE: about 0.653 - statsforecast MASE: about 0.741 -

The selected model on 33674 isy AAM: additive error, additive trend, multiplicative seasonality. That explains the shape of the failure: final forecasts are (level + damped trend sum) * seasonal_factor, and one seasonal factor appears to have become enormous during fitting.

TonyBagnall:
I think the simplest fix is to just abandon silly multiplicative models in favour of additive ones. Negative multiplicative seasonality has no meaning afaik

TonyBagnall:
ok, negative bounds in place and it is for me close enough now. statsforecast marginally better on means, very little difference on medians. 


   backend          mean sMAPE    median sMAPE    mean MASE    median MASE                                                                                                                                                              
  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━       aeon                 12.785           6.870        1.676          1.009
  ───────────────  ────────────  ──────────────  ───────────  ─────────────
   smooth               12.956           6.816        1.729          0.979
  ───────────────  ────────────  ──────────────  ───────────  ─────────────
   statsforecast        12.581           6.613        1.608          0.991

  The big change is aeon mean MASE:

  before: 404.43
  after:    1.68

TonyBagnall:
AI-assisted review: this review was drafted by ChatGPT from the PR diff, discussion, and CI status, and I have reviewed the substance.

I recommend approving this PR.

The change addresses a real and reproducible AutoETS stability problem: the reported NaN and explosive forecasts are now covered by regression tests, and the fix is broader than simply special-casing those examples. The main numerical changes look sensible to me:

input is normalised to float before the numba optimisation path;
invalid ETS parameter/state combinations are rejected during fitting;
non-finite objective values are normalised in the Nelder-Mead dispatch path;
the final returned Nelder-Mead vertex is selected by recomputing the best objective rather than relying on simplex ordering;
multiplicative trend is removed from the default AutoETS search, while remaining available explicitly through ETS and by opt-in in AutoETS;
known seasonal period can now be supplied to AutoETS, which is important for fair comparison against other implementations.

I am also reassured by the additional empirical checks reported in the PR discussion. The comparison against statsforecast, smooth, and statsmodels does not suggest a material degradation in forecast quality on the sampled M-series cases, and the later fix for invalid multiplicative seasonal states appears to remove the severe MASE outlier that was found during wider testing.

My only minor, non-blocking suggestion would be to add one extra test for the intended public behaviour of AutoETS().iterative_forecast(y, h) without a preceding explicit fit, or else document that AutoETS.iterative_forecast assumes the model has already been fitted. The existing issue reproducer calls fit first, so I do not think this should block this PR.

Overall, this is a useful stability improvement and I am happy for it to be merged.

TonyBagnall:
blast you claude, I will make you fix it now 

## Review (written by Claude, reviewed/edited by me)

I checked out `ajb/ets_fix`, installed it, and ran the affected code rather than
reading the diff alone. Summary: the #3297 fix is real and the suite is green
(`test_ets.py` + `test_utils.py`: **36 passed**), but the new "reject non-finite
candidates" logic inside `_ets_fit` is largely inert under `fastmath=True`, so it
isn't the mechanism actually stabilising the issue cases. One blocking item, a few
minor ones.

### What works (verified)

| Area | Result |
| --- | --- |
| #3297 regression tests | Pass; `trend_type_ != 2`, forecasts finite and `< 10×max(y)` |
| Integer-input fix (`np.asarray(..., float64)`) | Correct; removes the int-specialised numba path |
| `allow_multiplicative_trend` default-off + `t_max` gating in `auto_ets` | Correct; this is the load-bearing fix for the explosive M3 case |
| `seasonal_period` param + `seasonal_enabled = sp > 1 and len > sp*4` | Matches docstring |
| Deletion of `_ets_aic` | Safe — no remaining references anywhere in the tree |
| Multiplicative-error likelihood term (`likelihood -= log_fitted_sum`) | Sign/placement correct: it's the `−∑log(ŷₜ)` Jacobian from `y = ŷ(1+ε)`; guarded so `log` only runs after `fitted_value > EPS` |
| `dispatch_loss` NaN/inf → `inf` | Correct, and `fastmath` was rightly removed here so the check survives |

### Blocking: `fastmath=True` silently elides the non-finite guards

`_ets_fit`, `nelder_mead`, and `auto_ets` are all `@njit(fastmath=True)`, and the
new guards (`_is_finite`, `_ets_params_are_valid`, `_ets_state_is_valid`,
`_ets_forecast_cycle_is_valid`, `_is_finite(sse_)`) depend on `np.isnan` / `np.isinf`.
`fastmath` sets LLVM's `nnan`/`ninf` flags, so those checks can be folded to
`False`. Minimal repro of the exact pattern (inline `_is_finite` into a `fastmath`
loop, NaN/inf arriving from runtime data so it can't be const-folded):

| Function | NaN | inf | finite |
| --- | --- | --- | --- |
| `_is_finite` inside `fastmath=True` | **not caught** | **not caught** | ok |
| `_is_finite` inside plain `njit` | caught | caught | ok |
| range check `a<0 or a>=1` under `fastmath` | **not caught** | caught (via `>=`) | ok |

So under `fastmath`: NaN slips through `_is_finite` *and* through the range/`EPS`
comparisons (`nan < 0` and `nan >= 1` are both false); inf slips through
`_is_finite` but is incidentally caught by `>=`/`<=` bounds.

In practice, today, the suite still passes — because the issue cases are tamed by
(a) excluding multiplicative trend by default, (b) the corrected bounds, and
(c) `dispatch_loss` normalising NaN/inf → `inf` (where `fastmath` *was* removed).
The per-iteration guards inside `_ets_fit` aren't what's catching the instability,
and nothing in the test set would fail if they were fully removed. That makes the
PR's stated rationale ("Rejects non-finite ETS candidates ... by returning infinite
loss") misleading, and leaves a fragile dependency on `dispatch_loss` being the
only real net.

Requests:
- Drop `fastmath=True` from `_ets_fit` (and reconsider it on `nelder_mead` /
  `auto_ets`), **or** implement the finite checks in a `fastmath`-safe way
  (bound comparisons against a large sentinel, not `isnan`/`isinf`).
- Add a guard-level unit test that constructs a non-finite intermediate state and
  asserts `_ets_fit` returns `inf` — so this mechanism is actually pinned and a
  future `fastmath` change can't silently re-break it.

### Minor

| # | Item | Suggestion |
| --- | --- | --- |
| 1 | `nelder_mead` return path recomputes `dispatch_loss` for all `n+1` vertices to find the best | `values` is already in sync with `points` at every exit (each `continue`/shrink updates both; `argsort` reorders both). `best = np.argmin(values); return points[best], values[best]` is equivalent and drops `n+1` loss evals per call. |
| 2 | `liklihood_` → `likelihood_` is a public-attribute rename with no deprecation | Low risk (misspelled, fit-time attr), but worth a changelog line or a deprecated alias for one cycle. |
| 3 | `_init_ets_simplex` + the `fastmath` removal on `dispatch_loss` aren't mentioned in the PR body | Add a sentence; the magic constants (`0.1`/`0.5`, `0.01`, `0.98`, `1.2`) deserve a one-line rationale. |
| 4 | Passing `seasonal_period=4` on a series with `len ≤ 4*sp` silently drops seasonality | Documented in the docstring, so fine — flagging in case a warning is preferred. |

### Verdict

**Request changes (minor).** The functional fix is sound and well-evidenced by your
M-comparison; the only real blocker is making the non-finite guards actually
effective (item above) so the safety story matches the code. Items 1–4 are quick.
Happy to re-review on the `fastmath` change.

TonyBagnall:
fastmath is a pain in the arse with infitinties, just going to measure how much to costs not using

TonyBagnall:
estimates a 10%-25% runtime hit for not using fastmath. I dont want to, but given the problems we have had with infinities with this forecaster, I guess we have to

TonyBagnall:
image

TonyBagnall:
just me and my bots chatting on the PR

image

TonyBagnall:
so tl:dr, hit is small, but contract option is smallest.

TonyBagnall:
but claude code disagrees with chatgpt! oooh, what to do, oh ran out of tokens again

image

TonyBagnall:
more curious fastmath data, we are going with contract

ith 10 runs (medians, which strip warm-up outliers) the picture is even more decisive:

  
image

  Two important observations:

  1. fastmath=True is the slowest variant on every workload. This is not what fastmath is supposed to do. Strongly suggests the nnan/ninf flags are causing the optimiser to mishandle NaN simplex vertices, take more iterations, or converge
  to slower-to-fit model classes. The "speed" fastmath=True was buying us was a measurement artefact from the older bench where cache pollution let stale binaries shadow recompiles.
  2. {"contract"}, {"reassoc"}, and {"reassoc","contract"} are statistically indistinguishable and all ~30–40% faster than True. Any of them is correctness-safe (NaN/inf checks still work) AND a major speed win.

  Caveat about model selection

  I have not verified all five variants are selecting the same (error_type, trend_type, seasonality_type). If fastmath=True is picking a different (more expensive) model class because its guards are dead, these comparisons aren't
  apples-to-apples — but in that case True was also producing wrong/unstable models in production, which is even more reason to drop it.

hadifawaz1999 (dismissed):

TonyBagnall:
claude codes reply to claude chat

Thanks for the careful read on this — you're right that the finite guards were dead code under fastmath=True. The nnan/ninf flags let LLVM fold np.isnan(x)/np.isinf(x) to constant False, so the _is_finite(sse_) check inside _ets_fit 
  ▎ and the six np.isnan/np.isinf branches inside nelder_mead were not actually catching anything. The suite was passing only because the other fixes in this PR — excluding multiplicative trend by default, the tightened bounds, and the 
  ▎ NaN→inf normalisation in dispatch_loss (which is already off fastmath) — kept the bad cases away from the per-iteration guards.
  ▎
  ▎ Fixed by switching _ets_fit and nelder_mead from @njit(fastmath=True, cache=True) to @njit(fastmath={"contract"}, cache=True). {"contract"} keeps FMA fusion (which is where the speedup in this hot path actually comes from) but drops
  ▎ nnan/ninf, so the runtime finite checks evaluate correctly. The three inlined helpers (_ets_params_are_valid, _ets_state_is_valid, _ets_forecast_cycle_is_valid) are inline="always" with no decorator flags of their own, so they inherit
  ▎ the caller's flags — fixing _ets_fit fixes them transitively.
  ▎
  ▎ Benchmarked across n=20k–50k series at seasonal periods 1/12/24/52 with per-variant NUMBA_CACHE_DIR isolation and aggressive __pycache__ cleanup between runs (caches were the source of confusing earlier numbers). {"contract"} was 27–40
  ▎  % faster than {"True"} on every workload; {"reassoc"} and {"reassoc","contract"} were statistically indistinguishable from {"contract"}. The True slowdown likely reflects the dead guards letting NaN simplex vertices propagate through
  ▎ the optimiser instead of being rejected, so it does more iterations.
  ▎
  ▎ Holding off on the regression test that proves the guards fire under the unsafe flag — agree it should land, but it needs a constructed parameter vector that reliably produces a non-finite sse_ on this branch's tightened bounds, and
  ▎ I'd rather get that into its own follow-up PR than block this one. The six other ETS-path decorators still at fastmath=True (_arima_fit, _ets_initialise, _ets_update_states, _ets_predict_value, _init_ets_simplex, and the two _ets.py 
  ▎ kernels) have no finite checks in scope, so they're safe from this anti-pattern. Moving them to {"contract"} for the speed win is a separate change.

hadifawaz1999 (approved):

Issue #3453, opened 2026-06-16
[ENH] iterative_predict mixin


TonyBagnall:
### Describe the feature or idea you want to propose

tidying up forecasting, and the persistent confusion about use cases of when to fit or not. Resolved to this

| Method | Requires existing fit? | Fits internally? | Meaning |
|---|---:|---:|---|
| `fit(y)` | No | Yes | Estimate model state from `y` |
| `predict(y)` | Yes | No | Use fitted model to predict from context `y` |
| `forecast(y)` | No | Yes | Fit on `y`, then predict |
| `iterative_forecast(y, h)` | No | Yes, once | Fit on `y`, then recursively forecast `h` steps |

so 
f.fit(y)
f.iterative_forecast(y, 12)

calls fit twice, in that the logic is that iterative_forecast fits the model then forecasts. This is described in the documentation. 

### Describe your proposed solution

Occurred to me we could have 

iterative_predict( y, 12) for the cases where we want to fit then later forecast. Then you have

single -> horizon: forecast -> interative_forecast: no fit first
single -> horizon: predict -> interative_predict: fit required

PR #3454, opened 2026-06-16
[ENH] Refactor iterative_forecast for fit/predict differentiation and exog support

### LINKED ISSUE #3078: [ENH] Make fit in ARIMA iterative_forecast optional

Hi,

First, thank you for the great work on implementing a Python-native and efficient version of ARIMA.

The current implementation of `aeon.forecasting.stats.ARIMA.iterative_forecast` includes fitting the forecaster as its first step:

https://github.com/aeon-toolkit/aeon/blob/c80baf772dcf0e60aaaae87c27d1befdef90a7b4/aeon/forecasting/stats/_arima.py#L210

If the ARIMA model has already been fitted, this additional fit step is unnecessary. All the required information for subsequent predictions is already stored in the fitted model.

An option to skip this initial fit will speed up the iterative forecast when the model is already fitted.

A possible solution would be to include an optional argument (e.g., `fit=True`) in iterative_forecast.

- If `fit=True`, the method behaves as it currently does.
- If `fit=False`, the method skips the redundant fitting step and uses the already-fitted model directly.

```python
def iterative_forecast(self, y, prediction_horizon, fit=True):
    if fit:
        self.fit(y)
```

_No response_

### Additional context

_No response_

TonyBagnall:
See #3453 fixes #3078

This PR clarifies the contract for iterative forecasting and exogenous variables.

currently iterative_forecast does a check as to the equivalence of the input y. This PR removes this brittle test, and always fits in iterative_forecast, basically adopt the model:

| Method | Requires existing fit? | Fits internally? | Meaning |
|---|---:|---:|---|
| `fit(y)` | No | Yes | Estimate model state from `y` |
| `predict(y)` | Yes | No | Use fitted model to predict from context `y` |
| `forecast(y)` | No | Yes | Fit on `y`, then predict |
| `iterative_forecast(y, h)` | No | Yes, once | Fit on `y`, then recursively forecast `h` steps |

this has triggers a bigger redesign of forecasters to sort out the use of exogenous variables, the usage of which is mismatched between Regression and ARIMA. The main change is that iterative_forecast now has an explicit split between training exogenous data and future exogenous data:

```python
iterative_forecast(y, prediction_horizon, exog=None, *, future_exog=None)
```

The intended iterative_forecast logic is:

1. iterative_forecast always fits internally on the supplied y.
2. It does not reuse any previous fitted model state.
3. exog contains training exogenous values aligned with y.
4. future_exog contains future exogenous values aligned with the forecast horizon.
5. exog and future_exog must either both be provided, or both be None.
6. Combined past-and-future exog in a single exog argument is no longer supported.

This gives a clearer API. For example:

```python
forecaster.iterative_forecast(
    y,
    prediction_horizon=h,
    exog=train_exog,
    future_exog=future_exog,
)
```
means: fit once on (y, train_exog), then recursively forecast h steps using one row of future_exog for each forecast target.

### RegressionForecaster exogenous variables

This PR also updates RegressionForecaster so that exogenous variables follow the same target-time convention.

Previously, RegressionForecaster treated exogenous variables as additional lag-window channels by stacking y and exog before constructing regression windows. Conceptually, this meant training on features like:

[y[t-w], ..., y[t-1], x[t-w], ..., x[t-1]] -> y[t]

Now this may be a good way of doing things, but it is non standard compared to other packages and adds complexity involving storing internal state in predict.  The revised design treats exogenous variables as target-time regression features:

[y[t-w], ..., y[t-1], x[t]] -> y[t]

For horizon h, the exogenous row is aligned with the target value being predicted. This means _predict(y, exog=...) should accept the exogenous values needed for the prediction target, rather than requiring a full exogenous history of length window.

This aligns RegressionForecaster with the new iterative_forecast contract, and with the usual ARIMAX/SARIMAX-style convention used by statistical forecasting APIs: historical exog is used at fit time, and future exog is supplied for the forecast horizon.

Historical exogenous effects can still be represented by explicitly constructing lagged exogenous features before passing them to the forecaster. The important change is that exog itself now has one consistent meaning across fit, predict, and iterative_forecast.


aeon/forecasting/stats/_arima.py
```diff
@@ -17,6 +17,16 @@
...
+    """Convert exog to ``(n_timepoints, n_features)`` rows."""
+    exog = np.asarray(exog)
+    if exog.ndim == 0:
+        raise ValueError("exog must be one- or two-dimensional.")
+    if exog.ndim == 1:
+        exog = exog.reshape(-1, 1)
+    return exog
+
+
 class ARIMA(BaseForecaster, IterativeForecastingMixin):
```
hadifawaz1999, line 30: The docstring should also mention about exog no ? like in the regressor forecaster (docstring added in this PR)

hadifawaz1999 (changes requested):
just small docstring comment

hadifawaz1999 (approved):

Issue #3451, opened 2026-06-16
[ENH] Implement a DynamicOptimisedThetaForecaster model (DOTM)


TonyBagnall:
### Describe the feature or idea you want to propose

The Dynamic Optimised Theta Model (DOTM) is a generalisation of the Theta method. It is useful as a standalone univariate forecaster and is also needed as one of the component models for implementing SCUM, the Simple Combination of Univariate Models, see #3377.
Primary reference for DOTM:

Fiorucci, J. A., Pellegrini, T. R., Louzada, F., Petropoulos, F., & Koehler, A. B. (2016).
"Models for optimising the theta method and their relationship to state space models."
International Journal of Forecasting, 32(4), 1151-1161.
doi:10.1016/j.ijforecast.2016.02.005

### Summary of the algorithm

The classical Theta method decomposes a deseasonalised series into two theta lines:

a long-term component, often represented by the theta = 0 line, which removes curvature and captures the long-term trend;
a short-term component, traditionally using theta = 2, which emphasises local curvature and is usually extrapolated using simple exponential smoothing.

The forecasts from these components are then recombined.

DOTM extends the classical Theta method in two main ways:

Optimised theta: instead of fixing the short-term theta value to 2, DOTM estimates or selects the theta value that gives the best forecasting performance according to an objective function.
Dynamic formulation: DOTM revises the long-term theta line dynamically through a state-space formulation, rather than treating the long-term line as fixed after initial estimation.

We should be able to use the existing fast implementation of theta for this, although irrc we didnt use a state space formulation, will have to revisit.

### Reference implementations

The forecTheta R package includes a dotm implementation, alongside the Standard Theta Model, Optimised Theta Model and Dynamic Standard Theta Model.

statsforecast has native Theta, OptimizedTheta, DynamicTheta, DynamicOptimizedTheta, and AutoTheta implementations.

TonyBagnall:
closed by #3455

PR #2978, opened 2025-07-30
[ENH] Theta forecaster

TonyBagnall:
resolves #2977 

This is a stripped down implementation of theta forecast that matches the version used in the M3 competitions. 

Its very fast. Forecasts closely match statsforecast version, which is strangely slow at scale. Not sure why statsmodels forecasts are a bit different, it could be the config, but these results are I feel sufficient to validate the implementation.  

image

PR #3455, opened 2026-06-16
[ENH] Dynamic Optimised Theta Model (DOTM) forecaster

TonyBagnall:
#### Reference Issues/PRs
#3451 

as part of our ensemble package we are adding Dynamic Optimised Theta Model forecaster. This is to sit alongside our existing Theta model which is basic, true to the original and fast. DOTM is an extension of the Theta forecaster, which I have already implemented in aeon in PR #2978 and described in this blog https://medium.com/@aeon.toolkit/the-aeon-theta-forecaster-afaa190d6dbb

## From the Theta method to DOTM

The Dynamic Optimised Theta Model (DOTM) is the most flexible member of the Theta family. It's easiest to understand by building up from the classical method.

### Classical Theta (Assimakopoulos & Nikolopoulos, 2000)

The Theta method rests on one trick: you can change a series' local *curvature* while keeping its long-run trend fixed, by scaling its second differences. The resulting "theta lines" have a clean closed form — a blend of the data and its OLS trend line:

$$Z_t(\theta) = \theta\, y_t + (1-\theta)(a + b t)$$

So $\theta$ is a curvature knob:

- $\theta = 0$ → the straight OLS trend line (all curvature removed),
- $\theta = 1$ → the original series,
- $\theta = 2$ → deviations from the trend doubled.

The classical method forecasts by handling two lines separately and averaging them: extrapolate the $\theta=0$ line linearly, smooth the $\theta=2$ line with simple exponential smoothing (SES) and hold it flat, then combine with equal weight. Hyndman & Billah (2003) showed this is equivalent to **SES with a drift of half the OLS slope** — which is why it's cheap and very hard to beat (it won M3).

### Optimised Theta — OTM (Fiorucci et al., 2016)

Fiorucci et al. made two changes:

1. **Stop fixing the knobs.** Instead of $\theta = 2$ and equal weights, *estimate* the initial level $\ell_0$, the SES smoothing $\alpha$, and $\theta$ by minimising in-sample SSE.
2. **Rewrite as one recursion.** The two-line decomposition collapses into a single state-space form: an SES level plus a trend term scaled by

$$\omega = 1 - \tfrac{1}{\theta}.$$

The one-step forecast is $\ \mu_{t+1} = \ell_t + \omega \,(\text{extrapolated trend})$. When $\theta \to 1$, $\omega \to 0$ and the model reduces to plain SES; classical Theta ($\theta = 2$) is recovered as the special case $\omega = \tfrac12$.

### Dynamic Optimised Theta — DOTM

OTM estimates the trend line $(A, B)$ once, on the whole sample. **DOTM re-estimates it at every time step** via recursive OLS of $y$ on time, so the model carries an *evolving* local trend rather than one global fit. That's the only difference from OTM, and it's what "dynamic" refers to.

This PR adds a Dynamic Optimised Theta Model, DOTM, forecaster to aeon.forecasting.stats.

DOTM is the most flexible member of the Theta family. It extends the classical Theta method by estimating the initial level, smoothing parameter and theta parameter, and by updating the underlying linear trend dynamically through a recursive state-space formulation. This implementation is intended to sit alongside the existing Theta forecaster, which remains a fast implementation of the classical Theta method.

Main features

This implementation includes:

a new DOTM forecaster exported from aeon.forecasting.stats;
fitting of initial_level, alpha and theta, unless fixed by the user;
bounded Nelder-Mead optimisation of the scaled in-sample SSE objective;
recursive multi-step forecasting through iterative_forecast;
optional seasonal adjustment;
support for additive and multiplicative seasonal decomposition;
automatic, forced or disabled seasonality handling;
re-seasonalised fitted values and forecasts on the original data scale;
finite-output checks and guard paths for short, constant and non-finite input cases;
tests for fixed-parameter mode, forecast shape, one-step forecast consistency, constant series, invalid inputs and seasonal behaviour.
Seasonality

The implementation now supports seasonal DOTM through an outer seasonal adjustment layer:

original series
  -> optional seasonal adjustment
  -> non-seasonal DOTM state-space core
  -> recursive adjusted forecasts
  -> seasonal naive forecast of the seasonal component
  -> forecasts returned on the original scale

The DOTM state-space recursion and optimiser remain the core model. Seasonality is handled as a pre/post-processing step, which keeps the validated non-seasonal implementation stable while allowing the model to handle seasonal series.

The seasonal interface allows the user to specify a seasonal period and decomposition type. Multiplicative decomposition is intended for strictly positive series. Where multiplicative decomposition is unsafe, for example for non-positive data, the implementation falls back to a safer additive treatment or avoids seasonal adjustment, depending on the validation path.

Validation

The non-seasonal implementation was benchmarked against R forecTheta::dotm and StatsForecast DynamicOptimizedTheta. On a 132-series benchmark the results were very close:

backend         mean sMAPE   median sMAPE   mean MASE   median MASE
aeon             18.1383       15.2052        2.77337      1.53783
forecTheta       18.1656       15.1967        2.76891      1.56632
statsforecast    18.2049       15.2019        2.77229      1.63325

Runtime totals on the same benchmark were:

aeon:          5.58s
forecTheta:   12.81s
statsforecast: 6.47s

The seasonal extension should be interpreted as a practical seasonal wrapper around the validated DOTM core. Exact equality to R or StatsForecast is not expected in all seasonal cases, because seasonal tests and decomposition details differ between implementations. The important checks are that the non-seasonal path remains unchanged, seasonal forecasts are finite and on the original scale, and forced seasonal decomposition behaves sensibly on controlled synthetic examples.

Deliberate limitations

This PR does not add:

exogenous variables;
prediction intervals;
OTM, STM or DSTM variants;
automatic selection across the Theta family;
SCUM integration.

Those are better handled in follow-up PRs. The immediate aim is to add a reliable standalone DOTM forecaster, including practical seasonal support, so it can later be used as one component in the planned SCUM-style ensemble

**Reference:** Fiorucci, J. A., Pellegrini, T. R., Louzada, F., Petropoulos, F., & Koehler, A. B. (2016). Models for optimising the theta method and their relationship to state space models. *International Journal of Forecasting*, 32(4), 1151–1161.


TonyBagnall:
The non seasonal version seems very similar overall to the R version and statsforecast

Headline summary, 132 series:

   backend          mean sMAPE    median sMAPE    mean MASE    median MASE                                                                                                               
  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━━━━
   aeon                18.1383         15.2052      2.77337        1.53783
  ───────────────  ────────────  ──────────────  ───────────  ─────────────
   forectheta          18.1656         15.1967      2.76891        1.56632
  ───────────────  ────────────  ──────────────  ───────────  ─────────────
   statsforecast       18.2049         15.2019      2.77229        1.63325

  Runtime totals:

  - aeon: 5.58s
  - forectheta: 12.81s
  - statsforecast: 6.47s

TonyBagnall:
non seasonal version seems to work fine, will try speed it up then add the seasonal version on top

image

TonyBagnall:
ran some optimisations on the code, now about 5x faster

TonyBagnall:
pleased with that for seasonal!
image

TonyBagnall:
Final repeat run with latest version



# aeon `DOTM` vs StatsForecast `DynamicOptimizedTheta`

I ran aeon `DOTM` against StatsForecast `DynamicOptimizedTheta` on 72-series M1/M3 used in development for correctness. I will run the full 100k M1/M3/M4 later local comparison sample: the first 10 series per type, plus **M1:83** and **M3:113** as problem series. Both backends completed all 72 series with finite forecasts.

## Non-seasonal DOTM core

Paired differences are **aeon minus StatsForecast**.

| metric | mean diff | median diff | t-test p | Wilcoxon p |
|:-------|----------:|------------:|---------:|-----------:|
| sMAPE  |   -0.0976 |    -1.17e-8 |    0.485 |      0.827 |
| MASE   |   0.00634 |   -1.03e-10 |    0.663 |      0.968 |
| MAE    |     234.1 |    -1.74e-8 |    0.426 |      0.936 |
| RMSE   |     272.0 |    -4.62e-8 |    0.225 |      0.866 |

No statistically significant paired accuracy difference was detected.

W/D/L counts (lower is better), draw tolerance relative `1e-6` or absolute `1e-9`:

| metric | aeon W | draw | aeon L |  n |
|:-------|-------:|-----:|-------:|---:|
| sMAPE  |     33 |   10 |     29 | 72 |
| MASE   |     32 |   10 |     30 | 72 |
| MAE    |     32 |   10 |     30 | 72 |
| RMSE   |     32 |    9 |     31 | 72 |

Timing:

- aeon median: `0.000843s`
- StatsForecast median: `0.001839s`
- aeon median ratio: `0.512`
- excluding first call, aeon steady median: `0.000841s`, StatsForecast steady median: `0.001859s`

Obviously very fast, really only interesting comparison for much longer series, but thats future work.

## Seasonal DOTM, multiplicative decomposition, forced seasonal test

Both aeon and StatsForecast deseasonalised the same **40/72** series.

| metric | mean diff | median diff | t-test p | Wilcoxon p |
|:-------|----------:|------------:|---------:|-----------:|
| sMAPE  |    -0.274 |    -0.00210 |    0.505 |      0.264 |
| MASE   |  -0.00212 |   -0.000335 |    0.928 |      0.366 |
| MAE    |   -1653.5 |     -0.0511 |    0.325 |      0.469 |
| RMSE   |   -1716.7 |     -0.0569 |    0.348 |      0.897 |

Again, no statistically significant paired accuracy difference was detected, although nominally aeon doing better here,

W/D/L counts:

| metric | aeon W | draw | aeon L |  n |
|:-------|-------:|-----:|-------:|---:|
| sMAPE  |     40 |    2 |     30 | 72 |
| MASE   |     40 |    2 |     30 | 72 |
| MAE    |     40 |    2 |     30 | 72 |
| RMSE   |     38 |    2 |     32 | 72 |

Timing:

- aeon total includes a large first-call Numba compile cost: `9.75s` cold start.
- aeon steady median: `0.000689s`
- StatsForecast steady median: `0.001750s`
- paired median time ratio aeon / StatsForecast: `0.400`

## Interpretation

Forecast accuracy is statistically indistinguishable from StatsForecast on this sample. The non-seasonal core is very close. Seasonal forecasts show larger per-series forecast deltas than CES because decomposition/reseasonalisation details differ, but the paired accuracy tests still do not show a significant difference.

aeon/forecasting/stats/_dotm.py
```diff
@@ -0,0 +1,1051 @@
...
+        future_exog=None,
+    ):
+        """Fit DOTM on ``y`` and recursively forecast ``prediction_horizon`` steps.
+
+        ``exog`` and ``future_exog`` are accepted for signature compatibility
+        with :class:`~aeon.forecasting.base.IterativeForecastingMixin` but are
+        not supported by DOTM; passing either raises
+        :class:`NotImplementedError`.
+        """
+        if prediction_horizon < 1:
```
chrisholder, line 306: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**`iterative_forecast` skips the `prediction_horizon` type check the mixin enforces.** This override only checks `< 1`, then does `h = int(prediction_horizon)`. The base `IterativeForecastingMixin._check_iterative_forecast_inputs` raises `TypeError` for `bool`/non-int and rejects floats, so here `iterative_forecast(y, True)` silently runs with `h=1` and `iterative_forecast(y, 2.5)` silently truncates to 2 — a small behavioural divergence from the documented mixin contract. Consider mirroring the mixin's type guard (or calling `_check_iterative_forecast_inputs` for validation before branching).

aeon/forecasting/stats/_dotm.py
```diff
@@ -0,0 +1,1051 @@
...
+        y = _prepare_dotm_y(y)
+        if not self.deseasonalised_:
+            return float(
+                _dotm_forecast(y, 1, self.initial_level_, self.alpha_, self.theta_)[0]
+            )
+
+        n = y.shape[0]
+        m = self.season_length_
+        factors = self.seasonal_factors_
+        rep = factors[np.arange(n) % m]
```
chrisholder, line 270: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Seasonal phase assumption is sharp (documented, just flagging).** Factors are applied by absolute position `np.arange(n) % m`, which assumes the context passed to `predict` shares the training series' seasonal phase. The docstring is honest about this and the rolling-predict regression test covers the in-phase case, so this is fine for an experimental module — noting only that an out-of-phase context will silently misalign the re-seasonalisation.

aeon/forecasting/stats/_dotm.py
```diff
@@ -0,0 +1,1051 @@
...
+    # Multiplicative viability check.
+    use_mult = requested_mult
+    if use_mult:
+        for i in range(n):
+            if y[i] <= 0.0:
+                use_mult = False
+                break
+    if use_mult:
+        for t in range(half, n - half):
+            if ma[t] <= 0.0:
```
chrisholder, line 585: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Coverage gap (non-blocking).** `test_dotm_multiplicative_falls_back_to_additive_when_y_non_positive` exercises the `y <= 0` branch, but the fallback triggered by a non-positive centred MA (this line) and by below-floor / non-finite normalised factors further down isn't directly tested. A small synthetic case would lock these paths in if it's easy.

aeon/forecasting/stats/_dotm.py
```diff
@@ -0,0 +1,1051 @@
...
+
+    ell = level
+    a = a_state
+    b = b_state
+    mean_y = mean_y_state
+
+    # The first forecast is computed at i = n - 1, so power = (1 - alpha) ** n.
+    # We then update power = power * (1 - alpha) each iteration to avoid the
+    # ``**`` call that the original recurrence performed inside the loop.
+    power = one_minus_alpha**n
```
chrisholder, line 733: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Nit.** `power` is precomputed as `(1 - alpha)**n` then advanced by repeated multiplication under `fastmath=True`. Fine for the small horizons DOTM targets; just noting it trades a little floating-point accuracy for speed versus recomputing `**` each step.

aeon/forecasting/stats/_dotm.py
```diff
@@ -0,0 +1,1051 @@
...
+    b = 0.0
+    sse = 0.0
+    for i in range(0, y.shape[0] - 1):
+        power = (1.0 - alpha) ** (i + 1)
+        slope_term = b * (1.0 - (1.0 - alpha) ** (i + 2)) / alpha
+        fitted = ell + omega * (a * power + slope_term)
+        residual = y[i + 1] - fitted
+        if not np.isfinite(fitted) or not np.isfinite(residual):
+            return 1e300
+        if i + 1 >= 2:
```
chrisholder, line 871: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Question, not a change request.** The SSE objective starts accumulating at the third fitted point (`i + 1 >= 2`), skipping `fitted[1]`. Is that a deliberate match to the `statsforecast` initialisation rather than an off-by-one? The numerical benchmark in the PR description suggests it's intentional — just confirming.

chrisholder (commented):
> 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

## Summary

Adds a `DOTM` forecaster to `aeon.forecasting.stats`, implementing Fiorucci et al. (2016)'s Dynamic Optimised Theta Model as a numba-accelerated state-space recursion with bounded Nelder–Mead estimation, plus an outer seasonal-decomposition layer. It does what #3451 asks for (a standalone DOTM, intended as a future SCUM component), is well-documented, numerically validated against `forecTheta` and `statsforecast`, and CI-green. The core recursion is internally consistent across the three forecast paths (`_dotm_fitted_values`, `_dotm_forecast`, `_dotm_forecast_from_state`), and edge-case handling (short/constant/non-finite series, multiplicative→additive fallback, ACF auto seasonal test) is thorough. `[ENH]` prefix and `forecasting` label are correct, and as `forecasting` is experimental no deprecation cycle is needed.

**Recommendation: Approve with nits.** One required-by-guidelines doc fix below; the rest are small robustness/consistency points left as inline comments.

## Should-fix

**API reference docs.** The new public class isn't added to `docs/api_reference/forecasting.rst` (CI doesn't catch this, but aeon's guidelines require new public functionality to appear in the API reference). `aeon/forecasting/stats/__init__.py` and `__all__` are correctly updated, so this is the only registration gap:

```rst
    TAR
    AutoTAR
    Theta
    DOTM
    TVP
```

## Inline comments

I've left line-specific notes on: the `iterative_forecast` type-check divergence from the mixin contract, a seasonal-phase assumption in `_predict`, a fallback-path coverage gap, a `fastmath` precision nit, and a question about the SSE start index.

## Minor / optional

- No test asserts `fit`/`predict` leave the input `y` unmutated. Implementation looks clean (out-of-place ops only); an explicit assertion would lock it in.
- Consider defining `_get_test_params` with a fixed-parameter set (e.g. `{"alpha": 0.5, "theta": 2.0}`) so the general estimator checks skip the optimiser and stay deterministic — `{}` works but runs the full Nelder–Mead.

Nice work — clean, well-validated, and the docstrings/`References`/`Examples` are all in good shape.

chrisholder (approved):

aeon/forecasting/stats/_dotm.py
TonyBagnall: And codex replies
Yes, that is deliberate.

  The DOTM objective skips the first two observations because those early points are used to initialise the dynamic line states. The first fitted value is the trivial anchor, and the second fitted point is still affected by the initial
  state construction rather than the steady DOTM recursion. Scoring starts from the third fitted point to match the StatsForecast/forecTheta-style objective used in the validation runs.

  So i + 1 >= 2 is intentional, not an off-by-one. The forecast recursion and fitted-value recursion still compute the full in-sample path; only the optimisation SSE excludes the first two fitted points

human contribution: what a time to be alive!

aeon/forecasting/stats/_dotm.py
TonyBagnall: I FEEL THE NEED FOR SPEED: or, as codex says

Yes, that is intentional.

  The recurrence only uses the iterative update in the forecast tail, after fitting has already produced the final state. Advancing

  power *= 1.0 - alpha

  avoids repeated exponentiation in multi-step forecasts and keeps iterative_forecast cheap. The expected horizons for DOTM are small enough that accumulated rounding drift should be negligible relative to the model error, and the
  validation against StatsForecast did not show material forecast differences.

  I agree it is a speed/rounding tradeoff rather than an exact algebraic recomputation at every step.

aeon/forecasting/stats/_dotm.py
TonyBagnall: Good point. The below-floor multiplicative-factor fallback is now covered by:

  `test_dotm_multiplicative_falls_back_to_additive_for_tiny_factors`

  That uses strictly positive data with a very small seasonal position, so it exercises the fallback after multiplicative seasonal factors are normalised.

  For the non-positive centred-MA branch: with finite strictly positive `y`, the centred moving average is a weighted average with non-negative weights, so it cannot be non-positive. If `y` contains a non-positive value, the earlier `y <=
  0` guard already sends the path to additive before the MA check. So that line is defensive rather than a practically reachable branch under valid finite input.

aeon/forecasting/stats/_dotm.py
TonyBagnall: codex:

  Agreed. This is an explicit assumption in the current local-forecaster API shape: `predict(y)` receives only the context values, not an origin timestamp or phase offset, so DOTM has no reliable way to infer whether an arbitrary supplied
  context is phase-aligned with the training series.

  The implementation therefore applies fitted factors by absolute position (`np.arange(n) % m`) and documents that the context is assumed to share the training phase. The rolling-predict regression test covers the intended in-phase use
  case. Supporting out-of-phase contexts would need an explicit phase/origin parameter or a richer forecasting index contract, so I would leave that out of this experimental implementation.

PR #3456, opened 2026-06-16
[BUG] Fix TimeSeriesCLARA sampling

### LINKED ISSUE #3420: [BUG] Fix sampling in TimeSeriesCLARA so samples are drawn from the full dataset
### Describe the bug

TimeSeriesCLARA is intended to repeatedly sample subsets of the dataset, run PAM on each subset, and keep the best set of medoids. However, the current implementation appears to initialise the candidate sample indices using:

sample_idxs = np.arange(n_samples)

and then, when n_samples < n_cases, samples from this array:

sample_idxs = self._random_state.choice(
    sample_idxs,
    size=n_samples,
    replace=False,
)

This means the sample is drawn only from the first n_samples cases, rather than from all n_cases cases. Since the sample size equals the size of the candidate index array, each sampling iteration just permutes the same first n_samples indices. As a result, CLARA does not properly sample different subsets from the full dataset.

In aeon/clustering/_clara.py, change the sampling logic so that candidate samples are drawn from all case indices:

```python
if n_samples < n_cases:
    sample_idxs = self._random_state.choice(
        np.arange(n_cases),
        size=n_samples,
        replace=False,
    )
else:
    sample_idxs = np.arange(n_cases)
```

### Expected results

samples from n_cases

### Actual results

samples from n_samples

### Versions

_No response_

adam-goose:
#### Reference Issues/PRs

Fixes #3420 

This fixes the sampling logic in `TimeSeriesCLARA`.

Previously, when `n_samples < n_cases`, candidate sample indices were drawn from `np.arange(n_samples)`, meaning CLARA could only sample from the first `n_samples` cases.

This change updates the sampling logic so indices are drawn from `np.arange(n_cases)` instead, allowing samples to be selected from the full dataset.

The previous expected outputs in the CLARA tests were based on the buggy sampling behaviour. Since the corrected implementation samples from the full set of cases, the deterministic outputs for those tests changed and have been updated accordingly.

A regression test has also been added to verify that CLARA can sample cases beyond the first `n_samples`.

No.

#### Any other comments?

I ran:

```bash
python -m pytest aeon/clustering/tests/test_clara.py -vv
```
and the tests passed.

##### For all contributions

### DIFF
--- a/aeon/clustering/_clara.py
+++ b/aeon/clustering/_clara.py
@@ -165,13 +165,17 @@ class TimeSeriesCLARA(BaseClusterer):
         best_pam = None
         best_labels = None
         for _ in range(self.n_sampling_iters):
-            sample_idxs = np.arange(n_samples)
+
             if n_samples < n_cases:
                 sample_idxs = self._random_state.choice(
-                    sample_idxs,
+                    np.arange(n_cases),
                     size=n_samples,
                     replace=False,
                 )
+
+            else:
+                sample_idxs = np.arange(n_cases)
+
             pam = TimeSeriesKMedoids(
                 n_clusters=self.n_clusters,
                 init=self.init,
--- a/aeon/clustering/tests/test_clara.py
+++ b/aeon/clustering/tests/test_clara.py
@@ -34,18 +34,18 @@ def test_clara_uni():
     proba = clara.predict_proba(X_test)
     assert np.array_equal(
         test_medoids_result,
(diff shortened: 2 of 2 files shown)


TonyBagnall (approved):
thanks for this, all LGTM. One related follow-up, not a blocker for this PR: array-valued `init` in CLARA may need clarification, I'll raise an issue

PR #3457, opened 2026-06-16
[ENH] Vectorize OOB divisor counting in RotationForest with np.bincount

islem-esi:
#### Reference Issues/PRs

None. This is a small self-contained performance change using numpy.

`RotationForestClassifier.fit_predict_proba` and `RotationForestRegressor.fit_predict`
count out-of-bag (OOB) predictions per case with a nested Python loop, then normalize
with a second per-case Python loop. Both are pure-Python over every OOB prediction and
every case.

This counts with `np.bincount` over the concatenated OOB indices and vectorizes the
normalization with masked array ops. This keeps same behavior including the
`divisors == 0 → label_average / uniform` fallback  and no divide-by-zero warning is
introduced (zero-count cases are masked before the divide).

Validation: differential test, old vs new, on random `(y_preds/y_probas, oobs)` cases
incl. duplicate indices, empty OOB lists and zero-divisor cases. Micro-benchmark at
`n_cases=5000, n_estimators=200`: ~43 ms down to ~1.5 ms (~29 times faster)...

The same `divisors[inst] += 1` idiom also appears in `base_interval_forest.py` (2 sites)
and a weighted variant in `_arsenal.py`; I left those out since I only verified the
RotationForest path, happy to follow up in a separate PR.

No. `numpy` is already a core dependency.

#### Any other comments?

None

### DIFF
--- a/aeon/classification/sklearn/_rotation_forest_classifier.py
+++ b/aeon/classification/sklearn/_rotation_forest_classifier.py
@@ -282,17 +282,17 @@ class RotationForestClassifier(ClassifierMixin, BaseEstimator):
         y_probas, oobs = zip(*p)
 
         results = np.sum(y_probas, axis=0)
-        divisors = np.zeros(self.n_cases_)
-        for oob in oobs:
-            for inst in oob:
-                divisors[inst] += 1
+        oob_arrays = [np.asarray(o, dtype=np.intp) for o in oobs if len(o) > 0]
+        if oob_arrays:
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
thanks very much for this, follow up PRs for the others would be most welcome, measured the speed up too, its real

TonyBagnall:
@all-contributors add @islem-esi for code

PR #3459, opened 2026-06-17
[BUG] Reject array init for TimeSeriesCLARA

### LINKED ISSUE #3423: [BUG] TimeSeriesCLARA passes full-dataset init indices to sampled PAM fit
### Describe the bug

`TimeSeriesCLARA` passes array initialisation indices directly to `TimeSeriesKMedoids` after subsampling the input data.

This is problematic because `TimeSeriesCLARA` fits PAM on a sampled subset, `X[sample_idxs]`, but an array passed through `init` appears to be interpreted as indices into the full original dataset. These full-dataset indices are then used unchanged inside PAM, where they are interpreted as indices into the smaller sampled subset.

For example, if `init=np.array([6, 7])` and `n_samples=4`, PAM is fitted on a dataset of length 4, so indices 6 and 7 are invalid.

This currently raises:

```python
ValueError: Values must be in the range [0, 4).

The behaviour is ambiguous. Either array initialisation should be unsupported for TimeSeriesCLARA, or the full-dataset indices should be mapped correctly into the sampled subset

```python
import numpy as np

from aeon.clustering import TimeSeriesCLARA

X = np.random.RandomState(0).random(size=(10, 1, 8))

clara = TimeSeriesCLARA(
    n_clusters=2,
    init=np.array([6, 7]),
    n_samples=4,
    n_sampling_iters=1,
    distance="euclidean",
    random_state=0,
)

clara.fit(X)
```

### Expected results

TimeSeriesCLARA should not pass full-dataset initialisation indices directly into TimeSeriesKMedoids when PAM is fitted on a sampled subset.

One of the following behaviours should be implemented:

Reject array initialisation for TimeSeriesCLARA, with a clear error message explaining that array initialisation is not currently supported because CLARA fits PAM on sampled subsets.

For example:

ValueError: Array initialisation is not supported for TimeSeriesCLARA because CLARA fits PAM on sampled subsets. Use a string initialisation method instead.
Support array initialisation explicitly, by treating init indices as full-dataset indices, ensuring those cases are included in each sampled subset, mapping the full-dataset indices to local subset indices, and passing those local indices to TimeSeriesKMedoids.

For a first fix, rejecting array initialisation is probably the simplest and safest behaviour. The TimeSeriesCLARA docstring should also be updated to match the implemented behaviour.

### Actual results

ValueError: Values must be in the range [0, 4).

### Versions

_No response_

ProfTrader:
## Summary
- Reject ndarray initialisation in TimeSeriesCLARA with a clear ValueError before fitting sampled PAM subsets.
- Update the init docstring to match the supported CLARA initialisation methods.
- Add a regression test for the invalid full-dataset init-index case from #3423.

## Validation
- python -m py_compile aeon/clustering/_clara.py aeon/clustering/tests/test_clara.py
- Could not run pytest locally because this Python environment is missing pytest/sklearn.

Closes #3423.

### DIFF
--- a/aeon/clustering/_clara.py
+++ b/aeon/clustering/_clara.py
@@ -29,7 +29,7 @@ class TimeSeriesCLARA(BaseClusterer):
     n_clusters : int, default=8
         The number of clusters to form as well as the number of
         centroids to generate.
-    init : str or np.ndarray, default='random'
+    init : str, default='random'
         Method for initialising cluster centers. Any of the following are valid:
         ['kmedoids++', 'random', 'first'].
         Random is the default as it is very fast and it was found in [2] to
@@ -38,8 +38,6 @@ class TimeSeriesCLARA(BaseClusterer):
         accurate than random. It works by choosing centroids that are distant
         from one another. First is the fastest method and simply chooses the
         first k time series as centroids.
-        If an np.ndarray is provided it must be of shape (n_clusters,) and contain
-        the indices of the time series to use as centroids.
     distance : str or Callable, default='msm'
         Distance method to compute similarity between time series. A list of valid
         strings for metrics can be found in the documentation for
@@ -154,6 +152,13 @@ class TimeSeriesCLARA(BaseClusterer):
         return self._kmedoids_instance.predict(X)
 
     def _fit(self, X: np.ndarray, y=None):
+        if isinstance(self.init, np.ndarray):
+            raise ValueError(
+                "Array initialisation is not supported for TimeSeriesCLARA "
+                "because CLARA fits PAM on sampled subsets. Use a string "
+                "initialisation method instead."
+            )
+
         self._random_state = check_random_state(self.random_state)
         n_cases = X.shape[0]
         if self.n_samples is None:
--- a/aeon/clustering/_clara.py
+++ b/aeon/clustering/_clara.py
@@ -115,7 +115,7 @@ class TimeSeriesCLARA(BaseClusterer):
     def __init__(
         self,
         n_clusters: int = 8,
-        init: str | np.ndarray = "random",
+        init: str = "random",
         distance: str | Callable = "msm",
         n_samples: int | None = None,
         n_sampling_iters: int = 10,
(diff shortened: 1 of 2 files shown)


TonyBagnall:
thanks for this, I just merged a PR that fixed #3420 that unfortunately conflicts with yours. Could you resolve the conflict then I'll review? ta

ProfTrader:
Thanks Tony! I’ve resolved the conflict by merging the latest `main` into this branch and pushed the merge commit.

I kept both pieces of coverage: the new CLARA sampling test from the recently merged fix and this PR’s array-initialisation rejection test. I also ran the focused local check:

```
uv run --extra dev python -m pytest aeon/clustering/tests/test_clara.py
```

That passes locally with 4 tests passing. Thanks again for the heads up and for taking a look!

aeon/clustering/_clara.py
```diff
@@ -154,6 +152,13 @@ def _predict(self, X: np.ndarray, y=None) -> np.ndarray:
         return self._kmedoids_instance.predict(X)
 
     def _fit(self, X: np.ndarray, y=None):
+        if isinstance(self.init, np.ndarray):
+            raise ValueError(
+                "Array initialisation is not supported for TimeSeriesCLARA "
+                "because CLARA fits PAM on sampled subsets. Use a string "
+                "initialisation method instead."
+            )
```
TonyBagnall, line 160: best test if not a string, otherwise List etc could get through

TonyBagnall: Could you also add a list case to test_clara_rejects_array_init (e.g. parametrize over np.array([6, 7]) and [6, 7])

aeon/clustering/_clara.py
TonyBagnall: this is now out of date, should just be a string

TonyBagnall (changes requested):
thanks for this, a couple of very minor changes requested

ProfTrader:
Thanks Tony, addressed those changes. I tightened `TimeSeriesCLARA` so `init` is typed as `str` and rejects any non-string value, then parametrized the regression test over both `np.array([6, 7])` and `[6, 7]`.

Focused validation passed locally:

```
uv run --extra dev python -m pytest aeon/clustering/tests/test_clara.py
uv run --extra dev python -m py_compile aeon/clustering/_clara.py aeon/clustering/tests/test_clara.py
git diff --check
```

TonyBagnall (approved):
lgtm, thanks

TonyBagnall:
@all-contributors add @ProfTrader for bug

Issue #3462, opened 2026-06-17
[ENH] Complex Exponential Smoothing (CES)


TonyBagnall:
### Describe the feature or idea you want to propose

Implement Complex Exponential Smoothing (CES) forecaster

## Describe the feature

For SCUM we now have, or are close to having:

* `ARIMA`
* `ETS`
* DOTM
* missing: `CES`

The SCUM paper used CES through R `smooth::auto.ces()`, selected by AICc. For aeon, this should be implemented in stages. Phase 1 should implement a non-seasonal CES forecaster, then later work can extend this to seasonal/automatic CES for closer SCUM reproduction.

## Background

Complex Exponential Smoothing extends exponential smoothing by using complex-valued smoothing parameters and state updates. The fitted model uses a complex smoothing parameter, commonly written as:

```text
alpha = alpha_0 + i alpha_1
```

The real part of the recursive state gives the forecast, while the imaginary part acts as an internal correction term. This gives CES a different inductive bias from ETS, ARIMA and DOTM, which is why it is useful as a component in a heterogeneous forecasting ensemble such as SCUM.

The main reference implementation is R `smooth::ces()` / `smooth::auto.ces()`. StatsForecast also exposes `AutoCES`, which can be used as a Python-side benchmark.

## Proposed implementation

Add a new forecaster:

```text
aeon/forecasting/stats/_ces.py
```

```python
class CES(BaseForecaster, IterativeForecastingMixin):
    """Complex Exponential Smoothing forecaster."""
```

## Forecasting behaviour

The forecaster should follow the local univariate forecasting pattern used by the existing statistical forecasters:

* `_fit(y, exog=None)` fits the CES parameters on a single 1D series;
* `_predict(y, exog=None)` uses the fitted parameters and the supplied context series to return a one-step-ahead forecast;
* `iterative_forecast(y, prediction_horizon, exog=None)` fits on `y` and recursively forecasts the requested horizon;

The output should always be real-valued, even if the internal state is complex.

## Tests

Add tests in something like:

```text
aeon/forecasting/stats/tests/test_ces.py
```

Minimum tests:

1. Fit works on a simple positive non-seasonal series.
2. Forecast horizon shape is correct.
3. Fitted values, residuals and forecasts are finite.
4. Estimated `alpha_real_` and `alpha_imag_` satisfy the configured bounds.
5. Fixed-parameter mode works when `alpha_real` and `alpha_imag` are supplied.
[...]

PR #3463, opened 2026-06-17
[ENH] Implement Complex Exponential Smoothing (CES) forecaster

TonyBagnall:
his PR adds an initial implementation of Complex Exponential Smoothing (CES) for forecasting, closing #3462.

CES is a state space exponential smoothing method that uses a complex-valued smoothing parameter and two real-valued state components. This first implementation adds a non-seasonal CES forecaster to aeon.forecasting.stats, with support for fixed or optimised smoothing parameters, fitted values, residuals, one-step forecasts, and recursive multi-step forecasting.

Implementation plan

We are implementing CES in two phases, both on this PR

Phase 1, non-seasonal CES

This PR implements the core non-seasonal CES model. The aim is to establish the base recurrence, aeon forecaster integration, parameter validation, deterministic behaviour, and basic test coverage before extending the model.

Included in this phase:

non-seasonal CES recurrence;
optional fixed alpha_real, alpha_imag, and initial_level;
optimisation of unfixed parameters by minimising SSE;
fitted values, residuals, and stored one-step forecast;
predict and iterative_forecast support;
validation for invalid series, horizons, bounds, fixed parameters, and unsupported exogenous variables;
tests for agreement between forecast_, predict(y), and iterative_forecast(y, 1);
tests that fixed parameters are honoured and optimisation improves over a fixed baseline;
finite forecast checks for constant and short series.

The deliberate limitations of phase 1 are:

no seasonal CES model;
no automatic model selection;
no exogenous variable support;
no prediction intervals;
a single initial_level is currently used to initialise both state components;
the current optimisation bounds are simple box constraints and do not yet enforce the full CES admissibility region.

Phase 2, seasonal and automatic CES

The follow-up phase will extend this into a fuller CES implementation suitable for comparison with the R smooth implementation and for use in SCUM-style forecasting ensembles.

Planned phase 2 work includes:

seasonal CES support;
separate control of real and imaginary initial states if needed;
improved admissibility constraints and/or model-stability handling;
AutoCES-style model selection where appropriate;
stronger reference tests against smooth::ces() / auto.ces() and other available implementations;
benchmark scripts for accuracy and runtime comparisons;
further optimisation of the numba hot paths.


TonyBagnall:
approximate equivalence on phase 1, there are speed optimisations  we can do, but will go back to it on phase 2, which is GO

interesting timing for statsforecast, median seconds lower, but mean higher, some outliers for them

image

TonyBagnall:
# aeon `AutoCES` vs StatsForecast `AutoCES`

tl;dr. it gives same forecasters but is faster for our use case (single series in numpy arrays) 


I ran aeon `AutoCES` against StatsForecast `AutoCES` on the existing 72-series M1/M3 local comparison sample: the first 10 series per type, plus the explicit reference cases **M1:83** and **M3:113**.

Both backends completed all 72 series with finite forecasts.

## Model selection

Model selection matched exactly across all series.

| model | aeon | StatsForecast |
|:------|-----:|--------------:|
| F     |    4 |             4 |
| N     |   37 |            37 |
| P     |   29 |            29 |
| S     |    2 |             2 |

## Accuracy

Accuracy was effectively equivalent. Paired differences are **aeon minus StatsForecast**:

| metric | mean diff | median diff | t-test p | Wilcoxon p |
|:-------|----------:|------------:|---------:|-----------:|
| sMAPE  |  -0.02882 |    -6.00e-7 |   0.1449 |     0.3875 |
| MASE   | -0.000661 |    -4.23e-8 |   0.1820 |     0.0945 |
| MAE    |    -2.733 |    -3.64e-6 |   0.1335 |     0.3124 |
| RMSE   |    -2.951 |    -4.80e-6 |   0.1468 |     0.5862 |

There is no statistically significant paired accuracy difference on this sample.

W/D/L counts (lower is better), using the script's default draw tolerance of relative `1e-6` or absolute `1e-9`:

| metric | aeon W | draw | aeon L |  n |
|:-------|-------:|-----:|-------:|---:|
| sMAPE  |     14 |   49 |      9 | 72 |
| MASE   |     16 |   50 |      6 | 72 |

Forecasts were very close numerically: median relative forecast difference was `4.89e-8`; mean relative forecast difference was `6.60e-4`.

## Timing

Timing favoured aeon.

| backend       | total s | mean s | median s | steady mean s | steady median s |
|:--------------|--------:|-------:|---------:|--------------:|----------------:|
| aeon          |  8.1907 | 0.1138 |  0.00188 |       0.00223 |         0.00188 |
| StatsForecast | 23.6574 | 0.3286 |  0.01032 |       0.01416 |         0.01006 |

Paired runtime ratio, **aeon / StatsForecast**:

|  n | mean ratio | median ratio | mean seconds diff | median seconds diff |
|---:|-----------:|-------------:|------------------:|--------------------:|
| 72 |     0.2969 |       0.1911 |           -0.2148 |            -0.00843 |

Excluding the first cold-start series, aeon remained faster: mean ratio `0.2960`, median ratio `0.1903`. Runtime W/D/L was aeon **W=70, D=0, L=2** over 72 series.

results/dotm_m1_m3_m4_limit25_fastdotm/summary.md
```diff
@@ -0,0 +1,119 @@
+# DOTM M1/M3/M4 comparison
```
chrisholder: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Blocking — unrelated artifacts committed.** This whole `results/dotm_m1_m3_m4_limit25_fastdotm/` directory (15 files: CSVs, 3 PNGs, this `summary.md`) looks like leftover **DOTM** benchmark output from #3455 — it's unrelated to CES, isn't tracked on `main`, and isn't gitignored. Please drop it from the PR:

```
git rm -r results/dotm_m1_m3_m4_limit25_fastdotm/
```

Worth adding `results/` (or wherever the benchmark scripts write) to `.gitignore` so this can't recur.

aeon/forecasting/stats/_ces.py
```diff
@@ -0,0 +1,1580 @@
+"""Complex Exponential Smoothing (CES) forecaster.
+
+Phase 1: non-seasonal CES (model="N").
```
chrisholder: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Should-fix — internal "Phase 1/Phase 2" planning language in shipped docstrings.** This module docstring (and a few attribute docs further down, e.g. "Phase 2 does not optimise the seasonal seed" around line 143, plus several test docstrings) describe the PR's phasing. Once merged there are no phases, and this renders into the API docs/`help()`. Suggest keeping the phase framing in the PR description only and stripping it from the numpydoc.

aeon/forecasting/stats/_ces.py
```diff
@@ -0,0 +1,1580 @@
...
+        """Return testing parameter settings."""
+        return {"alpha_real": 0.5, "alpha_imag": 0.5, "initial_level": 0.0}
+
+
+# ---------------------------------------------------------------------------
+# AutoCES
+# ---------------------------------------------------------------------------
+
+
+class AutoCES(BaseForecaster, IterativeForecastingMixin):
```
chrisholder, line 577: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Nit.** `AutoCES` has no `_get_test_params` (`CES` defines one above). The general estimator checks will instantiate `AutoCES()` and run the full optimiser across candidates; a lightweight test-param set would keep those checks fast and deterministic. Minor, since `AutoCES` can't fix the alphas the way `CES` can.

aeon/forecasting/stats/tests/test_ces.py
```diff
@@ -0,0 +1,497 @@
...
+    """Long-form aliases resolve to their canonical model codes."""
+    y = _synthetic_seasonal()
+    f = CES(model=alias, season_length=4 if expected != "N" else 1).fit(y)
+    assert f.model_ == expected
+
+
+@pytest.mark.parametrize("model", ["N", "S", "P", "F"])
+def test_ces_matches_statsforecast_reference_forecasts(model):
+    """Aeon CES forecasts should broadly match StatsForecast for each model."""
+    statsforecast_models = pytest.importorskip("statsforecast.models")
```
chrisholder: > 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

**Should-fix — the key correctness tests skip in CI.** This reference test and `test_autoces_matches_statsforecast_auto_reference` below are what actually validate the seasonal recurrences and AutoCES selection against the reference implementation, but they're guarded by `importorskip("statsforecast.models")` and `statsforecast` isn't in aeon's deps/test extras — so they **silently skip** in CI. The seasonal/Auto paths then have no executable correctness guard there (only the two hand-calc traces run). Either add `statsforecast` to the test extras, or hard-code the expected reference arrays (as #3455 did for DOTM) so the equivalence check runs without the soft dep.

chrisholder (commented):
> 🤖 *AI-assisted review (Claude Code), posted by @chrisholder — please sanity-check.*

## Summary

Adds `CES` and `AutoCES` to `aeon.forecasting.stats`, implementing the Svetunkov & Kourentzes complex exponential smoothing family with a StatsForecast-compatible state recurrence, bounded adaptive Nelder–Mead optimisation, a backfitting objective, and IC-based model selection. Correct base class, sklearn-compliant `__init__`/`_fit` split, solid validation, `[ENH]` prefix and `forecasting` label correct, API reference updated, and the recurrence has hand-calculated trace tests. The CES code itself looks good — my concerns are mostly hygiene/process.

**Recommendation: Request changes**, primarily for the committed artifacts and the description/scope and CI-coverage gaps below. I've left inline comments on the specific lines.

## Blocking

- **Unrelated benchmark artifacts committed.** 15 files under `results/dotm_m1_m3_m4_limit25_fastdotm/` (DOTM benchmark output from #3455 — CSVs, PNGs, `summary.md`) are unrelated to CES and not gitignored. `git rm -r` the directory and consider adding `results/` to `.gitignore`. (Inline comment on the directory.)

## Should-fix

- **PR description contradicts the delivered code.** The description says "Phase 1, non-seasonal CES" and lists "no seasonal CES model" and "no automatic model selection" as deliberate limitations — but the PR ships all four variants (`N`/`S`/`P`/`F`) **and** `AutoCES`, both tested. The code being more complete than advertised is fine, but the description drives reviewer focus and the changelog; please update it to match (or split out the seasonal/Auto work if that wasn't intended).
- **The core correctness tests skip in CI.** The StatsForecast equivalence tests for the seasonal models and AutoCES are guarded by `importorskip("statsforecast.models")`, and `statsforecast` isn't an aeon dep — so they silently skip and the seasonal/Auto paths have no executable correctness guard in CI (only the hand-calc traces run). Add `statsforecast` to the test extras, or hard-code expected reference arrays as #3455 did. (Inline comment in the test file.)
- **"Phase 1/Phase 2" planning language leaks into shipped docstrings** (module docstring, some attribute docs, test docstrings). Keep it in the PR description only. (Inline comment on the module docstring.)

## Nits

- `AutoCES` has no `_get_test_params`, so the general checks run the full optimiser on it. (Inline comment.)
- `_ces.py` is 1,580 lines; the ~700 lines of numba kernels could live in a sibling `_ces_numba.py`. Consistent with aeon's larger forecasters, so optional.

## Questions

- Is shipping the seasonal models + `AutoCES` here intended, or did the description fall behind? (Determines whether the seasonal kernels need a full line-by-line review in this PR.)
- Can `statsforecast` go into the test extras, or would you prefer pinned reference arrays so the equivalence tests run rather than skip?
- `_predict` re-runs the full backfit on the supplied context using the *training* initial-state seeds — for a seasonal model called on a context that differs from training, does the seasonal phase alignment of those seeds stay correct? (Same-`y` tests pass; asking about the extended-context case.)

*Verification note: I reviewed the diff and the fetched source/tests at the PR head; I did not run the suite locally. CI was green/in-progress when I looked, but the StatsForecast reference tests skip there regardless.*

results/dotm_m1_m3_m4_limit25_fastdotm/summary.md
TonyBagnall: done

aeon/forecasting/stats/tests/test_ces.py
TonyBagnall: done

TonyBagnall:
on questions

Answers:

  1. Yes, shipping seasonal CES models and `AutoCES` is intended in this PR. The PR description was behind the implementation. The intended scope is
  `CES(model="N"/"S"/"P"/"F")` plus `AutoCES`; I’ve removed the internal “Phase 1 / Phase 2” wording from shipped docstrings/tests so the code no longer
  reads like staged planning text.

  2. I’ve switched the StatsForecast equivalence tests to pinned reference arrays rather than adding StatsForecast to test extras. The tests now always
  run in CI and no longer use `pytest.importorskip("statsforecast.models")`. The hard-coded references cover `N`, `S`, `P`, `F`, and `AutoCES` selected
  model/forecast.

  3. For `_predict`, the current assumption is that the supplied context is aligned with the same seasonal origin as the training data. That is fine for
  the local-forecaster pattern used by `iterative_forecast`, including extended contexts created by appending predictions to the original training series.
  It is not phase-safe for an arbitrary out-of-phase context because the method receives only values, not an origin timestamp or phase offset. Supporting
  arbitrary shifted contexts would need an explicit phase/origin contract; for now the behaviour is documented and covered for the same-y / rolling
  extended-context case.

  Small correction on point 3: that phase-alignment answer is for DOTM-style seasonal decomposition. For CES, _predict re-runs the CES state recursion on
  the supplied context using the fitted initial-state seeds; the same practical limitation still applies for arbitrary contexts because seasonal state
  rows are seeded by seasonal position. The intended and tested use is same-origin context, including rolling extension from the training series.

chrisholder (approved):

PR #3464, opened 2026-06-17
[BUG] clone regressor in RegressionForecaster to prevent mutation (fixes #3448)

### LINKED ISSUE #3448: [BUG] RegressionForecaster mutates supplied regressor instead of cloning it
### Describe the bug

RegressionForecaster reuses the regressor object passed by the user instead of cloning it during fit. This means fitting a RegressionForecaster mutates the original estimator object supplied to the constructor. It also means two RegressionForecaster instances constructed with the same regressor object can share fitted state unexpectedly.
This is inconsistent with the usual estimator convention that constructor arguments are treated as unfitted configuration, and fitted estimator instances are stored separately during fit.

```python
import numpy as np
from sklearn.linear_model import Ridge

from aeon.forecasting import RegressionForecaster


reg = Ridge()

y1 = np.arange(20.0)
y2 = np.arange(20.0)[::-1]

f1 = RegressionForecaster(window=5, regressor=reg)
f2 = RegressionForecaster(window=5, regressor=reg)

f1.fit(y1)
coef_after_f1 = reg.coef_.copy()

f2.fit(y2)
coef_after_f2 = reg.coef_.copy()

print(f1.regressor_ is reg)
print(f2.regressor_ is reg)
print(f1.regressor_ is f2.regressor_)
print(coef_after_f1)
print(coef_after_f2)
```

The supplied reg object is fitted directly, and both forecasters keep a reference to the same fitted regressor object.

### Expected results

RegressionForecaster.fit should clone the supplied regressor before fitting it.
The original regressor object passed to the constructor should remain unfitted.
Separate RegressionForecaster instances using the same supplied regressor object should not share fitted state.

### Actual results

A fix is to clone the supplied regressor in _fit, for example:

from aeon.base._base import _clone_estimator

if self.regressor is None:
    self.regressor_ = LinearRegression()
else:
    self.regressor_ = _clone_estimator(self.regressor)

A regression test should check that the original estimator passed to RegressionForecaster(regressor=...) is not fitted after calling fit, and that two forecasters constructed with the same regressor object do not share regressor_.

### Versions

_No response_

gaoflow:
## Description
RegressionForecaster._fit stored the user's regressor object directly in `regressor_` instead of cloning it first. This caused:
1. The original regressor to become fitted as a side-effect of fitting the forecaster
2. Two forecasters constructed with the same regressor to share fitted state

## Fix
Import `_clone_estimator` from `aeon.base._base` and clone the supplied regressor before fitting, matching the convention used elsewhere in the codebase (e.g. forecasting/base.py, classifiers).

Closes #3448

This pull request includes code written with the assistance of AI under my direction.

### DIFF
--- a/aeon/forecasting/_regression.py
+++ b/aeon/forecasting/_regression.py
@@ -8,6 +8,7 @@ to form ``X`` and trains to predict the next ``horizon`` points ahead.
 import numpy as np
 from sklearn.linear_model import LinearRegression
 
+from aeon.base._base import _clone_estimator
 from aeon.forecasting.base import (
     BaseForecaster,
     DirectForecastingMixin,
@@ -75,7 +76,7 @@ class RegressionForecaster(
         if self.regressor is None:
             self.regressor_ = LinearRegression()
         else:
-            self.regressor_ = self.regressor
+            self.regressor_ = _clone_estimator(self.regressor)
         self._n_exog = 0
         # Combine y and exog for windowing
         if exog is not None:
--- a/aeon/forecasting/tests/test_regressor.py
+++ b/aeon/forecasting/tests/test_regressor.py
@@ -118,6 +118,7 @@ def test_regressor_cloned_not_mutated():
     """Test RegressionForecaster clones the regressor instead of mutating it."""
     import numpy as np
     from sklearn.linear_model import Ridge
+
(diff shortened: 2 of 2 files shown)


TonyBagnall (approved):
LGTM, thanks

PR #3465, opened 2026-06-17
[BUG] use len(X) instead of X.shape[0] for list input in ShapeletTransformClassifier (fixes #3447)

### LINKED ISSUE #3447: [BUG] ShapeletTransformClassifier.predict_proba crashes on unequal-length list input
### Describe the bug

ShapeletTransformClassifier.predict_proba raises AttributeError: 'list' object has no attribute 'shape' when the underlying estimator lacks a predict_proba method and the input is a list of  unequal-length series.

  The classifier declares "capability:unequal_length": True and "X_inner_type": ["np-list", "numpy3D"], so this input is valid. The bug is identical to the one fixed for Catch22Classifier and TSFreshClassifier in #3446: the fallback code path in _predict_proba uses X.shape[0], which is invalid for Python lists.

 Reproduction

```python
  import numpy as np
  from sklearn.svm import LinearSVC
  from aeon.classification.shapelet_based import ShapeletTransformClassifier

  X = [np.random.RandomState(i).rand(1, 8 + i) for i in range(6)]
  y = np.array([0, 1, 0, 1, 0, 1])

  clf = ShapeletTransformClassifier(estimator=LinearSVC(), random_state=0)
  clf.fit(X, y)
  clf.predict_proba(X[:2])  # AttributeError: 'list' object has no attribute 'shape'

  Root cause

  aeon/classification/shapelet_based/_stc.py, lines 249 and 251:

  proba = np.zeros((X.shape[0], self.n_classes_))
  preds = self._estimator.predict(X_t)
  for i in range(0, X.shape[0]):
```

### Expected results

should run.

### Actual results

Fix is replace both X.shape[0] with len(X), matching the fix applied in #3446.

### Versions

_No response_

gaoflow:
## Description
`ShapeletTransformClassifier.predict_proba` crashed with `AttributeError: 'list' object has no attribute 'shape'` when:
- X is a list of unequal-length series
- The underlying estimator lacks a `predict_proba` method (falling back to the `predict`-based code path)

## Root Cause
The fallback code path in `_predict_proba` used `X.shape[0]`, which is invalid for Python lists.

## Fix
Replace both `X.shape[0]` occurrences with `len(X)`, matching the same fix applied for Catch22Classifier and TSFreshClassifier in #3446.

Closes #3447

This pull request includes code written with the assistance of AI under my direction.

### DIFF
--- a/aeon/classification/shapelet_based/_stc.py
+++ b/aeon/classification/shapelet_based/_stc.py
@@ -273,8 +273,7 @@ class ShapeletTransformClassifier(BaseClassifier):
         ):
             if self.verbose:
                 print(  # noqa: T201
-                    "Fitting estimator and generating train set estimates "
-                    "(RotF OOB)..."
+                    "Fitting estimator and generating train set estimates (RotF OOB)..."
                 )
 
             proba = self._estimator.fit_predict_proba(X_t, y)
--- a/aeon/classification/shapelet_based/_stc.py
+++ b/aeon/classification/shapelet_based/_stc.py
@@ -246,9 +246,9 @@ class ShapeletTransformClassifier(BaseClassifier):
         if callable(m):
             proba = self._estimator.predict_proba(X_t)
         else:
-            proba = np.zeros((X.shape[0], self.n_classes_))
+            proba = np.zeros((len(X), self.n_classes_))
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
LGTM, thanks

PR #3467, opened 2026-06-18
[BUG] remap CLARA init indices from full-dataset to sample coordinates (fixes #3423)

### LINKED ISSUE #3423: [BUG] TimeSeriesCLARA passes full-dataset init indices to sampled PAM fit

gaoflow:
When `TimeSeriesCLARA` receives `init` as an ndarray of indices and
`n_samples < n_cases`, the indices were passed directly to
`TimeSeriesKMedoids`, which validates them against the sampled subset
`X[sample_idxs]`.  If `init` contained any index ≥ `n_samples`, the
init-validity check in PAM raised a `ValueError`.

**Fix**: when `init` is an array, reserve those indices in the sample so
they are always present, then remap them to sample-relative positions
before passing to PAM.  When `init` is a string (random, first, etc.)
the sampling behaviour is unchanged.

Closes #3423.

---

### DIFF
--- a/aeon/clustering/_clara.py
+++ b/aeon/clustering/_clara.py
@@ -167,18 +167,39 @@ class TimeSeriesCLARA(BaseClusterer):
         for _ in range(self.n_sampling_iters):
 
             if n_samples < n_cases:
-                sample_idxs = self._random_state.choice(
-                    np.arange(n_cases),
-                    size=n_samples,
-                    replace=False,
-                )
-
+                n_init_indices = len(self.init) if isinstance(self.init, np.ndarray) else 0
+                if n_init_indices > 0:
+                    init_indices = self.init
+                    n_random = n_samples - n_init_indices
+                    remaining = np.setdiff1d(np.arange(n_cases), init_indices)
+                    if n_random > 0:
+                        random_sample = self._random_state.choice(
+                            remaining, size=n_random, replace=False
+                        )
+                        sample_idxs = np.concatenate([init_indices, random_sample])
+                    else:
+                        sample_idxs = init_indices
+                    self._random_state.shuffle(sample_idxs)
+                else:
+                    sample_idxs = self._random_state.choice(
+                        np.arange(n_cases),
+                        size=n_samples,
+                        replace=False,
+                    )
             else:
                 sample_idxs = np.arange(n_cases)
 
+            # Remap init indices from full-dataset space to sample space
+            if isinstance(self.init, np.ndarray):
+                sample_init = np.array(
+                    [np.where(sample_idxs == i)[0][0] for i in self.init]
+                )
+            else:
+                sample_init = self.init
+
             pam = TimeSeriesKMedoids(
                 n_clusters=self.n_clusters,
-                init=self.init,
+                init=sample_init,
                 distance=self.distance,
                 n_init=self.n_init,
                 max_iter=self.max_iter,
--- a/aeon/clustering/_clara.py
+++ b/aeon/clustering/_clara.py
@@ -9,6 +9,7 @@ import numpy as np
 from numpy.random import RandomState
 from sklearn.utils import check_random_state
 
+from aeon.clustering._cluster_initialisation import resolve_center_initialiser
 from aeon.clustering._k_medoids import TimeSeriesKMedoids
 from aeon.clustering.base import BaseClusterer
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
Thanks, this is the right direction and implements the more useful of the two behaviours described in #3423: treat ndarray init as full-dataset indices, force those cases into each CLARA sample, and remap them to sample-relative coordinates before passing them to PAM.

I think there is one robustness issue to fix before merge. The PR currently uses `self.init` as an index array before validating it. For example, out-of-range, non-integer, or wrongly shaped ndarray init values can now fail at `X[sample_idxs]` or during the CLARA sampling logic, rather than going through the existing `resolve_center_initialiser` validation used by `TimeSeriesKMedoids`.

Could you validate ndarray init once near the start of `_fit`, before the sampling loop, ideally using the existing `resolve_center_initialiser(..., use_indexes=True)` path, then use the validated array for the reserve-and-remap logic? That should preserve the current aeon-style `ValueError`s for invalid init arrays.

It would also be good to add one or two negative tests, for example out-of-range init and non-integer init, to ensure invalid array initialisation still raises a clean `ValueError`.

Minor doc point: the CLARA `init` docstring should say explicitly that ndarray init values are interpreted as full-dataset indices and are remapped to sample-local indices when CLARA subsamples.

binggao1230:
Thanks for the review. All three points addressed in [be8044a](https://github.com/gaoflow/aeon/commit/be8044a66ab35dc8ffdd17e10a5cf1673b859d71):

- **Validation**: `resolve_center_initialiser(use_indexes=True)` is now called once near the start of `_fit`, before the sampling loop. Out-of-range, non-integer, and wrongly-shaped init arrays raise the same `ValueError`s as `TimeSeriesKMedoids` instead of failing at `X[sample_idxs]`.
- **Negative tests**: three parametrized cases added — negative index, out-of-range index, and non-integer array — all verified to raise `ValueError`.
- **Docstring**: the `init` parameter now explicitly says ndarray values are full-dataset indices that get remapped to sample-local coordinates.

binggao1230:
Superseded by #3459 which was merged while this PR was under review. Upstream decided to reject ndarray init for CLARA rather than support it with remapping. Closing in favor of the upstream approach.

PR #3473, opened 2026-06-18
[BUG] use actual shapelet transform in LearningShapeletClassifier (fixes #3380)

### LINKED ISSUE #3380: [BUG] tslearn warpper bug when save_transformed_data is True
### Describe the bug

In _fit, when save_transformed_data=True, we save X_t into self.transformed_data_, but X_t is just the tslearn-formatted input, not the shapelet transform. 

Then get_transform(X) ignores its X argument and returns self.transformed_data_, and get_locations(X) ignores its X argument and calls self.clf_.locate(self.transformed_data_). So we are not actually saving the transform, and we are not actually using the X passed to those methods.

this is an addition to #3379 where it crashes with version 0.8.0

```python
import numpy as np

from aeon.classification.shapelet_based import LearningShapeletClassifier

X_train = make_example_3d_numpy(
    n_cases=10, n_channels=1, n_timepoints=20, random_state=0, return_y=False
)
X_test = make_example_3d_numpy(
    n_cases=3, n_channels=1, n_timepoints=20, random_state=1, return_y=False
)
y_train = np.array([0, 1] * 5)

clf = LearningShapeletClassifier(
    max_iter=10, total_lengths=1, save_transformed_data=True, random_state=0
)
clf.fit(X_train, y_train)

# Bug 1: get_transform(X) ignores X and returns cached training data
Xt_test = clf.get_transform(X_test)
assert np.array_equal(Xt_test, clf.transformed_data_)

# Bug 2: returned object has training-set case count, not test-set case count
assert Xt_test.shape[0] == X_train.shape[0]
assert Xt_test.shape[0] != X_test.shape[0]

# Bug 3: get_locations(X) ignores X as well
loc_train = clf.get_locations(X_train)
loc_test = clf.get_locations(X_test)
assert np.array_equal(loc_train, loc_test)
```

### Expected results

save transform not X

### Actual results

saves X not transform

### Versions

_No response_

gaoflow:
In `_fit`, `save_transformed_data=True` stored the tslearn-formatted input `X_t` instead of the actual shapelet transform. `get_transform(X)` and `get_locations(X)` both ignored their `X` argument — returning or operating on stored training data rather than transforming the given `X`.

- Store the real transform (`clf_.transform(X_t)`) in `transformed_data_`
- `get_transform(X)` now calls `clf_.transform()` on the given `X`
- `get_locations(X)` now calls `clf_.locate()` on the given `X`

Fixes #3380

### DIFF
--- a/aeon/classification/shapelet_based/_ls.py
+++ b/aeon/classification/shapelet_based/_ls.py
@@ -147,7 +147,7 @@ class LearningShapeletClassifier(BaseClassifier):
         X_t = _X_transformed_tslearn(X)
         self.clf_.fit(X_t, y)
         if self.save_transformed_data:
-            self.transformed_data_ = X_t
+            self.transformed_data_ = self.clf_.transform(X_t)
 
         return self
 
@@ -181,7 +181,8 @@ class LearningShapeletClassifier(BaseClassifier):
                 "Set save_transformed_data=True in the constructor to save the "
                 "transformed data for later use"
             )
-        return self.transformed_data_
+        X_t = _X_transformed_tslearn(X)
+        return self.clf_.transform(X_t)
 
     def get_locations(self, X):
         """Compute shapelet match location for a set of time series.
@@ -205,7 +206,8 @@ class LearningShapeletClassifier(BaseClassifier):
                 "Set save_transformed_data=True in the constructor to save the "
                 "transformed data for later use"
             )
-        return self.clf_.locate(self.transformed_data_)
+        X_t = _X_transformed_tslearn(X)
+        return self.clf_.locate(X_t)
 
     @classmethod
     def _get_test_params(cls, parameter_set: str = "default") -> dict | list[dict]:


MatthewMiddlehurst:
Please use the requested title formatting.

I would probably just deprecate the class really. If not that, these functions and parameter at least. It is just a pure wrapper and not one anyone I know uses.

Thoughts @TonyBagnall ?

TonyBagnall:
this fix is correct, but I agree its probably better to just remove the classifier, it performs really badly and there are compatibility issues with tslearn. Had claude run this for checking correctness and it identified that the real `LearningShapelets.fit` currently can't run under keras 3 (an upstream tslearn `broadcast_to` failure, unrelated to this PR).

So I agree to deprecate LearningShapeletClassifier, @gaoflow thanks for this and apologies, feel free to make the deprecation after we have created the issue :)

TonyBagnall:
we are deprecating this class

PR #3477, opened 2026-06-19
[BUG] Force ETS to fit seasonal when given period,

### LINKED ISSUE #3476: [BUG] test fail on new ECS version
### Describe the bug

test fail here
https://github.com/aeon-toolkit/aeon/actions/runs/27809127323/job/82315589226?pr=3383

run CI

### Expected results

pass

### Actual results

FAILED aeon/forecasting/stats/tests/test_ets.py::test_autoets_uses_provided_seasonal_period - assert 0 != 0
 +  where 0 = AutoETS(seasonal_period=4).seasonality_type_

### Versions

_No response_

TonyBagnall:
fixes #3476

In the failing case, AutoETS(seasonal_period=4) still allowed the automatic search to return a non-seasonal model with seasonality_type_ == 0. That made the test fail here:

  assert forecaster.seasonality_type_ != 0

The intended contract of this test is: if the user explicitly supplies a valid seasonal period and the series is long enough for seasonal ETS candidates, AutoETS should use that period in a seasonal model rather than silently falling back to no seasonality.

The fix keeps the normal AIC search intact, but also tracks the best seasonal candidate separately. When seasonal_period was explicitly provided and at least one finite seasonal fit exists, AutoETS returns the best seasonal model. If the seasonal period is inferred automatically, the existing behaviour is unchanged.

This also avoids a fragile CI-dependent result where optimizer/AIC differences could make the selected model non-seasonal even for a deliberately seasonal test series.

### DIFF
--- a/aeon/forecasting/stats/_ets.py
+++ b/aeon/forecasting/stats/_ets.py
@@ -429,6 +429,7 @@ class AutoETS(BaseForecaster):
         data = np.asarray(y.squeeze(), dtype=np.float64)
         if self.seasonal_period is None:
             seasonal_period = 0
+            require_seasonality = False
         elif self.seasonal_period < 1:
             raise ValueError(
                 f"seasonal_period must be a positive integer or None, "
@@ -436,10 +437,12 @@ class AutoETS(BaseForecaster):
             )
         else:
             seasonal_period = self.seasonal_period
+            require_seasonality = True
         best_model = auto_ets(
             data,
             self.allow_multiplicative_trend,
             seasonal_period,
+            require_seasonality,
         )
         self.error_type_ = int(best_model[0])
         self.trend_type_ = int(best_model[1])
@@ -501,10 +504,16 @@ class AutoETS(BaseForecaster):
 
 
 @njit(fastmath=True, cache=True)
-def auto_ets(data, allow_multiplicative_trend=False, seasonal_period=0):
+def auto_ets(
+    data,
+    allow_multiplicative_trend=False,
+    seasonal_period=0,
+    require_seasonality=False,
+):
     """Calculate model parameters based on the internal nelder-mead implementation."""
     if seasonal_period < 1:
         seasonal_period = calc_seasonal_period(data)
+        require_seasonality = False
     # Technically only needs to be 2 * seasonal periods to calculate initial conditions,
     # but makes no sense to run a seasonal model with any less than 2 seasonal periods
     # worth of usable data even that might be a bit low
@@ -517,7 +526,17 @@ def auto_ets(data, allow_multiplicative_trend=False, seasonal_period=0):
             break
     model = np.empty(4, dtype=np.int32)
     best_model = np.empty(4, dtype=np.int32)
+    best_model[0] = 1
+    best_model[1] = 0
+    best_model[2] = 0
+    best_model[3] = 1
+    best_seasonal_model = np.empty(4, dtype=np.int32)
+    best_seasonal_model[0] = 1
+    best_seasonal_model[1] = 0
+    best_seasonal_model[2] = 1
+    best_seasonal_model[3] = seasonal_period
     best_aic = np.inf
+    best_seasonal_aic = np.inf
     t_max = 3 if allow_multiplicative_trend else 2
     for error_type in range(1, 3):
         if error_type == 2 and not all_pos:
@@ -538,4 +557,9 @@ def auto_ets(data, allow_multiplicative_trend=False, seasonal_period=0):
                 if aic < best_aic:
                     best_aic = aic
                     best_model[:] = model
+                if seasonality_type != 0 and aic < best_seasonal_aic:
+                    best_seasonal_aic = aic
+                    best_seasonal_model[:] = model
+    if require_seasonality and seasonal_enabled and best_seasonal_aic < np.inf:
+        return best_seasonal_model
     return best_model
--- a/aeon/forecasting/stats/_ets.py
+++ b/aeon/forecasting/stats/_ets.py
@@ -531,10 +531,6 @@ def auto_ets(
     best_model[2] = 0
     best_model[3] = 1
     best_seasonal_model = np.empty(4, dtype=np.int32)
-    best_seasonal_model[0] = 1
-    best_seasonal_model[1] = 0
-    best_seasonal_model[2] = 1
-    best_seasonal_model[3] = seasonal_period
     best_aic = np.inf
     best_seasonal_aic = np.inf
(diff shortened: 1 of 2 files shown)


aeon/forecasting/stats/_ets.py
```diff
@@ -538,4 +557,9 @@ def auto_ets(data, allow_multiplicative_trend=False, seasonal_period=0):
                 if aic < best_aic:
                     best_aic = aic
                     best_model[:] = model
+                if seasonality_type != 0 and aic < best_seasonal_aic:
+                    best_seasonal_aic = aic
+                    best_seasonal_model[:] = model
+    if require_seasonality and seasonal_enabled and best_seasonal_aic < np.inf:
```
chrisholder, line 559: The new override behaviour isn't deterministically tested. The existing `test_autoets_uses_provided_seasonal_period` uses a strongly seasonal series (`np.tile([10,30,12,28], 8)`) where the seasonal model likely wins on AIC anyway — so that test would pass *even without this fix* and doesn't exercise this guard. Suggest adding a regression test with a series where non-seasonal wins on AIC but a period is supplied (e.g. `np.linspace(10,30,80) + small noise`, `seasonal_period=4`) and assert `seasonality_type_ != 0`, to pin the actual fix.

🤖 _AI-generated review comment, posted on the maintainer's behalf._

aeon/forecasting/stats/_ets.py
```diff
@@ -429,17 +429,20 @@ def _fit(self, y, exog=None):
...
             seasonal_period = 0
+            require_seasonality = False
         elif self.seasonal_period < 1:
             raise ValueError(
                 f"seasonal_period must be a positive integer or None, "
                 f"but saw {self.seasonal_period}"
             )
         else:
             seasonal_period = self.seasonal_period
+            require_seasonality = True
```
chrisholder, line 440: This is where the new contract originates: an explicit `seasonal_period` now *forces* a seasonal model (when the series is long enough), overriding pure-AIC selection. The `AutoETS` `seasonal_period` parameter docstring (~line 371, unchanged in this PR) should be updated to state this, otherwise the behaviour change is invisible to users.

🤖 _AI-generated review comment, posted on the maintainer's behalf._

aeon/forecasting/stats/_ets.py
```diff
@@ -517,7 +526,17 @@ def auto_ets(data, allow_multiplicative_trend=False, seasonal_period=0):
             break
     model = np.empty(4, dtype=np.int32)
     best_model = np.empty(4, dtype=np.int32)
+    best_model[0] = 1
+    best_model[1] = 0
+    best_model[2] = 0
+    best_model[3] = 1
+    best_seasonal_model = np.empty(4, dtype=np.int32)
+    best_seasonal_model[0] = 1
```
chrisholder: Minor: these `best_seasonal_model` default values are effectively dead — they're never returned, because the final guard requires `best_seasonal_aic < np.inf`, which only becomes true after `best_seasonal_model[:] = model` has overwritten them. Harmless/defensive (njit wants an initialised array), but could be dropped or given a one-line comment noting that.

🤖 _AI-generated review comment, posted on the maintainer's behalf._

chrisholder (commented):
Thanks — clean, minimal fix for the flaky seasonal selection in #3476. I ran it locally: the 26 ETS tests pass, and I confirmed the change has teeth — on a series where AIC alone picks a non-seasonal model `(1,1,0,1)`, `auto_ets(..., require_seasonality=True)` correctly returns seasonal `(1,1,1,4)`, so it's not a no-op. The inferred-period path (`seasonal_period=None`) is untouched, so it's backward-compatible for the default case.

Nice side-effect: initializing `best_model[:] = (1,0,0,1)` also closes a latent path where, if no candidate produced a finite AIC, `best_model` was returned from an uninitialised `np.empty` (garbage). Good catch.

No blockers — overall **approve with nits**. Main asks are a deterministic regression test and a docstring update (inline). Two design questions:

- **A.** Is "explicit `seasonal_period` ⇒ force a seasonal model" the intended *permanent* contract for `AutoETS`, or just to stabilise the test? It's a real behaviour change — users who passed `seasonal_period` expecting AIC to remain free to drop seasonality will now get different models. Fine for an experimental module, just want it to be deliberate + documented.
- **B.** Short-series case: if a period is supplied but `len(data) <= seasonal_period * 4`, `seasonal_enabled` is `False`, so the guard is skipped and a non-seasonal model is returned **silently** despite `require_seasonality`. Intended? If so, should it warn that the requested seasonal period was dropped for lack of data?

🤖 _AI-generated review, posted on the maintainer's behalf._

chrisholder (dismissed):

aeon/forecasting/stats/_ets.py
TonyBagnall: your AI is working my AI hard

TonyBagnall: he dead initialisation is removed.

TonyBagnall:
Thanks, agreed with both nits.

  A. Yes, this is intended as the permanent contract for AutoETS(seasonal_period=...), not only a test stabilisation. The motivation is that an explicit seasonal period should be treated as a
  user request for seasonal modelling when there is enough data to estimate seasonal states. If users want pure AIC to decide whether seasonality is present, they can leave
  seasonal_period=None. I’ve updated the parameter docstring to make the behaviour visible.

  B. The short-series fallback is intended. If len(data) <= seasonal_period * 4, seasonal ETS candidates are not considered because the seasonal initialisation is too weak for reliable fitting,
  so AutoETS falls back to non-seasonal candidates. I would avoid warning here for now: this is automatic model selection and short M-style series would make warnings noisy. The docstring now
  states the fallback explicitly. If reviewers prefer, we could add a warning later, but I’d keep this PR focused on deterministic behaviour and documentation.

  I also removed the dead best_seasonal_model defaults and kept the regression test deterministic.

aeon/forecasting/stats/_ets.py
TonyBagnall: I added a deterministic regression test:

  test_autoets_provided_seasonal_period_overrides_nonseasonal_aic_winner

  It uses:

  y = np.linspace(10.0, 30.0, 80) + 0.01 * np.sin(np.arange(80))

  and asserts:

  - AutoETS() selects non-seasonal: seasonality_type_ == 0
  - AutoETS(seasonal_period=4) selects seasonal: seasonality_type_ != 0
  - the forced seasonal period is retained: seasonal_period_ == 4

chrisholder (approved):

PR #3482, opened 2026-06-20
[BUG] Fixes AutoETS.iterative_forecast 

TonyBagnall:
Summary

  Fixes AutoETS.iterative_forecast so it follows the same contract as the other iterative forecasters: an unfitted estimator can be given y and prediction_horizon, it fits once on y, then  returns the full forecast path.

Previously, AutoETS.iterative_forecast delegated directly to self.wrapped_model_, which only exists after fit. Calling AutoETS().iterative_forecast(y, h) therefore failed with an attribute error, and components such as EnsembleForecaster(iterative_strategy="component") could not use AutoETS directly.

  Changes:

  - Make AutoETS inherit IterativeForecastingMixin.
  - Validate iterative forecast inputs through the shared mixin helper.
  - Fit AutoETS once inside iterative_forecast.
  - Add an internal ETS._iterative_forecast_from_fitted helper so AutoETS can forecast from the selected fitted ETS state without re-entering the public fit-on-call path.
  - Update the AutoETS test to cover calling iterative_forecast from an unfitted instance.

  Validation:

  - pytest aeon\forecasting\stats\tests\test_ets.py -q passed.
  - check_estimator(AutoETS) passed 19/19 checks.
  - Manual smoke check confirmed AutoETS().iterative_forecast(y, 3) works.


hadifawaz1999 (approved):
LGTM !

PR #3483, opened 2026-06-20
[ENH] Soft distance sub-package and signed soft-DTW

chrisholder:
### Reference Issues/PRs

First of a planned stack of PRs adding soft elastic distances and gradient-based soft barycentre averaging. This base PR establishes the `soft/` sub-package and makes soft-DTW correct; soft-MSM, the soft barycentre averaging, and a `NearestCentroidClassifier` follow in stacked PRs.

**Restructure.** Creates `aeon.distances.elastic.soft` as the home for soft elastic distances and moves soft-DTW into it (`elastic/soft/_soft_dtw.py`), with a shared `_utils` softmin helper.

**Correctness fix (user-visible).** `soft_dtw_distance` now returns the raw **signed** soft-DTW value instead of `abs(...)`. The previous `abs` was incorrect: soft-DTW is not a proper distance and is legitimately negative for larger `gamma`. The signed value now matches the `tslearn` reference **exactly** for all `gamma` (including the negative regime), where the old `abs` version disagreed.

**Distance contract.** Adds a `proper_distance` flag (default `True`) to the distance registry; soft-DTW sets it `False`. A derived `NON_METRIC_DISTANCES` list now gates the non-negativity, zero-diagonal and sklearn-compat test skips, instead of hardcoding `"soft_dtw"` in each test. Future soft/non-metric distances inherit the correct test behaviour automatically.

**Surface.** Soft-DTW's gradient stays private (`_soft_dtw_grad_x`, used by the upcoming soft barycentre averaging). The unused public `soft_dtw_grad_x`, `soft_dtw_alignment_matrix`, and the unread registry `"gradient"` key are not exposed.

No. (`tslearn`, already an optional aeon dependency, is used only as a guarded test oracle.)

### Any other comments?

Behaviour change to a public function (`soft_dtw_distance` output sign) is deliberate and validated against `tslearn`. The `soft_dtw` entry in the `test_kasba` "barycenter does not update" skip list is temporary — the follow-up averaging PR removes soft distances from the discrete BA methods entirely and deletes that skip.

### Test plan

- New `test_soft_dtw.py`: `tslearn` oracle for the signed distance (guarded with `importorskip`) + a finite-difference check of the private gradient.
- Updated soft-DTW expected results to signed values.
- Verified locally: `test_distances`, `test_cost_matrix`, `test_pairwise`, `test_sklearn_compatibility`, `test_kasba` / `test_petitjean_ba` / `test_subgradient`, the new soft-DTW tests, soft-DTW doctests, and a full `import aeon`. Pre-commit passes.


TonyBagnall:
I thought there was an issue for this, but seems I didnt post it

Correctness fix (user-visible). soft_dtw_distance now returns the raw signed soft-DTW value instead of abs(...). The previous abs was incorrect: soft-DTW is not a proper distance and is legitimately negative for larger gamma. The signed value now matches the tslearn reference exactly for all gamma (including the negative regime), where the old abs version disagreed.

TonyBagnall:
not sure why the gradient is private, I think it would be better to expose it, since it is the selling point of sfot

aeon/distances/elastic/soft/_soft_dtw.py
```diff
@@ -86,15 +56,16 @@ def soft_dtw_distance(
...
-        is used.
     itakura_max_slope : float, default=None
         Maximum slope as a proportion of the number of time points used to create
         Itakura parallelogram on the bounding matrix. Must be between 0. and 1.
 
     Returns
     -------
     float
-        soft-DTW distance between x and y, minimum value 0.
+        soft-DTW distance between x and y. soft-DTW is not a proper distance: the
```
TonyBagnall, line 66: I prefer metric, or formal distance, "proper distance" is a bit vague

aeon/distances/elastic/soft/_soft_dtw.py
```diff
@@ -180,16 +150,46 @@ def soft_dtw_cost_matrix(
...
-           [ 30.,  14.,   5.,   1.,   0.,   1.,   5.,  14.,  30.,  55.],
-           [ 55.,  30.,  14.,   5.,   1.,   0.,   1.,   5.,  14.,  30.],
-           [ 91.,  55.,  30.,  14.,   5.,   1.,   0.,   1.,   5.,  14.],
-           [140.,  91.,  55.,  30.,  14.,   5.,   1.,   0.,   1.,   5.],
-           [204., 140.,  91.,  55.,  30.,  14.,   5.,   1.,   0.,   1.],
-           [285., 204., 140.,  91.,  55.,  30.,  14.,   5.,   1.,   0.]])
+    array([[ 0.00000000e+00,  1.00000000e+00,  5.00000000e+00,
+         1.40000000e+01,  3.00000000e+01,  5.50000000e+01,
+         9.10000000e+01,  1.40000000e+02,  2.04000000e+02,
+         2.85000000e+02],
```
TonyBagnall, line 156: this is too big and poorly formatted, round it?

aeon/distances/elastic/soft/_soft_dtw.py
```diff
@@ -210,18 +210,20 @@ def soft_dtw_cost_matrix(
...
 
 
 @njit(cache=True, fastmath=True)
 def _soft_dtw_cost_matrix(
     x: np.ndarray, y: np.ndarray, bounding_matrix: np.ndarray, gamma: float
 ) -> np.ndarray:
-    if gamma == 0.0 or np.array_equal(x, y):
+    # When gamma == 0 the soft-min collapses to a hard-min, so soft-DTW reduces
+    # to DTW. Delegating avoids the 1/gamma division by zero in ``_softmin3``.
+    # The original ``np.array_equal(x, y)`` short-circuit is intentionally NOT
```
TonyBagnall, line 224: could you make this a function docstring?

TonyBagnall:
It would be nice to have a direct comparison to tslearn implementation for correctness and efficiency

TonyBagnall:
> It would be nice to have a direct comparison to tslearn implementation for correctness and efficiency

aeon's signed soft_dtw_distance matches tslearn.metrics.soft_dtw to ~3e-14 for both univariate and 4-channel multivariate, across γ ∈ {0.1, 1, 3, 10} including the negative regime.

TonyBagnall:
# soft-DTW: aeon (PR #3483) vs tslearn — equivalence and runtime

Benchmark of the signed soft-DTW implementation in PR #3483
(`aeon-toolkit/aeon`, branch `soft-dtw-refactor`) against the `tslearn`
reference. Single-pair distance, `gamma=1.0`, univariate.

## Value equivalence

`aeon.distances.soft_dtw_distance` matches `tslearn.metrics.soft_dtw` to within
floating-point tolerance for both univariate and multivariate (4-channel)
inputs, across `gamma` in {0.1, 1.0, 3.0, 10.0} including the negative regime:

- univariate: `max |aeon - tslearn| = 2.8e-14`
- multivariate (4ch): `max |aeon - tslearn| = 2.8e-14`

This is the signed value; the previous `abs(...)` disagreed with tslearn
wherever the true soft-DTW value was negative.

## Runtime and memory

Time excludes JIT compilation (numba compiles per type signature, not per
length, so a single warmup covers all sizes). Rows up to n=800 are median of
50 calls; rows from n=4,000 are few-shot steady-state. Memory is the analytical
working set: aeon allocates one `(n+1) x (n+1)` float64 cost matrix and computes
squared distances inline; tslearn allocates two large matrices (a distance
matrix `D` of shape `(m, n)` plus an accumulated cost matrix `R` of shape
`(m+2, n+2)`), hence ~2x.

| n | aeon (ms) | tslearn (ms) | speedup | aeon mem (MB) | tslearn mem (MB) |
|------:|----------:|-------------:|--------:|--------------:|-----------------:|
| 50 | 0.12 | 1.09 | 9.3x | 0.02 | 0.04 |
| 100 | 0.49 | 1.85 | 3.8x | 0.08 | 0.16 |
| 200 | 1.96 | 4.18 | 2.1x | 0.32 | 0.65 |
| 400 | 7.88 | 13.99 | 1.8x | 1.29 | 2.57 |
| 800 | 31.94 | 48.72 | 1.5x | 5.13 | 10.27 |
| 4,000 | 870 | 1,090 | 1.3x | 128 | 256 |
| 8,000 | 3,510 | 4,230 | 1.2x | 512 | 1,024 |
| 12,000 | 7,900 | 9,480 | 1.2x | 1,152 | 2,304 |
| 15,000 | 12,220 | OOM | — | 1,800 | 3,600 |
| 18,000 | 20,620 | OOM | — | 2,592 | 5,185 |

## Notes

- **Memory model is empirically anchored.** The predicted tslearn footprint
  crosses the ~3.7 GB sandbox ceiling between n=12,000 (2.3 GB) and n=15,000
  (3.6 GB), which is exactly where tslearn began to OOM. aeon's predicted sizes
  match the measured cost-matrix sizes.
- **Speedup declines monotonically** (9.3x to ~1.2x). aeon's advantage is
  mostly lower per-call overhead, which dominates at short lengths; at scale
  both converge on the shared O(n^2) inner loop, leaving aeon ~1.2x ahead.
- **Scaling is clean O(n^2)** (n=4,000 -> 8,000 is 0.87 s -> 3.51 s = 4.0x).
- **A 30 s single call was not reachable in this environment** (~3.7 GB RAM).
  It lands at n ~= 21,700 for aeon (~3.8 GB cost matrix); tslearn would need
  ~7.6 GB at that size. Runnable on a box with >=8 GB (>=16 GB to keep tslearn
  alive for the comparison).

Generated against aeon 1.4.0, PR #3483 head (`0463d48`).

TonyBagnall:
there are now three property flags (symmetric, unequal_support, non_negative). The first two are defined true or false for all distances, but non_negative is not. For consistency, they should all be treated in the same way, I prefer populating non_negative for the other 22.

TonyBagnall:
also good to have the derivation not rely on the default at all once every row is explicit — i.e. if not d["non_negative"] — which will then KeyError if someone adds a new distance and forgets the flag, turning "silently assumed non-negative" into "you must declare it."  it forces the contract on new contributors.

TonyBagnall (changes requested):
all looks good, a few minor things but definitely improves the module

TonyBagnall:
@chrisholder any chance of revisiting this?

Issue #3481, opened 2026-06-20
[ENH] Deprecate LearningShapeletClassifier wrapper from tslearn


TonyBagnall:
### Describe the feature or idea you want to propose

there are some issues with this classifier
https://github.com/aeon-toolkit/aeon/issues/3379
https://github.com/aeon-toolkit/aeon/issues/3380

and some problems with the latest tensorflow

### Describe your proposed solution

deprecate for 1.5, remove 1.6

PR #3441, opened 2026-06-12
[DEP] deprecation for 1.5 see #3440

TonyBagnall:
closes #3440

PR #3485, opened 2026-06-20
[ENH] Deprecates LearningShapeletClassifier

### LINKED ISSUE #3380: [BUG] tslearn warpper bug when save_transformed_data is True

TonyBagnall:
see #3481 

Note this reintroduces the dependency on Deprecated>=1.2.13 which was mistakenly removed in #3441 (because there were no deprecations left in aeon). 

closes #3379 and closes #3380 as no longer worth doing


dguijo (approved):
LGTM, @TonyBagnall, thank you.

PR #3492, opened 2026-06-20
[BUG] Stratify TSelect validation split

### LINKED ISSUE #3488: [BUG] TSelect should use a stratified train/validation split for class labels
### Problem

`TSelect` builds a train/validation split using only case indices:

```python
train_idx, valid_idx = train_test_split(
    list(range(n_cases)),
    test_size=test_size,
    random_state=self.random_state,
)
```
The split is not stratified by y. This can make TSelect.fit fail on small, imbalanced, or class-ordered datasets because either the training fold or validation fold may be missing one or more classes.

This is particularly likely for archive-style datasets where cases may be ordered by class. The downstream AUC calculation explicitly requires all classes to be present in the validation set and raises:

solution: 
Change _train_validation_split to accept y, and call train_test_split(..., stratify=y) when each class has enough cases to support a stratified split. For pathological cases where stratification is impossible, either:

raise a clear early ValueError, or
fall back to an unstratified split with a warning.

The call site in _fit should pass y into _train_validation_split.

```python

import numpy as np

from aeon.transformations.collection.channel_selection import TSelect

rng = np.random.default_rng(0)

# 12 cases, 2 channels, 8 time points.
X = rng.normal(size=(12, 2, 8))

# Class-ordered and imbalanced labels.
# With random_state=1 and validation_size=0.25, the validation split contains
# only class 0, even though the training split contains both classes.
y = np.array([0] * 10 + [1] * 2)

selector = TSelect(
    validation_size=0.25,
    random_state=1,
)

selector.fit(X, y)

Current result:
```

### Expected results

TSelect.fit should avoid a validation split missing one or more classes where stratification is possible. If stratification is impossible because a class has too few cases, it should raise a clear early ValueError.

### Actual results

ValueError: Not all classes are present in the validation set. Increase validation_size to compute the AUC.

### Versions

_No response_

nightcityblade:
#### Reference Issues/PRs

Fixes #3488.

TSelect now passes labels into its train/validation split and uses a stratified split when possible, so small or class-ordered datasets keep all classes represented in validation. If sklearn cannot stratify a pathological split, it falls back to the previous unstratified behavior with a warning.

A regression test covers the imbalanced class-ordered case from the issue.

No.

#### Any other comments?

N/A

##### For all contributions

### DIFF
--- a/aeon/transformations/collection/channel_selection/_tselect.py
+++ b/aeon/transformations/collection/channel_selection/_tselect.py
@@ -137,7 +137,7 @@ class TSelect(BaseChannelSelector):
             raise ValueError("y must have the same number of cases as X.")
 
         X = self._preprocess(X)
-        train_idx, valid_idx = self._train_validation_split(n_cases)
+        train_idx, valid_idx = self._train_validation_split(n_cases, y)
 
         self.channel_scores_ = np.full(n_channels, np.nan)
         self.channel_correlations_ = np.full((n_channels, n_channels), np.nan)
@@ -254,7 +254,7 @@ class TSelect(BaseChannelSelector):
 
         return X_scaled
 
-    def _train_validation_split(self, n_cases):
+    def _train_validation_split(self, n_cases, y):
         if self.validation_size is not None:
             test_size = self.validation_size
         elif n_cases <= 100:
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
Thanks, this is correct. 

One small request: could you add a fit-level regression test using the same reproduction pattern, i.e. construct X, y = np.array([0] * 10 + [1] * 2), and assert that TSelect(validation_size=0.25, random_state=1).fit(X, y) completes? The issue originally failed during the downstream AUC calculation, so testing the full fit path would make the regression coverage stronger than testing _train_validation_split alone.

nightcityblade:
Thanks for the review — I updated the regression coverage to exercise the fit-level case with the imbalanced class-ordered labels and ran the relevant TSelect tests locally.

TonyBagnall (approved):
thanks you for the contribution

TonyBagnall:
@all-contributors add @nightcityblade for bug

TonyBagnall:
@all-contributors add @nightcityblade for bug

PR #3493, opened 2026-06-20
[BUG] Respect requested distance in elbow class selector

### LINKED ISSUE #3489: [BUG] `ElbowClassSum` and `ElbowClassPairwise` silently treat every non-`"euclidean"` distance as DTW
### Problem

The `distance` parameter in `_create_distance_matrix` is documented as the distance metric used to calculate distances between class prototypes. However, the implementation only special-cases `"euclidean"` and treats every other value as DTW:

```python
if distance == "euclidean":
    dis = np.linalg.norm(cls1_ch - cls2_ch, axis=1)
else:
    dis = np.apply_along_axis(
        lambda row: aeon_distance(..., method="dtw"),
        ...
    )
```

The selector should pass method=distance into aeon_distance instead of the literal "dtw". Since "euclidean" is itself a valid aeon distance method, the if/else can likely collapse to a single aeon_distance(..., method=distance) call — worth verifying numeric parity with the current euclidean path (np.linalg.norm(..., axis=1)) and adding a regression test that asserts two different distance values produce different distance frames.

```python
import numpy as np
from aeon.transformations.collection.channel_selection._elbow_class import (
    _create_distance_matrix, _ClassPrototype,
)
X = np.random.RandomState(0).random((20, 3, 30))
y = np.array([0, 1] * 10)
proto, labels = _ClassPrototype("mean")._create_prototype(X.copy(), y)
for d in ("dtw", "msm", "erp"):
    print(d, _create_distance_matrix(proto.copy(), labels, distance=d).values.ravel())
```

### Expected results

different distances

### Actual results

all output the same

### Versions

_No response_

nightcityblade:
#### Reference Issues/PRs

Fixes #3489.

`_create_distance_matrix` now passes the requested non-Euclidean distance method through to `aeon.distances.distance` instead of always using `dtw`. Added a regression test that verifies different requested distances produce different distance matrices.

No.

#### Any other comments?

N/A.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, I can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot after the PR has been merged.

- [ ] N/A - no new estimator/function was added to the online API documentation.
- [ ] N/A - this is a minor bug fix, so I have not added myself as a maintainer.

- [ ] N/A - I do not have write access and did not update CODEOWNERS.

### DIFF
--- a/aeon/transformations/collection/channel_selection/_elbow_class.py
+++ b/aeon/transformations/collection/channel_selection/_elbow_class.py
@@ -87,7 +87,7 @@ def _create_distance_matrix(
                     lambda row: aeon_distance(
                         row[: row.shape[0] // 2],
                         row[row.shape[0] // 2 :],
-                        method="dtw",
+                        method=distance,
                     ),
                     axis=1,
                     arr=np.concatenate((cls1_ch, cls2_ch), axis=1),
--- a/aeon/transformations/collection/channel_selection/tests/test_elbow_class.py
+++ b/aeon/transformations/collection/channel_selection/tests/test_elbow_class.py
@@ -31,6 +31,21 @@ def test_create_distance_matrix():
     assert res.shape == (1, 1)
 
 
+def test_create_distance_matrix_uses_requested_distance():
+    """Test distance matrix uses the requested non-Euclidean distance."""
+    X = np.array(
+        [
+            [[0, 0, 0], [1, 2, 3]],
+            [[0, 1, 2], [1, 3, 5]],
+        ]
+    )
+
+    dtw_res = _create_distance_matrix(X, np.array([0, 1]), distance="dtw")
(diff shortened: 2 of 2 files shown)


TonyBagnall (changes requested):
Thanks, this looks like the right minimal fix. 

One small request: could you also add a regression test for an invalid distance string, for example distance="not_a_distance" or a typo such as "eucledian", asserting that _create_distance_matrix raises ValueError? The original issue was partly that unsupported strings were silently accepted as DTW, so that would make the coverage complete.

nightcityblade:
Thanks for the review — I added the invalid-distance regression test asserting that `_create_distance_matrix` raises `ValueError`, and ran the relevant elbow class tests locally.

TonyBagnall (approved):

PR #3507, opened 2026-06-20
[BUG] fix(unequal_length): validate integer target lengths reject bool and non-positive values

### LINKED ISSUE #3497: [BUG] Unequal-length transformers don't validate integer target lengths (zero / negative / bool silently accepted)
### Describe the bug

Problem

The unequal-length transformers do not validate integer target lengths. Values such as 0, negative integers, and booleans are accepted and either produce invalid zero-length outputs, silently do the wrong thing, or fail later with an opaque low-level NumPy error. The failure surface spans all three of: silent-wrong-output (worst), silent no-op, and unclear error.

Affected:

```python
Padder(padded_length=...)
Resizer(resized_length=...)
Truncator(truncated_length=...)
```
Suggested fix

Add a small helper in _commons.py to validate length parameters. Note it must use numbers.Integral rather than int so that legitimate NumPy integers (e.g. np.int64) are accepted — isinstance(np.int64(5), int) is False — and it must guard bool first, since bool is Integral:

```python
import numpy as np
from aeon.transformations.collection.unequal_length import Padder, Resizer, Truncator

X = np.arange(10).reshape(2, 1, 5)

print(Truncator(truncated_length=0).fit_transform(X).shape)      # (2, 1, 0)
print(Truncator(truncated_length=-1).fit_transform(X).shape)     # (2, 1, 4)
print(Truncator(truncated_length=True).fit_transform(X).shape)   # (2, 1, 1)

print(Resizer(resized_length=0).fit_transform(X).shape)          # (2, 1, 0)
Resizer(resized_length=-1).fit_transform(X)                      # ValueError: negative dimensions are not allowed
Resizer(resized_length=True).fit_transform(X)                    # TypeError: an integer is required

Xt = Padder(padded_length=0, error_on_long=False).fit_transform(X)
print(type(Xt), Xt[0].shape)                                     # <class 'list'> (1, 5)
Padder(padded_length=-1, error_on_long=False).fit_transform(X)   # list, original (1, 5) — silent no-op
```

### Expected results

Integer target lengths should be strictly positive integers. Each transformer should raise a clear ValueError for zero, negative integers, and booleans (bool is a subclass of int in Python but is not a meaningful length):
ValueError: padded_length must be a positive integer, 'min', or 'max'.
ValueError: resized_length must be a positive integer, 'min', or 'max'.
ValueError: truncated_length must be a positive integer, 'min', or 'max'.

### Actual results

| Call | Result | Problem |
|---|---|---|
| `Truncator(truncated_length=0)` | `(2, 1, 0)` | silent zero-length output |
| `Truncator(truncated_length=-1)` | `(2, 1, 4)` | silent negative slicing |
| `Truncator(truncated_length=True)` | `(2, 1, 1)` | silent length-1 from a bool |
| `Resizer(resized_length=0)` | `(2, 1, 0)` | silent zero-length output |
| `Resizer(resized_length=-1)` | `ValueError: negative dimensions are not allowed` | opaque low-level error |
| `Resizer(resized_length=True)` | `TypeError: an integer is required` | opaque low-level error |
| `Padder(padded_length=0, error_on_long=False)` | `list`, original `(1, 5)` | silent no-op |
| `Padder(padded_length=-1, error_on_long=False)` | `list`, original `(1, 5)` | silent no-op |

### Versions

_No response_

axelray-dev:
Fixes #3497.

Padder, Resizer, and Truncator accept integer length parameters (padded_length, resized_length, truncated_length) without validating they are positive integers. Python bool is a subclass of int, so True/False are silently accepted. Zero and negative integers are also accepted, producing silent wrong output or opaque errors.

Changes

- Added shared validation in aeon/transformations/collection/unequal_length/_commons.py: _is_positive_integer_length (bool-safe Integral check) and _validate_positive_integer_length (raises ValueError on bool or int < 1).
- Wired _validate_positive_integer_length into Padder.__init__, Resizer.__init__, and Truncator.__init__.
- Replaced isinstance(..., int) length detection in set_tags and transform paths with the bool-safe _is_positive_integer_length predicate.
- Added tests/test_length_validation.py covering invalid (True, False, 0, -1) and valid (1, 'min', 'max') cases for all three transformers.

Verification

- pytest aeon/transformations/collection/unequal_length/tests/ -> 53 passed
- ruff check aeon/transformations/collection/unequal_length/ -> passed

This pull request includes code written with the assistance of AI. The code has not yet been reviewed by a human.

### DIFF
--- a/aeon/transformations/collection/unequal_length/_commons.py
+++ b/aeon/transformations/collection/unequal_length/_commons.py
@@ -4,7 +4,22 @@ These should ideally be incorporated into the collection data utilities in utils
 the future.
 """
 
+import numbers
+
 import numpy as np
+
+
+def _is_positive_integer_length(value):
+    return isinstance(value, numbers.Integral) and not isinstance(value, bool)
+
+
+def _validate_positive_integer_length(value, parameter_name):
+    """Validate an integer target length parameter."""
+    msg = f"{parameter_name} must be a positive integer, 'min' or 'max'."
+    if isinstance(value, bool):
+        raise ValueError(msg)
+    if isinstance(value, numbers.Integral) and value < 1:
+        raise ValueError(msg)
 
 
 def _get_min_length(X):
--- a/aeon/transformations/collection/unequal_length/_commons.py
+++ b/aeon/transformations/collection/unequal_length/_commons.py
@@ -10,7 +10,12 @@ import numpy as np
 
 
 def _is_positive_integer_length(value):
-    return isinstance(value, numbers.Integral) and not isinstance(value, bool)
+    """Return whether value is a positive non-bool integer length."""
+    return (
+        isinstance(value, numbers.Integral)
+        and not isinstance(value, bool)
+        and value >= 1
+    )
 
 
 def _validate_positive_integer_length(value, parameter_name):
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -10,6 +10,8 @@ from aeon.transformations.collection.base import BaseCollectionTransformer
 from aeon.transformations.collection.unequal_length._commons import (
     _get_max_length,
     _get_min_length,
+    _is_positive_integer_length,
+    _validate_positive_integer_length,
 )
 
 
@@ -75,6 +77,8 @@ class Padder(BaseCollectionTransformer):
         error_on_long=True,
         random_state=None,
     ):
+        _validate_positive_integer_length(padded_length, "padded_length")
+
         self.padded_length = padded_length
         self.fill_value = fill_value
         self.add_noise = add_noise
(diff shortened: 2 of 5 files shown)


TonyBagnall (changes requested):
Thanks for the PR, this is a good focused fix and the tests cover the main reported failure modes.

The constructor validation looks right: `bool`, zero and negative integer values are now rejected for `Padder`, `Resizer` and `Truncator`, while `1`, `"min"` and `"max"` remain valid. This addresses the bug from #3497.

One small change requested before approval: `_is_positive_integer_length` currently only checks “integer and non-bool”, so it still returns `True` for `0` and `-1`. Because the helper is named “positive” and is used in the tag/transform paths, I think it should include the positivity check as well:

```python
def _is_positive_integer_length(value):
    """Return whether value is a positive non-bool integer length."""
    return (
        isinstance(value, numbers.Integral)
        and not isinstance(value, bool)
        and value >= 1
    )

also no test covers the numpy-int improvement — a one-line np.int64(1) valid case would lock that behaviour in.

TonyBagnall (approved):

TonyBagnall:
@all-contributors add @axelray-dev for bug

PR #3508, opened 2026-06-20
[ENH] Soft-MSM distance

chrisholder:
### Reference Issues/PRs

Second of the soft-distance stack. **Stacked on #3483** (soft distance sub-package + signed soft-DTW) — please review/merge that first; this branch targets it and the diff will shrink once #3483 merges.

Adds **soft-MSM**, a differentiable squared-cost variant of move-split-merge, to `aeon.distances.elastic.soft`, with the standard `distance` / `pairwise_distance` / `cost_matrix` / `alignment_path` public surface and a private gradient (`_soft_msm_grad_x`) for the upcoming soft barycentre averaging PR. It is registered in the distance registry with `non_negative=False`, so it inherits the `SIGNED_DISTANCES` test handling added in #3483.

**Squared-cost, not a relaxation of aeon's MSM.** Soft-MSM uses squared move/split/merge costs (mirroring how soft-DTW relates to DTW), so as `gamma -> 0` it converges to a *squared-cost* MSM, **not** aeon's absolute-cost `msm_distance`. The prototype docstrings claimed MSM recovery; that is corrected. Like soft-DTW it is signed (can be negative, `d(x, x) != 0`).

**Multivariate is independent.** Multivariate series use the independent strategy (sum of per-channel univariate cost matrices), matching `msm_distance(independent=True)`. The prototype silently used channel 0 only; this is fixed and tested.

**Between-gate kept as the prototype (`1e-9`).** The smooth "between" gate uses the original fixed `1e-9` floor. That floor makes the gate amplitude-dependent (it degenerates for very small/large series); making it scale-invariant — and updating the matching hand-derived gradient — is deferred and tracked as a follow-up in #3518, so this PR stays faithful to the prototype.

No.

### Any other comments?

No external reference implementation exists for this squared variant, so correctness is established with a self-authored hard (`gamma -> 0`) squared-MSM reference and a finite-difference gradient check, rather than an external oracle. The private gradient is univariate; multivariate soft barycentre averaging is handled in the follow-up averaging PR. soft-DTW and soft-MSM are added to the distances API reference.

### Test plan

- New `test_soft_msm.py`: `gamma -> 0` convergence to the hard squared-MSM reference, finite-difference gradient check, multivariate == per-channel sum, signed-value/registry checks, and a finiteness-across-scales check.
- soft-MSM added to `expected_distance_results`; picked up by the general distance suite.
- Verified locally: full `aeon/distances` suite, `test_soft_msm`, and the averaging suite (soft-MSM correctly stays out of the discrete BA methods). Pre-commit passes.


TonyBagnall:
add a link to the arxiv paper in the PR description

TonyBagnall:
closing this as we have renamed the algorithm, hopefully can return in its new incarnation!

PR #3510, opened 2026-06-21
[BUG] Fix OHIT random state handling

### LINKED ISSUE #3499: [BUG] OHIT mutates self.k / self.kapa in _transform
### Describe the bug

_ohit.py:127–130 overwrite the constructor parameters when they're None. This corrupts get_params() and leaks the first dataset's values into later calls; reusing the instance on data with a smaller minority class crashes.

Fix: compute into locals (k = self.k if self.k is not None else int(...)) or store derived values as fitted attributes self.k_/self.kapa_; never write back to self.k/self.kapa. Same not-in-__init__/not-mutating principle as #3490.

```python
ohit = OHIT(random_state=49)
ohit.fit_transform(XA, yA)        # 16 minority -> k becomes 5
ohit.get_params()["k"]            # 5  (was None)
ohit.fit_transform(XB, yB)        # 4 minority -> ValueError: n_neighbors=6 > n_samples_fit=4
```

### Expected results

k/kapa remain the parameters set at construction; each fit recomputes its own defaults; get_params() is unchanged.

### Actual results

params are overwritten; a fresh OHIT handles dataset B fine, but the reused instance raises ValueError.

### Versions

_No response_

### LINKED ISSUE #3501: [BUG] OHIT ignores random_state (uses global np.random)
### Describe the bug

_ohit.py:276: _generate_synthetic_samples calls np.random.multivariate_normal(...) on the global RNG, despite _transform creating a seeded random_state at line 139. OHIT is therefore non-deterministic even with a fixed seed.

Fix: pass the check_random_state generator into _generate_synthetic_samples and use rng.multivariate_normal(...); also build it once in _fit from self.random_state rather than re-seeding inside _transform. Add a reproducibility test asserting two same-seed runs are equal.

```python
r1 = OHIT(random_state=49).fit_transform(X, y)[0]
r2 = OHIT(random_state=49).fit_transform(X, y)[0]
np.array_equal(r1, r2)   # False ; max|diff| ≈ 1.36
```

### Expected results

identical output for identical random_state.

### Actual results

 outputs differ by up to 1.36.

### Versions

_No response_

gaoflow:
Closes #3501.
Closes #3499.

This fixes two OHIT state handling issues:

- build the fitted random state once during `fit` and pass it through synthetic sample generation instead of using the global `np.random` generator
- compute default `k`/`kapa` values as local fitted values instead of writing them back to constructor parameters

The added tests cover same-seed reproducibility and verify that default `k`/`kapa` do not mutate `get_params()` across fits.

Testing:

- `uv run pytest aeon/transformations/collection/imbalance/tests/test_ohit.py -q` initially could not run because the default pytest config requires plugins not installed in the minimal uv environment
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/imbalance/tests/test_ohit.py -q`
- `uv run ruff check aeon/transformations/collection/imbalance/_ohit.py aeon/transformations/collection/imbalance/tests/test_ohit.py`
- `uv run ruff format --check aeon/transformations/collection/imbalance/_ohit.py aeon/transformations/collection/imbalance/tests/test_ohit.py`
- `git diff --check`

### DIFF
--- a/aeon/transformations/collection/imbalance/_ohit.py
+++ b/aeon/transformations/collection/imbalance/_ohit.py
@@ -103,6 +103,7 @@ class OHIT(BaseCollectionTransformer):
             if key != class_majority
         }
         self.sampling_strategy_ = OrderedDict(sorted(sampling_strategy.items()))
+        self.random_state_ = check_random_state(self.random_state)
 
         return self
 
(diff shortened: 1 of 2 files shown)


TonyBagnall:
Thanks for the PR, and sorry for the timing here.

I have reviewed the change and the implementation is sound: it addresses the two OHIT issues by avoiding mutation of `k`/`kapa`, using a fitted random state rather than the global `np.random`, and adding regression tests for reproducibility and parameter preservation.

However, #3505 has already been merged and fixes the same two issues, so this PR now overlaps almost entirely with code that is already on `main`.

The one useful incremental difference here is that this version initialises `self.random_state_` in `_fit`, which follows the original issue text slightly more closely. If you want to keep contributing this part, could you rebase against current `main` and narrow the PR to just that small follow-up change? Otherwise I think we should close this as superseded by #3505.

Thanks again for the contribution. The code itself was on the right track.

PR #3511, opened 2026-06-21
[BUG] Fix Padder add_noise for integer input

### LINKED ISSUE #3494: [BUG] Padder raises UFuncTypeError on integer-typed input when add_noise is set
### Describe the bug

In aeon/transformations/collection/unequal_length/_pad.py, _transform adds uniform noise in place at line 195 (p[series.shape[1]:] += rng.uniform(...)). p comes from np.pad(channel, ...) and keeps the channel's dtype, so for integer input the in-place float add fails the NumPy same_kind casting rule. Affects any integer collection with add_noise set, regardless of fill_value.

Suggested fix: build the padded array as float before adding noise — e.g. cast channel/p with astype(float, copy=False) in the add_noise is not None branch (or float-cast the channel up front when add_noise is set). Add a regression test with an integer np-list and add_noise.

```python
import numpy as np
Xint = [np.arange(4*L, dtype=int).reshape(4, L) for L in (5, 7, 9)]
Padder(padded_length=12, add_noise=0.5, random_state=0).fit_transform(Xint)
# UFuncTypeError: Cannot cast ufunc 'add' output from dtype('float64')
#   to dtype('int64') with casting rule 'same_kind'
```

### Expected results

The integer series are padded to length 12 and noise between 0 and add_noise is added to the padded positions, returning a float (3, 4, 12) array (the addition of float noise implies a float result).

### Actual results

fit_transform raises:
UFuncTypeError: Cannot cast ufunc 'add' output from dtype('float64')
to dtype('int64') with casting rule 'same_kind'
No array is returned; the transform aborts.

### Versions

_No response_

gaoflow:
## Summary
- Cast padded channels to float before adding `Padder` noise so integer-valued inputs can receive float noise.
- Add a regression test for integer np-list input with `add_noise` enabled.

Closes #3494.

## Testing
- `uv run ruff check aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `uv run ruff format --check aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests/test_pad.py::test_padding_integer_input_with_noise -q`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests/test_pad.py -q`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests -q`
- `git diff --check`


TonyBagnall (commented):
lgtm, thanks

TonyBagnall:
Thanks for this @gaoflow — the Padder noise fix is clean: correctly scoped to the `add_noise` branch so non-noise integer input keeps its dtype, good regression test, conformance passes. Nice work, and the same on the OHIT one (that one collided with #3505, which had already merged — no harm done, just the reason we ask people to comment on an issue to get it assigned before starting, so two contributors don't land on the same fix).

Id say you have moved beyond good first issue PRs now, so I'd like to leave those for people newer to the project — that's what the label is there to protect. The good news is the more interesting work is elsewhere: I'll point you at some larger open issues that aren't GFI-tagged if you like.

If you're up for it, the highest-value thing you could do is what I've been doing to generate these issues in the first place: pick one of the `transformations.collection` subpackages that hasn't been swept yet, run through it systematically, and *file* well-formed bug reports (actual vs expected behaviour, a standalone reproduction). You've shown you can find and fix these, so finding and reporting them should be well within reach — and it helps the whole contributor pipeline rather than just closing tickets. Happy to give you a couple of pointers on where to start if that sounds good. Personally Ive been using codex and claude to help do this, feel free to use AI.

Next system wide improvements I want to prioritise is improving coverage. Feel free to open PRs for this.

Keeping to one open PR at a time also helps keep the review queue moving. Appreciate the energy here — we would be glad to have you around as a regular contributor.

TonyBagnall (approved):

PR #3512, opened 2026-06-21
[BUG] Validate Padder fill_value inputs

### LINKED ISSUE #3495: [BUG] Padder fill_value as a NumPy array is silently misinterpreted (not a per-series fill)
### Describe the bug

```python
import numpy as np

```

### Expected results

Case 1 returns

  [[[5. 6. 2. 2.]
    [7. 8. 2. 2.]]]

Case 2 raises

### Actual results

Case 1 returns

  [[[5. 6. 2. 2.]
    [7. 8. 2. 2.]]]

Case 2 raises

### Versions

_No response_

gaoflow:
## Summary
- Remove the unsupported NumPy array claim from `Padder.fill_value` documentation.
- Validate non-string, non-callable `fill_value` inputs so array-like values raise a clear `ValueError` instead of being silently misinterpreted by `np.pad` or surfacing a broadcast error.
- Add regression coverage for length-2 and length-3 NumPy array fill values.

Closes #3495.

## Testing
- `uv run ruff check aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `uv run ruff format --check aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests/test_pad.py::test_padder_rejects_array_fill_value -q`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests/test_pad.py -q`
- `uv run --with pytest-xdist --with pytest-rerunfailures --with pytest-timeout pytest aeon/transformations/collection/unequal_length/tests -q`
- `git diff --check`

### DIFF
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -26,10 +26,9 @@ class Padder(BaseCollectionTransformer):
         shortest series seen in ``fit``. If "max", will pad to the longest series seen
         in ``fit``. If an integer, will pad to that length.
         Calling ``fit`` is not required if ``padded_length`` is an int.
-    fill_value : int, str or Callable, default=0
-        Value to pad with. Can be a float or a statistic string or an numpy array for
-        each time series. Supported statistic strings are "mean", "median", "max",
-        "min".
+    fill_value : scalar, str or Callable, default=0
+        Value to pad with. Can be a scalar, a statistic string, or a callable.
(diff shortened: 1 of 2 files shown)


TonyBagnall:
this is a good first issue, best someone else takes it

PR #3522, opened 2026-06-21
[ENH] Improved coverage for channel selectors

### LINKED ISSUE #3437: [ENH] Coverage for channel selectors
### Describe the issue

coverage is slipping, our AI overlords can surely help improve it. My bad with the channel selector package, many missed lines, and this package will expand soon, so best get near to 100%.

_No response_

### Additional context

_No response_

TonyBagnall:
fixes #3437 

coverage tests for channel selectors


alexbanwell1:
## Review — Improved coverage for channel selectors

This PR is test-only (413 additions, 0 deletions) adding coverage for `TSelect`, the `_elbow_class` prototypes/knee-point helper, and `ChannelScorer._get_test_params`. I ran the full new suite locally (55 passed) and hand-checked the non-obvious assertions — the rank/NaN-interpolation math, the clustering branch expectations, and the floating-point split case — all match the implementation. Nice, well-targeted coverage; no correctness issues found. A couple of minor, optional notes:

**Minor / optional**

- `tests/test_tselect.py` (`test_tselect_get_test_params`) and `tests/test_channel_scorer.py` (`test_channel_scorer_get_test_params`) — the only assertion is `isinstance(obj, Cls)`, which is tautologically true after `Cls(**params)` and never exercises the params beyond `__init__`. The docstrings say "valid and usable" but nothing is fitted. Consider a quick `fit`/`transform` on small data to make these assert something real (or drop the assert and rely on construction not raising).

- `tests/test_tselect.py` (`test_tselect_train_validation_split_explicit_and_large`) — `assert len(train_idx) == 99` is correct but subtle: it depends on `1 - 100/150` rounding up in `train_test_split`'s internal `ceil(test_size * n)`. It reads like it should be 100. A one-line comment explaining the 99 (or asserting a ratio range instead) would spare future readers the head-scratch.

Overall this looks good to merge.

alexbanwell1 (approved):

PR #3524, opened 2026-06-21
fix: validate fill_value type in Padder to prevent silent misinterpretation

### LINKED ISSUE #3495: [BUG] Padder fill_value as a NumPy array is silently misinterpreted (not a per-series fill)

BB0813:
## Fix

Fixes #3495

### Problem

The `Padder` docstring claimed `fill_value` accepts 'a numpy array for each time series', but `np.pad` interprets arrays as `(before, after)` constant pairs, causing silently wrong results:

- A length-2 array fills every padded position with its second element (the 1.0 is never used)
- Other array lengths raise opaque NumPy broadcast errors

### Solution

1. Updated the docstring to document the actual supported types: numeric scalar (int/float), statistic string, or callable
2. Added validation in `__init__` to raise clear errors for unsupported types (numpy arrays, lists, None, etc.)
3. Invalid statistic strings are also caught at init time

### Changes

- `_pad.py`: Updated docstring, added fill_value type validation in `__init__`
- `test_pad.py`: Updated test to expect error at construction time

### Before/After

| Input | Before | After |
|-------|--------|-------|
| `fill_value=np.array([1, 2])` | silently wrong result | `TypeError` |
| `fill_value=[1, 2]` | opaque broadcast error | `TypeError` |
| `fill_value=None` | opaque error | `TypeError` |
| `fill_value='FOOBAR'` | error in transform | `ValueError` at init |

### Testing

All 33 existing unequal_length tests pass.

### DIFF
--- a/aeon/transformations/collection/unequal_length/_commons.py
+++ b/aeon/transformations/collection/unequal_length/_commons.py
@@ -4,7 +4,43 @@ These should ideally be incorporated into the collection data utilities in utils
 the future.
 """
 
+import numbers
+
 import numpy as np
+
+
+def _validate_length_param(param, param_name):
+    """Validate that a length parameter is a positive integer, 'min', or 'max'.
+
+    Parameters
+    ----------
+    param : int, "min", or "max"
+        The parameter value to validate.
+    param_name : str
+        Name of the parameter (used in error messages).
+
+    Raises
+    ------
+    ValueError
+        If ``param`` is not a positive integer, "min", or "max".
+    """
+    if isinstance(param, str):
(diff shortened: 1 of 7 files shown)


BB0813:
Closing in favor of combining all fixes into a single PR

PR #3525, opened 2026-06-21
fix: move RandomChannelSelector p validation from __init__ to _fit

### LINKED ISSUE #3490: [BUG] RandomChannelSelector validates p in __init__ instead of _fit (violates sklearn clone contract)
### Describe the bug

call init

### Expected results

validate in fit

### Actual results

init check

### Versions

_No response_

BB0813:
## Fix

Fixes #3490

### Problem

`RandomChannelSelector.__init__` validates that `p` is in `(0, 1]`, but the sklearn clone contract requires that `__init__` only stores parameters without validation. This makes `RandomChannelSelector` inconsistent with other channel selectors (`ChannelScorer`, `ElbowClass*`) in the same package, which defer validation to `_fit`.

### Solution

Move the range check for `p` from `__init__` to `_fit`. Now `__init__` only stores `self.p = p` and `self.random_state = random_state`.

### Changes

- `_random.py`: Removed validation from `__init__`, added it to `_fit`
- `test_random.py`: Updated test to expect `ValueError` at `fit_transform` time instead of construction time

### Testing

Test passes with the fix applied.

### DIFF
--- a/aeon/transformations/collection/channel_selection/_random.py
+++ b/aeon/transformations/collection/channel_selection/_random.py
@@ -43,16 +43,16 @@ class RandomChannelSelector(BaseChannelSelector):
     }
 
     def __init__(self, p=0.4, random_state=None):
-        if p <= 0 or p > 1:
-            raise ValueError(
-                "Proportion of channels to select should be in the range (0,1]."
-            )
         self.p = p
         self.random_state = random_state
         super().__init__()
 
     def _fit(self, X, y):
         """Randomly select channels to retain."""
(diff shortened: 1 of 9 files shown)


TonyBagnall:
conflicted now @BB0813, if you fix the conflicts that would be great

BB0813:
I apologize for not adding the [BUG] label. I have added the label in the new PR.
Please have the contributor review the new PR
https://github.com/aeon-toolkit/aeon/pull/3541

PR #3526, opened 2026-06-21
[ENH] Gradient-based soft barycentre averaging

chrisholder:
### Reference Issues/PRs

Third of the soft-distance stack. **Stacked on #3508** (soft-MSM distance), which is stacked on #3483. Please review/merge those first; this branch targets #3508 and the diff shrinks as they merge.

Adds **gradient-based soft barycentre averaging**. `soft_barycenter_average` (new `_ba_soft.py`) minimises a differentiable soft elastic objective with `scipy.optimize.minimize` (L-BFGS-B), using the private soft-DTW / soft-MSM gradients from the earlier PRs. It is wired into `elastic_barycenter_average` as `method="soft"`, added to the averaging callable registry, and exported alongside `VALID_SOFT_BA_METHODS`.

**Clean separation — soft distances only via `method="soft"`.** Soft distances are not used by the discrete BA methods (a soft distance has an alignment *distribution*, not a single hard path, so DBA-style discrete realignment is ill-defined and was a no-op). `soft_dtw` is removed from `VALID_BA_DISTANCE_METHODS` and from `_get_alignment_path`; early validation rejects `method="soft"` with a hard distance and the discrete methods with a soft distance. The discrete-BA test list drops `soft_dtw` and the temporary `test_kasba` skip added in #3483 is reverted.

**Fixes over the prototype:**
- `return_distances_to_center` now recomputes the distances at the optimised barycentre instead of returning the initial (`inf`) values from setup.
- Multivariate soft-MSM averaging accumulates the gradient **per channel** (independent), matching the soft-MSM distance. The prototype assigned a 1D gradient into a 2D slot, so soft-MSM averaging never actually worked (it also had no tests).
- Threading uses `numba_thread_handler` (the prototype imported a module that does not exist on `main`); the docstring covers both soft distances.
- `_ba_setup` gains a `compute_previous_cost` flag so the soft path skips the upfront pairwise-cost computation.

No. (`tslearn`, already optional, is used only as a guarded test oracle.)

### Test plan

- New `test_soft_ba.py` parametrised over **soft-DTW and soft-MSM**: shape/finiteness (uni + multivariate), `return_cost`/`return_distances_to_center`, weights change the result, single-series passthrough, the `method="soft"` validation matrix, and dispatch-equals-direct.
- soft-DTW barycentre validated against tslearn's `softdtw_barycenter` (guarded with `importorskip`).
- Verified locally: full `aeon/clustering/averaging` suite + `test_k_means` (402 passed) with the discrete-BA changes, and the soft BA tests. Pre-commit passes.

### DIFF
--- a/aeon/clustering/averaging/__init__.py
+++ b/aeon/clustering/averaging/__init__.py
@@ -5,18 +5,22 @@ __all__ = [
     "mean_average",
     "petitjean_barycenter_average",
     "subgradient_barycenter_average",
+    "soft_barycenter_average",
     "VALID_BA_METHODS",
     "VALID_BA_DISTANCE_METHODS",
+    "VALID_SOFT_BA_METHODS",
     "shift_invariant_average",
     "kasba_average",
 ]
 
 from aeon.clustering.averaging._averaging import mean_average
 from aeon.clustering.averaging._ba_petitjean import petitjean_barycenter_average
+from aeon.clustering.averaging._ba_soft import soft_barycenter_average
 from aeon.clustering.averaging._ba_subgradient import subgradient_barycenter_average
 from aeon.clustering.averaging._ba_utils import (
     VALID_BA_DISTANCE_METHODS,
     VALID_BA_METHODS,
+    VALID_SOFT_BA_METHODS,
 )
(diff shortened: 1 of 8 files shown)


TonyBagnall:
converting to draft as stacked on another

PR #3527, opened 2026-06-21
[ENH] NearestCentroidClassifier

chrisholder:
### Reference Issues/PRs

Fourth and final PR of the soft-distance stack. **Stacked on #3526** (soft barycentre averaging) → #3508 → #3483. Please review/merge those first; this branch targets #3526 and the diff shrinks as they merge.

Adds **`NearestCentroidClassifier`**, a nearest-centroid (Rocchio) time series classifier. It computes one centroid per class by averaging that class's training series, then classifies a new series by the nearest class centroid under a chosen elastic distance.

- **Naming.** Called `NearestCentroidClassifier` to avoid colliding with `sklearn.neighbors.NearestCentroid` (the same reason aeon's KNN is `KNeighborsTimeSeriesClassifier`).
- **Defaults.** `distance="dtw"` and `average_method=None`, which resolves to DBA (`petitjean`) for ordinary distances and to the gradient-based soft barycentre for soft distances.
- **Soft coupling.** A soft distance auto-promotes to `average_method="soft"`; explicitly pairing a soft distance with a non-soft averaging method (or vice versa) raises `ValueError`. Centroids are computed through `elastic_barycenter_average`, so this validation is shared with the averaging module rather than re-implemented.
- Full numpydoc docstring, `_tags`, `_get_test_params`, a runnable `Examples` doctest, and an entry in the classification API reference.

No.

### Test plan

- New `test_nearest_centroid.py`: per-class centroid shapes, label prediction, the soft auto-promote, the validation matrix (soft+hard and hard+soft both raise), the DBA default, and that mean vs DBA centroids differ on warped data.
- Passes aeon's general estimator checks (`check_estimator`, 22/22) and the doctest.
- Verified locally with euclidean/dtw/soft-DTW/soft-MSM configurations. Pre-commit passes.

### DIFF
--- a/aeon/classification/distance_based/__init__.py
+++ b/aeon/classification/distance_based/__init__.py
@@ -3,11 +3,15 @@
 __all__ = [
     "ElasticEnsemble",
     "KNeighborsTimeSeriesClassifier",
+    "NearestCentroidClassifier",
     "ProximityTree",
     "ProximityForest",
 ]
 
 from aeon.classification.distance_based._elastic_ensemble import ElasticEnsemble
+from aeon.classification.distance_based._nearest_centroid import (
+    NearestCentroidClassifier,
+)
 from aeon.classification.distance_based._proximity_forest import ProximityForest
 from aeon.classification.distance_based._proximity_tree import ProximityTree
 from aeon.classification.distance_based._time_series_neighbors import (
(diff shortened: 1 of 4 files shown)


TonyBagnall:
soft stack, converting to draft until others are in

Issue #3496, opened 2026-06-20
[DOC] Typos in unequal length package


TonyBagnall:
# Docstring & typo fixes in `aeon.transformations.collection.unequal_length`

A batch of docstring, typo, and wording corrections in the `unequal_length` package. Grouped by file as a checklist; identifiers in code spans. None change behaviour.

## `_pad.py` (`Padder`)

- [ ] Class summary says it `"pads to that with zeroes"`, but the pad value is configurable via `fill_value` (zero is only the default) — reword to reflect the configurable fill.
- [ ] Class summary (shared with the other two classes) says it finds the `"max/min length series across all series and channels"`; length is a property of each case/series, not each channel (each 2D case is `(n_channels, length_i)`) — reword to `"across all series"`.
- [ ] `fill_value` docstring lists `"mean"`, `"median"`, `"max"`, `"min"` but the implementation also supports `"last"` — add it.
- [ ] `fill_value` docstring claims `"a numpy array for each time series"` is supported. It is passed to `np.pad(..., constant_values=...)`, which interprets an array as `(before, after)` constants, not a per-series fill — simplify to **scalar, supported string, or callable** (and fix the grammar: `"an numpy array"` -> `"a numpy array"`).
- [ ] `random_state` docstring says `"Only used if add_noise is True"`, but `add_noise` is a float or `None` — reword to `"Only used if add_noise is not None"`.
- [ ] `padded_length` is referred to as `pad_length` in docstrings and error text — use `padded_length` consistently.
- [ ] `_transform` return docstring references `self.pad_length_`; the fitted attribute is `_padded_length`.
- [ ] `_transform` return docstring states a `numpy3D` array is returned, but a list of numpy arrays is returned when `error_on_long=False` — document both.
- [ ] `_transform` comment `# Must call fit if pad_length is None` — `None` is not a documented option for `padded_length`.
- [ ] Example uses `fill_value =42`; remove the space -> `fill_value=42`.

## `_resize.py` (`Resizer`)

- [ ] **`_transform` summary says `"Fit a linear function on each channel of each series, then resample."`** The implementation uses `np.interp` (piecewise linear interpolation), not the fitting of a single linear function — reword, e.g. `"Linearly interpolate each channel of each series to the target length."` *(highest-value fix here — it misdescribes the algorithm.)*
- [ ] `_transform` input docstring: `"2D nump arrays"` -> `"2D numpy arrays"`.
[...]

PR #3528, opened 2026-06-21
[DOC] Fix docstring inaccuracies and typos in unequal_length package

vir-cipher:
Addresses part of #3496. A focused subset of docstring/typo fixes in the unequal_length package. None change behaviour.

_resize.py
- `_transform` summary said "Fit a linear function … then resample", but the implementation uses `np.interp` (piecewise linear interpolation); reworded to describe interpolation.
- "2D nump arrays" -> "2D numpy arrays".
- Return docstring referenced a non-existent `self.length`; corrected to `resized_length`.

_pad.py
- Grammar: "an numpy array" -> "a numpy array".
- `random_state`: said "Only used if add_noise is True", but `add_noise` is a float or None; reworded to "is not None".
- Example: removed stray space in `fill_value =42`.

_truncate.py
- Error text grammar: "truncated_length is str" -> "truncated_length is a string".

### DIFF
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -27,7 +27,7 @@ class Padder(BaseCollectionTransformer):
         in ``fit``. If an integer, will pad to that length.
         Calling ``fit`` is not required if ``padded_length`` is an int.
     fill_value : int, str or Callable, default=0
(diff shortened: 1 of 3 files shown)


TonyBagnall (approved):
thanks

TonyBagnall:
@all-contributors add @vir-cipher for docs

PR #3533, opened 2026-06-21
[MNT] Improve transformations/series test coverage

TonyBagnall:
## Coverage

Adds focused behavioral coverage for series transformers, including:

  - documented defaults and parameter validation;
  - mathematical equivalence with reference implementations;
  - transform/inverse-transform round trips;
  - edge cases with meaningful user-visible outcomes;
  - pipeline and output-option behavior.


patrickzib:
@TonyBagnall thank you tony. I will check.

patrickzib:
@TonyBagnall yes, ClaSP changes are fine. Thank you. The ClaSP tests seem to be a bit over-engineered, but are ok.

aeon/transformations/series/_stl.py
baraline: I can't seem to get the structure of this check with the changes you made.

Shouldn't this whole auto-detection block be in the resolve_use_numba_local instead ? It looks like we test the same thing twice

aeon/transformations/series/tests/test_clasp.py
```diff
@@ -1,8 +1,16 @@
...
 import numpy as np
+import pytest
 
 from aeon.transformations.series import ClaSPTransformer
+from aeon.transformations.series._clasp import (
+    _binary_f1_score,
+    _compute_distances_iterative,
+    _roc_auc_score,
+    _sliding_dot_product,
+    clasp,
```
baraline: There are some local variable already using "clasp" as name. While it shouldn't cause issue, it is a bit confusing while reading the code.

Renaming it to "import clas as clasp_fn" or something similar would help to clarify

aeon/transformations/series/_acf.py
```diff
@@ -76,7 +76,7 @@ def _transform(self, X, y=None):
             raise ValueError(
                 f"The number of lags is too large for the length of the "
                 f"series, autocorrelation would be calculated with just"
```
baraline: Need a space after 'just' or at the start for the next string, otherwise the message will would display as "with justX observations"

baraline (commented):
Non-blocking comments, but still good to adress for quality if you can

aeon/transformations/series/_stl.py
TonyBagnall: fyi, Ive split out the STL and CLASP stuff from this branch, it was a mess, this now is just coverage for series transformers

aeon/transformations/series/_acf.py
TonyBagnall: same here, moving to a different PR

aeon/transformations/series/tests/test_clasp.py
TonyBagnall: move to clasp b ranch

alexbanwell1 (dismissed):
## Overview

Pure test-only PR adding behavioral coverage across `transformations/series` — no production code changes. Adds one new file (`test_log.py`) and extends 11 existing test modules covering documented defaults, parameter validation/error paths, mathematical equivalence with reference implementations (SciPy, NumPy, R), and transform/inverse-transform round trips.

## Correctness — verified against source

I cross-checked each non-trivial assertion against the implementation. All hold:

- **BoxCox**: error strings (`not recognized`, `seasonal periodicity`, `tuple of length 2`) match `_boxcox.py`; `method="all"` order is `[pearsonr, mle]`, matching `_all()`.
- **ACF zero-variance**: at lag 5 exactly one window has zero variance, and `_acf.py` returns `0` in that branch — correct, and meaningfully distinct from the both-zero → `1` branch.
- **ACF default n_lags**: source `int(max(1, X.shape[1]/4))` agrees with the test.
- **MSTL**: all six parametrized error substrings map to real raises in `_mstl.py`.
- **Diff**, **Dobin** (univariate warning), **Warping** (`0 or 1`), **RMS** (`[3,5,7]` default), **PLA**, **Log** (`log(scale*(x+offset))`) — all match.
- Round-trip / equivalence tests (Log, BoxCox, ScaledLogit, YeoJohnson) use appropriate tolerances and reference implementations.

## Code quality

- Consistent with repo style: one-line docstrings, `np.testing.assert_*`, `pytest.raises(match=...)`, standard fixtures.
- Good determinism hygiene: seeded RNG in the Dobin test; `stl_use_numba=False` in MSTL equivalence tests.
- `match=` on error assertions so tests fail loudly if the wrong `ValueError` fires.

## Minor observations (non-blocking)

- `test_acf_default_n_lags_is_quarter_series_length` uses `n_timepoints=20` (divisible by 4), so it can't distinguish floor from truncation — a length like 22 would exercise the rounding path more. Harmless.
- Some `match` substrings are broad (e.g. `integer`, `output`), but acceptable given the constrained inputs.

## Verdict

Approve. Well-targeted, accurate coverage that pins down documented defaults and validation behavior. The tests assert what the code actually does — the thing that most often goes wrong in coverage PRs, and here does not.

alexbanwell1 (approved):
Reviewed the full diff: purely test additions (12 files, +270/-2, no source changes).

The new tests are idiomatic and target meaningful behavior — inverse-transform round-trips (boxcox/log/scaled_logit/warping), validation/error paths (invalid methods, bounds, orders, `series_index`, MSTL option matrix), default-parameter contracts (RMS sieve windows, ACF default `n_lags`, MSTL `s_windows=[None]`), and edge cases (univariate DOBIN passthrough, ACF zero-variance window → 0). Each has a clear, specific assertion.

Merge conflicts with main (test_acf.py, test_mstl.py) resolved as the union of both sides' new tests; the obsolete numba-STL smoke test was dropped in line with main removing that code path. Full `aeon/transformations/series/tests/` suite passes locally (104 passed).

PR #3535, opened 2026-06-21
[MNT] Improve test coverage for transformations/collection/dictionary_based (PAA, SAX, SFAWhole)

TonyBagnall:
Raises test coverage for three modules in
  `aeon/transformations/collection/dictionary_based` towards 100%, measured with
  `NUMBA_DISABLE_JIT=1` (coverage.py can't trace into compiled `@njit` code).

  | File             | Before | After |
  |------------------|--------|-------|
  | `_paa.py`        | 82%    | 100%  |
  | `_sax.py`        | 91%    | 99%   |
  | `_sfa_whole.py`  | 74%    | 100%  |

  ### What was added

  - **`_paa.py`**: tests for the length-divisible-by-`n_segments` fast path
    (forward and inverse), the case where `n_segments` exceeds the number of
    timepoints (empty trailing segments), and `_get_test_params`.
  - **`_sax.py`**: tests for the unsupported-distribution `NotImplementedError`,
    `znormalized=False`, and `_get_test_params`. The one remaining uncovered
    line (`_generate_breakpoints`'s `if distribution == "Gaussian"` fallthrough)
    is unreachable in practice — the constructor already rejects any other
    distribution before `_generate_breakpoints` is ever called — so it's left
    as-is rather than forced via a contrived direct call.
  - **`_sfa_whole.py`**: new test file (`test_sfa_whole.py`) — there was no
    dedicated test for `SFAWhole` at all; it was only exercised indirectly via
    `.fit()` inside `test_sfa.py`. Added `fit_transform`, `transform` after a
    separate `fit`, and `_get_test_params`.

  ### What was intentionally not touched, and why

  - **`_sfa.py`** (63%) and **`_sfa_fast.py`** (58%): work was stopped on both
    after finding that `get_words()` crashes with a numba `TypingError` for any
    normal sliding-window configuration (only works in the degenerate case
    `window_size == series length`, i.e. exactly one word per series). The same
    bug is duplicated in both classes. Per direction, these are being filed as
    separate good-first-issues rather than fixed here, and no further coverage
    work was done on either class once the bug surfaced.
  - **`_borf.py`** (18%): `BORF.__init__` requires the `sparse` soft dependency,
    which isn't installed in this environment, so the class can't even be
    constructed. Left untouched, consistent with how `statsmodels`/`stumpy`
    were handled for other modules.


aeon/transformations/collection/dictionary_based/tests/test_sax.py
```diff
@@ -31,6 +31,27 @@ def test_equal_length_univariate_sax(n_segments, alphabet_size):
...
+    with pytest.raises(NotImplementedError):
+        SAX(distribution="bogus")
+
+
+def test_sax_znormalized_false_normalizes_internally():
+    """Test znormalized=False applies normalization before the PAA step."""
+    X = np.random.RandomState(0).normal(loc=5, scale=2, size=(5, 1, 40))
+    sax = SAX(n_segments=8, alphabet_size=4, znormalized=False)
+    X_sax = sax.fit_transform(X=X)
+    assert X_sax.shape[-1] == 8
```
patrickzib: The test does not perform the action described in the docstring? 
imo it should test, if mean=0 and std-dev=1?

patrickzib (changes requested):
The docstring does not represent the action performed

TonyBagnall:
thanks @patrickzib 

Fix (committed as 5d06825e on ajb/cov3) — the test now actually verifies the behaviour:
  - asserts the per-series z-normalization gives mean ≈ 0, std ≈ 1 (his literal suggestion);
  - checks znormalized=False on raw data reproduces hand-normalized data with znormalized=True (assert_array_equal);
  - checks it differs from skipping normalization (raw data with znormalized=True, which buckets everything into the top symbol).

aeon/transformations/collection/dictionary_based/tests/test_sax.py
TonyBagnall: good catch, as our AI overlords like to say

patrickzib (approved):
LGTM

PR #3539, opened 2026-06-21
[ENH] Improve coverage for similarity search

TonyBagnall:
Improves test coverage for `aeon/similarity_search` towards 100%, measured
  with `NUMBA_DISABLE_JIT=1` (coverage.py can't trace into compiled `@njit`
  code). Three real bugs were found along the way; per the maintainer's
  direction, they are **not fixed here** — work was stopped on each affected
  class as soon as the bug surfaced, and they'll be filed as separate
  good-first-issues instead.

  ### Coverage

  | File | Before | After |
  |------|--------|-------|
  | `collection/_base.py` | 94% | 100% |
  | `series/_base.py` | 94% | 100% |
  | `series/_commons.py` | 86% | 98% |
  | `series/motifs/_stomp.py` | 58% | 99% |

  - **`collection/_base.py` / `series/_base.py`**: added tests for the
    channel-count-mismatch `ValueError` in `_check_predict_series_format`,
    using the existing `MockCollectionSimilaritySearch` /
    `MockSeriesSimilaritySearch` test doubles.
  - **`series/_commons.py`**: added tests for `_check_X_index`'s valid/None/
    out-of-range branches, the empty-distance-list handling in
    `_extract_top_k_motifs`, and the threshold early-exit branch in
    `_extract_top_k_from_dist_profile`.
  - **`series/motifs/_stomp.py`**: there were previously no tests for the
    `StompMotif` class itself (only its internal `_stomp`/`_stomp_normalized`
    numba kernels were tested directly). Added tests for `fit_predict`,
    `predict` against a different series, both `k_motifs`/`r_motifs` ranking
    methods, the invalid-extraction-method `ValueError`, `_get_test_params`
    (including its `NotImplementedError` branch), and the self-matrix-profile
    exclusion-zone branch in both numba kernels.

  ### Dead code removed

  - `series/_commons.py`: removed an unreachable `if search_k == 0: break`
    guard in `_extract_top_k_from_dist_profile`. The surrounding `while`
    loop's `np.any(mask)` condition already guarantees `available_indices` is
    non-empty whenever that line is reached, so `search_k` (`= min(k,
    len(available_indices))`) can never be `0` there.


baraline (approved):
Still need to work on the Old PR that rework similarity search to be more clearly defined as a module to not mix in motif discovery, but no hurt getting this in. LGTM

PR #3541, opened 2026-06-22
[BUG] fix: move RandomChannelSelector p validation from __init__ to _fit

### LINKED ISSUE #3490: [BUG] RandomChannelSelector validates p in __init__ instead of _fit (violates sklearn clone contract)

BB0813:
## Fix

Fixes #3490

### Problem

`RandomChannelSelector.__init__` validates that `p` is in `(0, 1]`, but the sklearn clone contract requires that `__init__` only stores parameters without validation. This makes `RandomChannelSelector` inconsistent with other channel selectors in the same package.

### Solution

Move the range check for `p` from `__init__` to `_fit`.

### Changes

- `_random.py`: Removed validation from `__init__`, added it to `_fit`
- `test_random.py`: Updated test to expect `ValueError` at `fit_transform` time

### Testing

Test passes with the fix applied.

### DIFF
--- a/aeon/transformations/collection/channel_selection/_random.py
+++ b/aeon/transformations/collection/channel_selection/_random.py
@@ -43,16 +43,16 @@ class RandomChannelSelector(BaseChannelSelector):
     }
 
     def __init__(self, p=0.4, random_state=None):
-        if p <= 0 or p > 1:
-            raise ValueError(
-                "Proportion of channels to select should be in the range (0,1]."
-            )
         self.p = p
         self.random_state = random_state
         super().__init__()
 
     def _fit(self, X, y):
         """Randomly select channels to retain."""
+        if self.p <= 0 or self.p > 1:
+            raise ValueError(
+                "Proportion of channels to select should be in the range (0,1]."
+            )
         rng = check_random_state(self.random_state)
         to_select = math.ceil(self.p * X.shape[1])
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
Thanks for the PR, it all looks good and as a new contributor Im happy to let you have this issue. However, one small test change requested before approval. The current test now uses:

```python
with pytest.raises(
    ValueError, match="Proportion of channels to select should be in the range."
):
    RandomChannelSelector(p=0).fit_transform(X)
```
That would still pass if the old behaviour came back, because a ValueError raised by __init__ would also be caught by the context manager. Since the bug is specifically about validation happening at fit time rather than construction time, please split construction from fitting:

```python
selector = RandomChannelSelector(p=0)

with pytest.raises(
    ValueError, match="Proportion of channels to select should be in the range."
):
    selector.fit_transform(X)
```

BB0813:
OK. I will modify the original code according to this requirement, and after the modification, I will submit the new code and initiate a new CI test.

TonyBagnall (approved):
thanks, LGTM

TonyBagnall:
@all-contributors add @BB0813 for bug

PR #3543, opened 2026-06-22
[BUG] fix SFA/SFAFast get_words for sliding window configurations

### LINKED ISSUE #3537: [BUG] SFA.get_words() and SFAFast.get_words() crash for any normal sliding-window configuration
### Describe the bug

  ### Root cause

  `_sfa.py`:
  ```python
  def get_words(self):
      return np.array(
      )

  _sfa_fast.py:
  def get_words(self):
      return np.array(
      )

  type inference with:

```python
import numpy as np

  y = np.array([0])

  for cls in (SFA, SFAFast):
      p.transform(X)

```

### Expected results

  array([[[0, 0, 3, 0],
          [1, 0, 3, 0],
          [1, 0, 3, 0],
          [1, 0, 3, 0],
          [1, 0, 3, 0]]])

### Actual results

### Versions

_No response_

axelray-dev:
Fixes #3537

`SFA.get_words()` and `SFAFast.get_words()` crash when `window_size < n_timepoints` (the normal sliding-window configuration). The root cause is `np.squeeze(self.words)` which only removes size-1 axes. When multiple sliding windows exist, `self.words` has shape `(n_cases, n_windows)` with `n_windows > 1`. The squeeze operation does not collapse the `n_windows` dimension, so iterating over the result yields arrays instead of scalars. This causes a numba `TypingError` in `_get_chars()` which expects scalar integers.

## Changes

- `aeon/transformations/collection/dictionary_based/_sfa.py`: Replace `np.squeeze(self.words)` with nested iteration over both `n_cases` and `n_windows` dimensions
- `aeon/transformations/collection/dictionary_based/_sfa_fast.py`: Same fix for `SFAFast`
- `aeon/transformations/collection/dictionary_based/tests/test_sfa.py`: Add 3 regression tests covering sliding window, single window, and `SFAFast` configurations

The fix returns a `(n_cases, n_windows, word_length)` object array, correctly handling all window configurations.

## Testing

All 83 tests in `test_sfa.py` pass (80 existing + 3 new regression tests).

---


TonyBagnall:
thanks for this, I'll wait for @patrickzib to review this

patrickzib:
Thank you. I'll have a look. I am not exactly sure how it was intended. Will come back.

aeon/transformations/collection/dictionary_based/_sfa.py
```diff
@@ -476,9 +476,15 @@ def get_words(self):
         -------
         Array of words
         """
-        words = np.squeeze(self.words)
         return np.array(
```
patrickzib, line 479: Thank you for the addition. 

I am not sure, if this is the way to go. I think extending `_get_chars(words, ...)` to be accept multiple words and have everything in numba with njit is the way to go.

aeon/transformations/collection/dictionary_based/_sfa.py
```diff
@@ -476,9 +476,15 @@ def get_words(self):
...
-            [_get_chars(word, self.word_length, self.alphabet_size) for word in words]
+            [
```
patrickzib, line 480: Also you might want to check ,if `save_words=True` was set?

aeon/transformations/collection/dictionary_based/_sfa_fast.py
```diff
@@ -769,9 +769,15 @@ def get_words(self):
         -------
         Array of words
         """
-        words = np.squeeze(self.words)
         return np.array(
-            [_get_chars(word, self.word_length, self.letter_bits) for word in words]
+            [
```
patrickzib, line 773: Same as _sfa.py

aeon/transformations/collection/dictionary_based/tests/test_sfa.py
```diff
@@ -286,3 +286,55 @@ def test_sfa_dynamic(alphabet_allocation_method):
...
+    assert len(words[0][0]) == 4  # word_length
+
+
+def test_sfa_get_words_single_window():
+    """Test get_words with single window (window_size == n_timepoints)."""
+    X = np.random.rand(2, 1, 15)
+    y = np.array([0, 1])
+
+    # Single window: window_size == n_timepoints
+    p = SFA(word_length=6, alphabet_size=4, window_size=15, save_words=True)
```
patrickzib, line 315: please move all hyper-parameters to variables and check in the assert statements for exactly these. I.e. `word_length = word_length` and below change to `assert len(words[0][0]) == word_length`

aeon/transformations/collection/dictionary_based/tests/test_sfa.py
```diff
@@ -286,3 +286,55 @@ def test_sfa_dynamic(alphabet_allocation_method):
...
+    assert len(words[0][0]) == 6  # word_length
+
+
+def test_sfa_fast_get_words_sliding_window():
+    """Test SFAFast.get_words with sliding window."""
+    X = np.random.rand(4, 1, 25)
+    y = np.array([0, 1, 0, 1])
+
+    # Sliding window
+    p = SFAFast(word_length=5, alphabet_size=4, window_size=12, save_words=True)
```
patrickzib, line 332: please move all hyper-parameters to variables and check in the assert statements for exactly these. I.e. `word_length = word_length` and below change to `assert len(words[0][0]) == word_length`

aeon/transformations/collection/dictionary_based/tests/test_sfa.py
```diff
@@ -286,3 +286,55 @@ def test_sfa_dynamic(alphabet_allocation_method):
...
     assert np.mean(np.log2(p.alphabet_sizes)) <= np.log2(alphabet_size)
+
+
+def test_sfa_get_words_sliding_window():
+    """Test get_words with sliding window (multiple windows per case)."""
+    X = np.random.rand(3, 1, 20)
+    y = np.array([0, 1, 0])
+
+    # Sliding window: window_size < n_timepoints
+    p = SFA(word_length=4, alphabet_size=4, window_size=10, save_words=True)
```
patrickzib, line 297: please move all hyper-parameters to variables and check in the assert statements for exactly these. I.e. `word_length = word_length` and below change to `assert len(words[0][0]) == word_length`

patrickzib (changes requested):
Thank you for the addition. I would prefer a different approach with everything in numba, though.

axelray-dev:
Closing: this repo falls below the 5000-star threshold for external contributions.

TonyBagnall:
what times we live in, bots prowling github for easy wins

PR #3544, opened 2026-06-22
[BUG] fix(padder): validate fill_value to reject arrays and unsupported types

### LINKED ISSUE #3495: [BUG] Padder fill_value as a NumPy array is silently misinterpreted (not a per-series fill)

axelray-dev:
Fixes #3495. Add `fill_value` validation in `__init__`, fix the misleading docstring, and add regression tests for array, list, and `None` inputs.

The root cause was that unsupported non-scalar values could reach `np.pad`, where NumPy arrays such as `np.array([1., 2.])` are interpreted as before/after constant pairs instead of being rejected as unsupported per-series fill values. Constructor-time validation now rejects those inputs before transform time.

Validation run:
- `pytest aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `pre-commit run black --files aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `pre-commit run isort --files aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`
- `pre-commit run ruff --files aeon/transformations/collection/unequal_length/_pad.py aeon/transformations/collection/unequal_length/tests/test_pad.py`

Careful review: `fill_value=True` is now rejected despite being an `int` subclass, matching the scalar numeric contract and avoiding bool-as-constant ambiguity.

### DIFF
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -28,10 +28,9 @@ class Padder(BaseCollectionTransformer):
         shortest series seen in ``fit``. If "max", will pad to the longest series seen
         in ``fit``. If an integer, will pad to that length.
         Calling ``fit`` is not required if ``padded_length`` is an int.
-    fill_value : int, str or Callable, default=0
-        Value to pad with. Can be a float or a statistic string or a numpy array for
-        each time series. Supported statistic strings are "mean", "median", "max",
-        "min".
+    fill_value : int, float, str or Callable, default=0
+        Value to pad with. Scalar (int/float), statistic string (mean, median, max,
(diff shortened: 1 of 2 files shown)


axelray-dev:
Closing: this repo falls below the 5000-star threshold for external contributions.

Issue #3545, opened 2026-06-22
[ENH] Guard ETS likelihood/AIC calculation against zero residual variance


TonyBagnall:
### Describe the feature or idea you want to propose

ETSForecaster/AutoETS should guard the likelihood/AIC calculation against zero residual variance. At present, a perfectly fitted model can produce variance == 0, which then sends the log-likelihood/AIC path through log(0) and produces non-finite values. These are subsequently treated as invalid losses, so a valid perfect or near-perfect ETS fit can be discarded.

This surfaced in the Codecov workflow because that job runs the full test suite with Numba disabled:

NUMBA_DISABLE_JIT=1 python -m pytest -n logical --cov=aeon --cov-report=xml --timeout 1800

The observed failure was:

FAILED aeon/forecasting/stats/tests/test_ets.py::test_autoets_uses_provided_seasonal_period
assert 0 != 0
 + where 0 = AutoETS(seasonal_period=4).seasonality_type_

The test uses a simple repeated seasonal series:

y = np.tile(np.array([10.0, 30.0, 12.0, 28.0]), 8)
forecaster = AutoETS(seasonal_period=4)
forecaster.fit(y)
assert forecaster.seasonality_type_ != 0

The series is deliberately seasonal, but it is also exactly periodic. This can lead to a zero residual variance in the ETS likelihood calculation.

Cause

In _ets_fit, the likelihood/AIC calculation currently uses the raw residual variance:

```python
variance = sse_ / n_timepoints
likelihood = -0.5 * n_timepoints * (
    np.log(2 * np.pi) + np.log(variance) + 1
)
aic = -2 * likelihood + 2 * k_
```

If sse_ == 0, then variance == 0, so np.log(variance) is -inf. This leads to a non-finite likelihood/AIC. The loss dispatch logic then treats non-finite objective values as invalid and returns np.inf.

For AutoETS(seasonal_period=4), this can mean that seasonal candidates are tried but not retained, even though the provided seasonal period is valid and the seasonal model is appropriate. AutoETS can then fall back to a non-seasonal model, giving seasonality_type_ == 0.

### Describe your proposed solution

Clamp the residual variance used for likelihood/AIC to a small positive value before taking the logarithm.

For example:

variance = max(sse_ / n_timepoints, EPS)

or, if EPS is not appropriate for this scale:

variance = max(sse_ / n_timepoints, np.finfo(float).tiny)

[...]

PR #3546, opened 2026-06-22
[ENH] Add noise to ets test

TonyBagnall:
Temporary solution to the problem identified in #3545 in order to get overnights passing with codcov

ETS fails when seasonality fit is perfect, and when you turn numba off this causes a numeric error. Solution is a guard against zero variance we use throughout the toolkit, but I want to do that carefully and standardise better across tool kit, hence this short term fix to get code cov working


dguijo (approved):
GJ, @TonyBagnall

PR #3547, opened 2026-06-22
[BUG] Validate RandomChannelSelector p during fit

### LINKED ISSUE #3490: [BUG] RandomChannelSelector validates p in __init__ instead of _fit (violates sklearn clone contract)

ProfTrader:
## Summary

Moves `RandomChannelSelector.p` validation from construction to `_fit`, matching aeon's estimator convention that constructors store parameters unchanged and validation happens during fitting.

## Related issue

Fixes #3490

## Changes

- Remove the `p` range check from `__init__`.
- Validate `p` at the start of `_fit` before selecting channels.
- Update the existing test so invalid `p` can be constructed but raises on `fit`.

## Testing

- `uv run pytest aeon/transformations/collection/channel_selection/tests/test_random.py -q`
- `git diff --check`

### DIFF
--- a/aeon/transformations/collection/channel_selection/_random.py
+++ b/aeon/transformations/collection/channel_selection/_random.py
@@ -43,16 +43,16 @@ class RandomChannelSelector(BaseChannelSelector):
     }
 
     def __init__(self, p=0.4, random_state=None):
-        if p <= 0 or p > 1:
-            raise ValueError(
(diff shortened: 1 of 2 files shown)


TonyBagnall:
hi, this was solved by a new contributor, thanks anyway though

PR #3548, opened 2026-06-22
[BUG] Validate p in _fit instead of __init__ for RandomChannelSelector

### LINKED ISSUE #3490: [BUG] RandomChannelSelector validates p in __init__ instead of _fit (violates sklearn clone contract)

LeoLee0812:
#### Reference Issues/PRs

Fixes #3490

`RandomChannelSelector.__init__` raised `ValueError` when `p` was outside
`(0, 1]`. Validating in the constructor violates the scikit-learn clone
contract: `__init__` should only store parameters unchanged and defer
validation to `fit`. This is also inconsistent with the other selectors in
the same package (`ChannelScorer`, `ElbowClass*`), which validate at the
start of `_fit`.

This PR:
- Moves the `p` range check from `__init__` to the start of `_fit`, leaving
  `__init__` to only assign `self.p` / `self.random_state`.
- Updates `test_random` so an invalid `p` is accepted at construction and
  the `ValueError` is asserted at `fit` time instead.

No.

#### Any other comments?

Behaviour is unchanged except for *when* the error is raised (at `fit`
rather than at construction). The error message is preserved. All 16 tests
in `channel_selection/tests/` pass locally and pre-commit (ruff, black,
isort, codespell) is clean.

### DIFF
--- a/aeon/transformations/collection/channel_selection/_random.py
+++ b/aeon/transformations/collection/channel_selection/_random.py
@@ -43,16 +43,16 @@ class RandomChannelSelector(BaseChannelSelector):
     }
 
     def __init__(self, p=0.4, random_state=None):
-        if p <= 0 or p > 1:
-            raise ValueError(
-                "Proportion of channels to select should be in the range (0,1]."
-            )
         self.p = p
         self.random_state = random_state
         super().__init__()
 
     def _fit(self, X, y):
         """Randomly select channels to retain."""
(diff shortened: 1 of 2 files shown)


TonyBagnall:
hi, this was fixed by someone before you, sorry. More good first issues soon

PR #3552, opened 2026-06-22
[BUG] Fix AutocorrelationFunctionTransformer crash with default n_lags=None

### LINKED ISSUE #3551: [BUG] AutocorrelationFunctionTransformer crashes with default n_lags=None (float used as array dimension)
### Describe the bug

`AutocorrelationFunctionTransformer` crashes for **any** input when used with its
documented default `n_lags=None`. In `_transform`, the default branch computes
`lags = n_timepoints / 4` (true division), producing a `float`. That float is then
used as an array dimension in `np.zeros((n_cases, n_channels, lags))` and as the
`max_lag` argument of the numba kernel, which raises `TypeError`.

The transformer is therefore unusable in its default configuration. The bug is
masked because the only docstring/test example passes an explicit integer
(`n_lags=10`), so the `None` path is never exercised.

The sibling `AutoCorrelationSeriesTransformer` (`aeon/transformations/series/_acf.py`)
handles the same default correctly with `int(max(1, X.shape[1] / 4))`.

`aeon/transformations/collection/_acf.py:69`

```python
from aeon.transformations.collection import AutocorrelationFunctionTransformer

X = make_example_3d_numpy(n_cases=4, n_channels=2, n_timepoints=20,
                          random_state=0, return_y=False)

# default n_lags=None
AutocorrelationFunctionTransformer().fit_transform(X)
```

### Expected results

An array of shape `(4, 2, 5)` (with `5 == n_timepoints // 4`), identical to passing
`n_lags=5` explicitly.

### Actual results

```
TypeError: 'float' object cannot be interpreted as an integer
```

### Versions

aeon 1.4.0 (main), numpy 2.3.5, Python 3.12

### Suggested fix

Cast the default to an integer, mirroring the series sibling:

```python
lags = int(max(1, n_timepoints / 4))
```

---

This report was written with the assistance of AI. The diagnosis was reproduced and
verified by a human against the current source.

glitch-ux:
#### Reference Issues/PRs

Fixes #3551.

`AutocorrelationFunctionTransformer._transform` computed the default number of
lags as `lags = n_timepoints / 4`, which is a `float`. That value is then used as
an array dimension (`np.zeros((n_cases, n_channels, lags))`) and as the numba
kernel's `max_lag`, so the transformer raised `TypeError: 'float' object cannot be
interpreted as an integer` for **every** input under its documented default
`n_lags=None`. The `None` path was never covered because the existing docstring and
test only pass an explicit integer `n_lags`.

The fix casts the default to an int the same way the sibling
`AutoCorrelationSeriesTransformer` already does
(`aeon/transformations/series/_acf.py`):

```python
lags = int(max(1, n_timepoints / 4))
```

`max(1, ...)` keeps at least one lag for very short series, matching the sibling's
behaviour. For the common case the result is unchanged numerically: the default now
equals passing `n_lags = n_timepoints // 4` explicitly (verified in the new test).

Added `test_acf_default_n_lags` covering the previously-crashing `n_lags=None` path.

No.

#### Any other comments?

The pre-existing docstring example (`n_lags=10`) is unchanged; its doctest output is
identical.

---

### DIFF
--- a/aeon/transformations/collection/_acf.py
+++ b/aeon/transformations/collection/_acf.py
@@ -66,7 +66,7 @@ class AutocorrelationFunctionTransformer(BaseCollectionTransformer):
     def _transform(self, X, y=None):
         n_cases, n_channels, n_timepoints = X.shape
         if self.n_lags is None:
-            lags = n_timepoints / 4
+            lags = int(max(1, n_timepoints / 4))
         else:
             lags = self.n_lags(X) if callable(self.n_lags) else self.n_lags
         if lags > n_timepoints - self.min_values:
--- a/aeon/transformations/collection/tests/test_acf.py
+++ b/aeon/transformations/collection/tests/test_acf.py
@@ -12,3 +12,13 @@ def test_acf():
     acf = AutocorrelationFunctionTransformer(n_lags=100)
     with pytest.raises(ValueError, match=r"must be smaller than n_timepoints - 1"):
         acf.fit_transform(X)
+
+
+def test_acf_default_n_lags():
+    """Test the default n_lags=None resolves to an integer n_timepoints // 4."""
+    X = np.random.random((4, 2, 20))
+    default = AutocorrelationFunctionTransformer().fit_transform(X)
+    # default lags = n_timepoints / 4 = 5
+    assert default.shape == (4, 2, 5)
+    explicit = AutocorrelationFunctionTransformer(n_lags=5).fit_transform(X)
+    np.testing.assert_allclose(default, explicit)


TonyBagnall (approved):
Thanks, this is a good bug report and solution to go with it. I reproduced the crash on main (the default really does fail for every input, since `min_values=0` means the clamp at line 72 never fires and `lags` stays a float), and confirmed the fix resolves it: the default now resolves to `n_timepoints // 4` and is numerically identical to passing `n_lags` explicitly.

TonyBagnall:
@all-contributors add @glitch-ux for bug

Issue #3261, opened 2026-01-24
[ENH] Optimize memory complexity for elastic distances


SebastianSchmidl:
### Describe the feature or idea you want to propose

@raphaelgimenezneto reduced the memory complexity of the DTW distance computation by using a standard dynamic programming optimization. This optimization approach is applicable to almost all of our elastic distances (with minor or major additions).

The optimization drastically reduces the memory usage (from quadratic to linear complexity) and, due to higher data locality, also leads to a runtime speedup of ~2x. 

### Describe your proposed solution

We should apply the dynamic programming optimization described in #3257 to our other elastic distances as well.

For each distance,

- implement the optimization for the distance computation
- test whether the new implementation produces the same results as the original one
- perform a performance benchmark to demonstrate the memory reduction and runtime speedup

The following distances (also see sub-issues) should be optimized (in order from easy to more complex):

- [x] DTW #3258
- [x] DDTW (for free with #3258 because it calls `_dtw_distance()` internally)
- [ ] ADTW (additional `warp_penalty` parameter)
- [ ] WDTW (additional `g` parameter and different cost function)
- [ ] SoftDTW (additional `gamma` parameter and different cost function)
- [ ] EDR (different margin initialization and cost function, additional `epsilon` parameter)
- [ ] ERP (different margin initialization and cost function, additional `g` parameter)
- [ ] LCSS (different margin initialization, uses other matrix references to compute cost, additional `epsilon` parameter)
- [ ] MSM (also dynamic programming, but requires a couple of additional modifications)
- [ ] TWE (also dynamic programming, but requires a couple of additional modifications)
- ~~ShapeDTW~~ (not applicable because it operates on the full cost matrix)

MatthewMiddlehurst:
Hi @raphaelgimenezneto, could you assign yourself to this if you want to tackle these after the DTW one? Would be best to wait for that to merge first though so we can discuss any minor style or framework changes which would apply to all the others as well most likely.

raphaelgimenezneto:
@aeon-actions-bot assign @raphaelgimenezneto

raphaelgimenezneto:
Hi @MatthewMiddlehurst,

Thank you so much for the clear guidance and the vote of confidence. I truly appreciate the team's commitment to maintaining a high standard for contributions, and I'd be honored to take ownership of this initiative.

[...]

PR #3258, opened 2026-01-20
[ENH] Implement O(N) space complexity for dtw_distance

raphaelgimenezneto:
#### Reference Issues/PRs

Resolves #3257 

---

#### What does this implement/fix? Explain your changes.

This PR refactors the core `_dtw_distance` helper function to resolve a major performance and scalability bottleneck. It replaces the `O(N*M)` space complexity implementation with a space-efficient `O(M)` algorithm.

**The Problem:**
The previous implementation calculated the full cost matrix, leading to massive memory consumption (several GBs) and `MemoryError` for long time series. This made DTW unfeasible for large-scale datasets.

**The Solution:**
This change implements a space-efficient algorithm that calculates the distance iteratively using only two row vectors. This eliminates the need to allocate the full matrix, providing two key benefits:
1.  **Massive Memory Reduction:** Unlocks the ability to process massive time series.
2.  **Consistent ~2x Speedup:** Achieved through drastically improved CPU cache locality.

The public API remains unchanged, and the numerical output is identical to the baseline.

---

### 🚀 Primary Benefit: >99.99% Memory Reduction

This enhancement makes DTW viable for datasets of virtually any size. The memory footprint is now constant and negligible, regardless of series length. For a series of length N=20,000, the memory usage was reduced from over **3 GB** to just **0.3 MB**.

**Peak Memory Scaling (Log Scale):**
graph1
**Benchmark Data (Peak Memory Allocation):**
| Series Length (N) | Original Peak Memory | Optimized Peak Memory | Reduction Factor |
| :--- | :--- | :--- | :--- |
| 1,000 | 7.65 MB | 0.01 MB | ~500x |
| 5,000 | ~190 MB (estimated) | 0.07 MB | ~2,700x |
| 10,000 | ~763 MB (estimated) | 0.15 MB | ~5,000x |
| 20,000 | **~3,052 MB** | **0.30 MB** | **~10,000x** |
*Note: The O(NM) implementation consistently failed with an Out-of-Memory error on a 16GB RAM test machine for N > 9000.*

---

### ⚡️ Secondary Benefit: ~2x Speedup

As a direct result of improved cache locality, the new implementation is also consistently **~2x faster** for any non-trivial series length.

**Execution Time Scaling (up to N=10,000):**
graph2
**Benchmark Data (Mean of 20 runs):**
| Size | Original (s) | Optimized (s) | Speedup |
|---|---|---|---|
[...]

Issue #3257, opened 2026-01-20
[ENH] Reduce dtw_distance memory complexity from O(N^2) to O(N)


raphaelgimenezneto:
### Describe the feature or idea you want to propose

Yes, this request is related to a significant performance and scalability problem in the current dtw_distance implementation.
The function currently allocates the full O(N*M) cost matrix in memory. For long time series (e.g., N > 10,000), this leads to very high memory consumption (several gigabytes) and can cause MemoryError. This makes it unfeasible to use DTW on large-scale datasets, especially in parallel environments where multiple workers would amplify the memory pressure. This also results in poor cache performance, limiting execution speed even for moderately sized series.

### Describe your proposed solution

My proposed solution is to refactor the internal helper function responsible for the distance calculation (_dtw_distance) to use a space-efficient algorithm with O(M) memory complexity.
This approach calculates the distance iteratively using only two row vectors to store the costs of the previous and current rows, avoiding the allocation of the full matrix entirely. This is a standard dynamic programming optimization pattern that fits perfectly here.

The key benefits are:
- Massive Memory Reduction: Benchmarks show a >99.9% decrease in peak memory usage, making DTW viable for datasets of virtually any size.
- Consistent Speedup (~2x): The improved data locality results in a consistent ~2x speedup for non-trivial series lengths.

This change would maintain the existing API and numerical correctness while drastically improving the performance and scalability of all estimators that rely on distance="dtw".

### Describe alternatives you've considered, if relevant

The alternative is to maintain the current implementation. This is simpler from a code-sharing perspective (as dtw_distance reuses dtw_cost_matrix), but it comes at a very high performance and memory cost that limits the practical application of the library for many real-world use cases involving long time series.

### Additional context

I have already implemented and benchmarked a prototype of this solution. It successfully passes the entire local test suite (pytest aeon/), confirming numerical equivalence with the baseline.
I am ready to submit a Pull Request with the implementation and detailed benchmarks for review. Opening this issue first to ensure this is a welcome enhancement.

raphaelgimenezneto:
@aeon-actions-bot assign @raphaelgimenezneto

PR #3562, opened 2026-06-24
[ENH] msm memory and speed improvements

### LINKED ISSUE #3268: [ENH] Optimize memory complexity for MSM distance
Apply the dynamic programming optimization described in #3257 to the MSM distance function.

MSM has two code paths: one for computing the distance for independent channels and one for computing it for dependent channels. Both paths can be optimized but require some additional considerations.

TonyBagnall:
closes #3268 
part of #3261 

following #3257 and #3610 this implements MSM distance without the full m x m matrix, by just storing the last seen rows in the cost matrix. As with DTW, huge memory reduction on calls to msm_distance and 2x speed up due to less time spent allocating. It also deprecates the bounding box capability, which is non standard and memory intensive, see #3610 

The distance is now computed iteratively with two row buffers (prev and curr) of length min(n, m), reducing space from O(n·m) to O(min(n, m)) while producing bit-identical results.

- _msm_distance now dispatches to two new memory-light helpers instead of slicing a full cost matrix:
- _msm_dependent_distance — two-row buffer over the multivariate recurrence.
 - _msm_independent_distance — sums per-channel results from a new _univariate_msm_distance helper. Because the independent cost matrix is the element-wise sum of the per-channel matrices, the bottom-right cell of the sum equals the sum of the per-channel bottom-right cells, so each channel reduces to a scalar independently (no n × m accumulator needed).
  - Both helpers reuse the swap-and-transpose trick from #3258: if x is shorter than y they are swapped and the bounding matrix transposed, so the row buffers are sized by the smaller dimension. This is valid because MSM is symmetric — its move cost is symmetric and both cost functions (_cost_independent, _cost_dependent) are symmetric in their last two arguments, so the recurrence is invariant under transposition.

- The full cost matrix is intentionally retained where it is genuinely needed:
    - _msm_independent_cost_matrix / _independent_cost_matrix and _msm_dependent_cost_matrix are kept as-is.
    - These still back msm_cost_matrix (which by definition returns the whole matrix) and msm_alignment_path (which needs the full matrix to backtrack the alignment via compute_min_return_path).
  - The cost functions _cost_independent / _cost_dependent, the pairwise / multiple-to-multiple drivers, and the bounding-matrix / windowing / Itakura logic are untouched. We might want to revisit this. 

note there is now a bipass around the deprecated bounding box which temporarily increases code, but will give the benefits of optimisation prior to deprecation in the stanard non-bounding case


TonyBagnall:
image

TonyBagnall:
image

TonyBagnall:
image
independent had an extra memory allocation, hence improvement over theoretical

TonyBagnall:
dependent not faster, pointwise distance dominates, but memory same. 
image

TonyBagnall:
note there are other possible tweaks to gain some time

TonyBagnall:
exact equivalence with old implementation shown in all experiments

TonyBagnall:
will revise after #3610 to remove dense bounding box

MatthewMiddlehurst:
Could you explain the deprecations a bit? I wont say to keep them because i generally let this module be, but I may have missed the reason for why.

When talking about this with Claude found some issues with the DTW change which may be relevant. #3827 and #3826

MatthewMiddlehurst:
Feel free to ignore this as well if it doesn't fit with the code structure we currently have or doesn't actually work, but I was convinced enough that this seems to be a real improvement so posting it just as an FYI really. There are a few functions that are expected and testing in the module which I am not completely aware of. Not sure if ditching `_univariate_squared_distance` in the method is something we would want to do.

Claude thinks dependent (`independent=False`) doesn't benefit much speed wise at lest as "Almost all of the time goes on memory allocation in `_cost_dependent`. It is called about twice for every cell, and each call creates a new array." It's suggested change is to compute the four squared distances in a single loop without the temporary array:

```diff
 @njit(cache=True, fastmath=True)
 def _cost_dependent(x: np.ndarray, y: np.ndarray, z: np.ndarray, c: float) -> float:
-    diameter = _univariate_squared_distance(y, z)
-    mid = (y + z) / 2.0
-    distance_to_mid = _univariate_squared_distance(mid, x)
+    diameter = 0.0
+    distance_to_mid = 0.0
+    dist_to_q_prev = 0.0
+    dist_to_c = 0.0
+    for i in range(x.shape[0]):
+        difference = y[i] - z[i]
+        diameter += difference * difference
+        difference = (y[i] + z[i]) / 2.0 - x[i]
+        distance_to_mid += difference * difference
+        difference = y[i] - x[i]
+        dist_to_q_prev += difference * difference
+        difference = z[i] - x[i]
+        dist_to_c += difference * difference
 
     if distance_to_mid <= (diameter / 4.0):
         return c
-    else:
-        dist_to_q_prev = _univariate_squared_distance(y, x)
-        dist_to_c = _univariate_squared_distance(z, x)
-        if dist_to_q_prev < dist_to_c:
-            return c + dist_to_q_prev
-        else:
-            return c + dist_to_c
+    return c + min(dist_to_q_prev, dist_to_c)
```

The loop reads `y` and `z` over `x`'s channels without bounds checks, so it suggests the channel check should go once at the top of `_msm_dependent_distance_unbounded`, `_msm_dependent_distance` and `_msm_dependent_cost_matrix`:

```python
    if x.shape[0] != y.shape[0]:
        raise ValueError(
            "x and y must have the same number of channels for dependent MSM."
        )
```

At the moment, mismatched channel counts only fail by accident, with numba's `unable to broadcast` error. It would also be worth adding a test that compares `_cost_dependent` with a plain NumPy version of the formula, and one for the channel check.

Results match this PR to within 5.3e-16 relative difference for the distance and the full cost matrix, with and without parameters. Experiments:

**Dependent MSM, 3 channels (seconds, mean ± std)**

| N | PR | PR + fix | Speedup |
|---|---|---|---|
| 250 | 0.01542 ± 0.00029 | 0.00101 ± 0.00002 | 15.3x |
| 500 | 0.06049 ± 0.00111 | 0.00406 ± 0.00004 | 14.9x |
| 1,000 | 0.24641 ± 0.00331 | 0.01614 ± 0.00035 | 15.3x |
| 2,000 | 0.97847 ± 0.00753 | 0.06520 ± 0.00138 | 15.0x |
| 4,000 | 3.92418 ± 0.02257 | 0.26550 ± 0.01345 | 14.8x |
| 8,000 | 15.70287 ± 0.06896 | 1.05371 ± 0.01056 | 14.9x |

With the fix, dependent MSM takes about the same time as independent MSM on the same data.

msm_dependent_runtime_vs_length

**Speedup against number of channels (N = 1,000)**

| Channels | 1 | 2 | 3 | 5 | 10 | 20 | 50 |
|---|---|---|---|---|---|---|---|
| Speedup | 23.6x | 19.0x | 15.6x | 12.3x | 3.6x | 3.2x | 2.5x |

**Test x train distance matrix** (`msm_pairwise_distance(..., independent=False, n_jobs=1)`)

| Dataset | PR | PR + fix | Speedup |
|---|---|---|---|
| BasicMotions | 6.66 s | 0.74 s | 9.1x |
| JapaneseVowels | 14.11 s | 4.05 s | 3.5x |

TonyBagnall:
> Could you explain the deprecations a bit? I wont say to keep them because i generally let this module be, but I may have missed the reason for why.
> 
> When talking about this with Claude found some issues with the DTW change which may be relevant. #3827 and #3826

sure, the deprecations remove the capability to have bounding boxes with MSM. They are not part of the normal definition, chris put them in when first implementing this as an added feature, but it introduces a mxm memory cost of masks which we want to remove. The context aware nature of the algorithm make it a bit odd to have bounding boxes, and Chris and I agreed it best to just get rid of them. Given its a parameter, I thought we had to deprecate, but if not I'll just get rid!

TonyBagnall:
nice,

  
  Faster dependent MSM: _cost_dependent no longer allocates                                                                                                                                                         
                                                                                                                                                                                                                    
  Following @MatthewMiddlehurst's suggestion, _cost_dependent now makes a single pass over the channels. It no longer calls _univariate_squared_distance four times, and no longer creates a temporary mid = (y + z)
  / 2.0 array. That allocation happened about twice for every cell of the cost matrix and dominated the runtime of dependent MSM. The formula is unchanged, and _univariate_squared_distance is still used for the  
  move cost.                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  The new loop has no bounds checks, so dependent MSM now checks channel counts explicitly. The check runs once, before the loops, in _msm_dependent_distance_unbounded, _msm_dependent_distance and                
  _msm_dependent_cost_matrix, and raises:                                                                                                                                                                           
                                                                                                                                                                                                                    
  ValueError: Dependent MSM requires x and y to have the same number of channels.                                                                                                                                   
                                                                                                                                                                                                                    
  Previously, mismatched channel counts only failed by accident, with numba's unable to broadcast error. The check is not in _cost_dependent itself, which runs on every cell. The error also comes through cleanly 
  from msm_pairwise_distance, including inside its parallel loop. Independent MSM is unchanged.      

 Correctness                                                                                                                                                                                                       
                                                                                                                                                                                                                    
  - Numerically equivalent to the old cost function. A new test compares _cost_dependent with a NumPy copy of the old formula on random vectors with 1, 3 and 10 channels (rtol=1e-12); these reach both branches.  
    Results are expected to be numerically equivalent rather than guaranteed bit-identical, because of fastmath and the different order of summation.                                                               
  - Identical results on real data. Test × train distance matrices are the same before and after in every cell (maximum absolute difference 0.0) for BasicMotions (40 × 40) and JapaneseVowels (370 × 270).         
  - Existing tests pass. All distance tests pass, including the dependent MSM parameter tests and the elastic correctness, cost matrix, alignment path and bounding tests.                                          
  - New channel-check test. It covers the default distance path, the deprecated bounded path and the cost matrix.                                                                                                   
                                                                                                                                                                                                                    
  Timing                                                                                                                                                                                                            
                                                                                                                                                                                                                    
  msm_pairwise_distance(X_test, X_train, independent=False, n_jobs=1), timed after a warm-up run, mean ± std over 5 repeats:                                                                                        
                                                                                                                                                                                                                    
  ┌────────────────┬────────────────┬───────────────┬──────────┐                                                                                                                                                    
  │    Dataset     │     Before     │     After     │ Speed-up │                                                                                                                                                    
  ├────────────────┼────────────────┼───────────────┼──────────┤                                                                                                                                                    
  │ BasicMotions   │ 6.79 ± 0.17 s  │ 1.69 ± 0.17 s │ 4.0x     │                                                                                                                                                    
  ├────────────────┼────────────────┼───────────────┼──────────┤                                                                                                                                                    
  │ JapaneseVowels │ 15.07 ± 1.09 s │ 4.48 ± 0.04 s │ 3.4x     │                                                                                                                                                    
  └────────────────┴────────────────┴───────────────┴──────────┘                                                                                                                                                    
                                                                                                                                                                                                                    
  The "before" times are close to Matthew's (6.66 s and 14.11 s), but the BasicMotions speed-up is smaller than his 9.1x. That's probably down to hardware or OS; I haven't looked into it

MatthewMiddlehurst:
Sounds good. Deprecation version needs to be updated.

We should also probably try to resolve https://github.com/aeon-toolkit/aeon/pull/3828 or the dtype issue first, but wont say thats a must do.

PR #3563, opened 2026-06-24
[ENH] twe memory and speed improvements

TonyBagnall:
Part of #3261 (memory optimization for elastic distances). Applies the two-row dynamic-programming optimization established for DTW in #3258 (and MSM in #3562) to TWE. Deprecate bounding box. 

closes #3269 

  - Public API and outputs are unchanged. Signatures, defaults (nu=0.001, lmbda=1.0), docstrings, and return values of twe_distance, twe_pairwise_distance, twe_cost_matrix, and twe_alignment_path are all the same.
  - The full cost matrix is intentionally retained. _twe_cost_matrix is untouched and still backs twe_cost_matrix (which by definition returns the whole matrix) and twe_alignment_path (which needs the full matrix to backtrack the alignment via
  compute_min_return_path).
  - _pad_arrs, the pointwise Euclidean cost, the pairwise / multiple-to-multiple drivers are untouched. All callers of _twe_distance (the pairwise functions) already pass padded arrays and are unaffected.

  Validation

  - Equivalence: matches the previous full-matrix result to floating-point tolerance — max absolute difference ~3.5e-14 over 400 randomized cases (univariate/multivariate, varying nu/lmbda, windows, and unequal lengths), growing to ~1e-9 at n=8000
  (≈1e-13 relative). 

  Benchmark (univariate, n = m, no window)
| n | Runtime baseline → opt | Peak mem baseline → opt | Mem reduction |
|---:|:------------------------|:--------------------------|--------------:|
| 1,000 | 15.6 ms → 13.6 ms (1.1×) | 9.0 MB → 1.0 MB | 9.0× |
| 2,000 | 67.9 ms → 60.1 ms (1.1×) | 36.0 MB → 4.0 MB | 9.0× |
| 4,000 | 0.26 s → 0.21 s (1.3×) | 144.1 MB → 16.1 MB | 9.0× |
| 8,000 | 1.89 s → 0.85 s (2.2×) | 576.3 MB → 64.3 MB | 9.0× |

  The optimization eliminates the float64 cost matrix (n²·8 bytes); the ~9× reduction reflects that the remaining memory of the public call is the O(n²) boolean bounding matrix (n²·1 byte, window=None).

PR #3565, opened 2026-06-24
[ENH] wdtw memory and speed improvements

### LINKED ISSUE #3263: [ENH] Optimize memory complexity for WDTW distance
Apply the dynamic programming optimization described in #3257 to the WDTW distance function.

TonyBagnall:
This is the same optimisation pattern as #3562, #3563 and #3566, applied to WDTW.

The aim is to replace the full dynamic-programming cost matrix used by wdtw_distance with a rolling-row implementation, reducing the DP memory requirement from (O(nm)) to (O(\min(n,m))).

For the standard unbounded case, the implementation will also bypass construction of the dense bounding matrix entirely. The existing bounded route can remain for deprecated window/itakura_max_slope calls, while wdtw_cost_matrix and alignment-path functions will continue to construct the full matrix where it is actually required.

The reduced-memory route will also be passed through to pairwise distance calculations.

Code and updated benchmarks to follow.

Closes #3263.


TonyBagnall:
will revise after https://github.com/aeon-toolkit/aeon/pull/3610 to remove dense bounding box

TonyBagnall:
going to make draft until #3562 is in

PR #3566, opened 2026-06-24
[ENH] Erp memory and speed improvements

TonyBagnall:
his PR reduces the memory usage of ERP distance calculations by replacing the full dynamic-programming cost matrix with two rolling rows.

Previously, erp_distance called erp_cost_matrix and returned the final cell, requiring an (O(nm)) matrix even though only the final ERP distance was needed. The new implementation evaluates the same recurrence using two rows of length equal to the shorter series, reducing the DP memory requirement to (O(\min(n,m))).

For the standard unbounded ERP calculation, the implementation also bypasses creation of the dense bounding matrix entirely. The deprecated window and itakura_max_slope routes retain the existing bounded implementation for backwards compatibility. erp_cost_matrix and erp_alignment_path are unchanged in principle, since these functions require the full cost matrix.

The optimisation is also passed through to erp_pairwise_distance, including unequal-length collections, so repeated pairwise calculations use the reduced-memory implementation when no bounding parameters are supplied.

The recurrence and ERP gap penalties are unchanged, including support for both scalar g and multivariate g_arr.

Part of #3261.


TonyBagnall:
image

TonyBagnall:
will revise after https://github.com/aeon-toolkit/aeon/pull/3610 to remove dense bounding box

TonyBagnall:
going to make draft until #3562 is in

PR #3573, opened 2026-06-27
[DOC] Fix unequal length transformer docstrings

kiwoongyoon:
Title: Fix unequal length transformer docstrings

## Summary

Updates stale documentation in the unequal-length collection transformers to match the
current implementation:

- Describe min/max length as being found across series, not channels.
- Document `Padder.fill_value` as scalar/string/callable and include the supported
  `"last"` string.
- Use `padded_length` consistently in the Padder error message.
- Clarify list return cases for Padder and Truncator when unequal-length output may
  remain.
- Note that Truncator `_transform` accepts equal-length 3D numpy input.

## Test evidence

- `python -m compileall -q aeon/transformations/collection/unequal_length` passed.
- Targeted `rg` stale-string check found no remaining matches for the corrected issue
  text.
- `git diff --check` passed.

`pytest` was attempted but could not run because the active Python environment does not
have `pytest` installed. A runtime smoke check also could not run because `numpy` is not
installed in the active environment.

## Risks or notes for maintainers

This is documentation-only apart from one error-message string. Please review whether
`scalar` is the preferred docstring type wording for `Padder.fill_value`.

### DIFF
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -19,7 +19,7 @@ class Padder(BaseCollectionTransformer):
     """Pad unequal length time series to equal, fixed length.
 
     Pads the input dataset to either a fixed length or finds the max/min length
-    series across all series and channels and pads to that with zeroes.
+    series across all series and pads shorter series with ``fill_value``.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


TonyBagnall (approved):
lgtm, part of #3496

PR #3574, opened 2026-06-28
[ENH] speed up RandomShapeletTransform 

TonyBagnall:
Not dependent on #3631, but obviously STC sill be dependent on both that and this, 

Speed up RST without changing outcome. May also redesign and change  threading

 ### Stage 1: _online_shapelet_distance scalar rewrite

Replace traverse, sums, and sums2 list state with scalar left/right state

compute the normalised value on demand, avoid unnecessary slicing. Move unchanging values out of loop

### Stage 2: code restructure

I feel it could benefit from a redesign,. _process_fit_batch has too many responsibilities, fixed and contracted has duplication, _class_dictionary is populated but never read etc

### Stage 3: optimisations 

threading at the moment makes things worse. n_jobs=2/4 is slower than n_jobs=1 for both fit and transform, consistent 
with per-candidate Joblib overhead and Numba entry   points not being configured to release the GIL

1. nogil=True on the scoring/transform kernels — the big one: fit and transform ~3.3x on 4 cores (was ~1.3x; the threads were GIL-bound).
2. Transform kernel njit'd — one kernel over the whole block instead of per-pair Python dispatch.
3. Per-candidate fusion — z-norm + argsort folded into the scoring kernel (one numba crossing per candidate; np.argsort(kind="mergesort") reproduces the old sort exactly).
4. Single reused RNG — reseed-per-candidate instead of 10k RandomState allocations (~16x on that line, bit-exact), plus a clean sampling/scoring split.
5. Hard code IG constants


aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -821,54 +821,76 @@ def _online_shapelet_distance(series, shapelet, sorted_indicies, position, lengt
...
-        subseq = (subseq - mean) / std
-    else:
-        subseq = np.zeros(length)
+    use_std = std > AEON_NUMBA_STD_THRESHOLD
 
     best_dist = 0
-    for i, n in zip(shapelet, subseq):
-        temp = i - n
+    for i in range(length):
+        val = (series[position + i] - mean) / std if use_std else 0.0
```
patrickzib: Is this fast? I.e. calling the 'if' on every value? 

maybe first do normalization in a different loop if use_std is set ?

or divide by std=1.0 ? and set std = 1 if use_std is set?

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -821,54 +821,76 @@ def _online_shapelet_distance(series, shapelet, sorted_indicies, position, lengt
...
-    while traverse[0] or traverse[1]:
-        for n in range(2):
-            mod = -1 if n == 0 else 1
-            pos = position + mod * i
-            traverse[n] = pos >= 0 if n == 0 else pos <= len(series) - length
+            dist = 0
+            use_std = std > AEON_NUMBA_STD_THRESHOLD
+            for j in range(length):
+                idx = sorted_indicies[j]
+                val = (series[pos + idx] - mean) / std if use_std else 0
```
patrickzib: maybe more efficient to do this once outside and use the normalized values here?

patrickzib (commented):
Not a review, just a comment. Maybe calling if on every value is not super efficient? normalization once outside the loops or using / std=1.0 might be faster?

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: no its better taken out, ta

TonyBagnall:
back to this draft properly. Priorities

1. Clean ups (e.g. std dev in loops)
2. njit whole transform block
3. sort out threading
4. optimise

TonyBagnall:
almost done, modest gains on sequential but significant gains on parallel, nogil ftw

image

TonyBagnall:
picture improves with n_cases, and especially with n_jobs

image

done enough on this

aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
```diff
@@ -0,0 +1,103 @@
...
+        X = [
+            rng.normal(
+                size=(_N_CHANNELS, _MIN_UNEQUAL_TIMEPOINTS + i % _N_UNEQUAL_LENGTHS)
+            )
+            for i in range(_N_CASES)
+        ]
+    else:
+        X = rng.normal(size=(_N_CASES, _N_CHANNELS, _N_TIMEPOINTS))
+    y = np.arange(_N_CASES) % _N_CLASSES
+    return X, y
```
baraline: Don't we have functions in the testing package that can generate such data already ?

aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
```diff
@@ -0,0 +1,103 @@
...
+    ]
+    for first_shapelet, second_shapelet in zip(first.shapelets_, second.shapelets_):
+        assert_array_equal(first_shapelet[_VALUES], second_shapelet[_VALUES])
+
+
+def test_fit_transform_matches_fit_without_stale_distance_indices():
+    """Fit-transform caching should not change fitted state or retain cache indices."""
+    X, y = _make_test_data()
+    cached_transformer = RandomShapeletTransform(**_RST_TEST_PARAMS)
+    fitted_transformer = RandomShapeletTransform(**_RST_TEST_PARAMS).fit(X, y)
```
baraline, line 49: Seems to work without it, but won't we need to set the same random seed for this kind of test ?

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -161,7 +179,7 @@ def __init__(
         self.n_channels_ = 0
         self.max_n_timepoints_ = 0
         self.classes_ = []
-        self.shapelets = []
+        self.shapelets_ = []
```
baraline: Doesn't that kind of change fall into deprecation policy?

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -568,51 +589,102 @@ def _log_fit_progress(
...
+
+        ``self.shapelets_`` stays a list of tuples for external use. The ragged
+        shapelet values and sorted indices are packed into single flat arrays
+        with a CSR-style ``offsets`` array (shapelet ``n`` spans
+        ``offsets[n]:offsets[n + 1]``). Flat arrays are plain NumPy, so the
+        fitted estimator pickles cleanly (a Numba typed-list attribute does
+        not) and each transform block still runs in one ``njit`` call.
+        """
+        if len(self.shapelets_) == 0:
+            return
```
baraline: Edge case, but if you first fit a collection, and then call fit again and no shapelets are found, this will keep the previous fit data in. You might want to clear the attribute bellows here.

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -21,25 +21,41 @@
...
 
     Implementation of the binary shapelet transform along the lines of [1]_, [2]_, with
-    randomly extracted shapelets. A shapelet is a subsequence from the train set. The
+    randomly extracted shapelets. A shapelet is a subsequence from the training set. The
     transform finds a set of shapelets that are good at separating the classes based on
     the distances between shapelets and whole series. The distance between a shapelet
-    and a series (called sDist in the literature) is defined as the minimum Euclidean
-    distance between shapelet and all windows the same length as the shapelet.
+    and a series (called sDist in the literature) is the minimum z-normalised squared
+    distance between the shapelet and every window of the same length in the series.
```
baraline: Isn't sDist squared distance divided by series length ?

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
```diff
@@ -471,27 +449,70 @@ def _process_fit_batch(
...
 
         return batch_elapsed, current_kept
 
+    def _set_fit_attributes(self, X, y):
+        """Set fitted collection metadata and resolve shapelet parameters."""
+        self._n_jobs = check_n_jobs(self.n_jobs)
+
+        self.classes_, self._class_counts = np.unique(y, return_counts=True)
+        self.n_classes_ = self.classes_.shape[0]
+        self._class_dictionary = {}
```
baraline: Variable is not used anywhere I think.

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: I wondered if anyone would notice that, it just annoyed me because its not consistent and rdist correctly uses .shapelets_. I can change it back though, its just cosmetic

aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
TonyBagnall: yes we do, I wrote it and forgot, caught most of the equal length ones :)

aeon/transformations/collection/shapelet_based/tests/test_shapelet_transform.py
TonyBagnall: its set in 


_RST_TEST_PARAMS = {
    "n_shapelet_samples": 20,
    "max_shapelets": 6,
    "batch_size": 10,
    "random_state": 0,
}

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: this is true, now
```python
        if len(self.shapelets_) == 0:
            self._transform_values = np.empty(0, dtype=np.float64)
            self._transform_sorted_indices = np.empty(0, dtype=np.int32)
            return
```

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: just a matter of scale for finding the min val, since shapelet length is constant, but you are correct and we do divide by shapelet length, I guess for scale between shapelets (a lot of divisions though, could do it once at the end ....)

TonyBagnall: docstring corrected "is the minimum mean squared distances between the z-normalised shapelet"

TonyBagnall: The confirmation run is decisive: deferred scaling changes fit by effectively 0%, transform is within run-to-run noise, and fit_transform is about 11% slower because it adds a
  feature-matrix scaling pass. It also changes feature bit patterns. I’ve reverted the experiment; the original scaling location is the better tradeoff.

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: quite correct, good spot, removed it. I think that was left from a long time ago

aeon/transformations/collection/shapelet_based/_shapelet_transform.py
TonyBagnall: reverted

TonyBagnall:
thanks for the review @baraline

baraline (approved):
All good for me, comment adressed, nothing else to add.

PR #3580, opened 2026-06-29
[MNT] Enhanced forecasting tests

TonyBagnall:
Better coverage and much faster tests for the forecasting module

### DIFF
--- a/aeon/forecasting/ensembles/tests/test_ensemble.py
+++ b/aeon/forecasting/ensembles/tests/test_ensemble.py
@@ -56,32 +56,6 @@ class _WrongLengthTrajectoryForecaster(_TrajectoryForecaster):
         return self.start + np.arange(prediction_horizon - 1, dtype=float)
 
 
-class _RecursiveRuleForecaster(BaseForecaster):
-    """Test forecaster with predictions based on the latest observed value."""
-
-    def __init__(self, multiplier=1.0, intercept=0.0):
-        self.multiplier = multiplier
-        self.intercept = intercept
-        super().__init__(horizon=1, axis=1)
-
-    def _fit(self, y, exog=None):
-        return self
-
-    def _predict(self, y, exog=None):
-        last_value = np.asarray(y, dtype=float).reshape(-1)[-1]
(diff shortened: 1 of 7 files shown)


alexbanwell1:
## Review

This PR is a test-only maintenance change to the forecasting module: replaces bespoke test-double forecaster classes with `NaiveForecaster` instances, consolidates parametrized exog-validation tests, and reduces sample sizes for speed. All 129 affected tests pass; every numerical assertion and exception-match regex checked out against production source.

### Findings (most severe first)

**`aeon/forecasting/stats/tests/test_tar.py:428-465`** — `test_autotar_parameter_recovery_is_reasonable` was weakened to a tautology and dropped nonzero-threshold coverage. Forcing `ar_order=(1,1)` and `delay=d` makes `p_below_==1`/`p_above_==1`/`delay_==d` guaranteed by construction rather than testing order/delay selection, and the only nonzero-threshold case (r=0.25) was deleted. A regression in threshold search/order selection, especially for r≠0, could land undetected.

**`aeon/forecasting/stats/tests/test_scum.py:235-273`** — `test_scum_default_pool_uses_all_four_components`'s monkeypatch lambdas silently discard the `seasonal_period`/`season_length` kwargs, so the test doesn't actually verify `SCUM` forwards `season_length` to its components. A bug breaking that propagation wouldn't be caught.

**`aeon/forecasting/stats/tests/test_tar.py:428-465`** — `@pytest.mark.parametrize` converted to a manual loop, losing per-case isolation and pytest's per-case reporting.

**`aeon/forecasting/stats/tests/test_scum.py:125-136`** vs **`aeon/forecasting/ensembles/tests/test_ensemble.py:59-64`** — duplicated naive-forecaster-pool pattern with no shared conftest fixture; minor maintainability cost.

Nothing else survived verification — imports, constructor signatures, and dropped test-double coverage were all confirmed fine elsewhere.

---

alexbanwell1:
Doesn't look to me that any of those are particularly problematic, so approving

alexbanwell1 (approved):

PR #3581, opened 2026-06-29
[BUG] Fix SCUM predict refitting

### LINKED ISSUE #3579: [BUG] SCUM.predict(y) does not use the model fitted by SCUM.fit().
### Describe the bug

this is my bad, and it conflates with the whole issue of whether ets should have fit_is_empty as true and make it stateless.

SCUM.predict(y) does not use the model fitted by SCUM.fit(). Instead, SCUM._predict rebuilds a brand-new EnsembleForecaster with brand-new, untrained AutoETS/AutoCES/AutoARIMA/DOTM instances and refits all four from scratch on whatever y is passed to predict, completely ignoring self.ensemble_ and self.forecasters_ that were set during fit.

This means calling fit(y_train) then predict(y_new) gives exactly the same result as never calling fit() at all and just doing a fresh SCUM().fit(y_new) — fit() has no effect on subsequent predict() calls with different data. This breaks the standard aeon/sklearn forecasting contract used by every other forecaster in the module (e.g. ARIMA, RegressionForecaster, AutoETS), where fit() learns parameters once and predict(y_new) applies the already-learned parameters to a new context without re-running model selection.

It also contradicts SCUM's own docstring, which describes ensemble_/forecasters_ as "Fitted ... used to combine the component forecasts" — but these attributes are set and then never read by _predict.

A practical consequence: every call to predict() silently re-runs the full AutoETS/AutoCES/AutoARIMA/DOTM order-selection and optimisation pipeline, which is far more expensive than the lightweight "apply fitted model to new context" inference every other forecaster does, and can select different component model orders than the ones chosen at fit time if predict is called with a different series.

This isn't caught by check_estimator(SCUM) or by the existing test_scum.py tests, because the only fit-then-predict test (test_scum_predict_matches_horizon_one_iterative_forecast) calls predict() with the same y used in fit(), so the discrepancy never surfaces.

mport numpy as np
from aeon.forecasting.stats import SCUM

np.random.seed(0)
y_fit = np.cumsum(np.random.randn(60)) + 100
y_other = np.cumsum(np.random.randn(40)) + 5000  # unrelated series, different scale

scum = SCUM()
scum.fit(y_fit)
print(scum.forecast_)            # forecast learned from y_fit

pred = scum.predict(y_other)
print(pred)                      # 5001.38...

# Compare: fit a brand-new SCUM directly on y_other, bypassing the original fit entirely
fresh = SCUM().fit(y_other)
print(fresh.forecast_)    

### Expected results

scum.predict(y_other) should use the components fitted in scum.fit(y_fit) (i.e. scum.ensemble_) to produce a forecast informed by what was learned from y_fit, the same way ARIMA.predict(y_other) reuses self.phi_/self.theta_/self.beta_ rather than re-running the Nelder-Mead fit.

### Actual results

scum.predict(y_other) is numerically identical to constructing and fitting a brand-new SCUM directly on y_other. The model learned during fit(y_fit) is discarded.

### Suggested fix

SCUM._predict should call self.ensemble_.predict(y) (using the already-fitted ensemble from _fit) instead of calling self._fit_predict(y, 1), which constructs a new EnsembleForecaster and a new set of untrained component forecasters every time.

### Versions

_No response_

TonyBagnall:
fixes #3579 

Fixed SCUM prediction to reuse the fitted ensemble instead of rebuilding/refitting during predict. Added regression tests confirming
  predict does not call _fit_predict after fit, and that negative fitted-ensemble predictions are still clipped when clip_negative=True


alexbanwell1 (commented):
## Review

This PR fixes `SCUM._predict` so it reuses the fitted `ensemble_` (and its already-fitted component forecasters) instead of calling `_fit_predict` again, which previously refit all four models from scratch on every `predict()` call. The new implementation calls `self.ensemble_.predict(y, exog)` and applies the same `clip_negative` logic used in `_fit_predict`.

I traced the change through `EnsembleForecaster.predict()`/`_predict()` and `BaseForecaster.predict()`: `ensemble_.is_fitted` is set `True` by the prior `iterative_forecast()` call inside `_fit_predict`, so the public `predict()` call on `ensemble_` passes its fitted check and forwards `y` to each component's own `predict()` without refitting — this matches the existing pattern used elsewhere in the framework (e.g. `EnsembleForecaster._predict` itself calls `f.predict(y)` on fitted components). `exog` stays `None` throughout since SCUM declares `capability:exogenous: False`, so the pass-through to `ensemble_.predict(y, exog)` is inert. No other callers of `_fit_predict` exist outside this module, so the change is well-contained.

The new tests directly target the regression: `test_scum_predict_uses_fitted_ensemble_not_fit_predict_again` asserts `_fit_predict` is called exactly once across `fit()` + `predict()`, and `test_scum_predict_clips_negative_fitted_ensemble_prediction` confirms clipping still applies on the fitted-ensemble path.

No correctness, cleanup, or convention issues found — the fix is minimal, correctly aligned with the codebase's fit-once/predict-many convention, and the added tests cover the exact regression described in the linked issue. LGTM.

alexbanwell1 (dismissed):

alexbanwell1 (approved):
Approve as above, very minor changes since.

PR #3583, opened 2026-06-30
[BUG] AutoARIMA: respect max_d in the differencing loop (#3577)

### LINKED ISSUE #3577: [BUG] AutoARIMA(max_d=0) can still select one order of differencing
### Describe the bug

Problem

    self.d_ += 1

import numpy as np

forecaster.fit(y)

print(forecaster.d_)

### Expected results

1

### Actual results

forecaster.d_ == 0

### Suggested fix

### Versions

_No response_

CedricConday:
Fixes #3577

## What was wrong

`AutoARIMA._fit` selects the non-seasonal differencing order with:

```python
    differenced_series = np.diff(differenced_series, n=1)
    self.d_ += 1
```

Because the guard is `self.d_ <= self.max_d`, the loop can enter when `self.d_ == self.max_d`, apply one more difference, and increment `self.d_` past the configured maximum. With `max_d=0` (differencing disabled) a non-stationary series is still differenced once, ending at `d_ = 1`.

## Fix

Use `self.d_ < self.max_d`, so at most `max_d` differences are applied.

## Test

Adds `test_autoarima_respects_max_d_zero`: fits `AutoARIMA(max_d=0)` on a trending (non-stationary) series and asserts `d_ == 0`. Verified it fails on `main` (`d_ == 1`) and passes with this change.

---
*AI-assisted, human-reviewed — I'm an AI engineer; I find, fix, and test with AI, then review and verify before opening.*

### DIFF
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -431,7 +431,10 @@ class AutoARIMA(BaseForecaster, IterativeForecastingMixin):
         series = np.array(y.squeeze(), dtype=np.float64)
         differenced_series = series.copy()
         self.d_ = 0
-        while not kpss_test(differenced_series)[1] and self.d_ <= self.max_d:
+        # Stop before exceeding max_d: with `<=`, the loop could enter when
+        # ``self.d_ == self.max_d`` and apply one more difference, pushing the
+        # selected order past the configured maximum (e.g. max_d=0 -> d_=1).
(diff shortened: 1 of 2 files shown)


TonyBagnall:
hi, its failing for formatting, could you run pre-commit before submitting please? thanks

CedricConday:
Thanks! Ran pre-commit — the formatting is clean now.

TonyBagnall:
thanks for the PR, but I am giving this to #3593 because it is a GFI and a new contributor issue

PR #3585, opened 2026-06-30
[BUG] Padder: validate fill_value and reject array-likes (#3495)

### LINKED ISSUE #3495: [BUG] Padder fill_value as a NumPy array is silently misinterpreted (not a per-series fill)

CedricConday:
Fixes #3495

## What was wrong

`Padder`'s docstring stated `fill_value` could be "a numpy array for each time series", but `_transform` passes it straight to `np.pad(..., constant_values=fill_value)`. `np.pad` interprets an array as a `(before, after)` pair, so a length-2 array silently pads every position with its second element, and other lengths raise an opaque broadcast `ValueError`.

## Fix

- Validate that `fill_value` is a scalar, a supported statistic string (`mean`/`median`/`min`/`max`/`last`), or a callable — raising a clear `ValueError` for array-likes.
- Correct the docstring to drop the unsupported array claim (and add the already-supported `last`).

## Test

Adds `test_padder_rejects_array_fill_value`. Verified it fails on `main` (no error raised) and passes with this change.

---
*AI-assisted, human-reviewed — I'm an AI engineer; I find, fix, and test with AI, then review and verify before opening.*

### DIFF
--- a/aeon/transformations/collection/unequal_length/_pad.py
+++ b/aeon/transformations/collection/unequal_length/_pad.py
@@ -150,6 +150,19 @@ class Padder(BaseCollectionTransformer):
                     "padded_length is 'min' or 'max')."
                 )
 
+        # fill_value must be a scalar, a supported statistic string, or a callable.
+        # An array-like is silently misinterpreted by np.pad's constant_values
+        # (treated as a (before, after) pair), so reject it with a clear error.
+        if (
+            not callable(self.fill_value)
+            and not isinstance(self.fill_value, str)
+            and np.ndim(self.fill_value) != 0
+        ):
+            raise ValueError(
+                "fill_value must be a scalar, one of {'mean', 'median', 'min', "
+                "'max', 'last'}, or a callable; an array-like is not supported."
+            )
+
         # Determine if fill value is a function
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
thanks for this, looks like the right fix.

TonyBagnall (changes requested):
sorry, https://github.com/aeon-toolkit/aeon/pull/3573 introduced docstring changes and now conflicts. If you could resolve that then this is good to go

CedricConday:
Thanks! Resolved the conflict with #3573 — I kept your reworded `fill_value` docstring. Should be good to go now.

TonyBagnall (approved):

TonyBagnall:
@all-contributors add @CedricConday for bug

PR #3591, opened 2026-06-30
[ENH] Changes left stampi so fit_is_empty

### LINKED ISSUE #2819: [BUG] `LeftSTAMPi` output is incorrect shape

TonyBagnall:
Fixes #2819

Fixes LeftSTAMPi so it behaves as an unsupervised series anomaly detector with empty fit. Im not sure this is the best way, but this is a genuine bug, so this solution stops unexpected behaviour for this wrapper.

predict(X) and fit_predict(X) now treat X as the complete stream being scored: STUMPi is initialized from the first n_init_train points, then updated over the remaining points. This avoids carrying state from a previous fit call into later predictions, returns scores with the correct length for the prediction series, and lets the standard series anomaly detector output check run again.

LeftSTAMPi exclusion from check_series_anomaly_detector_output was removed.

The behavioral change means fit(train).predict(test) no longer appends test onto the fitted training stream;
  predict(test) scores test independently. Again, not sure this is best, we may need to add an update method to allow genuine streaming?

### DIFF
--- a/aeon/anomaly_detection/series/distance_based/_left_stampi.py
+++ b/aeon/anomaly_detection/series/distance_based/_left_stampi.py
@@ -22,6 +22,11 @@ class LeftSTAMPi(BaseSeriesAnomalyDetector):
 
     LeftSTAMPi supports univariate time series only.
 
+    This estimator is unsupervised and ``fit`` is a no-op. Calls to ``predict(X)``
+    and ``fit_predict(X)`` treat ``X`` as an independent stream: the initial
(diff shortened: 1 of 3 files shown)


MatthewMiddlehurst (approved):
Agreed do not know what the best ending for this is, but i see this as better than the current state which seems buggy and breaks the contract.

PR #3592, opened 2026-07-01
[ENH] Better examples for CES and AutoCES

### LINKED ISSUE #3590: [ENH] Better example for CES

CES forecaster example is currently this

 65(8), 685-704.

    Examples
    --------
    >>> import numpy as np
    >>> from aeon.forecasting.stats import CES
    >>> y = np.array([2.1, 2.4, 2.8, 3.0, 3.6, 4.1, 4.4, 4.9, 5.3, 5.9])
    >>> f = CES()
    >>> f.iterative_forecast(y, prediction_horizon=2).shape
    (2,)

which is not very good, an example forecast would be better. 

get output rounded to a few DPs and have the expected output match that

GiGiKoneti:
#### Reference Issues/PRs

Fixes aeon-toolkit/aeon#3590.

- Replaced the `CES` docstring example forecast returning `.shape` with a concrete array of actual predictions (`array([6.38..., 7.05...])`) to show the expected output values.
- Added a corresponding example showing usage and forecasts for the `AutoCES` forecaster.

No.

#### Any other comments?

None.

##### For all contributions


GiGiKoneti:
@TonyBagnall 
please review this

.github/workflows/update_contributors.yml
```diff
@@ -10,6 +10,9 @@ on:
 jobs:
   generate-markdown-and-commit:
     runs-on: ubuntu-24.04
+    permissions:
```
TonyBagnall: could you not change this workflow in the PR?

TonyBagnall (changes requested):
Thanks for this. Before merging, can you round the forecast values in both examples rather than printing full precision, e.g. f.iterative_forecast(y, prediction_horizon=2).round(4) giving array([6.3830, 7.0506]). CES/AutoCES fit via numerical optimisation, so full-precision output like 6.38295867 is fragile across platforms and numpy/scipy versions and is likely to cause doctest failures down the line (the ellipsis-matching fix commit earlier in this PR is a symptom of exactly that). Rounding to 3-4 decimal places keeps the example informative without pinning it to exact floating point output.

TonyBagnall (approved):

TonyBagnall:
thanks

PR #3593, opened 2026-07-01
[BUG] AutoARIMA: never select differencing order above max_d

### LINKED ISSUE #3577: [BUG] AutoARIMA(max_d=0) can still select one order of differencing

kratos0718:
#### Reference Issues/PRs
Fixes #3577.

#### What does this implement/fix?

`AutoARIMA._fit` could select a non-seasonal differencing order greater than `max_d`. The loop condition was:

```python
    differenced_series = np.diff(differenced_series, n=1)
    self.d_ += 1
```

Because of `self.d_ <= self.max_d`, the loop could enter when `self.d_ == self.max_d`, apply one more difference, and set `self.d_ = self.max_d + 1`. For `max_d=0` this produced `d_ == 1`, i.e. differencing was applied even though the user disabled it.

Changed the condition to `self.d_ < self.max_d`, so a new difference is only applied while `d_` is strictly below the configured maximum. The fitted order now always satisfies `0 <= d_ <= max_d`.

#### Test

Added `test_autoarima_respects_max_d_zero`: `AutoARIMA(max_d=0).fit(np.arange(50.))` now yields `d_ == 0` (it returned `1` before this change).

### DIFF
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -431,7 +431,7 @@ class AutoARIMA(BaseForecaster, IterativeForecastingMixin):
         series = np.array(y.squeeze(), dtype=np.float64)
         differenced_series = series.copy()
         self.d_ = 0
-        while not kpss_test(differenced_series)[1] and self.d_ <= self.max_d:
+        while not kpss_test(differenced_series)[1] and self.d_ < self.max_d:
             differenced_series = np.diff(differenced_series, n=1)
             self.d_ += 1
         include_constant = 1 if self.d_ == 0 else 0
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
Thanks, this is the right fix for #3577 and the regression test is useful.

One small suggestion: it may be cleaner to write the loop as

while self.d_ < self.max_d and not kpss_test(differenced_series)[1]:

so that kpss_test is not called at all when max_d=0. it better reflects the meaning of max_d=0.

TonyBagnall:
hi, closing this as corrections were not made and an earlier contributor had solved the issue

PR #3594, opened 2026-07-01
Fix AutoARIMA(max_d=0) still differencing once

### LINKED ISSUE #3577: [BUG] AutoARIMA(max_d=0) can still select one order of differencing

Mubashir78:
The while loop used `self.d_ <= self.max_d`, so when max_d=0 the loop could still enter (d_=0 <= 0 is true), apply np.diff, and set d_=1. Changed to `<` so the loop stops before exceeding the limit.

Closes #3577

### DIFF
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -431,7 +431,7 @@ class AutoARIMA(BaseForecaster, IterativeForecastingMixin):
         series = np.array(y.squeeze(), dtype=np.float64)
         differenced_series = series.copy()
         self.d_ = 0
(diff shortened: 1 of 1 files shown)


TonyBagnall:
thanks for the PR. We already have #3593 open for the same issue, and that PR includes a regression test as well as the code change, so I’m going to close this as a duplicate.

PR #3598, opened 2026-07-02
[BUG] Relegate flat test to a warning

### LINKED ISSUE #3570: [BUG] Intermittent notbook fail with hybrid
### Describe the bug

overnight fails intermittently with hive cote in the notebook 

https://github.com/aeon-toolkit/aeon/actions/runs/28142499609/job/83342631710

caused by the guard against zero variance,

ValueError: Input collection has too little variation: std <= 1e-07 for 1 case/channel pair(s): (case=15, ch=0). Rescale (e.g., multiply by a constant) or normalise your data.

`check_collection_variance` is called unconditionally inside `BaseCollectionEstimator._check_X` (`aeon/base/_base_collection.py`). Because `_check_X` runs on the input of every collection transformer, the guard also runs on the short sub-series slices that interval-based methods extract internally. These slices can legitimately be near-constant, so the guard raises and aborts the whole fit.

This is a false positive: near-constant interval slices are expected behaviour for interval and shapelet pipelines, not a data problem the user can fix by rescaling.

The periodic notebook tests caught this in `examples/classification/hybrid.ipynb` (HIVE-COTEv2 fit cell). Failing run: https://github.com/aeon-toolkit/aeon/actions/runs/28142499609/job/83342631710

PR #3315, "[BUG] Handling series with extremely small scale"

So the guard was added deliberately to handle tiny-scale series, and the regression is a side effect of placing the check in the shared _check_X path, where it also runs on the inner interval slices rather than only the top-level input. 

The call site forces the error:

```python
# aeon/base/_base_collection.py, end of _check_X
check_collection_variance(X)
```

`check_collection_variance` accepts a `raise_error` argument, but the call passes no arguments other than `X`, so it raises on every nested transform input rather than only validating the top-level estimator input.

```python
from aeon.classification.hybrid import HIVECOTEV2
from aeon.datasets import load_italy_power_demand

X_train, y_train = load_italy_power_demand(split="train")
HIVECOTEV2(time_limit_in_minutes=0.2).fit(X_train, y_train)
```

The same failure is reachable more directly through `DrCIFClassifier.fit_predict`, since the offending path is the interval transform, not HC2 specifically.

### Expected results

The fit completes. Near-constant sub-series produced by random-interval extraction should not trigger a hard error, since this is normal for interval-based estimators.

### Actual results

```
ValueError: Input collection has too little variation: std <= 1e-07
for 1 case/channel pair(s): (case=15, ch=0). Rescale (e.g., multiply by a constant)
or normalise your data.
```

Abridged call stack (full trace in the linked run log):

```
HIVECOTEV2._fit
  _BaseHIVECOTE._fit                       # est.fit_predict for CAWPE weights
    DrCIFClassifier._fit_predict
      BaseIntervalForest._fit_predict_proba -> _fit_forest
        RandomIntervals._fit_transform -> _generate_interval
          feature.fit_transform(X[:, dim, start:end:dilation])   # interval slice
            BaseCollectionTransformer.fit_transform
              _preprocess_collection -> _check_X
                check_collection_variance(X)   # raises ValueError
```

The trigger is RNG-dependent (the offending interval depends on the random state), although it is reproducible here given the fixed seeds. ItalyPowerDemand is z-normalised and length 24, so a short interval for `(case=15, ch=0)` falls below the `1e-07` variance threshold.

### Suggested fix

1. Do not run the variance check on the inner transform path. Restrict it to the top-level estimator input, for example only when `store_metadata=True` in `_preprocess_collection`, or behind a capability tag that interval and shapelet transformers can opt out of. This is the cleanest option, since near-constant slices are normal for these methods.
2. Demote the guard to a warning at this call site, or pass `raise_error=False`, so low-variance inner slices do not abort the fit.
3. Relaxing the threshold alone is not sufficient, as it only narrows the window for the same class of failure.

A regression test driving an interval-based estimator (for example `DrCIFClassifier.fit_predict` on ItalyPowerDemand with a fixed seed) would guard against reintroduction.

### Versions

Periodic tests run against latest dependencies; failure observed on `main` at commit `54bc654` (run #1099, 2026-06-25), Python 3.12, macOS arm64 runner.

TonyBagnall:
Fixes overnight test fail for hybrid notebook, fixes #3570 

 - aeon/base/_base_collection.py:199 — _check_X now calls check_collection_variance(X, raise_error=False) and emits a UserWarning instead of raising, so near-constant internal slices no longer abort fits. The standalone check_collection_variance() function still raises by default for anyone calling it directly.
  - aeon/base/tests/test_base_collection.py — added test_check_X_low_variance_warns as a regression test.

Verified three ways: the exact input that crashed the notebook (first difference of ItalyPowerDemand case 15, interval [13:16], std =  5.3e-9) now completes Catch22.fit_transform with a warning; the notebook's HC2 cell (HIVECOTEV2(time_limit_in_minutes=0.2) on  ItalyPowerDemand) runs end-to-end with accuracy 1.0; and the 122 base/validation tests all pass. One caveat on my earlier message: the
  "0/10 seeds failed" DrCIF repro was accidentally run against a stale aeon 1.5.0 install in .venv, so it was meaningless — the  deterministic slice-level evidence is what confirms the diagnosis and the fix.

### DIFF
--- a/aeon/base/_base_collection.py
+++ b/aeon/base/_base_collection.py
@@ -19,6 +19,7 @@ State:
     fitted state inspection - check_is_fitted()
 """
 
+import warnings
 from abc import abstractmethod
 
 import numpy as np
@@ -196,7 +197,16 @@ class BaseCollectionEstimator(BaseAeonEstimator):
             )
             raise ValueError(msg)
 
-        check_collection_variance(X)
+        if not check_collection_variance(X, raise_error=False):
+            warnings.warn(
+                f"Data seen by instance of {type(self).__name__} has one or more "
+                "case/channel pairs with very little variation (std <= 1e-07). "
+                "Some aeon methods may treat these series as effectively constant. "
+                "If this is unintended, rescale (e.g., multiply by a constant) or "
+                "normalise your data.",
+                UserWarning,
+                stacklevel=2,
+            )
 
         return metadata
 
--- a/aeon/base/tests/test_base_collection.py
+++ b/aeon/base/tests/test_base_collection.py
@@ -230,6 +230,22 @@ def test_preprocess_collection(data):
     assert meta == cls.metadata_
 
 
+def test_check_X_low_variance_warns():
+    """Test near-constant data warns rather than raises in _check_X.
+
+    Interval and shapelet based estimators call transformers on small subseries
+    internally, which are often near-constant. A hard error here aborts fits on
+    valid data, see issue #3570.
+    """
+    dummy = MockClassifier()
+    X = np.ones((5, 1, 10))
+    X[0][0][0] += 1e-8
+
+    with pytest.warns(UserWarning, match="very little variation"):
+        meta = dummy._check_X(X)
+    assert meta
+
+
 def test_convert_np_list():
     """Test np-list of 1D numpy converted to 2D."""
     x1 = np.random.random(size=(1, 10))
--- a/aeon/classification/tests/test_sklearn_compatability.py
+++ b/aeon/classification/tests/test_sklearn_compatability.py
@@ -108,8 +108,15 @@ def test_sklearn_parameter_tuning(parameter_tuning_method):
     X, y = make_example_3d_numpy(
         n_cases=20, n_channels=2, n_timepoints=30, min_cases_per_label=10
     )
+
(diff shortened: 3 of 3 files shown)


patrickzib (approved):
LGTM

PR #3601, opened 2026-07-02
[BUG] Fix sporadic fail for Inception time file save

### LINKED ISSUE #3595: [BUG] Intermittent CI fail with deep learning saving models
### Describe the bug

persistent occasional CI fail caused by file writing

https://github.com/aeon-toolkit/aeon/actions/runs/28488732035/job/84440686147?pr=3581

Fixed for InceptionTime  in #3601 since that was the one that failed, but this applies similarly for other deep learners.

TonyBagnall:
Reference Issues/PRs

Fixes #3595.

IndividualInceptionRegressor._fit loads the temporary best-model checkpoint after training and previously guarded this with except FileNotFoundError.

However, Keras raises ValueError, not FileNotFoundError, when load_model is called on a missing .keras path. With ModelCheckpoint(save_best_only=True), the checkpoint file is occasionally never written. When that happened, the ValueError escaped and fit crashed, causing the intermittent CI failure in check_regressor_random_state_deep_learning seen in #3595:

ValueError: File not found: filepath=./1782872340055533900.keras.
Please ensure the file is an accessible `.keras` zip file.

This PR replaces the try/except with an explicit os.path.exists check on the checkpoint path:

if the checkpoint exists, it is loaded with compile=False and removed when save_best_model=False, preserving existing behaviour;
if the checkpoint does not exist, model_ falls back to a copy of the trained in-memory model;
no broad ValueError is caught, so genuine model-loading or deserialisation errors still surface.

This PR also adds a fast regression test that simulates the missing-checkpoint case by preventing the ModelCheckpoint callback from being added, then asserts that fit completes, no .keras file was written, and predict works.

Note: the same except FileNotFoundError pattern exists in other deep-learning modules, including classification, clustering, forecasting, and self-supervised transformers, so they may be susceptible to the same intermittent failure. This PR deliberately only fixes the regressor that failed in CI; aligning the others is left for a follow-up.

Verified locally:

both tests in test_inception_time.py pass;
the exact failing CI check, check_regressor_random_state_deep_learning on IndividualInceptionRegressor, passes;
pre-commit is clean on the modified files.

### DIFF
--- a/aeon/regression/deep_learning/_inception_time.py
+++ b/aeon/regression/deep_learning/_inception_time.py
@@ -752,13 +752,14 @@ class IndividualInceptionRegressor(BaseDeepRegressor):
             callbacks=self.callbacks_,
         )
 
-        try:
-            self.model_ = tf.keras.models.load_model(
-                self.file_path + self.file_name_ + ".keras", compile=False
-            )
+        # with save_best_only=True the checkpoint file is not guaranteed to have
+        # been written, fall back to the trained model if it is missing
+        file_path = self.file_path + self.file_name_ + ".keras"
+        if os.path.exists(file_path):
+            self.model_ = tf.keras.models.load_model(file_path, compile=False)
             if not self.save_best_model:
-                os.remove(self.file_path + self.file_name_ + ".keras")
-        except FileNotFoundError:
+                os.remove(file_path)
+        else:
             self.model_ = deepcopy(self.training_model_)
 
         if self.save_last_model:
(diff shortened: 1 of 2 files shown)


hadifawaz1999:
might be a good fix, checked the code should be good
do you prefer we merge this first and then make an issue to do the rest or we keep this one up and do all changes here ?

hadifawaz1999 (approved):
Should make an issue for the other ones

PR #3602, opened 2026-07-03
[MNT] Forecasting test coverage

TonyBagnall:
coverage for forecasting which highlighted a couple of minor bugs

### DIFF
--- a/aeon/forecasting/deep_learning/_deepar.py
+++ b/aeon/forecasting/deep_learning/_deepar.py
@@ -305,7 +305,7 @@ class DeepARForecaster(
             horizon capability, returns single step prediction.
         """
         if y is None:
-            if not hasattr(self, "last_window_"):
+            if self.last_window_ is None:
                 raise ValueError("No fitted data available for prediction.")
             y_inner = self.last_window_
         else:
@@ -324,7 +324,7 @@ class DeepARForecaster(
         if self.use_probabilistic:
             prediction = pred[0, :, :]  # return both mu and std
         else:
-            prediction = float(pred[0, 0, :])  # return only mu
+            prediction = float(pred[0, 0, 0])  # return only mu
         return prediction
 
     @classmethod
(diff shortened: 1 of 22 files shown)


alexbanwell1 (commented):
### Review — [MNT] Forecasting test coverage

Reviewed the diff (mostly new tests; small production changes). Overview: this PR adds substantial test coverage across the forecasting package and, in the process, makes a handful of small production fixes. I focused the review on the production changes and on whether the new tests exercise real code paths / assert against messages that actually exist.

**Production changes — all look correct:**

- `deep_learning/_deepar.py`, `_tcn.py`, `_nbeats.py`: `hasattr(self, "last_window_")` → `self.last_window_ is None`. Safe — `last_window_` is initialized to `None` in `deep_learning/base.py` (`__init__`), so the unfitted case still raises `ValueError("No fitted data available...")` rather than `AttributeError`. This is a genuine improvement.
- `deep_learning/_deepar.py`: `float(pred[0, 0, :])` → `float(pred[0, 0, 0])`. This is a real bug fix — the network always emits `(mu, sigma)` (last dim = 2), so `float(pred[0, 0, :])` on a 2-element array would have raised `TypeError` in the non-probabilistic path.
- `machine_learning/_setar.py`: removal of the `num_obs <= 0` and `grid.size == 0` guards. Both were unreachable: `len(y) > maxlag` guarantees `num_obs >= 1`, and the `end <= start` check already returns `None` before the slice, so `np.unique(...)` always has ≥1 element. Dead-code removal, and the new `test_fit_setar_returns_none_for_degenerate_inputs` covers the surviving `end <= start` path.
- `stats/_ces.py`: removal of `_ces_fit_objective` and `_ces_fit_pass_sse`. Confirmed both are unreferenced anywhere in the tree.

**Tests:** spot-checked the `pytest.raises(match=...)` strings against the actual source messages (deepar/tcn/nbeats/setar/setarforest/ensemble) — all match. The `use_probabilistic` toggle-after-fit path in `test_deepar_save_options_and_predict_paths` is valid because the model always outputs both mu and sigma.

No actionable defects found. Nice coverage addition. 👍

alexbanwell1 (approved):

PR #3609, opened 2026-07-06
[BUG] SFA.get_words() and SFAFast.get_words() crash for any normal sliding-window configuration

### LINKED ISSUE #3537: [BUG] SFA.get_words() and SFAFast.get_words() crash for any normal sliding-window configuration

patrickzib:
Fixes #3537

`SFA.get_words()` and `SFAFast.get_words()` crashed when `window_size < n_timepoints` (the normal sliding-window configuration). Now fixed

## Changes

- `aeon/transformations/collection/dictionary_based/_sfa.py`: 
- `aeon/transformations/collection/dictionary_based/_sfa_fast.py`
- `aeon/transformations/collection/dictionary_based/tests/test_sfa.py`: Add new regression tests covering sliding window

## Testing

All tests in `test_sfa.py` pass


TonyBagnall (dismissed):
thanks for the fix, but new get_chars helper mutates the input words array while decoding it, by shifting each word in place. A getter shouldnt mutate the state of the object imo

aeon/transformations/collection/dictionary_based/_sfa.py
```diff
@@ -1178,16 +1183,17 @@ def _get_test_params(cls, parameter_set="default"):
...
+    for j in range(chars.shape[0]):
+        for i in range(word_length):
+            # Extract the last bits
+            char = words[j] & mask
+            chars[j][-i - 1] = char
 
-        # Right shift by to move to the next group of bits
-        word >>= letter_bits
+            # Right shift by to move to the next group of bits
+            words[j] >>= letter_bits
```
TonyBagnall: doesnt this change words in place?

TonyBagnall: equivalent to 
words[j] = words[j] >> letter_bits

patrickzib: True. Should be ok now.

TonyBagnall (changes requested):
only small things, but issue #3537 covers both SFA.get_words() and SFAFast.get_words(), and this PR changes both implementations. The new sliding-window regression test only SFA. Could you add the equivalent SFAFast case? Also I don't think get_chars needs to be added to __all__, probably better private

patrickzib:
> only small things, but issue #3537 covers both SFA.get_words() and SFAFast.get_words(), and this PR changes both implementations. The new sliding-window regression test only SFA. Could you add the equivalent SFAFast case? Also I don't think get_chars needs to be added to **all**, probably better private

Thanks for the feedback! 

Sure, will add a test to SFAFast, and remove it from **all**.

PR #3567, opened 2026-06-24
[ENH] edr lcss memory and speed improvements

TonyBagnall:
Part of #3261. Rewrites _edr_distance and _lcss_distance to use two-row buffers instead of the full O(n·m) cost matrix, reducing space to linear.

closes #3265 and closes #3267

PR #3564, opened 2026-06-24
[ENH] adtw memory and speed improvements

TonyBagnall:
ADTW
closes #3262 

Part of https://github.com/aeon-toolkit/aeon/issues/3261 (memory optimization for elastic distances). Applies the two-row dynamic-programming optimization established for DTW in https://github.com/aeon-toolkit/aeon/pull/3258 (and MSM in https://github.com/aeon-toolkit/aeon/pull/3562) to ADTW.

more of the same 

| n | Runtime base → opt | Mem base → opt |
|---:|:-------------------|:----------------|
| 1,000 | 9.6 → 6.2 ms (1.5×) | 9.0 → 1.0 MB (9.0×) |
| 4,000 | 0.25 → 0.12 s (2.1×) | 144 → 16 MB (9.0×) |
| 8,000 | 1.00 → 0.42 s (2.4×) | 576 → 64 MB (9.0×) |
| 16,000 | not run → 1.63 s | ~2.3 GB → 256 MB |

Issue #2563, opened 2025-02-22
[DOC] Missing Docstrings in Fit, Predict, Forecast Regression


kavya-r30:
### Describe the issue linked to the documentation

Missing Docstring in aeon/forecasting/_regression.py

current 
``` 
def _fit(self, y, exog=None):
        """Fit forecaster to time series.

        Split X into windows of length window and train the forecaster on each window
        to predict the horizon ahead.

        Parameters
        ----------
        X : Time series on which to learn a forecaster

        Returns
        -------
        self
            Fitted estimator
        """

def _predict(self, y=None, exog=None):
        """Predict values for time series X."""
  
def _forecast(self, y, exog=None):
        """Forecast values for time series X.

        NOTE: deal with horizons
        """
```

kavya-r30:
@aeon-actions-bot assign @kavya-r30

PR #3610, opened 2026-07-06
[ENH] Remove dense bounding box from DTW

TonyBagnall:

This PR rewrites the `dtw_distance` computation path (single pair and pairwise) to compute
Sakoe-Chiba / Itakura band bounds directly, per row, instead of using an (n,m) 
boolean bounding matrix and testing it cell-by-cell.  

1. Every `dtw_distance` call allocates this matrix including calls where there is no bound, where `create_bounding_matrix` returns `np.full((n, m), True)`. That is O(n·m) memory per call (e.g. ~156 MB
   for a pair of length-12800 series) plus the time to build it, which is a measurable fraction of total call time.

2. Windowing reduces no work. The distance iterates all `n·m` cells and tests `bounding_matrix[i, j]` inside the innermost loop, so a Sakoe-Chiba window only changes which cells *count*, not how many are *visited*. A 20% window therefore runs no faster than an unconstrained call (slightly slower, from the added branch). For unequal-length series it is worse still: the windowed stepping-band mask is expensive to build, so a windowed call can take over twice as long as an unconstrained one on the same pair.

Results are unchanged — the new path is verified to match the existing implementation (see Testing) — and
there no public API changes.

basically if you only use Sakoe-Chiba (fixed interval) or Itakura parallelogram bounds, you don't need to store the mask. We do lose the ability to make completely custom bounding boxes, but I think that is a price worth paying. 

no

#### Any other comments?

To keep the first comment manageable I will describe the code below  then the extensive benchmarking vs reference packages which I will summarise below. This was written with Fables help. 

If this gets in I will roll it out using the PRs that already address part of the problem #3261, these are PRs

#3562, #2563. #3564, #3565. #3566, #3567

##### For all contributions


TonyBagnall:
First: on bounding boxes currently work like this:

Three bounding strategies are available, selected by the two keyword arguments on `dtw_distance` (and `dtw_pairwise_distance`)

| you pass | bounding used |
|---|---|
| neither (`window=None`, `itakura_max_slope=None`) | **Full** — every cell allowed |
| `window=w` with 0 ≤ w ≤ 1 | **Sakoe-Chiba band** — half-width `int(w * min(x_size, y_size))` around the (interpolated) diagonal. `window=1.0` is treated as full; `window=0.0` is the thinnest band; outside [0, 1] raises `ValueError` |
| `itakura_max_slope=s` with 0 ≤ s ≤ 1 | **Itakura parallelogram** — slope-constrained region; equal-length series only (pre-existing limitation, unchanged) |

**Precedence**: `itakura_max_slope` is checked *first* — if both are passed, Itakura silently wins and `window` is ignored. This is inherited from`create_bounding_matrix` and deliberately mirrored, since changing it would be a behaviour change outside this PR's scope.

### How the switch happens

Two levels:

1. **`_dtw_distance_direct`** (in `_dtw.py`) makes a *structural* choice first: if   `itakura_max_slope is None` **and** the series are equal length, the  Sakoe-Chiba/full bounds have a closed form, so it computes the band thickness inline (`y_size` for full, `int(window * y_size)` otherwise) and calls `_dtw_distance_sakoe_equal` — no bounds arrays are allocated at all. Everything
   else (unequal lengths, or any Itakura) falls through to step 2.

2. **`create_band_bounds`** (in `_bounding_matrix.py`) does the strategy dispatch proper, with the same if-chain and validation as `create_bounding_matrix`:  `itakura_max_slope is not None` → `_itakura_bounds`; `window is not None and window != 1.0` → `_sakoe_chiba_bounds`;
   else full bounds (`j_start = 0`, `j_end = y_size` for every row). The result
   feeds `_dtw_distance_banded`.


The public `dtw_distance` docstring describes `window` as a fraction of the *longest* series, but the construction actually uses `min(x_size, y_size)` for the band thickness — identical for equal lengths, and for unequal lengths the band follows the interpolated diagonal, so the docstring is a bit loose, tightened it up

TonyBagnall:
## Implementation details 

- **`_bounding_matrix.py`** has `create_band_bounds(x_size, y_size, window,
  itakura_max_slope)`, returning two int32 arrays `(j_start, j_end)` such that row `i` of the equivalent dense bounding matrix is True exactly on  `j_start[i] <= j < j_end[i]`. The Sakoe-Chiba bounds replicate the existing interpolated-diagonal stepping construction (including the transposed unequal-length branch) with two O(x_size + y_size) passes; the Itakura bounds
  reuse the per-row bound computation the dense builder already performs internally. `create_bounding_matrix` itself is unchanged and still used by  `dtw_cost_matrix` and by other elastic distances.

- **`_dtw.py`**: `dtw_distance` now routes through a new `_dtw_distance_direct` dispatcher. For the common case — equal-length series with a Sakoe-Chiba window  (or no window) — the bounds have a closed form  (`[max(0, i - t), min(n, i + t + 1))` with `t = int(window * n)`, identical to  the stepping construction at equal lengths), so no bounds arrays are allocated  at all: the only O(n) memory is the two rolling cost rows. Unequal lengths and Itakura go through `create_band_bounds` plus a banded kernel that iterates only  the in-band range per row.

- The kernels keep the existing two-row rolling-buffer scheme but hand rows over by **buffer swap instead of an O(m) copy**, and take a **univariate fast path**  that reads scalars from flat views rather than creating two strided column  views and running the channel loop for every DP cell.

- The pairwise functions (`_dtw_pairwise_distance`,  `_dtw_from_multiple_to_multiple_distance`) call the same per-pair dispatch, so  no dense mask is built in pairwise workloads either; their now-unused  `unequal_length` parameter is removed (private functions, no other importers).

Correctness of the buffer reuse relies on the band bounds being non-decreasing per row (true for both Sakoe-Chiba and Itakura); this invariant, and an explicit +inf guard for empty Itakura rows (no valid warping path), are documented in the
kernel docstrings.

TonyBagnall:
Note several docstring errors were identified. 

What changed in _dtw.py:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                         
  1. dtw_cost_matrix now correctly describes the accumulated DP cost matrix — entry (i,j) is the optimal-path cost aligning the first i+1 points of x with the first j+1 of y, with pointwise cost (x_i − y_j)² (was described as the pairwise           
  "Euclidean" matrix (x_i − x_j)²).                                                                                                                                                                                                                      
  2. Removed the stray duplicated "is used." line in its window description.                                                                                                                                                                             
  3. dtw_pairwise_distance Returns section: X-vs-y shape corrected to (n_cases, m_cases).                                                                                                                                                                
  4. dtw_distance formula corrected to d_dtw(x, y) = D_P*(x, y, M).                                                                                                                                                                                      
  5. "realligning" → "realigning".

TonyBagnall:
Now the fun bit. Experiments. We compare the new dtw implementation to: 

 - aeon_old -- the pip-installed aeon 1.5.0 (dense bounding-matrix implementation)                                                                                                                                      
 - aeon_new -- the local banded version on this PR                                                                                                                                                
 - sktime       -- sktime 1.0.1 DtwDist (the legacy aeon code from 2020, wrapped into pairwise-transformer objects);                                                                                                                      
 - tslearn      -- tslearn 0.9.0 metrics.dtw;                                                                                                                                                  
 - dtw-python   -- dtw-python 1.7.5 (default symmetric2 step pattern, its own distance convention); univariate only here                                                                                                                                                           
 - pyts         -- pyts 0.13.0 metrics.dtw(method="classic");                                                                                                                               
 - dtaidistance -- dtaidistance 2.4.0 C extension;                                                                                                                                          
 - tsdistances  -- local Rust build (../tsdistances, maturin develop --release),                                                                                                                                 

I'm running all these experiments on my desktop which is old and slow

TonyBagnall:
first though aeon old (1.5) vs new (this PR). Univariate full window and 20% window. All results checked for complete equivalence. 

Runtime average speed up 1.64x for full window, 3x for 20%
  - Univariate: new 0.50s/0.18s (full/20%) vs old 0.79s/0.71s, at 0.14MB vs 156MB                                                                                                                                                

image
Memory usage (approx) quadratic -> linear
image

TonyBagnall:
Now test on dependent multivariate, 3 channel. For the full-window multivariate runtime was a tie (0.870s old vs 0.871s new at n=12800) while univariate full-window showed a 1.6× win. That's because aeon_new's univariate speed  advantage at full window comes mostly from the univariate fast path — reading scalars from flat views instead of creating two strided column views and running a channel loop per cell. Multivariate can't use that path (it genuinely needs the channel loop), so at full window nothing distinguishes the two on work done: same cells visited, same per-cell cost. The mask allocation and row-copy savings that remain are small next to multivariate's heavier per-cell arithmetic, hence the tie.

Note independent will gain as univariate does

image

TonyBagnall:
unequal length is curious. 

1. The old implementation's window makes it slower on unequal lengths at scale. Its windowed call (0.558s) takes 1.4× longer than its own unconstrained call (0.392s) at n=12800. The cause is specific to unequal lengths: the
  Sakoe-Chiba mask isn't the trivial |i−j|≤w band but a stepping construction along the interpolated diagonal, and building that dense mask costs more than the DP work the window saves. In the runtime chart this is the blue  
  dashed line crossing above the blue solid line — the crossover happens around n=6400 (below that, the old windowed call is still mildly faster than old full).                                                                 
  2. The banded version behaves as a windowed algorithm should: its 20% window runs at ~36% of its own full-window time (0.090 vs 0.252s), matching the band's share of the cost matrix, giving the 6.2× headline over old       
  windowed.                                                                                                                                                                                                                      
  3. Memory tells the same story as the other cases, scaled by the rectangle: the old mask is n × n/2 bytes (78MB at n=12800, half the equal-length 156MB), identical for windowed and unconstrained, growing O(n²). The banded  
  version stays flat at ~0.13–0.27MB — this is the one case that actually exercises the int32 bounds arrays (equal-length series use the closed-form path with no arrays at all), and they're negligible.                        
  4. Correctness on unequal lengths is the most heavily tested part of the whole change: beyond the identical distances in every benchmark cell, 1800 randomized (x_size, y_size, window) combinations — including extreme ratios
  like 2×500 and 1000×10 — verified the banded bounds reproduce the dense stepping construction row-for-row exactly, in both orientations.     
<img width="1255" height="452" alt="image" src="https://github.com/user-attachments/assets/75965c40-68c2-4fcc-a8ee-229553e7f06e" />


  Set up│ aeon 1.5.0 │ aeon banded │ ratio │                                                                                                                                                            
  │ full window                  │     0.392s │      0.252s │  1.6× │                                                                                                                                                            
  │ 20% window                   │     0.558s │      0.090s │  6.2× │                                                                                                                                                            
  │ peak memory (either variant) │     78.2MB │ 0.24–0.27MB │ ~300×

TonyBagnall:
so for bake off I will restrict attention to univariate equal length only, since this is not a formal comparison. In terms of capabilities, this is what the oracle says 


| Package / API | Multivariate support | Unequal-length support | Notes |
|---|---:|---:|---|
| `aeon_old` / aeon 1.5.0 | Yes | Yes | Dense bounding-matrix implementation. Supports univariate, dependent multivariate, equal-length and unequal-length DTW. |
| `aeon_new` / PR #3610 | Yes | Yes | Same public DTW API as `aeon_old`; replaces dense bounding matrices with row-wise band bounds / closed-form band handling. Itakura remains equal-length only. |
| `sktime` 1.0.1 `DtwDist` | Partial | No for this estimator | Native DTW distance supports 2D dependent multivariate input, but the performant `DtwDist` path is effectively equal-length only. Unequal-length use needs the aligner-based API. |
| `tslearn` 0.9.0 `metrics.dtw` | Yes | Yes | Supports univariate and multivariate series. Series lengths may differ, but dimensions must match. |
| `dtw-python` 1.7.5 | Yes | Yes | Supports unequal-length series. Multivariate data can be passed as matrices with time points as rows, or through a precomputed local-distance matrix. |
| `pyts` 0.13.0 `metrics.dtw(method="classic")` | No direct support | Yes | Direct API is for 1D arrays, but the two arrays may have different lengths. A multivariate cost matrix could be precomputed externally, but that is not direct multivariate DTW support. |
| `dtaidistance` 2.4.0 | Yes | Yes | Univariate DTW is in `dtaidistance.dtw`; multivariate dependent DTW is in `dtaidistance.dtw_ndim`. Supports unequal lengths. |
| `tsdistances` local Rust build | Probably no direct support | Yes for 1D pairs | Treat as univariate unless the Python wrapper exposes channels separately from batches. Ragged batch support needs local verification. |

TonyBagnall:
ok, didnt think I would spend all day doing this but its all good. Here we go with headline comparison :aeon banded is this PR. 

### runtime

<img width="580" height="398" alt="image" src="https://github.com/user-attachments/assets/301aea39-e56b-47e0-b80b-ff877a67a04b" />

<img width="665" height="402" alt="image" src="https://github.com/user-attachments/assets/ad1185c7-c07c-498a-a4a4-5713dba1a80c" />


At n=12800, univariate:                                                                                                                                          
                                                                                                                                                                                                                                 
  Runtime — aeon banded is the fastest of all eight at both settings:                                                                                                                                                            
  - Full window: aeon_new 0.506s → tsdistances 0.617s → dtaidistance 0.736s → aeon_old 0.879s → tslearn 1.04s → dtw-python 5.38s → sktime 5.63s → pyts 10.2s                                                                     
  - 20% window: aeon_new 0.199s → dtaidistance 0.257s → tsdistances 0.270s → aeon_old 0.808s → tslearn 0.917s → pyts 2.13s → sktime 5.87s → dtw-python 7.68s

sktime shows the start up cost of their approach. Its also worth noting, given our legacy, that we are an order of magnitude faster than sktime which is based on the first implementation we contributed in 2020. 

The flat ~0.15–0.19s floor at n≤800 is the design choices bolted onto the core code — mtype detection, panel conversion, scikit-base tag plumbing per transform call. It stops dominating only around n=3200. For KNN-style workloads
  (millions of short-series calls) this overhead would be the whole story. We have used their DtwDist pairwise-transformer API, because thats what they recommend afaik.

TonyBagnall:
### Memory
this is harder to measure for rust and c versions.

Where the measurement is sharp, it decomposes beautifully. Every large reading resolves into near-exact multiples of the obvious arrays at n=12,800 (n² bool = 156MB, n² float64 = 1250MB): aeon 1.5.0 = 156MB (precisely the mask); tslearn = 1406MB = one float64 cost matrix + one boolean mask; pyts = 2500MB full (two float64 matrices) → 3750MB windowed (a third); dtw-python = 3750/6250MB (three, then five); sktime ≈ 5000MB (four). The numbers landing on round multiples is itself evidence the harness measures what we think it does.

dtaidistance — the floor problem. RSS only grows when the allocator commits new pages from the OS; sub-MB requests are served from free pages the heap already owns, so they never appear. dtaidistance's true working set (two rolling cost rows, ~0.2MB at n=12,800 — the same scheme as aeon banded) sits entirely under that ~0.5MB floor, which is why it reads 0.004–0.02MB everywhere. We verified rather than assumed: scaling n until the allocation must exceed any plausible slack made it surface at exactly theory (0.78MB observed vs 0.82MB predicted at n=51,200). So its column should be read as "≈ aeon banded", not "zero" — the 0.00-vs-0.23MB gap in the chart is measurement floor, not a real order of magnitude.

tsdistances.** The Rust extension allocates inside its own                                                                                                                                     runtime, behind the FFI boundary — there is no Python-side visibility at all, so                                                                                                                                      
external RSS is the *only* available probe, and what it shows is the Rust                                                                                                                                             
allocator's page-level footprint rather than live bytes. Its readings are                                                                                                                                      
plausible but slightly lumpy



image
image

TonyBagnall:
and finally, pushed it to the point only three algorithms are feasible. tsdistances, aeon new and dtaidistance. Still running but 

aeon_new full-window is topping out at 148.9s for n=204,800 with 3.1MB peak memory (O(n) holding at 16× the earlier max length; a dense mask at that n would have been 39G.

TonyBagnall:
and finally .... these differences may be down to minor differences my end . Memory gets more accurate with size

image

image

image

image

chrisholder (approved):

PR #3611, opened 2026-07-07
[MNT] Use Agg backend in shapelet plotting tests

TonyBagnall:
Fixes flaky _tkinter.TclError: invalid command name tcl_findLibrary in overnight CI runs by switching to the non-interactive Agg backend, matching every other visualisation test module.


patrickzib (approved):
LGTM

PR #3612, opened 2026-07-07
[BUG] Fix nightly tests related to SFA failing

patrickzib:
Nightly tests occasionally fail due to errors with sfa.
See: https://github.com/aeon-toolkit/aeon/actions/runs/28651558930/job/84970580199?pr=3602

The root cause is likely related to division by zero and numba handing NaN values incorrectly when using fastmath.

Fixes:
- Change to fastmath={"contract"} for MCB discretization
- Fix division by zero for dynamic alphabet computation


TonyBagnall:
I'll just run this with the full suite

TonyBagnall (approved):
LGTM thanks

PR #3553, opened 2026-06-22
[ENH] add time series agglomerative clusterer

SotonSweetXss:
#### Reference Issues/PRs
Fixes #3433 . See also #1241 

#### What does this implement/fix? Explain your changes.
A `TimeSeriesAgglomerative` has been implemented by using `aeon` to compute a pairwise time series distance matrix (e.g. Euclidean distance, DTW or MSM), and then applying scikit-learn’s `AgglomerativeClustering` algorithm to this pre-computed distance matrix. It supports single-linkage, complete-linkage and average-linkage clustering, but does not support Ward’s linkage clustering, as this method is incompatible with any pre-computed time series distance.

#### Does your contribution introduce a new dependency? If yes, which one?
no, Based entirely on sklearn and Aeon

#### Any other comments?
I will first submit the interface for this level of clustering; if it is approved, I will subsequently add code for reliability verification in `aeon\clustering\test`.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/aeon-toolkit/aeon/blob/main/.all-contributorsrc). Alternatively, you can use the [@all-contributors](https://allcontributors.org/docs/en/bot/usage) bot to do this for you **after** the PR has been merged.
- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] indicating whether the PR topic is related to enhancement, maintenance, documentation, bugs, refactoring, deprecation or governance.

##### For new estimators and functions
- [ ] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).
- [ ] (OPTIONAL) I've added myself as a `__maintainer__` at the top of relevant files and want to be contacted regarding its maintenance. Unmaintained files may be removed. This is for the full file, and you should not add yourself if you are just making minor changes or do not want to help maintain its contents.

##### For developers with write access
- [ ] (OPTIONAL) I've updated aeon's [CODEOWNERS](https://github.com/aeon-toolkit/aeon/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

Issue #2477, opened 2024-12-31
[ENH] Implementing DBSCAN


aryanpola:
### Describe the feature or idea you want to propose

We currently do not have a sub-module for density-based clustering. This issue is for implementing Density-Based Spatial Clustering of Applications with Noise (DBSCAN).

### Describe your proposed solution

Paper Reference: https://cdn.aaai.org/KDD/1996/KDD96-037.pdf

Scikit-learn for reference: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/cluster/_dbscan.py

### Additional context

The `scikit-learn` implementation cannot `predict`. This does not fit with our interface.

aryanpola:
@aeon-actions-bot assign @aryanpola

SebastianSchmidl:
DBSCAN is already implemented in sklearn, and aeon depends on sklearn. Why do we need to re-implement it directly in aeon?

aryanpola:
The clustering examples had a graph contained all the clustering sub-modules which includes density based, so figured why not implement it in aeon.
I'll close this if it's unnecessary.
Thanks for checking!

MatthewMiddlehurst:
Agreed that unless there is some reason the sklearn version wont work (or is insufficient in any other way) that theres no need to have it here really. Any thoughts @chrisholder ?

TonyBagnall:
the problem with dbscan is that it does not have a predict, i.e. it can only cluster train data. I'm not sure of the best way to deal with this. I'll have a go with the scikit learn version

aryanpola:
> the problem with dbscan is that it does not have a predict, i.e. it can only cluster train data. I'm not sure of the best way to deal with this. I'll have a go with the scikit learn version

Yep, not having a predict function makes it a lot more difficult to cluster the newer points into the existing clusters (requires reevaluation of the model).
I'll close the issue then.

TonyBagnall:
we cannot use DBSCAN in scikit directly, as it requires 2D numpy. We could wrap for unvariate and squash, but it would not work with multivariate. once #3613 is in we can port this in

Issue #1241, opened 2024-02-24
[DOC] Hierarchical, spectral, or density-based clustering using sklearn and aeon distance metrics


CodeLionX:
### Describe the issue linked to the documentation

The [clustering component](https://www.aeon-toolkit.org/en/stable/api_reference/clustering.html) in aeon currently supports only partition-based methods. However, there are also hierarchical, spectral, and density-based clustering methods [1].

### Suggest a potential alternative/fix

Using the distance metrics in aeon, we can pre-compute the distance matrix for traditional clustering methods. Some methods are already implemented in sklearn, which is a core dependency of eaon and, thus, available to users. I think we should at least link to the sklearn-clusterers in the documentation. With a bit more effort, we could provide examples on how to use sklearn's clusterers with aeon's distance measures ([here](https://www.aeon-toolkit.org/en/stable/examples/clustering/partitional_clustering.html)).

- hierarchical clustering:
  - [`sklearn.cluster.AgglomerativeClustering`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html) with `metric="precomputed"`
- density-based clustering:
  - [`sklearn.cluster.DBSCAN`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html) with `metric="precomputed"`
  -  [`sklearn.cluster.OPTICS`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.OPTICS.html)  with `metric="precomputed"`
  - Density Peaks? https://doi.org/10.1007/s10115-018-1189-7
- spectral clustering:
  - [`sklearn.cluster.SpectralClustering`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html) with `affinity="precomputed"` and the inverse of the distance matrix (large values indicate greater similarity)

> I did not yet test this approach.

[1]: Paparrizos, John, and Luis Gravano. "Fast and Accurate Time-Series Clustering." ACM Transactions on Database Systems 42, no. 2 (2017): 8:1-8:49. https://doi.org/10.1145/3044711.

TonyBagnall:
thanks for this, we have some examples I think of using precomputed with scikit, but if its not clear it would be great if it was clearer. I would like to get density peaks in, iirc we have a java implementation.

SalmanDeveloperz:
Hey, Can i work on this issue?

SebastianSchmidl:
[...]

PR #3613, opened 2026-07-07
[ENH] Add capability:predict tag for Clustering

TonyBagnall:
Motivation

many clusterers have no  default use for fittting on unseen data (transducive clusterers). scikit simply has no predict method for these. We need to work it a bit more because of the predict/_predict model. This PR introduces a tag and tests for it in the base class. Needed for #3553 

see #1241 #2477 

What was changed:                                                                                                                                                                  
                                                                                                                                                                                     
  - Tag registry (aeon/utils/tags/\_tags.py): new capability:predict tag for the clusterer class — True means fit(X\_train); predict(X_new) is supported; False means the clusterer is 
  transductive and only fit(X)/labels_ and fit_predict(X) work.                                                                                                                      
  - BaseClusterer (aeon/clustering/base.py): defaults "capability:predict": True so every existing clusterer is unaffected. predict and predict\_proba now call a  \_check_predict\_capability() helper after the fitted-state check and raise NotImplementedError with your a message when the tag is False. fit_predict now passes y through to fit and its docstring says it is equivalent to fitting on X and returning labels_ — it never touches predict. The fitted-state check deliberately runs before the capability check so unfitted estimators still raise the standard NotFittedError that common tests expect.       
  - Tests: new MockTransductiveCluster (its\_predict raises RuntimeError if ever reached, proving fit\_predict doesn't route through predict); MockCluster.\_fit now sets labels\_ to match the base-class contract. test_base.py covers fit\_predict/labels_ equivalence, fit\_predict independence from _predict (via a call-counting clusterer), the error for tag-False clusterers, unchanged behaviour for tag-True ones, and the preserved NotFittedError. test\_agglomerative.py keeps the PR's sklearn-equivalence tests and adds fit(X).labels\_,  fit_predict determinism, and the predict/predict\_proba error tests.                                                                                                                
                                                                                                                                                                                     
  Verification: the full check_estimator suite passes for TimeSeriesAgglomerative, DummyClusterer, and TimeSeriesKMeans; the whole clustering module (575 tests, deep learning       
  excluded), aeon/utils (556 tests), estimator-checking/mock tests (304), and the new class's docstring example all pass locally.


aeon/clustering/tests/test_base.py
```diff
@@ -63,3 +64,80 @@ def test_base_clusterer():
...
+def test_fit_predict_returns_labels():
+    """Test fit_predict returns the same labels as fit(X).labels_."""
+    X = np.random.random(size=(10, 1, 20))
+
+    fitted = MockTransductiveCluster().fit(X)
+    labels = MockTransductiveCluster().fit_predict(X)
+    assert np.array_equal(labels, fitted.labels_)
+
+    fitted = MockCluster().fit(X)
+    labels = MockCluster().fit_predict(X)
```
patrickzib: Just a comment: 

What is the general contract of calling `fit_predict` after `fit`? 
Should the second call to `fit_predict` refit a clusterer or keep the fitted one? 

And if refitted: should both always yield the same clustering? For k-means this would not be true.

aeon/clustering/tests/test_base.py
```diff
@@ -63,3 +64,80 @@ def test_base_clusterer():
...
+    assert np.array_equal(labels, fitted.labels_)
+
+
+def test_fit_predict_does_not_call_predict():
+    """Test fit_predict does not route through predict/_predict."""
+    X = np.random.random(size=(10, 1, 20))
+
+    clst = _CountingClusterer()
+    clst.fit_predict(X)
+    assert clst.predict_calls == 0
```
patrickzib: Is there a general contract that `_predict` must  not be called whenever `fit_predict` is called?

I do not understand what is tested here. The code tests the concrete implementation of `_CountingClusterer`, which was, however, defined just before.

aeon/clustering/tests/test_base.py
```diff
@@ -63,3 +64,80 @@ def test_base_clusterer():
...
+    clst = MockTransductiveCluster()
+
+    assert not clst.get_tag("capability:predict")
+
+    # unfitted estimators still raise the standard fitted-state error
+    with pytest.raises(NotFittedError, match="has not been fitted"):
+        clst.predict(X)
+
+    clst.fit(X)
+    msg = (
```
patrickzib: maybe just check for just `does not support out-of-sample prediction` and not the full message like below in the other tests?

aeon/clustering/tests/test_base.py
TonyBagnall: fit_predict should refit, for non deterministic clusterers they could be different. But this is not what the test is for. It is calling fit and fit_predict on two separate instances of a deterministic clusterer. Now could the test be more tightly defined? I guess so, I pay less attention to the test definitions than you. comment should say
 
 """Test base fit_predict returns labels_ from the fit it performs.""

aeon/clustering/tests/test_base.py
TonyBagnall: well there is an unnecessary first bit for sure, but the test is valid.

aeon/clustering/tests/test_base.py
TonyBagnall: really dont like coverage tests?

TonyBagnall: Done — it now matches just does not support out-of-sample prediction, consistent with the other tests and robust to any rewording of the guidance text that follows.

TonyBagnall:
so my bot say this below and has rewritten the test, hopefully more to your liking, Are you ok with the actual changes @patrickzib? Its a design issue more than a testing style issue 


Good questions, and you're right the test was loosely defined. The contract is that fit_predict(X) is fit(X) then return labels_, so it always refits — and for a non-deterministic clusterer (k-means being the obvious case) two separate fits need not agree, which is exactly why comparing two separate instances was fragile. I've rewritten it to test the contract directly on a single instance: fit_predict(X)    returns labels_ from the fit it just performed, never predict(X). MockTransductiveCluster makes it discriminating — its labels_ are a known alternating 0/1 pattern and its _predict raises, so matching that  pattern shows fit_predict returned labels_ rather than routing through predict. 

Agreed the _CountingClusterer was unnecessary. The property under test is the one documented on fit_predict: it must not route through _predict, so it stays valid for transductive clusterers. I've removed the counting clusterer and now test that directly with MockTransductiveCluster, whose _predict raises — fit_predict completing instead of erroring proves _predict was never called. That's the whole test now, with a docstring saying exactly that.


What changed in test_base.py                                                                                                                                                                                     
                                                                                                                                                                                                                   
  - test_fit_predict_returns_labels → rewritten as test_fit_predict_returns_labels_from_its_own_fit: tests the real contract on a single instance (returns labels_ from the fit it performs, not predict(X)), using
  the transductive mock's known 0/1 pattern as the discriminating check — no reliance on "two separate fits agree."                                                                                                
  - test_fit_predict_does_not_call_predict → replaced by test_fit_predict_valid_without_predict_capability: _CountingClusterer deleted; now uses MockTransductiveCluster (whose _predict raises), so fit_predict   
  completing proves _predict was never called.                                                                                                                                                                     
  - test_predict_without_capability_raises: matches only does not support out-of-sample prediction.                                                                                                                
  - Every test got a docstring explaining what it verifies; redundant import numpy.random removed.

patrickzib:
> Good questions, and you're right the test was loosely defined. 
> 
> Agreed the _CountingClusterer was unnecessary. 

Love it that agents are so affirmative :)

Thank you. I'll check again this afternoon.

aeon/clustering/tests/test_base.py
patrickzib: No, the coverage test was fine. 

My comment was just about checking for a full string, which is not very flexible in case of rewording, as you noted :) 

I'll check the full PR later on.

chrisholder (approved):

PR #3614, opened 2026-07-07
[BUG] AutoARIMA: respect max_d in the differencing loop

### LINKED ISSUE #3577: [BUG] AutoARIMA(max_d=0) can still select one order of differencing

CedricConday:
Fixes #3577.

The non-seasonal differencing loop used `self.d_ <= self.max_d`, so it could enter when `d_ == max_d`, apply one more `np.diff`, and set `d_` past the configured maximum — with `max_d=0` it still produced `d_=1`. Changed the guard to `self.d_ < self.max_d`.

Added `test_autoarima_respects_max_d_zero`: a trending series with `max_d=0` must keep `d_ == 0`. Verified it fails on the old `<=` and passes on the fix; full `test_arima.py` green (38 passed); pre-commit clean.

*AI disclosure: written with AI assistance (Claude); I reviewed, ran, and verified every line and the test behaviour myself.*

### DIFF
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -431,7 +431,10 @@ class AutoARIMA(BaseForecaster, IterativeForecastingMixin):
         series = np.array(y.squeeze(), dtype=np.float64)
         differenced_series = series.copy()
         self.d_ = 0
-        while not kpss_test(differenced_series)[1] and self.d_ <= self.max_d:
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
thanks, lgtm

PR #3615, opened 2026-07-07
[BUG] NaiveForecaster: validate seasonal_period for strategy='seasonal_last'

### LINKED ISSUE #3576: [BUG] NaiveForecaster should validate seasonal_period for strategy="seasonal_last"
### Describe the bug

NaiveForecaster._predict uses:

period = y_squeezed[-self.seasonal_period :]
idx = (self.horizon - 1) % self.seasonal_period
return period[idx]

There is no validation that seasonal_period is a positive integer or that there are enough observations to index the requested season. That can give low-level ZeroDivisionError or IndexError rather than a clear parameter error

import numpy as np
from aeon.forecasting import NaiveForecaster

y = np.arange(5.0)

NaiveForecaster(strategy="seasonal_last", seasonal_period=0).predict(y)
# ZeroDivisionError: integer modulo by zero

NaiveForecaster(strategy="seasonal_last", seasonal_period=10, horizon=10).predict(y)
# IndexError, because period has length 5 but idx can be 9

### Expected results

```python
import numpy as np
from aeon.forecasting import NaiveForecaster

y = np.arange(5.0)

NaiveForecaster(strategy="seasonal_last", seasonal_period=0).predict(y)
# ZeroDivisionError: integer modulo by zero

NaiveForecaster(strategy="seasonal_last", seasonal_period=10, horizon=10).predict(y)
# IndexError, because period has length 5 but idx can be 9
```

### Actual results

ZeroDivisionError: integer modulo by zero

### Suggested fix

check and raise value error

### Versions

_No response_

CedricConday:
Fixes #3576.

`NaiveForecaster._predict` with `strategy='seasonal_last'` did no validation on `seasonal_period`, so `seasonal_period=0` raised a low-level `ZeroDivisionError` (modulo by zero) and a period larger than the series raised `IndexError`. Added two guards that raise a clear `ValueError`: reject non-positive/non-integer periods, and reject a period exceeding the number of observations.

Added `test_naive_seasonal_last_validates_seasonal_period` covering `0`, `-1`, `None`, and an over-long period. Verified it fails on the unfixed code (`ZeroDivisionError`) and passes with the guards; full `test_base.py` green (24 passed); pre-commit clean.

*AI disclosure: written with AI assistance (Claude); I reviewed, ran, and verified every line and the test behaviour myself.*

### DIFF
--- a/aeon/forecasting/_naive.py
+++ b/aeon/forecasting/_naive.py
@@ -55,6 +55,19 @@ class NaiveForecaster(
         elif self.strategy == "mean":
             return np.mean(y_squeezed)
         elif self.strategy == "seasonal_last":
+            if (
+                not isinstance(self.seasonal_period, (int, np.integer))
+                or self.seasonal_period < 1
+            ):
+                raise ValueError(
+                    "seasonal_period must be a positive integer for "
+                    f"strategy='seasonal_last', got {self.seasonal_period!r}."
+                )
+            if self.seasonal_period > len(y_squeezed):
+                raise ValueError(
+                    f"seasonal_period ({self.seasonal_period}) cannot exceed the "
+                    f"number of observations ({len(y_squeezed)})."
+                )
             period = y_squeezed[-self.seasonal_period :]
             idx = (self.horizon - 1) % self.seasonal_period
             return period[idx]
--- a/aeon/forecasting/_naive.py
+++ b/aeon/forecasting/_naive.py
@@ -27,7 +27,6 @@ class NaiveForecaster(
             - "last" predicts the last value of the input series for all horizon steps.
             - "mean": predicts the mean of the input series for all horizon steps.
             - "seasonal_last": predicts the last season value in the training series.
-              Returns np.nan if the effective seasonal data is empty.
     seasonal_period : int, default=1
         The seasonal period to use for the "seasonal_last" strategy.
         E.g., 12 for monthly data with annual seasonality.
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
Thanks, this is fixes #3576.

One small validation issue before merge: bool is a subclass of int, so the current check accepts seasonal_period=True and treats it as 1. I think we should reject bools explicitly, consistent with the existing horizon validation tests.

Something like:

if (
    isinstance(self.seasonal_period, (bool, np.bool_))
    or not isinstance(self.seasonal_period, (int, np.integer))
    or self.seasonal_period < 1
):
    raise ValueError(...)

Please also add True to the invalid-period test cases.

Two minor cleanups: the over-long-period test would better reproduce the original issue if it used a horizon large enough to trigger the old IndexError, and the seasonal_last docstring still says it returns np.nan for empty effective seasonal data, which no longer matches the new validation behaviour.

CedricConday:
Thanks @TonyBagnall — all four addressed:

- **bool subclass:** validation now rejects `bool`/`np.bool_` explicitly before the int check, so `seasonal_period=True` no longer sneaks through as `1`.
- **test:** added `True` to the invalid-period cases.
- **over-long-period test:** now uses `horizon=15`, which would have overflowed the effective period and raised the original `IndexError` under the old code — so it reproduces #3576 before hitting the new `cannot exceed` guard.
- **docstring:** dropped the stale `np.nan`-on-empty line, since invalid/empty seasonal data now raises rather than returning `np.nan`.

(Heads up: I couldn't run the suite locally due to a NumPy 2.5 / numba ABI clash in my env, so I'm leaning on CI here.)

TonyBagnall (approved):
lgtm, thanks

Issue #3617, opened 2026-07-08
[ENH] Speed up DrCIF with exact changes


TonyBagnall:
### Describe the feature or idea you want to propose

part of #361

Deterministic speed up. DrCIF runtime is dominated by catch22 feature creation. 

## Setup

- Dataset: ArrowHead (211 cases, 1 channel, length 251), 3 classes.
- `DrCIFClassifier(n_estimators=100, n_jobs=1)`, numba JIT pre-warmed.
- Scripts: `profile_drcif.py` (cProfile, whole pipeline), `profile_catch22.py`
  (per-feature timing + length sweep).

## Headline numbers (clean, unprofiled wall time)

| phase   | time  |
|---------|-------|
| fit     | 68.6s |
| predict | 64.6s |
| total   | 133s  | (100 trees; defaults are 200 trees ⇒ ~4.5 min)

Predict is nearly as costly as fit: it re-runs the interval transforms on every tree.

## Where the time goes

catch22 is the overwhelming bottleneck. In the cProfile run,
`Catch22._transform_case` accounts for **~76% of cumulative runtime**; the 7
summary-stat functions (mean/std/slope/median/iqr/min/max) are cheap numba
row-ops. So DrCIF runtime ≈ catch22 runtime.

### Within catch22 (per-feature, clean wall-clock on 211×128)

The two outlier-timing features alone are **~43% of catch22 compute**:

| feature                          | share |
|----------------------------------|-------|
| `DN_OutlierInclude_p` (idx 13)   | ~21%  |
| `DN_OutlierInclude_n` (idx 14)   | ~21%  |
| welch/FFT (`SP_Summaries`, 15/20)| ~9%   |
| `SC_FluctAnal` rs_range/dfa (18/19) | ~10% |
| `SB_TransitionMatrix` (8)        | ~6%   |
| autocorrelation kernel (shared)  | ~5%   |
| everything else                  | ~26%  |

Measured directly: the two `_outlier_include` numba kernels take 45ms of a
105ms full-transform (43%); the shared autocorrelation kernel is only 5%; the
Python/joblib per-transform dispatch floor is ~2% (small in clean runs, but
heavily inflated by cProfile because catch22 is called ~5000 times, each
spinning up a joblib `Parallel(delayed(...))` over all cases).

Issue #3616, opened 2026-07-08
[ENH] Speed up HIVE-COTEv2.0 classifier


TonyBagnall:
### Describe the feature or idea you want to propose

summer project is to improve HC2 components, which doesnt get used enough because its too slow. 
this is then setting the scene for a redesign moving towards HC3.

Four components will have own issue
 - DrCIF #3618 #3620 
 - TDE #3625 
 - STC
- Arsenal

### Describe your proposed solution

For this issue, we will only look at exact or within tolerance improvements, and possibly the contract mechanism. Will leave threading alone for now

Issue #3474, opened 2026-06-18
[ENH] fastmath = True decorator for numba may not be fast ....


TonyBagnall:
### Describe the feature or idea you want to propose

deep in forecasting, and testing the effect of fastmath =true. It causes all sorts of problems with infinites and our Nelder-MEad optimiser. Anyway, surprising result of the tests is that it may be slower than other options

With 10 runs (medians, which strip warm-up outliers) the picture is even more decisive:

<img width="566" height="188" alt="Image" src="https://github.com/user-attachments/assets/cb614693-ca4b-4704-966d-aea8d4348a28" />

  1. fastmath=True is the slowest variant on every workload. This is not what fastmath is supposed to do. Strongly suggests the nnan/ninf flags are causing the optimiser to mishandle NaN simplex vertices, take more iterations, or converge
  to slower-to-fit model classes. The "speed" fastmath=True was buying us was a measurement artefact from the older bench where cache pollution let stale binaries shadow recompiles.
  2. {"contract"}, {"reassoc"}, and {"reassoc","contract"} are statistically indistinguishable and all ~30–40% faster than True. Any of them is correctness-safe (NaN/inf checks still work) AND a major speed win.

this needs some heavy duty experimental testing, but making the issue because its interesting

### Describe your proposed solution

test fastmath variants in slowest areas of the toolkit

PR #3618, opened 2026-07-08
[ENH] Speed up DrCIF part 1: Optimise Catch22 transformer

TonyBagnall:
part of #3617 and #3616 


this adds speed ups that do not change the output at all. This is done with claude code under my guidance. This PR is improvements to catch22, since it dominates the run time for DrCIF. Aiming for 2x speed up. All the changes are bitwise identical except the last, which could potentially give differences <1e-13 difference, but on tests doesnt. 


<img width="943" height="327" alt="image" src="https://github.com/user-attachments/assets/52d2f45f-809d-4030-bec6-969859d1d566" />

### catch 22 features

note all changes verified identical to the previous. If reviewer requires can provide more info. 

Validation: full-transform output is bit-for-bit identical to a reference saved before the change across 12 configs (univariate/multivariate, several lengths); the predict-time skip-mask path matches; the catch22 and  interval-forest test suites pass; the DrCIF doctest predictions are unchanged. 


| feature                          | share |
|----------------------------------|-------|
| `DN_OutlierInclude_p` (idx 13)   | ~21%  |
| `DN_OutlierInclude_n` (idx 14)   | ~21%  |
| welch/FFT (`SP_Summaries`, 15/20)| ~9%   |
| `SC_FluctAnal` rs_range/dfa (18/19) | ~10% |
| `SB_TransitionMatrix` (8)        | ~6%   |
| autocorrelation kernel (shared)  | ~5%   |
| everything else                  | ~26%  |

The two outlier-timing features alone are **~43% of catch22 compute**: These are all on outlier_include

## 1. _outlier_include

 Speed up catch22 _outlier_include without changing output                                                                                                                                                                        
                                                                                                                                                                                                                                   
  Replace the per-threshold rescan (O(num_thresholds*n) with per-iteration                                                                                                                                                        allocations) with a single-pass approach: bucket each position by its                                                                                                                                                            
  highest qualifying threshold, then sweep thresholds while maintaining                                                                                                                                                            
  count/min/max incrementally and reading the median from a Fenwick tree.                                                                                                                                                          
  Bit-for-bit identical to the previous implementation; ~8.7x faster.

More detail available if required, but this is very clever, would have taken some deep investigation to find this. 

## 2. welch-feature FFT

Batch catch22 welch-feature FFT across cases without changing output                                                                                                                                                             
                                                                                                                                                                                                                                   
  The two SP_Summaries welch features need np.fft.fft of the mean-centred                                                                                                                                                         
  series; _transform_case computed one small FFT per case, so numpy's fixed                                                                                                                                                        
  per-call overhead dominated. Precompute the FFT for the whole batch in a                                                                                                                                                         
  single np.fft.fft(..., axis=1) call and pass it into _transform_case.                                                                                                                                                            
                                                                                                                                                                                                                                   
  Bit-for-bit identical: a batched FFT row equals the per-case 1D FFT, and                                                                                                                                                         
  the subtracted mean still uses the same numba mean() (np.mean differs by                                                                                                                                                         
  ULPs under fastmath). Falls back to per-case for unequal-length input and                                                                                                                                                        
  honours the predict-time attribute-skip mask.    

at this point benchmark arrowhead 133s -> 84s. 

## 3. _sb_coarsegrain — remove redundant sorts

_sb_coarsegrain computes num_groups + 1 quantile thresholds by calling _quantile, and each _quantile call sorted the input array from scratch — so the same array was sorted 4 times per call. It now sorts once and reads every threshold from the sorted copy via a new _quantile_sorted helper (_quantile keeps its old signature by sorting then delegating). Results are bit-for-bit identical. This speeds the two features that use coarse-graining: entropy_pairs (SB_MotifThree) by ~21% and transition_matrix (SB_TransitionMatrix), reducing the full 22-feature transform from ~55 ms to ~50 ms (211×128).

## 4. Feature 12 whiten_timescale — compile the wrapper                                                                                                                                                                                
                                                                                                                                                                                                                                   
  _FC_LocalSimple_mean1_tauresrat was a plain-Python @staticmethod that made three separate numba calls (_local_simple_mean, _compute_autocorrelations, _ac_first_zero), marshalling the intermediate arrays back to Python between  each. It now carries @njit like its sibling feature functions, keeping the whole computation inside numba. Output is bit-for-bit identical (verified over 6.4k fuzz cases). This is a minor consistency/cleanup change — its     
  runtime is dominated by the autocorrelation FFT, so the measurable speedup is negligible.  


## 5 — bespoke least squares in _fluct_prop (rs_range + dfa)

This was my spot not claude, based on forecasting work, so I'm taking credit although we may ditch these features in part 2. Replaced the per-window _linear_regression (which re-summed the fixed grid d=[1..tau] every window) with closed-form grid moments computed once per tau, and for DFA computed the fluctuation as SSR-from-sums (Σy² − c1·Σxy − c2·Σy) instead of materialising and detrending the window.

- Correctness: bit-identical — 0.0 diff across 4,200 fuzz cases (including ties and near-perfect-linear inputs where SSR≈0 and the negative-clamp matters), plus the full-transform harness. The closed-form moments are exact integers, and the SSR-from-sums rounds identically here — so this one turned out exact, not just within tolerance.
- Speed: ~1.6× on rs_range + dfa (10.3→6.6 ms @128, 31.8→19.4 @512, 59.8→37.3 @1024). Those two are ~30% of Catch22, so ~1.1–1.15× on the full transform.

### 6: twiddle precompute — the FFT twiddle table in _compute_autocorrelations is rebuilt on every call but depends only on the (constant-within-a-transform) series length. 

Cant claim this, all claude. Basically features 2, 3, 8, 10 and 12 all need the autocorrelation, whose FFT twiddle factors depend only on the (common) series length. Build them once here and reuse across every case instead of rebuilding per autocorrelation call.

this is more benefit for full feature transform, but worth doing. 


## 6 — twiddle precompute (shared autocorrelation)
  
  Refactored _compute_autocorrelations into reusable pieces (_ac_nfft, _ac_twiddles, _autocorrelations_with_tw) and precompute the FFT twiddle table once per transform (via a new _ac_twiddle_cache, threaded
  through like the welch cache), reusing it across all cases for the shared ac used by features 2/3/8/10/12. The public _compute_autocorrelations(X) stays a thin wrapper, so anything else (incl. feature 12's own
  autocorrelation) is unchanged. 
  
  - Correctness: bit-identical — 0.0 diff on the full-transform harness and multivariate; the twiddles are deterministic np.exp, so this is exact (a "common function" reuse, not a numerical change).
  - Speed: ~1.12–1.17× on the acf-cluster features in isolation (the twiddle loop was ~17% of each autocorrelation, now computed once instead of per case).
          
Looking at fastmath. When True it causes problems with infinities. We found on #3474 that we can probably trade to contract with no loss of runtime and safer infinities. First test with catch22

<img width="1039" height="267" alt="image" src="https://github.com/user-attachments/assets/91b93c4d-916b-4bb8-81a8-039dd80ac846" />
<img width="564" height="235" alt="image" src="https://github.com/user-attachments/assets/d3df9d52-ed59-45b0-b140-8307d653291a" />

so 14% fastmath stays here at least, given our need for speed.


TonyBagnall:
image

once this is in we should deprecate the use_catch22 option

TonyBagnall:
so finished on Catch22. Enhanced the notebook to include an overview of the changes, which primarily involves caching transforms used in different feature sets. 

Ran a comparison against the version on main, the c version and the legacy sktime code, python below with a work around to get the version on main. New version checked for agreement in values to the old version. There are differences: pycatch22 internally normalises each series, whereas aeon doesnt: if we pre normalise the data aeon is the same as pycatch22. Sktime does not normalise, but it has not had any of the bug fixes aeon has:
- stretch_high / stretch_decreasing differ by 13.0 — the long-stretch counting fix;                                                                                                                                                                    
- acf_timescale differs by 0.47 — aeon's interpolated 1/e crossing vs the older integer crossing;                                                                                                                                                      
- outlier_timing_neg returns NaN on all 50 gaussian series in sktime — that's your "one-sided NaNs 50", and it looks like an outright sktime defect;                                 
sktime also carries a lot of boilerplate baggage. so is very slow

image

image

TonyBagnall:
```python
"""Compare catch22 implementations: aeon current vs old aeon vs C vs sktime.

Hard-coded: transforms random collections (length 200, univariate) of
increasing n_cases, checks agreement of the answers, times each
implementation (best of 3) and plots runtime + prints a speedup table.

    python local/drcif/compare_four.py
"""

import importlib.util
import os
import shutil
import sys
import time

import matplotlib

matplotlib.use("Agg")
import numpy as np
import pycatch22

from aeon.transformations.collection.feature_based import Catch22
from aeon.transformations.collection.feature_based._catch22 import feature_names
from sktime.transformations.panel.catch22 import Catch22 as SktimeCatch22

HERE = os.path.dirname(os.path.abspath(__file__))
N_CASES = [50, 100, 250, 500, 1000, 2000,4000]
LENGTH = 400

# old aeon: the true original module before any speedups (commit 0287deb1,
# exported to _catch22_true_orig.py). Clear the stale numba cache first, it is
# keyed to the module name used when the file was last loaded.
shutil.rmtree(os.path.join(HERE, "__pycache__"), ignore_errors=True)
spec = importlib.util.spec_from_file_location(
    "_c22_true_orig", os.path.join(HERE, "_catch22_true_orig.py")
)
_orig_mod = importlib.util.module_from_spec(spec)
sys.modules["_c22_true_orig"] = _orig_mod
spec.loader.exec_module(_orig_mod)

# pycatch22 and sktime both use the published catch22 feature order, aeon uses
# its own - map their output columns onto aeon's order by feature name
from sktime.transformations.catch22 import CATCH22_FEATURE_NAMES  # noqa: E402

_pyc_probe = pycatch22.catch22_all(list(range(30)))
_pyc_order = [_pyc_probe["names"].index(name) for name in feature_names]
_sk_order = [list(CATCH22_FEATURE_NAMES).index(name) for name in feature_names]


def run_current(X):
    return Catch22().fit_transform(X)


def run_old(X):
    return _orig_mod.Catch22().fit_transform(X)


def run_c(X):
    out = np.array(
        [pycatch22.catch22_all(X[i, 0].tolist())["values"] for i in range(len(X))]
    )
    return out[:, _pyc_order]


def run_sktime(X):
    return SktimeCatch22().fit_transform(X).to_numpy()[:, _sk_order]


IMPLS = [
    ("aeon current", run_current),
    ("aeon old", run_old),
    ("pycatch22 (C)", run_c),
    ("sktime", run_sktime),
]


def agreement(a, b):
    # max abs diff over positions where both are real, plus a count of
    # one-sided NaNs (one implementation NaN, the other not)
    nan_a, nan_b = np.isnan(a), np.isnan(b)
    one_sided = int(np.sum(nan_a != nan_b))
    mask = ~(nan_a | nan_b)
    return float(np.max(np.abs(a[mask] - b[mask]))), one_sided


def main():
    # agreement check on one small collection
    X, _ = make_example_3d_numpy(n_cases=50, n_channels=1, n_timepoints=LENGTH)
    ref = run_current(X)
    print("agreement with aeon current:")
    for name, fn in IMPLS[1:]:
        d, ns = agreement(fn(X), ref)
        print(f"  {name:14s} max abs diff {d:.3e}   one-sided NaNs {ns}")

    # timing loop
    times = {name: [] for name, _ in IMPLS}
    for n in N_CASES:
        X, _ = make_example_3d_numpy(n_cases=n, n_channels=1, n_timepoints=LENGTH)
        for name, fn in IMPLS:
            best = np.inf
            for _ in range(3):
                t0 = time.perf_counter()
                fn(X)
                best = min(best, time.perf_counter() - t0)
            times[name].append(best)
        print(f"n_cases={n} done")

    # speedup table (relative to aeon current)
    print(f"\nruntime in seconds (length {LENGTH}), speedup = impl / aeon current")
    header = f"{'n_cases':>8} " + "".join(f"{name:>16}" for name, _ in IMPLS)
    print(header)
    print("-" * len(header))
    for i, n in enumerate(N_CASES):
        row = f"{n:>8} "
        for name, _ in IMPLS:
            row += f"{times[name][i]:>16.3f}"
        print(row)
        row = f"{'':>8} "
        for name, _ in IMPLS:
            row += f"{times[name][i] / times['aeon current'][i]:>15.1f}x"
        print(row)

    # plot
    for name, _ in IMPLS:
        plt.plot(N_CASES, times[name], marker="o", label=name)
    plt.xscale("log")
    plt.yscale("log")
    plt.xlabel("n_cases")
    plt.ylabel("transform time (s)")
    plt.title(f"catch22 implementations, length {LENGTH}")
    plt.legend()
    plt.grid(True, which="both", alpha=0.3)
    out = os.path.join(HERE, "compare_four.png")
    plt.savefig(out, dpi=120, bbox_inches="tight")
    print(f"\nsaved plot to {out}")


if __name__ == "__main__":
    main()
```

TonyBagnall:
Final correctness status: this is output-preserving for DrCIF in the tested reference suite, but the Catch22 transform itself is no longer strictly bitwise identical in every path. Most feature rewrites are bit-identical; the batched autocorrelation path can shift `CO_f1ecac` at ~1e-14 scale due to pocketfft vs the previous numba FFT, and recompilation under fastmath can introduce sub-ULP differences in a couple of features. No class/regression predictions changed in the DrCIF reference checks.

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -287,28 +287,188 @@ def _transform(self, X, y=None):
...
+                f_arr = np.asarray(f_idx, dtype=np.int64)
+                # catch24's standard deviation stays a numpy call for exact
+                # reproducibility (numba's np.std can round differently)
+                stds = None
+                if 23 in f_idx:
+                    stds = np.empty((n_cases, n_channels))
+                    for i in range(n_cases):
+                        for c in range(n_channels):
+                            stds[i, c] = np.std(X[i, c])
+                # typed empty placeholders for absent caches; the matching
```
hadifawaz1999, line 340: i didnt get much why we need to do this for feature 23
do we need loops ?

"# catch24's standard deviation stays a numpy call for exact
                # reproducibility (numba's np.std can round differently)"
makes sense just wondering if we can't do np.std(X, axis=2)
wouldnt be much faster but better to vectorize i think

TonyBagnall: good question, it is to make sure it is bit exact to previous version, which processes series by series calling np.std. which did this
```python
elif feature == 23:                                                                                                                                                                                    
      c22[dim + n] = np.std(series) # numpy np.std, per 1-D series     
```
The problem with np.std(X, axis=2) is it is only exactly equivalent if the data is stored contiguously in memory - I checked, and data does come that way. We could change it to  
`np.std(np.ascontiguousarray(X), axis=2)`

but the difference is very very small in time,

hadifawaz1999 (changes requested):
Looks great, i think it would be best if we add a small test to make sure results are the same with pycatch22 just to avoid future come back to this and question it again

aeon/transformations/collection/feature_based/_catch22.py
hadifawaz1999: good to know ! given it doesnt make it slower its okay with me

TonyBagnall: just testing that actually

TonyBagnall:
> Looks great, i think it would be best if we add a small test to make sure results are the same with pycatch22 just to avoid future come back to this and question it again

done

aeon/transformations/collection/feature_based/_catch22.py
TonyBagnall: actually  does speed it up! Added, changed to np.std(np.ascontiguousarray(X), axis=2)

hadifawaz1999 (approved):
LGTM ! great addition to 1.5.1 i think :)

PR #3620, opened 2026-07-08
[ENH] Speed up interval-based forests (BaseIntervalForest / RandomIntervals)   

### LINKED ISSUE #3617: [ENH] Speed up DrCIF with exact changes

TonyBagnall:
part of #3616 closes #3617 

this is minor tweaks to the code behind DrCIF independent of Catch22. Removes the use of some list comprehensions, some unnecessary stacking and a couple of other things. 

This speeds up the machinery shared by every interval-based forest — BaseIntervalForest and the RandomIntervals transformer — independently of feature extraction. It touches no feature code (Catch22  etc.), so all interval forests inherit the gains: TimeSeriesForest, CanonicalIntervalForest, DrCIF, RISE and the generic IntervalForest, both classifier and regressor, plus the standalone RandomIntervals transformer.

## Three changes:                                                                                                                                                                                         
                                                                                                                                                                                                         
  1. Vectorise per-interval feature assembly. The interval transform built its output as an n_cases-long Python list-of-lists per feature ([[f] for f in feature(...)]) and then converted it. This is   
  replaced with np.asarray(feature(...)).reshape(-1, 1), keeping the data in NumPy throughout. This is the main win.                                                                                     
  2. Concatenate once instead of growing. Interval/feature blocks were accumulated with Xt = np.hstack((Xt, ...)) inside the loop, which reallocates and copies the whole array on every iteration       
  (quadratic in the number of intervals). Blocks are now collected into a list and joined with a single np.hstack(...).                                                                                  
  3. Skip joblib when n_jobs == 1. New _eval_estimators / _eval_intervals helpers run a pre-built list of delayed tasks in a plain loop when not parallelising, avoiding Parallel's per-task             
  wrapping/copy overhead, and via Parallel otherwise. Building the task list up front preserves the RNG draw order, so n_jobs > 1 results are byte-for-byte unchanged


## Correctness                                                                                                                                                                                            
                                                                                                                                                                                                         
  Bit-identical, verified: classifier predict_proba, regressor predict, and train-set estimates are unchanged, including with n_jobs=2 (the pre-built task list preserves interval/attribute selection). All interval-based tests pass.                                                                                                                                                                         
                                                                                                                                                                                                         
## Docs                                                                                                                                                                                                   
                                                                                                                                                                                                         
  Adds examples/classification/interval_based/drcif.ipynb — a DrCIF walkthrough (the three representations, fitting/evaluation vs TSF, reproducing a tree's feature matrix and the ensemble vote by hand,
  temporal importance curves, parameters, runtime) — and links it from the examples gallery.        


## Speed ups

DrCIF gets the least benefit, because its time is dominated by catch22, about 5%, but this helps a lot of estimators. See below


        

.


TonyBagnall:
running black box tests on UCR, but seems about a 2.3x speed up, which is nice

image

TonyBagnall:
Equivalence on UCR confirmed, 30 resamples of all rerun on HALI cluster, 2.4x speed up
image

image

TonyBagnall:
speed ups without sktime version, which seems fundamentally broken, running four hours others done in minutes

image

TonyBagnall:
note the above speed ups included the Catch22 improvements, so the benefit of this PR are more modest for DrCIF, but help a lot of estimators

TonyBagnall:
<img width="488" height="398" alt="image" src="https://github.com/user-attachments/assets/a118afdd-86e4-455f-a943-682ddf507a30" />

 Reading this:                                                                                                                                                                                                                             
  - The win is in predict for summary-statistic interval forests — TimeSeriesForest and IntervalForest go 2.1–3.5× faster, classifier and regressor. predict is pure interval feature extraction, so the framework overhead the PR removes  
  (per-interval Python list building and repeated hstack copies) was a large fraction of it there.                                                                                                                                          
  - fit gains are modest (1.1–1.35×) because fit is dominated by decision-tree building, which this PR doesn't touch.                                                                  
  - Catch22-based forests (CIF, DrCIF) barely move (~1.0–1.2×) — they're feature-extraction-bound, so the framework is a small slice. Their speed comes from the separate Catch22 PR, not this one.                                         
  - RISE and SupervisedTimeSeriesForest are unchanged, as expected: RISE uses whole-series intervals (the per-interval assembly barely applies) and STSF uses SupervisedIntervals, which this PR doesn't modify

TonyBagnall:
Regression trees cost ~1.5–1.8× more to build than classification trees, because the default regressor uses DecisionTreeRegressor(criterion="absolute_error") (line 374 of base_interval_forest.py) — MAE splitting, which is much slower 
  than the classifier's entropy. So for a regressor, tree building is a bigger slice of fit, feature extraction is a smaller slice, and this PR's extraction speedup gets diluted more. Concretely: extraction is ~25% of TSF-classifier fit
  but only ~16% of TSF-regressor fit, so the same 2.4× on extraction turns into 1.35× fit for the classifier and only 1.09× for the regressor.

TonyBagnall:
Ran three comparisons. Tried the sktime version too, but sktime's DrCIF ran 17+ hours on a single data point without producing a result — over 100× the ~9 minutes all three aeon versions took combined — so it was never going to yield a usable number, and its ~15–30x slowness versus aeon was already established from the Catch22 comparison anyway.

Not much benefit to DrCIF on this PR, but overall worth it for the wider impact

image

MatthewMiddlehurst (approved):

PR #3622, opened 2026-07-08
[ENH] Add SARIMA forecaster

TonyBagnall:
credited to @GiGiKoneti, done this way so I can benchmark and test it. Still wip

### DIFF
--- a/aeon/forecasting/utils/_nelder_mead.py
+++ b/aeon/forecasting/utils/_nelder_mead.py
@@ -3,7 +3,7 @@
 import numpy as np
 from numba import njit
 
-from aeon.forecasting.utils._loss_functions import _arima_fit, _ets_fit
+from aeon.forecasting.utils._loss_functions import _arima_fit, _ets_fit, _sarima_fit
 
 
 @njit(cache=True)
@@ -12,6 +12,8 @@ def dispatch_loss(fn_id, params, data, model):
         value = _arima_fit(params, data, model)
     elif fn_id == 1:
         value = _ets_fit(params, data, model)[0]
+    elif fn_id == 2:
+        value = _sarima_fit(params, data, model)
     else:
         raise ValueError("Unknown loss function ID")
     if np.isnan(value) or np.isinf(value):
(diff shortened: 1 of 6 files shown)


aeon/forecasting/utils/_nelder_mead.py
```diff
@@ -12,6 +12,8 @@ def dispatch_loss(fn_id, params, data, model):
         value = _arima_fit(params, data, model)
     elif fn_id == 1:
         value = _ets_fit(params, data, model)[0]
+    elif fn_id == 2:
```
GiGiKoneti, line 15: Adding `fn_id == 2` here conflicts with the existing unit test `test_dispatch_loss_unknown_id_raises` in `aeon/forecasting/utils/tests/test_utils.py`, which still expects `2` to be an unknown ID.

Since `2` is now mapped here, calling `dispatch_loss(2, ...)` in that test attempts to run `_sarima_fit` with an incompatible mock model array, causing an out-of-bounds access in Numba that segfaults the test runner on macOS and Linux.

We should update `test_dispatch_loss_unknown_id_raises` in `aeon/forecasting/utils/tests/test_utils.py` to use `3` instead:
```python
def test_dispatch_loss_unknown_id_raises():
    """An unknown loss function id raises a ValueError."""
    with pytest.raises(ValueError):
        dispatch_loss(
            3,
            np.array([0.5]),
            np.arange(10.0),
            np.array([1, 0, 0, 1], dtype=np.int32),
        )
```

TonyBagnall: yes seems correct, want to correct it?

GiGiKoneti:
Hi @TonyBagnall, I've opened #3629 targeting your branch with the test fix.

TonyBagnall:
unrelated fail, rerunning

GiGiKoneti:
@TonyBagnall 
anything else we have here ?

GiGiKoneti:
@TonyBagnall 
Can you review this please ?

Issue #3621, opened 2026-07-08
[ENH] Speed up DrCIF/Catch22 with non exact changes


TonyBagnall:
### Describe the feature or idea you want to propose

DrCIF runtime is dominated by catch22. Three ways to further improve speed I can think of, which will change behaviour and may require subtyping/version changing or something.

1. Improve the efficiency of existing catch22 features with non exact changes

some could be implemented with approximate, faster algorithms. For example,  periodicity (PD_PeriodicityWang) is10% of the 22 calculation and is O(n^2) with a bespoke fft. WE can replace with a nlogn version, but its not identical. Two are dependent on many tiny regressions which could be made better.

2. Ditch certain variables.

SC_FluctAnal and and FC_LocalSimple (rs_range and whiten) take up 25% and 11% of the total respectively. If we can just not use these, we reduce transform time by at least 20%.

3. Restructure intervals

the big speed up, just change how DrCIF works. Simple things about parameters eg.

1. Reduce the number of catch22 selected at each feature
2. Fix depth of trees
3. Reduce number of intervals per tree

that is simply making model smaller. We can also get speed up by sharing intervals to reduce the number of transforms. Basic idea is to batch trees, create one large pool of intervals, share proportion of intervals to each tree, then find intervals once. This is the biggest possible speed up, potentially 5x faster, depending on the level of sharing.

 

### Describe your proposed solution

All of this will need validating, and there is also the issue of whether we paramateris changes or just bake them in

TonyBagnall:
completed by #3623

PR #3623, opened 2026-07-08
[ENH] Speed up DrCIF part 3: Catch22 speed ups that may be numerically different

TonyBagnall:
Part 1 of #3621 

these are basically code changes using faster algorithms or implementations that do not necessarily produce identical output, but are conceptually the same (e.g. FFT instead of DFT). 

### Setup                                                                                                                                                                                                            
                                                                                                                                                                                                                   
  - Cloned _catch22.py → _catch22_fast.py (class Catch22Fast), exported from feature_based/__init__.                                                                                                               
  - DrCIF (classifier + regressor) and CIF (classifier + regressor) now use Catch22Fast.     

keep the original, err, canonical one, the same, have faster version.

### Speedup #1 — bespoke least squares in _fluct_prop (rs_range + dfa)

This was my spot not claude, based on forecasting work, so I'm taking credit although we may ditch these features in part 2. Replaced the per-window _linear_regression (which re-summed the fixed grid d=[1..tau] every window) with closed-form grid moments computed once per tau, and for DFA computed the fluctuation as SSR-from-sums (Σy² − c1·Σxy − c2·Σy) instead of materialising and detrending the window.

- Correctness: bit-identical — 0.0 diff across 4,200 fuzz cases (including ties and near-perfect-linear inputs where SSR≈0 and the negative-clamp matters), plus the full-transform harness. The closed-form moments are exact integers, and the SSR-from-sums rounds identically here — so this one turned out exact, not just within tolerance.
- Speed: ~1.6× on rs_range + dfa (10.3→6.6 ms @128, 31.8→19.4 @512, 59.8→37.3 @1024). Those two are ~30% of Catch22, so ~1.1–1.15× on the full transform.

### Speedup #2: twiddle precompute — the FFT twiddle table in _compute_autocorrelations is rebuilt on every call but depends only on the (constant-within-a-transform) series length. 

Cant claim this, all claude. Basically features 2, 3, 8, 10 and 12 all need the autocorrelation, whose FFT twiddle factors depend only on the (common) series length. Build them once here and reuse across every case instead of rebuilding per autocorrelation call.

this is more benefit for full feature transform, but worth doing. 


Speedup #2 — twiddle precompute (shared autocorrelation)
  
  Refactored _compute_autocorrelations into reusable pieces (_ac_nfft, _ac_twiddles, _autocorrelations_with_tw) and precompute the FFT twiddle table once per transform (via a new _ac_twiddle_cache, threaded
  through like the welch cache), reusing it across all cases for the shared ac used by features 2/3/8/10/12. The public _compute_autocorrelations(X) stays a thin wrapper, so anything else (incl. feature 12's own
  autocorrelation) is unchanged. 
  
  - Correctness: bit-identical — 0.0 diff on the full-transform harness and multivariate; the twiddles are deterministic np.exp, so this is exact (a "common function" reuse, not a numerical change).
  - Speed: ~1.12–1.17× on the acf-cluster features in isolation (the twiddle loop was ~17% of each autocorrelation, now computed once instead of per case).
  - Canonical Catch22 untouched.
  
  Cumulative so far (Catch22Fast vs canonical Catch22)
  
  Full 22-feature transform is now ~1.2× faster, entirely bit-identical:                                                                                                                                                 
                                                                                                                                                                                                                   
  Both #1 (fluct_prop bespoke LS) and #2 (twiddle precompute) turned out exact, so I will port them into the exact branch!

### DIFF
--- a/aeon/classification/interval_based/_cif.py
+++ b/aeon/classification/interval_based/_cif.py
@@ -11,7 +11,7 @@ import numpy as np
 from aeon.base._estimators.interval_based import BaseIntervalForest
 from aeon.classification import BaseClassifier
 from aeon.classification.sklearn import ContinuousIntervalTree
-from aeon.transformations.collection.feature_based import Catch22
+from aeon.transformations.collection.feature_based import Catch22Fast
 from aeon.utils.numba.stats import row_mean, row_slope, row_std
 
 
@@ -190,7 +190,7 @@ class CanonicalIntervalForestClassifier(BaseIntervalForest, BaseClassifier):
             replace_nan = 0
 
         interval_features = [
-            Catch22(outlier_norm=True, use_pycatch22=use_pycatch22),
+            Catch22Fast(outlier_norm=True, use_pycatch22=use_pycatch22),
             row_mean,
             row_std,
             row_slope,
--- a/aeon/classification/interval_based/_drcif.py
+++ b/aeon/classification/interval_based/_drcif.py
@@ -14,7 +14,7 @@ from aeon.base._estimators.interval_based import BaseIntervalForest
 from aeon.classification.base import BaseClassifier
 from aeon.classification.sklearn._continuous_interval_tree import ContinuousIntervalTree
 from aeon.transformations.collection import PeriodogramTransformer
-from aeon.transformations.collection.feature_based import Catch22
+from aeon.transformations.collection.feature_based import Catch22Fast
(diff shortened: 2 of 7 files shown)


TonyBagnall:
the differences are so small I will roll this into the other PR

TonyBagnall:
merged into #3621 as no practical difference

PR #3625, opened 2026-07-09
[ENH] Speed up TDE

### LINKED ISSUE #3418: [ENH] Change TDE and IndividualTDE parameter max_dims

both these classifiers have a parameter to cap the max number of channels the classifier uses. This is currently called max_dims, because it was written before we settled on channels. 

A simple refactor to max_n_channels would help. I dont think this requires deprecation by our standard policy @MatthewMiddlehurst ? 

_No response_

### Additional context

_No response_

### LINKED ISSUE #3624: [ENH] Speed up TDE classifier

the slowest of the lot especially on multivariate. Optimisation plan

Top level profile

Image

optimisation plan/things to explore
1. SFA: I'll work on a clone to avoid clashes with a range of other PRs
2. IndividualTDE: speed up transform if possible
3. Intersection/NN
4. TBC

_No response_

### Additional context

_No response_

TonyBagnall:
part of #3616 closes #3624 

also deprecates max_dims for max_channels, closes #3418 

Stages:

1.  Use a specialised SFA implementation
2. Optimise TDE
 
Every kept change was verified output-identical on univariate and multivariate train/test probabilities (stash-compare harness), and TDE tests + expected-output regression + ordinal + full estimator checks all pass. It may be ott, but people are nervous about AI code, so I have got it to write a notebook explaining the structure as I walked through the code with it..

## Stage 2: Use a specialised SFA implementation

This PR adds a TDE-specific SFA implementation and wires IndividualTDE to use it. The goal is to remove generic SFA overhead from TDE while preserving current TDE behaviour.

The new implementation keeps only the SFA features TDE actually uses: alphabet size 4, univariate channel input, numerosity reduction, MCB/IGB binning, bigrams, spatial pyramid levels, and reduced binning bags for multivariate  dimension selection.

###  What Changed

  Added aeon/classification/dictionary_based/_tde_sfa.py, a SFA rewrite specialised for TDE.

TDE repeatedly builds SFA bags and performs nearest-neighbour comparisons during parameter search. Generic SFA supports many behaviours TDE does not use, and its dictionary bag representation adds overhead. Moving directly to
  sorted arrays gives TDE a representation that is ready for fast merge intersections and avoids repeated dict-to-array conversion.

  Compared with generic SFA, the new version:

  - Uses fixed alphabet size 4 and 2-bit word letters.
  - Assumes univariate input; TDE slices channels before calling it.
  - Always applies numerosity reduction.
  - Stores bags as sorted array segments: (keys1, keys2, counts, offsets).
  - Avoids Python/numba dictionaries in the main TDE path.
  - Uses merge-based histogram intersection over sorted arrays.
  - Reuses the sliding-window MFT in fit_transform.
  - Uses MFT stride rows for MCB breakpoints.
  - Uses direct non-overlapping DFT for IGB and reduced binning bags to preserve exact old-SFA parity.
  - Reimplements IGB in numba as a best-first entropy tree instead of fitting one sklearn decision tree per coefficient.
  - Casts IGB DFT values through float32 to match sklearn tree threshold behaviour.
  - Supports TDE bigram and spatial-pyramid key encoding.

  ### Validation

  Ran:

  - pytest local\tde\test_sfa.py -q -> 42 passed
  - pytest aeon\classification\dictionary_based\tests\test_tde.py -q -> 6 passed
  - py_compile on touched modules
  - local\tde\bench_sfa.py
  - smoke run of local\tde\bench_tde_scaling.py

## Stage 2: TDE Improvements

  - Instantiate TDESFA instead of generic SFA.
  - Store transformed training data as array-backed bags.
  - Use array-backed LOOCV nearest-neighbour kernels.
  - Combine multivariate selected-dimension bags into the same array representation.
  - Clear retained transformer _fit_X and _fit_mft caches after bags are built to reduce memory use.
  - Preserve old dict-based fallback helpers for compatibility paths.
 - Symmetric LOOCV NN kernel (sim(i,j)=sim(j,i))                                                                                                                                                                                          
 - Replace KernelRidge/StandardScaler GP selection with numpy ridge                                                                                                                                                                         
 - Trig tables in _binning_dft_all and _mft_all first window                                                                                                                                                                                
 - Flat NN kernels for levels==1 (skip keys2)                                                                                                                                                                                              
 - Ensemble-level MFT/binning-DFT cache per (window_size, norm)
###  Observed Benchmark Direction

  The focused SFA benchmark showed large fit+transform speedups on tested configs, including MCB-heavy configs around 61x and 72x in one run. The scaling smoke benchmark also showed improved IndividualTDE timing and lower  retained memory after clearing transformer caches.

The notebook provides more detail on the implementation

Also added some skips for validation to gain a small speed up


patrickzib:
Not sure, I understand. So there is a new tde_sfa.py only intended for tde. 

Why are the changes to the file _sfa.py needed in this case?

TonyBagnall:
so will put the evaluation here to stop bloating the description. tl;dr, looking very promising

Early results on UCR show effective equivalence, reran old TDE from 1.5 release. This iis single sample and only 75

image


Currently running on all the EEG multivariate and the UCR,

seemingly 20x speed up fit and 5-10x on predict, which is remarkable. memory is higher, but possible to mitigate that in a few ways, may not impact speed

image

image

image

so there is a memory trade-off it is for example allocating a lot of memory in predict for small speed ups, still looking at order of magnitude speed up, I'll take that

TonyBagnall:
now comparing against 30 resamples on each on previously generated results, still on UCR. Only have 82 complete for old TDE, but its enough, nothing significant here 
image
 
very close mean, identical median 

image

Predict timing for 83 : Old TDE approx 20 hours, new TDE 2 hours. Not sure timings are comparable, but I had hoped for 10x

More on timing: on the new run:

fit and estimate time on 83: TDE_New: 26 mins, TDE_Old: 339 minutes
predict: TDE_New: 44 mins, TDE_Old: 360 minutes

interesting predict takes longer than fit (and produce train estimates), 1-NN classification for you, but might be interesting to try factor it out. Have tried trees before, but now we have some speed, worth revisiting

TonyBagnall:
just did a memory check and new branch appears to use less memory on UCR, although not convinced of the measurements. After black boxing on results, I will run a local time/memory experiment. EEG, not got as many to compare, but seems little difference

image

image

35x speed up on these data, multivariate for the win. Shame its not more accurate at EEG, but different issue.

TonyBagnall:
I'll share this walkthough with fable, might help a reviewer. Just TDE structure on this thread, will describe SFA separately. 

## Part 1: Storing the histograms

Everything else in the file follows from one data-structure decision. The old TDE stored each training case's word histogram as a numba typed Dict (word → count), and every operation — building bags, comparing them, pickling them — paid a Python↔numba boundary cost per dictionary operation. The new TDE stores all bags for a model in four flat numpy arrays:

- (keys1, keys2, counts, offsets)
-  keys1   int64  – the word (or packed bigram)
-  keys2   int64  – quadrant / dimension tag (0 when unused, -1 for bigrams)
-  counts  uint32 – the histogram count
-  offsets int64  – case i's bag is rows offsets[i] : offsets[i+1]

So self._transformed_data in a fitted IndividualTDE is one 4-tuple of arrays covering every training case, with each case's segment sorted lexicographically by (keys1, keys2). That sorting is the point: two sorted bags can 
  be compared with a merge intersection (two pointers walking both segments, like merge-sort's merge step) instead of hash lookups — and the whole nearest-neighbour search becomes a handful of long-running numba kernel calls 
  instead of hundreds of thousands of small ones.       

Its all packed into an int, but essentially key1 is the word, key2 is the location (channel/spatial pyramid level), counts in the number of occurrences, and offset unpacks all of this info into case information. I've added a notebook to explain this core datastructure. 

 If you know scipy sparse matrices, it's exactly the indptr array of CSR format — the bags are effectively one big sparse matrix (cases × possible words) stored row-compressed, except the "column index" is the compound      
  (keys1, keys2) pair.  

## Part 2: Predicting a new case

A single TDE member classifies a test case by comparing its bag against every training bag and taking the class of the most similar one (1-nearest-neighbour). The rewrite does this for  all test cases at once in one kernel call:          

```python
 _tde.py:911  (inside IndividualTDE._predict)                                                                                                                                                                                 
  sims = nn_similarities_all(                                                                                                                                                                                                    
      *self._transformed_data, keys1, keys2, counts, t_offsets                                                                                                                                                                   
  )                    
```  
`nn_similarities_all` takes the whole training model (four arrays) and the whole test transform (four arrays) and returns an `(n_test × n_train)` integer matrix — every pairwise merge intersection. Row i is test case i's similarity to each training case; the prediction is essentially `class_vals[argmax(row)].`

except "essentially" hides the interesting part: ties. Two training cases often reach the same similarity, and TDE's historical rule is a coin flip per tie event during a left-to-right scan — take the tied index if rng.random() < 0.5. Preserving that rule exactly while staying fast took two observations:

Observation 1 (nn_first_max, _tde_sfa.py): the running maximum of the scan never depends on the coin flips — ties don't change best_sim, only which index is remembered. So a numba kernel can find each row's first-maximum and simply flag whether any tie event occurred; tie-free rows are fully resolved with no randomness at all.

Observation 2 (nn_tie_break): with an integer random_state, the scan constructs a fresh, identically seeded generator per test case — meaning every case consumes the same draw sequence. So one pool of draws, computed once, serves all cases, and the entire tie-respecting scan moves into numba:

```python
if isinstance(self.random_state, (int, np.integer)) and not isinstance(
    self.random_state, bool
):
    # with an integer seed every case's tie-break generator yields
    # the same sequence, so one precomputed draw pool resolves all
    #cases inside numba, exactly as per-case generators would
    #draws = check_random_state(self.random_state).random(sims.shape[1])
    #nn_idx = nn_tie_break(sims, draws)
    #classes = [self._class_vals[nn_idx[i]] for i in range(n_cases)]
else:
    #unseeded or shared generators consume draws across cases, so
    #tie events must be resolved sequentially; the running maximum
    #is draw-independent, so tie-free cases are resolved in numba
    nn0, has_tie = nn_first_max(sims)
    classes = [
        self._nn_from_sims(sims[i]) if has_tie[i] else self._class_vals[nn0[i]]
        for i in range(n_cases)
    ]
```
This mattered more than it looks: on our benchmark 87% of test cases hit at least one tie event, and each one used to seed a fresh Mersenne Twister (~50µs of GIL-held Python). Eliminating that cut single-threaded predict from 495ms to 207ms — before any threads.

The last piece is the `n_jobs` block : with `n_jobs > 1` the test cases are chunked across threads, each chunk being one `nn_similarities_all` call on a slice of `t_offsets`. The kernels carry `nogil=True`, so threads genuinely run concurrently, and the chunk matrices are vstacked in order — identical results for any `n_jobs`.

## Part 3: the ensemble fit loop — TemporalDictionaryEnsemble._fit

The ensemble's job is to pick, from a big grid of member configurations, the `max_ensemble_size` (default 50) best ones. The grid is every combination of window size, word length, norm, pyramid levels and binning method      
  (`_unique_parameters`, _tde.py:545) — typically a few thousand candidates. Evaluating one candidate means fitting a member on a 70% subsample and estimating its accuracy by LOOCV, so TDE can't try them all; it samples        
 `n_parameter_samples` (default 250) in two phases:                                                                                                                                                                               
                                                                                                                                                                                                                                 
  Phase 1 — random (first `randomly_selected_params`, default 50): pick candidates uniformly.                                                                                                                                      
                                                                                                                                                                                                                                 
  Phase 2 — guided: model accuracy as a function of parameters from the members evaluated so far, and pick the candidate with the best predicted accuracy:                                                                       
                                                                                                                                                                                                                                 
  _tde.py:339                                                                                                                                                                                                                  
```python  
preds = _kernel_ridge_preds(                                                                                                                                                                                                   
      np.array(self._prev_parameters_x, dtype=np.float64),                                                                                                                                                                       
      np.array(self._prev_parameters_y, dtype=np.float64),                                                                                                                                                                       
      candidate_parameters,                                                                                                                                                                                                      
  )                                                                                                                                                                                                                              
  idx = rng.choice(np.flatnonzero(preds == preds.max()))                                                                                                                                                                         
```                                              

Phase 2 — guided: model accuracy as a function of parameters from the members evaluated so far, and pick the candidate with the best predicted accuracy:
  
   _tde.py:339
```python  
  preds = _kernel_ridge_preds(
      np.array(self._prev_parameters_x, dtype=np.float64),
      np.array(self._prev_parameters_y, dtype=np.float64),
      candidate_parameters,
  )
  idx = rng.choice(np.flatnonzero(preds == preds.max()))
```
   
  Two changes here versus the old code. First, the regressor: it used to build a sklearn `StandardScaler + KernelRidge(kernel="poly", degree=1)` from scratch on every iteration — and profiling showed a quarter of total fit time going into sklearn's per-call validation machinery, not maths. The maths itself is ten lines of numpy, now in `_kernel_ridge_preds` (_tde.py:38):

```python
mean = x_hist.mean(axis=0)
std = x_hist.std(axis=0)
std[std == 0.0] = 1.0
xs = (x_hist - mean) / std              # StandardScaler
cs = (candidates - mean) / std

gamma = 1.0 / xs.shape[1]
k = xs @ xs.T * gamma + 1.0             # degree-1 polynomial kernel
k.flat[:: k.shape[0] + 1] += 1.0        # alpha = 1 regularisation
dual = np.linalg.solve(k, y_hist)
return (cs @ xs.T * gamma + 1.0) @ dual
```

There's a unit test asserting this matches sklearn's output to numerical precision.

Second, a small one: `candidate_parameters` is a float-array mirror of the candidate list, kept in sync as candidates are popped (`np.delete`, _tde.py:346) — previously the whole list was re-converted to an array on every guided iteration.

## Part 4: the LOOCV accuracy estimate — _individual_train_acc (_tde.py:557) and loocv_train_acc (_tde_sfa.py).

Every candidate member gets a quality score: leave-one-out 1NN accuracy on its own training subsample — for each case i, find its nearest neighbour among the other cases and check the class matches. That's inherently O(n²) 
  histogram intersections, and it runs once per candidate (250 times per fit), which made it the single largest cost in the whole ensemble — about a third of fit time even after the bag rewrite.                               
                                                                                                                                                                                                                                 
  The old shape of the loop was one Python call per held-out case, each scanning all other bags. Two things were wrong with that: per-case call overhead, and — the real prize — every pair was computed twice, because sim(i, j)
  was evaluated while predicting case i and again as sim(j, i) while predicting case j. The histogram intersection is symmetric, so half of all that work was pure waste.      

TThe whole LOOCV now happens in one kernel call:
  
 _tde.py:562
```python
  if train_size <= _SYMMETRIC_LOOCV_MAX_N:
      _, y_codes = np.unique(y, return_inverse=True)
      n_done, correct, preds = loocv_train_acc(
          *tde._transformed_data, y_codes.astype(np.int64), required_correct
      )
```          
  Inside loocv_train_acc, an n×n similarity matrix is filled upper-triangle only, each value mirrored:
  
```python    
  for i in range(n):
      ... 
      for j in range(i + 1, n):            # j > i only
          sim = _histogram_intersection(...)
          sims[i, j] = sim
          sims[j, i] = sim                
```
When row i is processed, similarities to j < i are already sitting in the matrix from earlier rows — so each pair is merged exactly once. The matrix is int32 and a typical subsample is ~100–500 cases, so it's kilobytes, not
  a memory problem (and there's a guard: above _SYMMETRIC_LOOCV_MAX_N = 4096 cases it falls back to the per-case search rather than allocate gigabytes).                                                                         
                                                                                                                                                                                                                                 
  The subtle part was preserving early abandoning. The ensemble passes required_correct — once it's full, a candidate only matters if it beats the current worst member, so the original code bailed out the moment that became  
  arithmetically impossible. The check now sits before each row, exactly where the old per-case loop had it, so an abandoned candidate does the same amount of work (roughly half, thanks to the triangle) and returns the same -1 accuracy that marks it for rejection. This mattered for exactness: abandonment changes what accuracy gets recorded, which feeds the kernel-ridge regressor, which changes future parameter choices — so we couldn't just "compute everything and check afterwards".                                                                                                                                                                                                                   
                                                                                                                                                                                                                                 
  Two more details worth knowing:                                                                                                                                                                                                
                                                                                                                                                                                                                                 
  - preds comes back too: when `fit_predict_proba` needs train estimates `(keep_train_preds=True)`, the per-case nearest-neighbour predictions recorded during LOOCV are reused directly — no second pass.                           
  - Ties: unlike predict, the LOOCV rule has always been deterministic first-maximum (no coin flips), so the kernel needs no randomness at all.                                                                                  
                                                                                                                                                                                                                                 
  The same trick serves multivariate channel selection: `_select_dims` scores each channel by exactly this LOOCV (with required_correct=0, so no abandoning), one kernel call per channel

## Parft 5: multivariate handling — `_select_dims` (_tde.py:969) and `_combine_dim_bags` (_tde.py:947).

TDE's multivariate strategy is per-member channel selection: each ensemble member independently decides which channels are worth using, then merges the survivors' words into one bag per case. Three steps.

Step 1 — score every channel cheaply. For each channel, a per-channel transform is fitted, but its usefulness is estimated on reduced bags: bags built from the handful of non-overlapping binning windows rather than every sliding window — a fraction of the words, so a fast proxy. The score is exactly the Part-4 LOOCV, one kernel call per channel:

_tde.py:1002  (inside _select_dims)
```python
transformers[i].fit(X_dim, y)
sfa = transformers[i].binning_bags()      # reduced bags, straight as arrays
...
if self.n_cases_ <= _SYMMETRIC_LOOCV_MAX_N:
    # whole LOOCV in one numba call, each symmetric pair
    # intersection computed once
    _, correct, _ = loocv_train_acc(*sfa, y_codes, 0)
```
(`required_correct=0` — channel scores are always computed in full, never abandoned.)

Step 2 — keep the good ones. Channels within dim_threshold (default 85%) of the best channel's score survive; if more than max_dims (default 20) survive, a random subset is kept:                                             
                                                                                                                                                                                                                                 
  _tde.py:1013                                                                                                                                                                                                                 
 ```python
if accs[i] >= max_acc * self.dim_threshold:                                                                                                                                                                                    
      dims.append(i)                                                                                                                                                                                                             
```                                                                                                                                                                                                                                 

So different members typically use different channel subsets — channel selection is itself ensembled.                                                                                                                          
                                                                                                                                                                                                                                 
  Step 3 — build and merge the real bags. Each kept channel's full sliding-window bags are built (reusing the MFT cached during fit — `transform(self._transformers[i]._fit_X)` hits the identity check, so nothing is recomputed),
  then merged into one bag per case:          

_tde.py:817  (inside `IndividualTDE._fit`)      
```python                                                                                                                                                                               
  dim_words = [                                                                                                                                                                                                                  
      self._transformers[i].transform(self._transformers[i]._fit_X)                                                                                                                                                              
      for i in range(len(self._dims))                                                                                                                                                                                            
  ]                                                                                                                                                                                                                              
  self._transformed_data = self._combine_dim_bags(dim_words, self._dims, self.n_cases_)                                                                                                                                          
 ```
                                                                                                                                                                                                                  
The merge is where the array representation pays off again. Each channel's bags are already sorted, and the channel rewrite of keys2 — (`quadrant << highest_dim_bit) | dim`, or just `dim for levels=1` — is order-preserving     
  within a channel. And since the channel is baked into `keys2`, keys from different channels can never be equal. Two consequences: no re-sorting, and no count aggregation — the combined bag is a pure k-way merge of sorted     
  streams, done in one numba kernel:                                                                                                                                                                                             
                                                                                                                                                                                                                                 
  _tde.py:947                                                                                                                                                                                                                  
 ```python
 def _combine_dim_bags(self, dim_bags, dims, n_cases):                                                                                                                                                                          
      # per-dimension bags are already sorted, so a numba k-way merge                                                                                                                                                            
      # builds the combined sorted bags without any re-sorting                                                                                                                                                                   
      all_k1 = np.concatenate([b[0] for b in dim_bags])                                                                                                                                                                          
      ...                                                                                                                                                                                                                        
      return combine_dim_bags(all_k1, all_k2, all_v, ...)                                                                                                                                                                        
```
                                                                                                                                                                                                                               
  (The previous version concatenated then re-sorted every case with two stable argsorts — the merge replaced that.)                                                                                                              
                                                                                                                                                                                                                                 
  After the merge, `_clear_transformer_fit_cache()` (_tde.py:920) drops the cached MFT arrays so fitted models don't carry megabytes of intermediate Fourier data — only breakpoints and bags persist. Predict then follows the    
  identical path — per-channel transform, same k-way combine, and the combined bags flow into the standard `nn_similarities_all` machinery from Part 2 with no multivariate special-casing at all.

One historical footnote a reviewer might wonder about: the old code rebuilt each channel's transform from scratch in both `_select_dims` and `_fit`; now the transformers fitted during selection are reused directly, with the    
  reduced-bag DFT discarded once scoring is done (`transformers[i]._binning_dft = None`).                                                                                                                                          
                                                                                                                                                                                                                                 
## Overview 

The tour so far: sorted-array bags → merge intersection → batched 1NN predict with exact tie-breaks → two-phase ensemble selection → symmetric LOOCV with early abandoning → per-member channel selection and k-way merging.   
  The last part of the story is the evidence that this rewrite is the same classifier, and what it bought.                                                                                                                       
                                                                                                                                                                                                                                 
  Three layers of verification, coarsest to finest:                                                                                                                                                                              
                                                                                                                                                                                                                                 
  1. Exact unit-level parity. test_tde_sfa.py asserts the new transform produces bags exactly equal to the generic SFA across representative configurations (flat/pyramid, bigrams, norm, both binning methods) — plus           
  hand-checkable unit tests for every kernel we've met: loocv_train_acc (including the abandon sentinel), nn_tie_break's draw consumption, combine_dim_bags' key shifting.                                                       
  2. Behavioural invariants. test_tde.py pins the properties that must survive any future change: n_jobs never changes predictions; pickled models predict identically after round-trip; _kernel_ridge_preds matches sklearn to  
  1e-10; the >4096-case LOOCV fallback agrees with the symmetric kernel; the deprecated parameters (alphabet_size, typed_dict — sentinel defaults, FutureWarning, removal v1.7.0) warn exactly when used. And notably,           
  IndividualTDE's stored expected-results check passes unchanged — the new implementation reproduces the archived probabilities on unit_test.                                                                                    
  3. Statistical equivalence. Bit-identity was explicitly not the goal (float-level differences exist — most fundamentally in members where word_length ≥ window_size, whose trailing letters are roundoff noise in both         
  versions). The claim is the one you validated on EEG and UCR: accuracies not significantly different. 

  What it bought, against origin/main on like-for-like benchmarks:                                                                                                                                                               
  
| Change | Fit | Predict |
|---|---:|---:|
| Algorithmic rewrite | 10–23×, growing with dataset size | 4–11× |
| + tie-break pooling | — | Further ~2.4× single-threaded |
| + threading (`n_jobs=8`) | Fit stays single-threaded by design | ~3× more |                                                                                                               
                                                                                                                                                                                                                                 
  The remaining runtime floor is the merge intersections themselves — the actual O(n²) similarity work — which is about as close to "the algorithm, and nothing else" as it gets. Fit remains sequential because the guided      
  parameter selection is inherently so: each choice depends on every accuracy before it.                                                                                                                                         
                                                                                                                                                                                                                                 
  And the quiet win that motivated the whole structure: _tde.py went from ~1,420 lines (with pickling workarounds, typed-Dict conversions and threading that couldn't scale) to ~1,050 where every line is reachable by a newly  
  fitted model.

TonyBagnall:
this is only on one resample, hence the variation, I'll set it off on 30, but the speed up is x13!

image
image

aeon/classification/dictionary_based/tests/test_tde.py
```diff
@@ -105,10 +133,165 @@ def test_histogram_intersection():
...
+        )
+
+    st = ensemble(1).fit(X, y)
+    mt = ensemble(2).fit(X, y)
+    np.testing.assert_array_equal(st.predict_proba(X), mt.predict_proba(X))
+
+    Xm, ym = make_example_3d_numpy(n_cases=16, n_channels=3, n_timepoints=40)
+    it_st = IndividualTDE(window_size=12, random_state=0, n_jobs=1).fit(Xm, ym)
+    it_mt = IndividualTDE(window_size=12, random_state=0, n_jobs=3).fit(Xm, ym)
+    np.testing.assert_array_equal(it_st.predict(Xm), it_mt.predict(Xm))
```
MatthewMiddlehurst: This should go. We dont really want to thread in normal testing, this should be covered in the generic multithread tests

aeon/classification/dictionary_based/tests/test_tde.py
```diff
@@ -105,10 +133,165 @@ def test_histogram_intersection():
...
+    assert ensemble.max_channels == 1
+    with pytest.warns(FutureWarning, match="dim_threshold"):
+        ensemble = TemporalDictionaryEnsemble(dim_threshold=0.0)
+    assert ensemble.channel_threshold == 0.0
+    ensemble = TemporalDictionaryEnsemble(max_channels=1)
+    assert ensemble.max_channels == 1
+    with pytest.warns(FutureWarning, match="typed_dict"):
+        IndividualTDE(typed_dict=True)
+    with pytest.warns(FutureWarning, match="typed_dict"):
+        TemporalDictionaryEnsemble(typed_dict=False)
```
MatthewMiddlehurst: Not needed

aeon/classification/dictionary_based/_tde_sfa.py
MatthewMiddlehurst, line 1: I would put this in the TDE file itself rather than an individual file as its not a classifier.

Not going to review this in full TBH, but having another SFA implementation is a bit grim.

Fine as a temporary measure to improve one algorithm which has been demonstrated.

aeon/classification/dictionary_based/_tde.py
```diff
@@ -33,10 +63,12 @@ class TemporalDictionaryEnsemble(BaseClassifier):
...
-    Overview: Input 'n' series length 'm' with 'd' channels
-    TDE searches 'k' parameter values selected using a Gaussian processes
-    regressor, evaluating each with a LOOCV. It then retains 's'
-    ensemble members.
+    Overview: Input 'n' series of length 'm' with 'd' channels.
+    TDE searches 'k' parameter values, using kernel ridge regression over
+    previously evaluated parameter combinations to predict the accuracy of
+    candidate parameter sets, and evaluates each selected set with a LOOCV.
+    (The reference paper [1] describes this step as a Gaussian process
+    regressor.) It then retains 's' ensemble members.
```
MatthewMiddlehurst: If we want this detail should probably be a note rather than in the top explanation. Its a guided search/bayesian optimization/whatever.

aeon/classification/dictionary_based/_tde.py
```diff
@@ -68,33 +100,36 @@ class TemporalDictionaryEnsemble(BaseClassifier):
     min_window : int, default=10
         Minimum window length.
     randomly_selected_params : int, default=50
-        Number of parameters randomly selected before the Gaussian process parameter
-        selection is used.
+        Number of parameters randomly selected before the kernel ridge regression
+        guided parameter selection is used.
```
MatthewMiddlehurst: same as above

aeon/classification/dictionary_based/_tde.py
```diff
@@ -170,14 +206,16 @@ def __init__(
...
+        max_dims="deprecated",
         time_limit_in_minutes=0.0,
         contract_max_n_parameter_samples=np.inf,
-        typed_dict=True,
+        typed_dict="deprecated",
         train_estimate_method="loocv",
         n_jobs=1,
         random_state=None,
+        max_channels=20,
+        channel_threshold=0.85,
```
MatthewMiddlehurst: these are out of order

aeon/classification/dictionary_based/_tde.py
```diff
@@ -331,17 +397,20 @@ def _fit(self, X, y, keep_train_preds=False):
                 if len(np.unique(y_subsample)) > 1:
                     break
 
+            # members are kept single threaded: the ensemble parallelises
+            # over members in predict, so member-level threads would only
+            # oversubscribe
```
MatthewMiddlehurst, line 391: revisitng threading is a whole different thing but there is probably a better solution to this that removing fit threading altogether.

MatthewMiddlehurst (commented):
Wouldn't merge the notebook

MatthewMiddlehurst:
Also going to add two points from my AI summary since apparently this is taking over the world now. If you can refute fine by me, I didnt come up with it but it seems like a reasonable point:

1. Important: IGB does not exactly reproduce sklearn splitting
[_best_split](https://github.com/aeon-toolkit/aeon/blob/5e0ba9e0cb2189991076335dcaf5962b8ea579c6/aeon/classification/dictionary_based/_tde_sfa.py#L576-L622) considers any strictly different adjacent values splittable.
The previous DecisionTreeClassifier implementation treats feature values within approximately 1e-7 as equal. For example, sklearn does not split 0 and 5e-8, while this implementation will after the float32 conversion because they remain distinct.
That can change IGB breakpoints and therefore bags and predictions. The midpoint-guard unit test does not test this sklearn tolerance.
Suggested fix: reproduce sklearn’s feature-equality threshold and add differential tests against the old SFA using closely spaced float32 values and equal-gain trees.
2. Important: numerically unstable variance in direct DFT
[_binning_dft_all](https://github.com/aeon-toolkit/aeon/blob/5e0ba9e0cb2189991076335dcaf5962b8ea579c6/aeon/classification/dictionary_based/_tde_sfa.py#L373-L410) calculates variance as:
square_sum / window_size - mean * mean
This suffers catastrophic cancellation for large-offset, low-variance series. For 1e9 + np.arange(8), that expression evaluates to zero in float64, although the actual variance is 5.25. Other inputs can produce a negative result and therefore a NaN standard deviation.
The old direct binning path used np.std, so this is a new equivalence problem affecting IGB and multivariate channel-selection bags.
Suggested fix: use a stable two-pass or Welford calculation. Add large-offset, low-variance parity tests.

aeon/classification/dictionary_based/_tde_sfa.py
TonyBagnall: ok, will move it, but I dont think its grim at all, its much safer like this. SFA itself is used in many places and its beyond the scope of this to rewrite the main SFA class

MatthewMiddlehurst: Yes not against it as a temporary private replacement to the slow SFA, but including Alis one this would be the 4th. Moving it is fine for now.

aeon/classification/dictionary_based/_tde.py
TonyBagnall: what do you suggest? threading individual members protects against over threading, since they themselves are threaded. Nested threading is not great, and given as you say the whole thing needs revisiting, this seems the best move for now,

TonyBagnall:
> Also going to add two points from my AI summary since apparently this is taking over the world now. If you can refute fine by me, I didnt come up with it but it seems like a reasonable point:
> 
> 1. Important: IGB does not exactly reproduce sklearn splitting
>    [_best_split](https://github.com/aeon-toolkit/aeon/blob/5e0ba9e0cb2189991076335dcaf5962b8ea579c6/aeon/classification/dictionary_based/_tde_sfa.py#L576-L622) considers any strictly different adjacent values splittable.
>    The previous DecisionTreeClassifier implementation treats feature values within approximately 1e-7 as equal. For example, sklearn does not split 0 and 5e-8, while this implementation will after the float32 conversion because they remain distinct.
>    That can change IGB breakpoints and therefore bags and predictions. The midpoint-guard unit test does not test this sklearn tolerance.
>    Suggested fix: reproduce sklearn’s feature-equality threshold and add differential tests against the old SFA using closely spaced float32 values and equal-gain trees.
> 2. Important: numerically unstable variance in direct DFT
>    [_binning_dft_all](https://github.com/aeon-toolkit/aeon/blob/5e0ba9e0cb2189991076335dcaf5962b8ea579c6/aeon/classification/dictionary_based/_tde_sfa.py#L373-L410) calculates variance as:
>    square_sum / window_size - mean * mean
>    This suffers catastrophic cancellation for large-offset, low-variance series. For 1e9 + np.arange(8), that expression evaluates to zero in float64, although the actual variance is 5.25. Other inputs can produce a negative result and therefore a NaN standard deviation.
>    The old direct binning path used np.std, so this is a new equivalence problem affecting IGB and multivariate channel-selection bags.
>    Suggested fix: use a stable two-pass or Welford calculation. Add large-offset, low-variance parity tests.

are these blockers for you? They seem pretty minor to me, but I will look into it. On point 2. low variance series is a problem throughout, currently poorly handled for DrCIF by raising on the release version and excessive warnings on main, not sure if my proposed fix got in, its been a while

TonyBagnall:
added a zero variance check and moved SFA from its own private file to _tde. Want to stick to unthreaded single TDE and dont think the sklearn point really matters.

aeon/classification/dictionary_based/_tde.py
MatthewMiddlehurst: yeah not a blocker

MatthewMiddlehurst:
Missed push maybe? some bits seem missing. Not blockers, but if they are easy to correct may as well.

TonyBagnall:
> Missed push maybe? some bits seem missing. Not blockers, but if they are easy to correct may as well.

yes forgot to push, doh

TonyBagnall:
> Wouldn't merge the notebook

you dont like my notebook? Its the level of detail I wanted to see and wrote it to explain what we did. Was thinking it would help someone wanting to improve main SFA

MatthewMiddlehurst:
It is useful for explaining the changes and improvements in the PR, for developers and pointing out some differences, but does not really work as an example for users which is what goes in that folder IMO. Have not taken a look in a while but the vast majority are usage or framework examples rather than giving implementation details I believe. A lot of the comparisons from the previous code to new code get less relevant as time goes on really.

Blog with a link in docstring or changelog may be a more relevant format but I don't really do that stuff so could be wrong there 🤷.

TonyBagnall:
> It is useful for explaining the changes and improvements in the PR, for developers and pointing out some differences, but does not really work as an example for users which is what goes in that folder IMO. Have not taken a look in a while but the vast majority are usage or framework examples rather than giving implementation details I believe. A lot of the comparisons from the previous code to new code get less relevant as time goes on really.
> 
> Blog with a link in docstring or changelog may be a more relevant format but I don't really do that stuff so could be wrong there 🤷.

not sure what examples for users we provide beyond fit/predict, which are generic for all. I think notebooks should explain how classifiers work, maybe detail is extreme. I will remove it if blocking, but personally I like it.

MatthewMiddlehurst:
Maybe something to discuss. I think we removed all the basic fit/predict ones which I agree are pretty shit. My main issue is mainly the implementation/datatype stuff explaining improvements from the old method rather than the explanations on ensembling, multivariate handling etc.

"TDE is the most accurate classifier of the dictionary based family on the UCR archive" deleted this comment previously but also i though that was weasel 2 now

TonyBagnall:
> Maybe something to discuss. I think we removed all the basic fit/predict ones which I agree are pretty shit. My main issue is mainly the implementation/datatype stuff explaining improvements from the old method rather than the explanations on ensembling, multivariate handling etc.
> 
> "TDE is the most accurate classifier of the dictionary based family on the UCR archive" deleted this comment previously but also i though that was weasel 2 now

I'll take it out for now then and put it on a draft branch so I dont lose it

MatthewMiddlehurst (approved):
Thanks

PR #3627, opened 2026-07-09
[ENH] Add Swale (Sequence Weighted Alignment) distance (#425)

### LINKED ISSUE #425: [ENH] Implement Sequence Weighted Alignment model (Swale) distance function

Swale is a distance function used in some comparisons, for example here
https://www.cs.ucr.edu/~eamonn/vldb_08_Experimental_comparison_time_series.pdf
https://dl.acm.org/doi/10.1145/1247480.1247544
(there is another more recent comparison I cant find the link for right now, will post it anon)

We have not used Swale in any of our distance experiments, because it did not perform very well in the above study. However, that was 15 years ago and it would be timely to re-evaluate it through an sktime implementation.

A numba based function modelled along the lines of existing distances. I can probably track down a matlab implementation if someone wants to take this on.
there is a link to a java implementation here
http://users.ece.northwestern.edu/~hdi117/tsim.htm

_No response_

### Additional context

_No response_

CedricConday:
Fixes #425.

Adds the Swale (Sequence Weighted ALignmEnt) distance from Morse & Patel, "An efficient and accurate method for evaluating time series similarity" (SIGMOD 2007), as a numba elastic distance modelled on the existing ones (`_lcss.py`, `_erp.py`). It provides the full API — `swale_distance`, `swale_cost_matrix`, `swale_pairwise_distance`, `swale_alignment_path` — and is registered in `DISTANCES`, so it is covered by the parametrised distance tests.

Swale is an ε-threshold measure like LCSS and EDR, but it scores both a reward for matches and a penalty for gaps, using the recurrence from Definition 5.1 of the paper:

```
Swale(R,S) = n·g            if m = 0
             m·g            if n = 0
             r  + Swale(Rest(R), Rest(S))                        if |r_1 - s_1| ≤ ε
             max{ g + Swale(Rest(R), S), g + Swale(R, Rest(S)) } otherwise
```

with match reward `r` and gap cost `g`.

**Similarity → distance (the main thing to review).** The raw recurrence is a *similarity* (higher = more similar), which does not fit the distances framework (non-negativity, self-distance 0). I convert it the same way LCSS does: the maximum attainable score is `match_reward * min(m, n)`, so the distance is `1 - score / (match_reward * min(m, n))` — 0 for identical series, non-negative in general (the score can never exceed that maximum), and requiring `match_reward > 0`. Happy to change the convention if you'd prefer a different normalisation, or to expose the raw similarity instead.

**Defaults.** `epsilon=1.0` (as LCSS), and `match_reward=50`, `gap_penalty=-8` following the reward/penalty ratio the paper trains on its ASL data — also a reasonable thing to reconsider.

**Correctness.** I validated the DP against an independent naive recursive implementation of Definition 5.1: exact match (0 error) across 300 random unequal-length cases. The committed expected values were then generated from that verified implementation. The full `aeon/distances/tests/` suite passes (513 passed), including the non-negativity, symmetry, unequal-length and single-point checks, and `pre-commit` is clean.


CedricConday:
Checking back in — still current, Swale is not in `aeon/distances/elastic` on `main`.

The part most worth a reviewer's time is the similarity-to-distance conversion. The recurrence in the paper is a similarity, so I normalise it the way LCSS does, `1 - score / (match_reward * min(m, n))`. That gives 0 for identical series and stays non-negative, but it is a convention choice and I would rather match your preference than guess. Same for the `match_reward=50` / `gap_penalty=-8` defaults, which come from the paper's ASL experiments.

The DP is validated against an independent naive implementation of Definition 5.1, exact across 300 random unequal-length cases, and `aeon/distances/tests/` passes at 513.

CedricConday:
Closing this for now — it has been open a while without review bandwidth and I would rather not leave it lingering. The Swale distance implementation is complete on my side; happy to reopen and bring it up to date if there is interest.

PR #3628, opened 2026-07-09
[ENH] Speed up interval-based forest regressors

alexbanwell1:

Speeds up the interval-based forest regressors (`DrCIFRegressor`, `CanonicalIntervalForestRegressor`, `TimeSeriesForestRegressor`, `RandomIntervalSpectralEnsembleRegressor`, `IntervalForestRegressor`), which currently become infeasibly slow on larger datasets (e.g. windowed forecasting problems with 1000-10000 training cases).

**1. Change the default base estimator for regression from `DecisionTreeRegressor(criterion="absolute_error")` to `DecisionTreeRegressor(criterion="squared_error")` in `BaseIntervalForest`.**

sklearn's `absolute_error` criterion scales roughly quadratically with the number of cases (a long-standing sklearn limitation, see scikit-learn/scikit-learn#9626), while `squared_error` is `O(n log n)`. Fitting a single tree on DrCIF-transformed features (~210 features):

| n_cases | absolute_error | squared_error |
|---|---|---|
| 500 | 0.39s | 0.07s |
| 1000 | 1.23s | 0.17s |
| 2500 | 6.18s | 0.47s |
| 5000 | 22.5s | 1.03s |
| 10000 | 84.1s | 2.26s |

With the default 200 estimators, tree fitting alone at 10000 cases drops from ~4.7 hours to ~8 minutes per fit. Test regression quality is essentially unchanged in a quick check (DrCIF, 50 trees): CardanoSentiment test MAE 0.224 (absolute_error) vs 0.226 (squared_error); on a synthetic windowed forecasting problem squared_error was slightly more accurate (MAE 1.62 vs 1.82).

This is a behavioural change for anyone relying on the old default: results produced with defaults will differ slightly from previous versions (e.g. the TSER bake-off reference results). The old behaviour is available by passing `base_estimator=DecisionTreeRegressor(criterion="absolute_error")`, and this is noted in the docstrings. Classification defaults are untouched.

**2. Transform whole collections in a single numba call in `Catch22`.**

`Catch22` is called once per interval per tree inside DrCIF/CIF (classification and regression) with `n_jobs=1`. Previously each case was transformed by a Python-level `_transform_case` call (via a joblib `delayed` wrapper), paying Python branching and numba dispatch overhead tens of millions of times in a large forest fit. For single-threaded equal-length 3D input, the per-case loop now runs inside a single numba function per transform call (`_transform_collection_3d`), which mirrors `_transform_case` exactly. The FFT used by the two `SP_Summaries` features is precomputed for all cases with a vectorised `np.fft.fft` call (verified bit-identical to the per-case calls), as `np.fft` is unsupported in numba. The per-case path is kept for unequal length input, `n_jobs > 1` and `pycatch22`.

One caveat: numba compiles the feature functions a second time when they are called from the batch function, and with `fastmath=True` this can differ in the last float digits from the Python-called versions (max observed relative difference `3.5e-13` over a large grid of inputs, feature subsets, `catch24` and `outlier_norm` settings, NaN patterns identical). Care was taken to avoid this where it matters: the FFT input mean is computed exactly as the per-case path does, since the FFT DC term is a catastrophic cancellation that would otherwise meaningfully change the `SP_Summaries` features for short series. In testing, `DrCIFRegressor` predictions and the stored expected results outputs for `Catch22Classifier`, `Catch22Regressor`, `CanonicalIntervalForest` (both), `DrCIFClassifier` and `RISTClassifier` were byte-identical with and without the batch path. The multithreading and datatype consistency estimator checks (which use `np.allclose`) all pass.

**End-to-end `DrCIFRegressor.fit` (default params, univariate, series length 100, all runs in a single session):**

| n_cases | n_estimators | main | this PR | speedup |
|---|---|---|---|---|
| 1000 | 20 | 102.9s | 42.6s | 2.4x |
| 2500 | 20 | 308.0s | 108.7s | 2.8x |
| 10000 | 4 | 436.6s | 71.3s | 6.1x |

The advantage grows with n_cases as the quadratic tree fitting cost takes over on main. At 10000 cases with the default 200 estimators this works out to roughly 6 hours on main vs 1 hour with this PR per fit.

Expected test results for the four affected regressors with stored results were regenerated using the `_write_estimator_results.py` procedure. Doctest outputs are unchanged (the examples predict on their training data, where fully-grown trees reproduce the targets exactly regardless of criterion). All `check_estimator` checks pass for the five affected regressors plus `Catch22`, `Catch22Classifier` and `DrCIFClassifier`, and the interval-forest/regression/catch22 test suites pass.

No.

#### Any other comments?

The criterion change alters default behaviour, so flagging for maintainer opinion: `absolute_error` has been the regression default since the base class was introduced, but it makes the regressors unusable beyond a few thousand cases (a week-long fit on a windowed forecasting problem motivated this PR). sklearn's own `RandomForestRegressor` defaults to `squared_error`.

The two changes are independent — happy to split the PR if preferred.

##### For all contributions


TonyBagnall:
hi alex, this will conflict with this #3620 and #3619 , maybe wait until after that is in?

TonyBagnall:
those speed ups for decision trees regression are insane. I suggest rebase off #3620, make the change to squared and I will run the timing and accuracy comparison as main -> 3620 (x2.5 faster) -> squared error regression/ gini for classification

alexbanwell1:
## Accuracy + speed evaluation of the `absolute_error` → `squared_error` default

Following the same idea as the #3620 evaluation (accuracy + speed), adapted to the fact that this change is deliberately **not** bit-identical: rather than showing identical results, the goal here is to show the default change does **not significantly affect predictive quality**.

**Setup:** 5 TSER datasets (`NaturalGasPricesSentiment`, `CardanoSentiment`, `Covid3Month`, `AppliancesEnergy` [24 channels], `FloodModeling1`), the 3 affected Catch22/summary regressors (`DrCIFRegressor`, `TimeSeriesForestRegressor`, `CanonicalIntervalForestRegressor`), 3 resamples each, default `n_estimators=200`. 45 paired runs. Old default `DecisionTreeRegressor(criterion="absolute_error")` vs new `squared_error`. This is a local confirmatory subset, not the full archive — a cluster run over all of TSER (à la #3620) would obviously be more complete.

### 1. Accuracy — no significant difference

| Metric | Result |
|---|---|
| Paired runs: `squared` better / `absolute` better / tie | **25 / 20 / 0** |
| Mean test-RMSE ratio (squared / absolute) | **1.001** (median 0.997) |
| Wilcoxon signed-rank | **p = 0.40** |
| Paired t-test | **p = 0.078** |

Both tests are comfortably above 0.05 → no statistically significant RMSE difference. If anything `squared_error` is marginally *lower* on average (DrCIF −1.5%, TSF −1.9%, CIF −0.4%).

Per-dataset mean test RMSE (mean over resamples & regressors):

| dataset | absolute_error | squared_error | Δ% |
|---|---|---|---|
| AppliancesEnergy | 2.3912 | 2.3543 | −1.54% |
| CardanoSentiment | 0.28838 | 0.28867 | +0.10% |
| Covid3Month | 0.03666 | 0.03735 | +1.90% |
| FloodModeling1 | 0.00972 | 0.00967 | −0.52% |
| NaturalGasPricesSentiment | 0.05750 | 0.05774 | +0.42% |

All deltas <2%, mixed sign. On a scatter of per-run test RMSE (squared vs absolute) every point lies on the `y = x` line.

### 2. Speed — the reason for the change

Single decision-tree fit on a realistic ~210-feature matrix (isolating the criterion cost, everything else held constant):

| n_cases | squared_error | absolute_error | ratio |
|---|---|---|---|
| 250 | 0.06 s | 0.22 s | 3.5× |
| 500 | 0.09 s | 0.69 s | 7.8× |
| 1000 | 0.31 s | 2.42 s | 7.8× |
| 2000 | 0.61 s | 7.21 s | 11.8× |
| 4000 | 1.43 s | 23.9 s | **16.8×** |

The gap widens with `n`, confirming `absolute_error`'s ~O(n²) vs `squared_error`'s ~O(n log n) split cost (scikit-learn/scikit-learn#9626). At the default 200 trees this is what makes the difference between feasible and infeasible in the 1k–10k-case regime this PR targets.

End-to-end `fit()` on these (small) standard TSER datasets is roughly flat (~1.1× mean, within threading noise) — because `n` is small there and fit is dominated by *shared* Catch22 extraction, exactly as noted for DrCIF in #3620. So the criterion change neither helps nor hurts small datasets; the payoff appears at scale (the end-to-end table in the PR description: ~6 h → ~1 h at 10 000 cases).

### Summary

Predictive quality is statistically unchanged across 5 datasets × 3 regressors × 3 resamples (Wilcoxon p = 0.40), while the change removes a quadratic tree-building bottleneck that grows to >16× by 4 000 cases. @TonyBagnall — happy for this to feed into the `main → #3620 → squared-error` comparison you offered to run on the cluster; I can share the scatter plot and raw CSVs if useful.

cakedev0:
Hi, I have no context of this repo, but about:

> sklearn's absolute_error criterion scales roughly quadratically with the number of cases 

It's fixed in recent scikit-learn versions (1.9 and 1.10), it's now n log n. But it still has a higher constant than squared error criterion (3-6x slower).

TonyBagnall (approved):
forgot about this, lgtm

PR #3631, opened 2026-07-10
[ENH] Speed up rotation forest

### LINKED ISSUE #3630: [ENH] Speed up rotation forest classifier and regressor

both are built on scikit and could be tweaked to be faster without logical changes. After that, parameter changes could yield improvement also. All this is prior to introducing or at least experimentally testing algorithmic changes, good to optimise what we do first for easy wins

Note the regressor and classifier are currently independent and reproduce code, will move that into the base package to mirror pattern with interval based 

Optimisation plan, ranked by promise
  
  1. Replace sklearn PCA with a minimal in-house PCA (fit and transform). Groups are 3 columns wide by default; a mean + 3×3 covariance + np.linalg.eigh, storing just mean_ and
  components_ arrays, measures 8× faster per group than sklearn's PCA.fit + transform (0.052 ms vs 0.426 ms). This removes the PCA fit cost, all transform-time check_array/validate_data
  overhead in both fit and predict (thousands of calls), and lets the odd NaN-retry loop become a rare fallback (eigh on a covariance can't produce the SVD failure mode; keep the
  random-padding fallback for degenerate input). Bonus: models shrink a lot on disk — 200 trees × ~270 sklearn PCA objects is heavy for STC/HC2 pickles. Expected: ~1.3–1.5× fit, 3–5×
  predict.
  
  2. Skip redundant validation on the base tree: check_input=False and direct construction. When base_estimator is the default tree, pass check_input=False to
  tree.fit/predict_proba/predict (X_t is already a finite float32 array we just built), and construct the tree directly instead of going through _clone_estimator —
  clone/get_params/inspect.signature alone are ~4% of fit. Guard on the default so user-supplied estimators keep the safe path. Expected: another ~5–10% on fit, more on predict.
  
  3. Vectorise per-group training-data assembly (classifier _fit_estimator). The inner loop grows X_t with repeated np.concatenate per class (O(k²) copying) and fancy-indexes columns
  group by group. Precompute per-class row indices once in _fit_rotf, build the row selection first, then slice columns in one operation; similarly gather all groups' columns for the
  final transform with a single fancy index instead of ~270 small ones. _fit_estimator self-time was ~5% at 800 features and grows with width. Expected: ~5% on fit, more on wide data.
  
  4. Fix the O(n²) out-of-bag computation in fit_predict(_proba). oob = [n for n in indices if n not in subsample] scans a numpy array per element — quadratic in n_cases. At n=5,000 it's
  already ~4s across 200 estimators and grows quadratically (n=20,000 → ~a minute of pure waste). Replace with a boolean mask (mask[subsample] = True; oob = np.flatnonzero(~mask)). Only  
  affects fit_predict, but that's the path STC/HC2 use for train estimates, where n can be large. Also X_t[idx][subsample] double-indexes; minor.                                          
                                                                                                                                                                                           
  5. Parallelism hygiene. Items 1–3 remove most GIL-holding Python from the workers, so the existing prefer="threads" will actually scale (tree fit/predict release the GIL). On top: use  
  joblib batch_size/chunking to cut dispatch overhead for 200 small tasks, and the contract-time loop's "exactly n_jobs tasks per round" pattern could dispatch larger rounds. No      

_No response_

### Additional context

_No response_

TonyBagnall:
closes #3630 part of #3616 

part of the STC speed up, also for regression

### Stage 1: Restructure

There is a lot of duplication between classifier and regressor. Stage 1 is purely cosmetic: create a common code base in base class and switch behaviour on classifier or regressor, then have thin wrappers in the respective dir.

One design choice that reviewer may not like. Both algorithms have a pca_solver argument that was passed through to scikit learn pca. This has become a no-op deprecation now, see point 2 for reasons. It was actually ignored in the regressor.



### Stage 2: Replace sklearn PCA with a minimal in-house PCA

Rotation forest does many small group PCA transforms. These are computational, and using sckit gives a large validation burden. Because the groups are usually very small (three features), it is unlikely whether any of the solver options will be better. Instead, each attribute-group rotation is now a minimal _GroupPCA — mean + covariance + np.linalg.eigh with sklearn's sign convention — instead of a full sklearn PCA object.                                                                                                                                       
                                                                                                                                                                                           
It is bit-identical to scikit on "full" solver on simple tests, so comes for free. All docstring tests still pass. 

The eigh components match sklearn 1.8's "full" solver exactly (I verified the sign convention is v-based over 200 random trials), so predictions on default settings are unchanged down to the bit — docstring examples, STC/HC2/FreshPRINCE outputs, and the regressor's pinned test values all came back   identical. The only behavioural change is that pca_solver is now a deprecated no-op (documented as such); the classifier's pinned test used "randomized" so its expected values were
  regenerated, and the regressor's reverted to the original pre-refactor numbers.     

So each group rotation now stores only the two arrays actually needed at predict time. That means:

  - less data returned from each parallel fit task,
  - smaller fitted estimator object,
  - less memory pressure during training,
  - faster model serialization if users pickle/joblib-dump the fitted RotationForest,
  - less overhead if joblib ever uses process workers or if downstream code serializes the estimator.
See comment 1 for speed up for this change

### Stage 3: Internal data structures

scikit learn tree fitting is now 60-80% of runtime. One option would be to implement a stripped down version of C4.5 from Weka, which is what the original algorithm used. However, lets get the easy wins first


#### Preallocated-write transform + cheap finite check — biggest win, bit-identical.                                                                                                                  
  Replace the concatenate([...267 arrays...]) with one preallocated (n, n_comp) float32 array written per group, and replace nan_to_num's three full-array scans (isnan/isposinf/isneginf) with a single np.isfinite(X_t).all() guard that only falls back to nan_to_num in the (essentially never) non-finite case — _GroupPCA components are always finite, so the transform can't introduce NaN/inf. Measured 1.67× on the transform, diff 0.0. That's ~1.7× on predict, plus gains on the _fit_estimator tail and fit_predict. It also cuts the Python/GIL work, so it's the prerequisite for predict actually scaling with threads. Applies to both the _transform_for_estimator and the inline concatenate in _fit_estimator — factor them into one helper.  

there are some further tweaks, but not worth it at this point, given that tree building is now 90% of effort.

### Stage 4: Tree parameters

Largely leaving the trees alone, and thats where the time is, but we should make it possible to reconfigure to speed up, so I will expose parameters such as split criterion (gini seems much faster) etc

  1. criterion 
  2. splitter
  3. max_features
  4. max_depth and/or max_leaf_nodes.
  5. min_samples_leaf.
   

-


TonyBagnall:
Speed up from bespoke PCA about 2.4

image

TonyBagnall:
Stage 3 improvements, most impact still in predict. 

image

memory is about the same. We are still much slower than random forest, which, given how much time we spend building trees, is surprising. I think its tree aprameters

Stage 4 added and this is where the real speed up is: 
gini 1.5×, gini+random splitter 5×, gini+max_features='sqrt' 7.5×.

I have exposed these on this PR and I will test if they effect STC performance, but not changing them here. Could be a big speed up, particularly max_features

rerunning with this
```python
 RandomForestClassifier(
      n_estimators=n_estimators,
      criterion="entropy",
      max_features=None,
      bootstrap=False,   # arguable, closer to full-data tree fitting
      n_jobs=n_jobs,
      random_state=0,
  )
```

TonyBagnall:
always going to be slower, we do more work, but with a matching configuration, its much closer

image

aeon/base/_estimators/sklearn/_rotation_forest.py
```diff
@@ -0,0 +1,637 @@
...
+        """Run a sequence of ``delayed`` jobs, skipping joblib when single-threaded.
+
+        ``jobs`` is an iterable of ``delayed(func)(*args, **kwargs)`` tuples. When
+        ``n_jobs == 1`` these are called directly, avoiding joblib's per-task
+        dispatch overhead while preserving the order in which the jobs (and hence
+        any random draws in their arguments) are produced.
+        """
+        if self._n_jobs == 1:
+            return [func(*args, **kwargs) for func, args, kwargs in jobs]
+        return Parallel(n_jobs=self._n_jobs, prefer="threads")(jobs)
```
MatthewMiddlehurst: second time I have seen this format now, but with a different name. May be better as a general utility if it is more efficient.

aeon/classification/sklearn/tests/test_rotation_forest_classifier.py
```diff
@@ -9,42 +9,68 @@
...
 
     rotf = RotationForestClassifier(
         n_estimators=10,
-        pca_solver="randomized",
         random_state=0,
     )
     rotf.fit(X_train, y_train)
 
+    # expected values changed when the group PCA moved to an exact
+    # eigendecomposition
```
MatthewMiddlehurst: remove the comment

aeon/classification/sklearn/tests/test_rotation_forest_classifier.py
```diff
@@ -9,42 +9,68 @@
...
+        random_state=0,
+    )
+
+    rotf_full.fit(X_train, y_train)
+    rotf_randomized.fit(X_train, y_train)
+
+    np.testing.assert_array_equal(
+        rotf_full.predict_proba(X_test[:15]),
+        rotf_randomized.predict_proba(X_test[:15]),
+    )
```
MatthewMiddlehurst: does not seem like a useful test for both classes.

aeon/regression/sklearn/tests/test_rotation_forest_regressor.py
```diff
@@ -12,36 +12,50 @@
...
-    ]
-    np.testing.assert_array_almost_equal(expected, rotf.predict(X_test[:15]), decimal=4)
+    y_pred = rotf.predict(X_test)
+    assert y_pred.shape == y_test.shape
+
+    # exact per-case predictions are not pinned: scikit-learn 1.8 changed
+    # decision-tree handling of almost-constant features, which shifts them
+    # (see the class docstring). The aggregate error is stable across versions.
+    mse = mean_squared_error(y_test, y_pred)
+    np.testing.assert_almost_equal(mse, 0.0019, decimal=4)
```
MatthewMiddlehurst: Why have we moved to a score here but kept probabilities in the classifier?

aeon/regression/sklearn/tests/test_rotation_forest_regressor.py
```diff
@@ -12,36 +12,50 @@
...
+        random_state=0,
+    )
+
+    rotf_full.fit(X_train, y_train)
+    rotf_randomized.fit(X_train, y_train)
+
+    np.testing.assert_array_equal(
+        rotf_full.predict(X_test[:15]),
+        rotf_randomized.predict(X_test[:15]),
+    )
```
MatthewMiddlehurst: same as clf

aeon/classification/sklearn/_rotation_forest_classifier.py
```diff
@@ -107,24 +126,36 @@ def __init__(
         max_group: int = 3,
         remove_proportion: float = 0.5,
         base_estimator: BaseEstimator | None = None,
+        criterion: str = "entropy",
+        splitter: str = "best",
+        max_features=None,
+        max_depth: int | None = None,
+        max_leaf_nodes: int | None = None,
+        min_samples_leaf=1,
```
MatthewMiddlehurst: Do we need these? wouldn’t you just pass a different base_estimator?

aeon/regression/sklearn/_rotation_forest_regressor.py
```diff
@@ -112,24 +130,36 @@ def __init__(
         max_group: int = 3,
         remove_proportion: float = 0.5,
         base_estimator: BaseEstimator | None = None,
+        criterion: str = "squared_error",
+        splitter: str = "best",
+        max_features=None,
+        max_depth: int | None = None,
+        max_leaf_nodes: int | None = None,
+        min_samples_leaf=1,
```
MatthewMiddlehurst: same as clf

MatthewMiddlehurst (commented):
Few comments but fine with the restructure generally and the speed ups are nice.

Need to do a bit more to deprecate the parameter for both classes. If it is changed from its default value, put a warning. Put the next release in the docstring and warning if you can.

aeon/classification/sklearn/_rotation_forest_classifier.py
TonyBagnall: it models random forest, I think we should align with that ultimately, but I dont feel strongly

aeon/classification/sklearn/tests/test_rotation_forest_classifier.py
TonyBagnall: centralised and made better

aeon/regression/sklearn/tests/test_rotation_forest_regressor.py
TonyBagnall: short answer: scikit-learn. We could shift to random data if you prefer?

The regressor can't pin per-case predictions: we support scikit-learn 1.6–1.8, and 1.8 fixed decision-tree handling of almost-constant features, which covid_3month has — so individual predictions shift between supported versions even with the same random state (noted in the class docstring). The aggregate MSE is stable across versions, so that's what's checked. The classifier's probabilities on unit_test aren't affected, so they stay pinned exactly. I've added this explanation to the test docstring.

TonyBagnall:
> Few comments but fine with the restructure generally and the speed ups are nice.
> 
> Need to do a bit more to deprecate the parameter for both classes. If it is changed from its default value, put a warning. Put the next release in the docstring and warning if you can.

ok will do, bizarre they were different values in regression and classification in the first place

aeon/base/_estimators/sklearn/_rotation_forest.py
TonyBagnall: yes good idea, the drop through for n_jobs does help a bit, every little helps

TonyBagnall: actually three places, one of my goto first speed ups

_eval_intervals in RandomIntervals and _eval_estimators in BaseIntervalForest from #3620, plus _parallel here

TonyBagnall: makes this PR a bit bigger, but good to do

TonyBagnall: done, see what you think

aeon/classification/sklearn/_rotation_forest_classifier.py
TonyBagnall: I've reverted, I agree

aeon/regression/sklearn/tests/test_rotation_forest_regressor.py
MatthewMiddlehurst: Do we have a case in the CI where we use different versions? Will leave it for now as its still a general comparison to past results but not 100% sure its necessary.

MatthewMiddlehurst (approved):
Thanks for the changes.

PR #2844, opened 2025-05-26
[ENH]  Add Dynamic Alphabet Sizes for SFA

patrickzib:
This PR introduces the concept of dynamic alphabet sizes to SFA. 

The alphabet size is used as a budget and assigned over all coefficients to maximize tightness of lower bound. Alphabet sizes are assigned proportional to the variance using three 3 strategies:
- Linear-proportional  to variance
- Sqrt-proportional  to variance
- Log2-proportional to variance

# Illustration 

Example with Alphabet Sizes [4, 4, 2, 2] and variance-based feature selection:
image

# Example

E.g. Example for word length of 4 using 4 each, we have a budget of 16=4*4:
- Prior to this PR the alphabet has to be fixed for each coefficient:    [a-d, a-d, a-d, a-d] = [4, 4, 4, 4] = 16
- Now, the number of symbols gets assigned based on importance: [a-h, a-d, a-d, a-b] = [8, 4, 4, 2] = 16

# CD-Diagram for (average) alphabet-size 64
image

# Experiments

Using this kind of assignment is most beneficial for smaller alphabet sizes. TLB results (larger is better) show that for 2 to 8 alphabet sizes large improvements can be observed.

| Average Symbols      |   2 |   4 |  8 |   16 |   32 | 64 | 128 | 256 |
|:------------|-------------:|-------------:|-------------:|--------------:|--------------:|--------------:|---------------:|---------------:|
| SFA        |      37.515 |      56.694 |      69.425 |       77.726 |       82.2309 |       85.6476 |        86.8577 |        87.5971 |
| SFA+Linear |      48.474 |      63.373 |      72.769 |       79.669 |       83.8591 |       86.0971 |        87.1459 |        87.6656 |
| SFA+Log    |      46.017 |      60.966 |      72.265 |       79.352 |       83.8492 |       85.9773 |        87.075  |        87.628  |
| SFA+Sqrt   |      44.958 |      60.841 |      71.268 |       79.280 |       83.6312 |       86.0426 |        87.1275 |        87.6555 |
| iSAX        |      28.025 |      43.014 |      54.823 |       62.948 |       69.5433 |       75.366  |        78.3346 |        80.1139 |

PR #3634, opened 2026-07-11
[BUG] fix uncommented njit decorator in sfa_fast

patrickzib:
The previous PR #2844 accidentally removed an njit decorator in the method `_transform_words_case`. In parts because `PCA` could not be called in jnit mode. The purpose of this PR is to add the decorator back to `_transform_words_case` by externalizing the pca transform.


baraline (dismissed):
Nice catch, LGTM, just a small comment, if there is a valid reason feel free to ignore it :

Why cast the pca_means and components to float64 (line 625) ? Wouldn't that cause issues with the numba function if your input X is float32 ?

patrickzib:
> Nice catch, LGTM, just a small comment, if there is a valid reason feel free to ignore it :
> 
> Why cast the pca_means and components to float64 (line 625) ? Wouldn't that cause issues with the numba function if your input X is float32 ?

Thank you @baraline - and very good catch! 

The cast is not needed, and I have removed it. The internal variables in the PCA object are already stored as float64, so an explicit cast to float64 is unnecessary.

PCA is applied to the DFT features, not directly to the raw time series. Since the DFT features are also of type float64, I prefer to keep both representations in float64 for now.

TonyBagnall:
you might want to consider setting nogil=True on these two?

@njit(fastmath=True, cache=True, nogil=True)
def _apply_pca_transform(...):
    ...


@njit(fastmath=True, cache=True, parallel=True, nogil=True)
def _transform_words_case(...):
    ...

patrickzib:
Good point. I experimented with nogil=True in the past and didn’t observe much of a speedup, but I don’t think it hurts either.

nogil=True should be safe for _apply_pca_transform, and it should also be safe for _transform_words_case. The latter calls _mft, which uses objmode for the SciPy FFT path, so the entire transform won’t be GIL-free. 

That said, nogil=True is still a reasonable annotation for the compiled sections, thus I made the change.

baraline (approved):
Tested with float32 input and doesn't crash.
Having it fixed to float64 is not ideal as I think users should have control of the precision to manage memory better, but it's the case for most estimators so no issue for now.

patrickzib:
Thank you very much :)

PR #3637, opened 2026-07-12
[BUG] fix SFA.get_words() crash when window_size < n_timepoints

### LINKED ISSUE #3537: [BUG] SFA.get_words() and SFAFast.get_words() crash for any normal sliding-window configuration

BB0813:
## Fix

Fixes #3537

### Problem

Both `SFA.get_words()` and `SFAFast.get_words()` raise a numba `TypingError` whenever a series has more than one sliding window (i.e. `window_size < n_timepoints`). `np.squeeze` does not remove the `n_windows` axis, so `_get_chars` receives an array instead of a scalar integer.

### Solution

Reshape to 2D after squeeze and iterate over `(n_cases, n_windows)`:
```python
if words.ndim == 1:
    words = words.reshape(1, -1)
return np.array([
    [_get_chars(w, self.word_length, ...) for w in case]
    for case in words
])
```

### Changes

- `_sfa.py`: Fixed `get_words()` iteration
- `_sfa_fast.py`: Fixed `get_words()` iteration
- `tests/test_sfa.py`: Added `test_get_words_multiple_windows` regression test

### Testing

All 82 existing SFA tests pass, plus the new regression test.

### DIFF
--- a/aeon/transformations/collection/dictionary_based/_sfa.py
+++ b/aeon/transformations/collection/dictionary_based/_sfa.py
@@ -477,8 +477,13 @@ class SFA(BaseCollectionTransformer):
         Array of words
         """
         words = np.squeeze(self.words)
+        if words.ndim == 1:
+            words = words.reshape(1, -1)
         return np.array(
-            [_get_chars(word, self.word_length, self.alphabet_size) for word in words]
+            [
+                [_get_chars(w, self.word_length, self.alphabet_size) for w in case]
+                for case in words
+            ]
         )
 
     def _binning(self, X, y=None):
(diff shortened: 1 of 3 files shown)


patrickzib:
Thank you very much for addressing it. 

Unfortunately, this is a duplicate of #3609

Thus, I rather close it for now.

BB0813:
Got it, thanks for letting me know and for reviewing! I'll look into other issues to contribute to.

PR #3638, opened 2026-07-12
[BUG] fix outlier_norm using wrong feature indices in pycatch22 path

### LINKED ISSUE #3632: [BUG] Catch22 outlier_norm is silently ignored when use_pycatch22=True
### Describe the bug

n `Catch22._transform_case_pycatch22`, the `outlier_norm` handling is dead code due to a missing `elif`:                                                                                                                                              
                                                                                                                                                                                                                                                         
  https://github.com/aeon-toolkit/aeon/blob/main/aeon/transformations/collection/feature_based/_catch22.py  


The first `if` correctly computes the two `DN_OutlierInclude` features (indices 3 and 4 in the pycatch22 ordering) on the z-normalised series, but because the next statement is `if` rather than `elif`, execution falls through to the second chain, 
  whose `else` branch recomputes the feature on the raw series and overwrites the result.   
```python                                                                                                                                                                                                                                              
  if self.outlier_norm and feature in [3, 4]:                                                                                                                                                                                                            
      c22[dim + n] = features[feature](outlier_series)                                                                                                                                                                                                   
  if feature == 22:                                      # <-- should be elif                                                                                                                                                                            
      c22[dim + n] = np.mean(series)                                                                                                                                                                                                                     
  elif feature == 23:                                                                                                                                                                                                                                    
      c22[dim + n] = np.std(series)                                                                                                                                                                                                                      
  else:                                                                                                                                                                                                                                                  
      c22[dim + n] = features[feature](series)                                                                                                                                                                                                           
  ```     



Consequences when `use_pycatch22=True`:                                                                                                                                                                                                                
                                                                                                                                                                                                                                                         
  - `outlier_norm=True` has no effect on the output, so the pycatch22 path disagrees with the numba path (which handles this correctly by normalising the series before a single dispatch chain).                                                        
  - The z-normalisation performed earlier in the loop for these features is wasted work.                                                                                                                                                                 
                                                                                                                                                                                                                                                         
  This also affects `DrCIFClassifier`/`DrCIFRegressor` and `CanonicalIntervalForest*` with `use_pycatch22=True`, since they construct `Catch22(outlier_norm=True, ...)`.       


### Expected results

outlier_norm=True` should z-normalise the series before computing the two `DN_OutlierInclude` features, matching the numba implementation.

### Actual results

One-word change: `if feature == 22:` → `elif feature == 22:`.

### Suggested fix

_No response_

### Versions

_No response_

### PR DESCRIPTION
**Pull request by BB0813**
## Fix

Fixes #3632

### Problem

In `_transform_case_pycatch22`, the `outlier_norm` condition checked features `[3, 4]` (`CO_FirstMin_ac`, `CO_HistogramAMI_even_2_5`) instead of `[13, 14]` (`DN_OutlierInclude_p_001_mdrmd`, `DN_OutlierInclude_n_001_mdrmd`). This meant `DN_OutlierInclude` was always computed on the raw series when `use_pycatch22=True`, making `outlier_norm=True` ineffective for these features.

Additionally, `if feature == 22:` should be `elif feature == 22:` to prevent the `else` branch from overwriting the `outlier_norm` result.

The numba path (`_transform_case`) correctly uses features 13 and 14 for `outlier_norm`.

### Changes

- `_catch22.py`: Fixed feature indices `[3, 4]` → `[13, 14]` and changed `if` → `elif` for feature 22
- `tests/test_catch22.py`: Added `test_catch22_pycatch22_outlier_norm` regression test that uses mock to verify the z-normalised series is passed to `DN_OutlierInclude`

### Testing

All 4 existing catch22 tests pass, plus the new regression test.

### DIFF
--- a/aeon/transformations/collection/feature_based/_catch22.py
+++ b/aeon/transformations/collection/feature_based/_catch22.py
@@ -398,7 +398,7 @@ class Catch22(BaseCollectionTransformer):
             dim = i * len(f_idx)
             series = list(X[i])
 
-            if self.outlier_norm and (3 in f_idx or 4 in f_idx):
+            if self.outlier_norm and (13 in f_idx or 14 in f_idx):
                 outlier_series = list(z_normalise_series(X[i]))
 
             for n, feature in enumerate(f_idx):
@@ -406,9 +406,9 @@ class Catch22(BaseCollectionTransformer):
                 if not transform_feature[f_count]:
                     continue
 
-                if self.outlier_norm and feature in [3, 4]:
+                if self.outlier_norm and feature in [13, 14]:
                     c22[dim + n] = features[feature](outlier_series)
-                if feature == 22:
+                elif feature == 22:
                     c22[dim + n] = np.mean(series)
                 elif feature == 23:
                     c22[dim + n] = np.std(series)
--- a/aeon/transformations/collection/feature_based/tests/test_catch22.py
+++ b/aeon/transformations/collection/feature_based/tests/test_catch22.py
@@ -1453,3 +1453,50 @@ catch22wrapper_basic_motions_data = np.array(
         ],
     ]
 )
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies("pycatch22", severity="none"),
+    reason="skip test if required soft dependency pycatch22 not available",
+)
+def test_catch22_pycatch22_outlier_norm():
+    """Test that outlier_norm passes z-normalised series to DN_OutlierInclude.
+
+    Regression test for GH #3632: the outlier_norm condition checked the
+    wrong feature indices, so DN_OutlierInclude (features 13, 14) was
+    always computed on the raw series when use_pycatch22=True.
+    """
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by TonyBagnall — CHANGES_REQUESTED** [review]
Good spot on the misnumbering! However, the failing `test-no-soft-deps` job is because the new regression test imports `pycatch22` unconditionally. Since this PR tests the `use_pycatch22=True` path, the test should be skipped when the soft dependency is unavailable Our way is like this


```python
from aeon.utils.validation._dependencies import _check_soft_dependencies

@pytest.mark.skipif(
    not _check_soft_dependencies("pycatch22", severity="none"),
    reason="required soft dependency pycatch22 not available",
)
def test_catch22_pycatch22_outlier_norm():
    import pycatch22
    ...
```

BB0813:
@TonyBagnall Thanks for the review! I've added the `@pytest.mark.skipif` decorator and fixed the black formatting. Please take a look.

TonyBagnall (approved):

PR #3640, opened 2026-07-12
[ENH] Speedup BOSS/cBOSS (up to 7x speedup, average 2x-3x) by vectorizing sparse matrix distance computations 

patrickzib:
This PR vectorizes the computation of the BOSS distance over sparse matrices. This is one bottleneck when using BOSS, but also cBOSS.

The vectorization significantly improves both training and prediction times while preserving the original behavior. Accuracy remains unchanged; only the implementation of the distance computation has been optimized.

Benchmarking in progress...

## Results for BOSS (slowest datasets)

| Dataset                    | Accuracy (New / Old) | Fit Time (New / Old) | **Fit Speedup** | Predict Time (New / Old) | **Predict Speedup** |
|----------------------------|---------------------:|---------------------:|----------------:|-------------------------:|--------------------:|
| NonInvasiveFetalECGThorax2 | 0.867 / 0.867        | 2069.44 / 8455.25    | **4.09×**       | 81.34 / 471.35           | **5.80×**           |
| NonInvasiveFetalECGThorax1 | 0.836 / 0.836        | 2213.27 / 8647.56    | **3.91×**       | 62.54 / 312.22           | **4.99×**           |
| HandOutlines | 0.914 / 0.914 | 4326.68 / 16254.51 | **3.76x** | 140.82 / 444.63 |  **3.17x** |
| FordB | 0.819 / 0.819 | 7755 / 24131  | **3.11x** | 25.75 / 200.69 |  **7.7x** |

### BOSS (with new distance / this PR)

| Metric | Value |
|--------|------:|
| **Mean Accuracy** | **0.812** |
| Median Accuracy | 0.860 |
| Fit Time (s) | 41,304 |
| Prediction Time (s) | 2,355 |

### BOSS (original distance / aeon main)

| Metric | Value |
|--------|------:|
| **Mean Accuracy** | **0.812** |
| Median Accuracy | 0.860 |
| Fit Time (s) | 124,387 |
| Prediction Time (s) | 8,366 |

All runtimes using a single core.

### cBOSS (with new distance / this PR)

| Metric | cBOSS |
|--------|-------:|
| Mean Accuracy | 0.814 |
| Median Accuracy | 0.849 |
| Fit Time (s) | 26,494 |
| Prediction Time (s) | 1,617 |

### Twice as fast as BOSS with new distance


aeon/classification/dictionary_based/_boss.py
```diff
@@ -680,46 +680,92 @@ def _set_word_len(self, X, y, word_len):
...
+
+        # The BOSS distance is asymmetric: only words present in the left
+        # operand X contribute. CSR makes those left-hand rows cheap to scan.
+        if not isspmatrix_csr(X):
+            X = X.tocsr(copy=True)
+        X.eliminate_zeros()
+
+        # The kernel then needs all right-hand rows with a matching word.
+        # CSC makes that direct.
+        if not isspmatrix_csc(X):
```
TonyBagnall: should this not be a test on Y not X?

aeon/classification/dictionary_based/_boss.py
```diff
@@ -680,46 +680,92 @@ def _set_word_len(self, X, y, word_len):
...
-            for i in range(len(distance_matrix)):
-                distance_matrix[i] = boss_distance(X, Y, i, XX_row_norms, XY)
+        if X.shape[1] != Y.shape[1]:
+            raise ValueError("X and Y must have the same number of features.")
+
+        # The BOSS distance is asymmetric: only words present in the left
+        # operand X contribute. CSR makes those left-hand rows cheap to scan.
+        if not isspmatrix_csr(X):
+            X = X.tocsr(copy=True)
+        X.eliminate_zeros()
```
TonyBagnall: so if X is already a CSR, this mutates X in place. Does that have any side effects?

aeon/classification/dictionary_based/_boss.py
```diff
@@ -680,46 +680,92 @@ def _set_word_len(self, X, y, word_len):
...
+
+        # The kernel then needs all right-hand rows with a matching word.
+        # CSC makes that direct.
+        if not isspmatrix_csc(X):
+            Y = Y.tocsc(copy=True)
+        Y.eliminate_zeros()
+
+        distance_matrix = _boss_pairwise_distances(
```
TonyBagnall, line 705: just a question, no change required, but could the boss distance move to distances? Not sure if we have considered this

TonyBagnall (changes requested):
Fantastic, I love a good speed up :) One typo X vs Y I think, one query about in place mutation and a general query about the distance function

TonyBagnall:
also, can this benefit CBOSS and can that be quantified too?

patrickzib:
> also, can this benefit CBOSS and can that be quantified too?

Yes, it does - here are the numbers:

### BOSS (with new distance / this PR)

| Metric | Value |
|--------|------:|
| **Mean Accuracy** | **0.812** |
| Median Accuracy | 0.860 |
| Fit Time (s) | 41,304 |
| Prediction Time (s) | 2,355 |

### cBOSS (with new distance / this PR)

| Metric | cBOSS |
|--------|-------:|
| Mean Accuracy | 0.814 |
| Median Accuracy | 0.849 |
| Fit Time (s) | 26,494 |
| Prediction Time (s) | 1,617 |

### Twice as fast as BOSS with new distance


I do not have the speedup for cBOSS over its old non-vectorized version, but it is almost 2 times faster than BOSS.

aeon/classification/dictionary_based/_boss.py
patrickzib: Ops. Yes, that was a bug. Thank you for spotting it.

aeon/classification/dictionary_based/_boss.py
patrickzib: I think not - given it is a bag of word counts, and zero is equal to no word present. It is only a means of speeding up computations

GPT says:
> If you’re using SciPy or scikit-learn distance functions (e.g., pairwise_distances, 
> NearestNeighbors, cosine_similarity, sparse dot products), calling 
> `X.eliminate_zeros()`
> is generally safe and often beneficial for performance. The computed distances should 
> remain exactly the same.

aeon/classification/dictionary_based/_boss.py
patrickzib: Do we have any other use case for this? The BOSS distance is defined only for Bag-of-Words representations, so I don’t see where else this would apply.

aeon/classification/dictionary_based/_boss.py
patrickzib: But I made it more explicit and consistent now by swapping statements.

TonyBagnall (approved):
lgtm, nice speed up

aeon/classification/dictionary_based/_boss.py
TonyBagnall: yeah true, probably best living here

PR #3642, opened 2026-07-13
[MNT] Deprecate pycatch22

TonyBagnall:
After #3618 we no longer need the soft dependency for pycatch22, since our implementation a 6x+ faster and produces identical output. This change touches a lot of files, which is another argument for deprecation: this argument is passed around all the time.

Had to make a choice about warning. If standard approach used and warning raised in catch22, a single DrCIF call would raise a large number of warnings. Instead, we just raise once

This PR deprecates the use_pycatch22 parameter across Catch22 and the classifiers, regressors, clusterers, interval forests, and RIST estimators that expose it. The parameter is scheduled for removal in
  v1.7.0 because aeon’s native implementation is faster.

  Existing behavior remains compatible during the deprecation period:

  - Omitting use_pycatch22 uses aeon’s implementation without warning.
  - Explicitly setting either True or False raises a FutureWarning.
  - use_pycatch22=True continues to use pycatch22 until removal.
  - Each public estimator instance warns once, including across aeon’s internal fit() reset.
  - Internal Catch22 transformers and forest clones do not repeat the warning.

  The pycatch22 section has also been removed from the Catch22 example notebook.

  Tests now verify repeated public warnings, warnings-as-errors, dependency tags, and that CIF, DrCIF, and RIST workflows emit only one deprecation warning.

Validation completed:

  - 31 directly affected tests pass.
  - Estimator API checks pass for all 10 affected estimators.
  - All pre-commit hooks pass.
  - git diff --check passes.

  Areas for careful review are the per-instance warning tracking and the private clone-safe Catch22 implementation used by composed estimators.


hadifawaz1999 (approved):
LGTM !

PR #3644, opened 2026-07-14
[ENH] HC2 verbosity

### LINKED ISSUE #3616: [ENH] Speed up HIVE-COTEv2.0 classifier

TonyBagnall:
see #2815 
closes #3616 

Motivation

On main, HIVECOTEV1 and HIVECOTEV2 take a verbose parameter, documented as "for information only", but it produces almost no useful output:

- No ensemble-level output. _BaseHIVECOTE prints nothing itself. It only copies verbose onto components that have a verbose attribute, so there's no report of which component is fitting, how long it took, its train accuracy or its CAWPE weight.
- Most components have no verbose at all. Arsenal, TDE, DrCIF, cBOSS, TSF, RISE and RotationForestClassifier (STC's default classifier) neither take verbose nor print anything. BaseIntervalForest and BaseRotationForest have no logging either.
- STC is the only component that prints. Its verbose is a bool that prints fixed messages such as Fitting estimator..., with no timings, sizes or progress.
- RST is the only one with progress. RandomShapeletTransform prints per-batch progress, but skips the first batch and gives no start summary and no end summary beyond a final shapelet count.

As a result, a HIVE-COTE 2 fit, which can take hours, gives almost no feedback: nothing on progress, which component is running, or how much time is left.

This PR adds consistent, levelled verbose output across the HC2 stack:

- Components: Arsenal, TDE, DrCIF (through BaseIntervalForest), RotationForest, STC and RST. Level 1 prints the fit configuration, periodic progress and a summary; level 2 adds per-estimator or per-candidate lines with estimated or contract time remaining.
- HIVECOTEV1 and HIVECOTEV2: they now report per-component timings, train accuracy and weights, and pass verbose - 2 down to their components.

I'm using a bespoke logging, ultimately better to move to a proper logging tool, but this will be easy to adapt if we do, I'm more interested in the structure of logging. Logging levels for components 0,1,2

HIVECOTEV1 and HIVECOTEV2 now log their own fit progress. Components receive verbose - 2, so verbose=1 reports ensemble progress only, verbose=2 adds each component's parameters, and verbose=3 or higher is needed for the components' own output (previously any non-zero verbose was passed straight to the components).

image


TonyBagnall:
also do some profiling with the new HC2. The bottleneck now is DrCIF, TDE now the fastest (!) STC-Arsenal about equivalent. So 

──────────┬─────────┬───────────┬─────────────────┐                                                                                                        
  │ Component │  main   │  ajb/hc2  │     speedup     │                                                                                                        
  ├───────────┼─────────┼───────────┼─────────────────┤                                                                                                        
  │ STC       │ ~15.9 s │ 5.7 s     │ ~2.8x           │                                                                                                        
  ├───────────┼─────────┼───────────┼─────────────────┤                                                                                                        
  │ DrCIF     │ ~25.1 s │ 23.1 s    │ ~1.1x           │                                                                                                        
  ├───────────┼─────────┼───────────┼─────────────────┤                                                                                                        
  │ Arsenal   │ ~3.4 s  │ 3.3–5.7 s │ ~1x (untouched) │                                                                                                        
  ├───────────┼─────────┼───────────┼─────────────────┤                                                                                                        
  │ TDE       │ ~10.8 s │ 1.1 s     │ ~8.6x           │                                                                                                        
  └───────────┴─────────┴───────────┴─────────────────┘

TonyBagnall:
note  this has a lot of commits from other branches it is built upon, its pretty simple when they are in

TonyBagnall:
three CI failures share one root cause, operations are now faster :)
                                                                         
  Root cause: time.time() has 15.6 ms resolution on Windows. When a fit completes inside one clock tick,
  train_time is exactly 0.0:

  time.time()         resolution: 0.015625
  time.perf_counter() resolution: 1e-07                                  
                                                                                                    
  time.time():   997/1000 short operations measured exactly 0.0
  perf_counter:    0/1000

aeon/classification/hybrid/_base_hive_cote.py
```diff
@@ -94,18 +107,58 @@ def _fit(self, X, y):
...
         return self
 
+    def _log(self, message, level=1):
+        """Print a message when the configured verbosity reaches ``level``."""
+        if self.verbose >= level:
+            print(message, flush=True)  # noqa: T201
+
+    def _log_fit_configuration(self):
+        """Log subclass-specific fit configuration when verbosity is enabled."""
+
```
baraline: Not sure why we have an empty method here, not used anymore or forgot the log call for the fit config ?

aeon/base/_estimators/sklearn/_rotation_forest.py
```diff
@@ -268,30 +293,132 @@ def _fit_rotf(self, X, y, save_transformed_data: bool = False):
...
+            self._log(
+                f"[RotF] Finished fit: built={len(self.estimators_)}, "
+                f"elapsed={time.perf_counter() - start_time:.2f}s"
+            )
         return X_t
 
+    @staticmethod
+    def _log(message):
+        """Print a fit progress message after the caller checks verbosity."""
+        print(message, flush=True)  # noqa: T201
```
baraline, line 405: Wouldn't we be better of putting this function in the base estimator and using a proper loggin (can use standard library `logging`) that user can customize based on their need ? (https://docs.python.org/3/library/logging.html)

Using prints is not ideal for this

baraline (changes requested):

TonyBagnall:
I'm working on this atm, its changing :)

aeon/base/_estimators/sklearn/_rotation_forest.py
TonyBagnall: yes but that is a bigger change beyond the scope here, we have an existing print verbose mechanism I am adapting to make it work properlyI think we discussed in #2815

TonyBagnall: also made the point in the PR comment above about logging. FYI, scikit do this:

Estimators use print. verbose output is plain printing behind if self.verbose / if verbose > n checks. That covers the forest building tree i of n line, MLP iteration losses and similar. There are about 130  
    such checks and over 400 print( calls outside the tests.                                                                                                                                                        
  - Parallel progress bypasses logging. joblib writes the [Parallel(...)]: Done ... lines directly to stderr, or to stdout when verbose >= 50. Its source even has a note: # XXX: Not using the logger framework:   
    need to learn to use logger better.                                                                                                                                                                             
  - The logging module is only used by the dataset downloaders. The files are _california_housing, _covtype, _kddcup99, _lfw, _rcv1, _species_distributions and _twenty_newsgroups, which log downloads and         
    extraction with logger.info / logger.debug. sklearn/__init__.py also creates a package logger.

aeon/classification/hybrid/_base_hive_cote.py
TonyBagnall: its an optional abstract method, or an optional hook, but its only overridden in one place so I;ll get rid of it

TonyBagnall:
cahtgpt likes it, only point is 
_format_duration is now essentially duplicated in BaseIntervalForest, BaseRotationForest, Arsenal, TDE and RST. I wouldn't hold this PR up for that
and HC2 weird contract.

MatthewMiddlehurst:
going to try the copilot resolve.. if it cant do this then it is unusable

MatthewMiddlehurst:
@copilot resolve the merge conflicts in this pull request

MatthewMiddlehurst:
LMAO ok it went rogue. Note to self, never press that button again

baraline (approved):
Per Tony comments, considering the scope is only HC2 and not general aeon logging, I'm OK with the changes

TonyBagnall:
thanks @baraline I'll create an issue for logging, might be a good project for a more substantial contribution tie up.

PR #3648, opened 2026-07-14
[BUG] fix RandomShapeletTransform distance instability for high offsets

### LINKED ISSUE #3643: [BUG] RandomShapeletTransform distance is unstable for high-offset, low-variance series
### Describe the bug

RandomShapeletTransform calculates rolling window variance using:

  variance = (sum_of_squares - mean * mean * length) / length

  For series containing a large constant offset and comparatively small variation, this subtracts two nearly equal large values. The resulting catastrophic cancellation can make the estimated variance inaccurate or effectively zero.

  Shapelet distance is z-normalised and should therefore be invariant to adding a constant to the input. Currently, adding an offset can substantially change the calculated distance.

  This issue exists on main and is separate from the current RandomShapeletTransform efficiency work.

```python
import numpy as np

  from aeon.transformations.collection.shapelet_based._shapelet_transform import (
      _online_shapelet_distance,
  )
  from aeon.utils.numba.general import z_normalise_series

  rng = np.random.RandomState(0)
  series = rng.normal(scale=1e-5, size=100)

  position = 30
  length = 20
  shapelet = z_normalise_series(series[position : position + length])
  sorted_indices = np.argsort(np.abs(shapelet))[::-1].astype(np.int32)

  base_distance = _online_shapelet_distance(
      series, shapelet, sorted_indices, position, length
  )
  shifted_distance = _online_shapelet_distance(
      series + 1e8, shapelet, sorted_indices, position, length
  )

  print(base_distance)
  print(shifted_distance)

  np.testing.assert_allclose(
      shifted_distance,
      base_distance,
      rtol=1e-10,
      atol=1e-10,
  )
```

### Expected results

Adding a constant to a series should not change its z-normalised shapelet distance. Both distances should be approximately zero because the shapelet comes from the selected source window.

### Actual results

A representative result is:

  base_distance:    4.33e-32
  shifted_distance: 0.999985

  The shifted distance is close to 1 rather than zero. This can affect candidate quality scores, shapelet selection, and downstream predictions for high-offset, low-variance data.

### Suggested fix

Use a numerically stable variance calculation while retaining the incremental distance algorithm.

  One possible approach is to subtract a fixed reference value from series values before accumulating the rolling sum and sum of squares. This removes the large common offset without allocating a centred copy. A stable rolling mean/
  variance algorithm could also be considered.

  The change should include:

  - An additive-offset invariance test.
  - Comparison against a stable brute-force z-normalised distance implementation.
  - Tests for flat and nearly flat windows.
  - Benchmarks to ensure the fix does not materially regress transform performance.

### Versions

On main

BB0813:
## Fix

Fixes #3643

### Problem

`_online_shapelet_distance` computed rolling variance as:

```python
std = math.sqrt((sum2 - mean * mean * length) / length)
```

When a series has a large constant offset and comparatively small variation, this subtracts two nearly equal large values (catastrophic cancellation). The resulting std is wrong (often effectively zero or meaningless), so z-normalised shapelet distances are no longer invariant to adding a constant.

Issue reproducer on main:

- `base_distance ≈ 0`
- `shifted_distance ≈ 1` after `series + 1e8`

### Solution

Subtract a fixed reference value (`series[position]`) before accumulating the rolling sum and sum of squares. This removes the large common offset without allocating a centred copy, and keeps the incremental update structure.

When reconstructing the true window mean for z-normalisation, use `mean + ref`.

### Changes

- `_shapelet_transform.py`: stable rolling sum / sum-of-squares in `_online_shapelet_distance`
- `tests/test_shapelet_transform.py`: regression test for offset invariance

### Testing

- New regression test covers:
  - unit-scale series with large offset (`allclose`)
  - high-offset, low-variance series from the issue (distance stays near 0, not ~1)
- Verified the unstable formula yields a bogus std (~1.26) on the issue data, while the fixed path recovers the true std (~1e-5)

### DIFF
--- a/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
+++ b/aeon/transformations/collection/shapelet_based/_shapelet_transform.py
@@ -821,18 +821,23 @@ def _online_shapelet_distance(series, shapelet, sorted_indicies, position, lengt
     if position + length > len(series):
         position = int((len(series) - length) / 2)
 
+    # Subtract a fixed reference before accumulating sum/sum-of-squares.
+    # This avoids catastrophic cancellation when the series has a large
+    # constant offset relative to its variance (see GH #3643).
+    ref = series[position]
     subseq = series[position : position + length]
 
     sum = 0.0
     sum2 = 0.0
     for i in subseq:
-        sum += i
-        sum2 += i * i
+        v = i - ref
+        sum += v
+        sum2 += v * v
 
     mean = sum / length
     std = math.sqrt((sum2 - mean * mean * length) / length)
     if std > AEON_NUMBA_STD_THRESHOLD:
-        subseq = (subseq - mean) / std
+        subseq = (subseq - (mean + ref)) / std
     else:
         subseq = np.zeros(length)
 
@@ -855,8 +860,8 @@ def _online_shapelet_distance(series, shapelet, sorted_indicies, position, lengt
             if not traverse[n]:
                 continue
 
-            start = series[pos - n]
-            end = series[pos - n + length]
+            start = series[pos - n] - ref
+            end = series[pos - n + length] - ref
 
             sums[n] += mod * end - mod * start
             sums2[n] += mod * end * end - mod * start * start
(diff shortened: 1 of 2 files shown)


BB0813:
Note: this is intentionally separate from #3574 (RST speedup). That PR still uses the unstable `(sum2 - mean^2 * length) / length` form. Happy to rebase / port the reference-offset fix onto the rewritten distance function if #3574 lands first.

TonyBagnall:
hi, thanks for the PR but I assigned this to myself, this module is in a state of flux atm, this change is only to be made once changes are in

BB0813:
Got it, thanks for the clarification and for reviewing! Happy to revisit after the RST changes land.

PR #3649, opened 2026-07-14
[ENH] Enhancing the functionality of SAX

hadifawaz1999:
After getting #2972 now i finally added these changes.

## Summary

This PR extends the `SAX` transformer with several new features and improves its documentation and test coverage.

### Changes

- Added support for custom alphabets.
- Added support for windowed SAX through `window_size` and `stride`.
- Extended `inverse_sax` to support both standard and windowed SAX representations.
- Added Numba-accelerated SAX symbol generation.
- Improved the SAX example notebook with new examples and visualizations.
- Expanded the test suite with deterministic, windowing, custom alphabet, and inverse SAX tests.

The default behaviour remains unchanged, ensuring backward compatibility.


aeon/transformations/collection/dictionary_based/tests/test_sax.py
```diff
@@ -1,101 +1,465 @@
-"""Test for SAX transformations on time series."""
+"""Tests for the Symbolic Aggregate approXimation (SAX) transformer."""
```
patrickzib: Just a comment, not a blocker: "transformation"? transformer sounds like the architecture using attention.

aeon/transformations/collection/dictionary_based/tests/test_sax.py
```diff
@@ -1,101 +1,465 @@
...
 
-@pytest.mark.parametrize("n_segments", [8])
-@pytest.mark.parametrize("alphabet_size", [4])
-def test_equal_length_univariate_sax(n_segments, alphabet_size):
-    """Test if SAX maintains equal length for univariate time series.
+def _z_normalize(X):
+    """Z-normalize each channel of each series independently."""
+    means = np.mean(X, axis=-1, keepdims=True)
+    stds = np.std(X, axis=-1, keepdims=True)
+    stds[stds == 0] = 1.0
```
patrickzib: We are commonly using 
`from aeon.utils.numba.general import AEON_NUMBA_STD_THRESHOLD`

and checking against this one instead of `0`?

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -117,13 +164,10 @@ def _get_paa(self, X):
...
-            # Safe, if std is 0
-            X = (X - np.mean(X, axis=-1, keepdims=True)) / (
-                np.std(X, axis=-1, keepdims=True) + 1e-8
-            )
-
-            # Non-Safe is std is 0
-            # X = scipy.stats.zscore(X, axis=-1)
+            means = np.mean(X, axis=-1, keepdims=True)
+            stds = np.std(X, axis=-1, keepdims=True)
+            stds[stds == 0] = 1.0
```
patrickzib: Please see comment above about using `AEON_NUMBA_STD_THRESHOLD`, too

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -150,8 +194,46 @@ def _transform(self, X, y=None):
...
+        # (cases, channels, windows, window_size)
+        # into:
+        # (cases * windows, channels, window_size)
+        X_windows_3d = X_windows.transpose(0, 2, 1, 3).reshape(
+            n_cases * n_windows,
+            n_channels,
+            self.window_size,
+        )
+
+        X_paa = self._get_paa(X_windows_3d)
```
patrickzib: So currently, PAA is re-computed over each sliding window?
Wouldn't a sliding window PAA (with sliding mean, std) perform faster here?

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -260,6 +442,27 @@ def _get_test_params(cls, parameter_set="default"):
         return params
 
 
+@njit(cache=True, parallel=True)
+def _extract_windows(X, window_size, stride):
```
patrickzib, line 577: It might be worth considering to apply mean / std normalization here already, as you can compute sliding means and stds over windows quickly?

patrickzib (commented):
Nice PR! 

So, the purpose is to make SAX a BOP Model? Just some general thoughts about performance of the sliding window operations. 

I did not check the tests.

aeon/transformations/collection/dictionary_based/tests/test_sax.py
hadifawaz1999: makes sense !

aeon/transformations/collection/dictionary_based/tests/test_sax.py
hadifawaz1999: i will look at it, didnt know we had this !

aeon/transformations/collection/dictionary_based/_sax.py
hadifawaz1999: i will look at it, didnt know we had this !

aeon/transformations/collection/dictionary_based/_sax.py
hadifawaz1999: Not sure i understand

aeon/transformations/collection/dictionary_based/_sax.py
hadifawaz1999: Not sure i understand

aeon/transformations/collection/dictionary_based/_sax.py
patrickzib: If I understand correctly, you are first extracting windows, then transforming each separately via PAA? 

Thus, the complexity should be something like O(n * window_size)

However, there is quite an  overlap between consecutive windows. 

Thus, you can reduce it to O(n * n_segments) by exploiting overlap
.

hadifawaz1999: ah yes probably you're right, didnt think much about that. Might be a bit messy to handle given we have stride too. Happy to make an issue for it after this pr if you want

patrickzib: Sure, not a blocker, just a chance to improve the performance.

aeon/transformations/collection/dictionary_based/_sax.py
patrickzib: Same as above - not a blocker, just performance improvement possibilities.

patrickzib:
@hadifawaz1999 not sure how to do a review of the notebook (it does not seem to keep my comments)

The visualization is really cool. You might want to consider moving the plotting function 

```python
def plot_sax_representation(
    X,
    X_sax,
    X_inverse,
    sax,
    series_index=0,
    channel_index=0,
    window_index=None,
):
```

to a helper in aeon/visualisation/learning_task such that it can be re-used in other notebooks, etc.

hadifawaz1999:
> @hadifawaz1999 not sure how to do a review of the notebook (it does not seem to keep my comments)
> 
> The visualization is really cool. You might want to consider moving the plotting function
> 
> ```python
> def plot_sax_representation(
>     X,
>     X_sax,
>     X_inverse,
>     sax,
>     series_index=0,
>     channel_index=0,
>     window_index=None,
> ):
> ```
> 
> to a helper in aeon/visualisation/learning_task such that it can be re-used in other notebooks, etc.

Yes its always annoying to review notebooks on PRs i know ! Makes sense actually i will do that change then re-ping you !

patrickzib:
@hadifawaz1999 If not mistaking, I think, this paper contains the rough idea of sax over sliding windows using cumulative sums:
https://ieeexplore.ieee.org/document/8215500

imageimage

hadifawaz1999:
> @hadifawaz1999 If not mistaking, I think, this paper contains the rough idea of sax over sliding windows using cumulative sums: https://ieeexplore.ieee.org/document/8215500
> 
> imageimage

cool ! can check that out at some point to accelerate the code

hadifawaz1999:
@patrickzib checking this now i think i should znormalize the windows not the full series.
from sax-vsm cz they used windowed sax, https://jmotif.github.io/sax-vsm_site/algorithm/slidingwindowsax/
will make this update ! (just to be algorithmically correct)

aeon/transformations/collection/dictionary_based/tests/test_sax.py
patrickzib: This might still be missing?

aeon/transformations/collection/dictionary_based/tests/test_sax.py
```diff
@@ -203,6 +208,128 @@ def test_sax_znormalized_false_handles_constant_series():
...
+        alphabet_size=4,
+        znormalized=False,
+        window_size=window_size,
+        stride=stride,
+    ).fit_transform(X)
+
+    windows = X.reshape(1, 1, 2, window_size)
+    means = windows.mean(axis=-1, keepdims=True)
+    stds = windows.std(axis=-1, keepdims=True)
+    stds[stds == 0] = 1.0
```
patrickzib: And here you might want to use the constant, too?

aeon/transformations/collection/dictionary_based/_sax.py
```diff
@@ -150,6 +149,14 @@ def _validate_window_parameters(self, n_timepoints):
         if not isinstance(self.stride, int) or self.stride <= 0:
             raise ValueError("stride must be a positive integer")
 
+    @staticmethod
+    def _z_normalize(X):
+        """Z-normalize each series/channel independently along time."""
+        means = np.mean(X, axis=-1, keepdims=True)
+        stds = np.std(X, axis=-1, keepdims=True)
+        stds[stds == 0] = 1.0
```
patrickzib: Again, might want to change to the aeon constant?

patrickzib (commented):
You might still want to use 
`from aeon.utils.numba.general import AEON_NUMBA_STD_THRESHOLD`
instead of `0`?

aeon/transformations/collection/dictionary_based/_sax.py
patrickzib, line 383: I might be wrong on this, but don't you have to account somewhere for the PAA-normalization? i.e. add back the mean and stdev after _invert_sax_symbols?

hadifawaz1999: makes sense forgot about that, this means we have to save in memory the means and stds of every window, will fix it

aeon/transformations/collection/dictionary_based/_sax.py
hadifawaz1999: yes i should forgot, will fix it

hadifawaz1999:
@patrickzib let me know if its all good now

patrickzib (approved):
LGTM now. Thank you very much. Very nice addition.

There were a couple of minor issues left, so I fixed them instead of bothering you over and over.

- wording: `SAX transformer` vs `SAX transformation`
- re-implementing z-norm twice in the test case 
- an issue in `_paa_sax` when `y` was none
- changing Numba thread count without try/finally
- Out-of-range symbols for inverse transform
- Check for alphabet_size must be greater than 0

hadifawaz1999:
Thanks for all the changes @patrickzib !

PR #3650, opened 2026-07-15
[BUG] fix ClaSP edge cases

TonyBagnall:
Fixes two ClaSP edge cases affecting F1 scoring and neighbour-mask construction. Also removes an unused helper and simplifies scoring-metric selection.

  ## Bugs fixed

  - F1 scoring could raise ZeroDivisionError when a split contained no predicted or true positives for a class. Undefined precision and recall are now assigned 0.0, following the standard convention.
  - Neighbour masks were incorrectly populated when fewer than k candidates were available. The fallback attempted to assign fewer candidate indices to a fixed-length row. It now cycles through the available indices,
    producing a complete and deterministic mask.

  ## Cleanup

  - Removed the unused _sliding_window helper.
  - Simplified the complementary branch in _check_scoring_metric.

  ## Tests

  - Added a public regression using the GunPoint segmentation data that reproduces the pre-fix F1 crash under Numba.
  - Added a public regression asserting the complete neighbour mask when five neighbours are requested but only two candidates exist.
  - Removed coverage-only tests of private helpers and output shapes.


TonyBagnall:
Note repeated use of clasp name

PR #3651, opened 2026-07-15
[BUG] fix ACF, STL and MSTL

### LINKED ISSUE #3534: [ENH] STL and MSTL tranformer have unnecessary numba fallback

both test if numba is installed, but numba is a requirement,so this can be removed. 

remove non numba path. Not to be done before #3533 and will involve a general redesign. 

_No response_

### Additional context

_No response_

TonyBagnall:
closes #3534 

Fixes error handling in ACF, removes unnecessary non-Numba STL execution paths, and enables MSTL’s existing missing-value imputation through the public API.

  ## Changes

  - Fix the ACF insufficient-observations error so it reports the number of remaining observations instead of accessing the nonexistent `self._n_lags` attribute.
  - Use Numba unconditionally for STL. Numba is a required aeon dependency, so the optional import handling and duplicate NumPy implementation are unnecessary.
  - Remove the `use_numba` STL parameter and the corresponding `stl_use_numba` MSTL parameter.
  - Declare that MSTL accepts missing values, allowing inputs containing `NaN` to reach its existing interpolation logic.

  ## Tests

  - Verify that ACF raises the intended `ValueError` for a series that is too short for the requested lags.
  - Verify that MSTL’s built-in missing-value imputation produces the same decomposition as explicitly interpolating the input first.
  - Exercise STL and MSTL through the Numba implementation in the existing behavioural tests.
  - Verify that MSTL components reconstruct the input when knot jumps are enabled.

  ## Review focus

  Removing `use_numba` and `stl_use_numba` is a public API change. Code passing either keyword will need to remove it. However, I see no scenario where you would choose to not use numba. You would need to manually uninstall it.

  The MSTL capability tag allows missing values through the estimator validation so that MSTL can apply its documented `impute_missing` behaviour.


alexbanwell1 (dismissed):
### Overview
The PR does three things: fixes an `AttributeError` masking the intended `ValueError` in ACF's insufficient-observations check, removes the NumPy fallback path and `use_numba`/`stl_use_numba` parameters from STL/MSTL, and enables MSTL's existing NaN imputation through the public API via the `capability:missing_values` tag. Net −329 lines.

### What's correct and well done
- **ACF fix is right.** `self._n_lags` never existed, so a too-short series raised `AttributeError` instead of the intended `ValueError`. The new `X.shape[1] - n_lags` uses the local variable correctly, and the new test covers the default-lags path (`n=1` → `n_lags=1` → 0 remaining observations).
- **Removing the NumPy fallback is justified.** Numba is a hard aeon dependency, and the rest of the codebase (e.g. `_acf.py` itself) imports `njit` unconditionally. The try/except shim was dead weight.
- **The removal also fixes a latent perf bug.** In the old code, `_resolve_use_numba_local()` returned `bool(self.use_numba)`, so the default `use_numba=None` resolved to `False` and the subsequent `if use_nb is None:` "auto" branch was dead code — the default path always ran the slow NumPy implementation. Worth mentioning in the PR description: this change makes the default faster, not just cleaner. The old parity tests (1e-7 tolerance) support that the two paths were numerically equivalent, so behaviour is preserved.
- **MSTL imputation test is well constructed** — for a single interior NaN, the neighbour-average expectation exactly mirrors `np.interp`, so it genuinely validates the imputation path end-to-end through the public API.

### Issues

1. **(Minor, but it's the line being fixed) Missing space in the ACF error message.** The f-string still renders `"...calculated with just0 observations."` — `"just"` has no trailing space before the interpolation. Since this exact line is being touched, add the space.

2. **(Worth addressing) `impute_missing=False` + NaN input now fails confusingly.** Previously, NaN input was rejected up front by base-class validation (`_base_series.py`). With the tag set to `True`, MSTL accepts NaNs unconditionally, but only handles them when `impute_missing=True`. With `impute_missing=False`:
   - No Box–Cox: the NaNs flow to the internal `STLSeriesTransformer.transform()`, which raises `"Missing values not supported by STLSeriesTransformer"` — a confusing error naming an internal component the user never touched.
   - With `boxcox_lambda` set: worse — NaN comparisons are `False`, so `np.any(z <= 0.0)` passes and Box–Cox silently produces NaNs before STL errors.

   Suggest an explicit check in `_compute_components_`: if NaNs are present and `impute_missing` is False, raise a clear `ValueError` telling the user to enable `impute_missing`.

3. **(Maintainer decision — already flagged in the PR description) Breaking API change without a deprecation cycle.** STL shipped in v1.4.0 and MSTL in v1.5.0, so `use_numba`/`stl_use_numba` are released public API; aeon's policy would normally deprecate for a cycle before removal. Mitigating: the parameter's default never actually enabled Numba (the bug above), and results are numerically equivalent either way, so the removal changes behaviour for essentially nobody who didn't explicitly pass `use_numba=False`. If removed outright, it should be called out as breaking in the release notes.

4. **(Very minor) `test_mstl_with_jumps_reconstructs_input` is a smoke test in disguise.** With `output="all"`, the remainder is computed as `deseas - trend`, so `components.sum(axis=1) == input` holds *by construction* — the assertion can only catch NaNs or shape errors, not decomposition quality. Fine to keep, but the name overstates what it verifies.

### Test coverage note
The removed parity tests were the only numerical cross-check for STL, but their reference (the NumPy implementation) is what's being deleted, so removing them is consistent. The remaining behavioural tests exercise the Numba path adequately. A future improvement (out of scope here) would be a tolerance comparison against `statsmodels.tsa.STL` as an independent reference.

### Verdict
Solid cleanup PR — all three fixes are correct, and it incidentally repairs the fact that the default STL path never used Numba. Points 1 and 2 are actionable before merge; point 3 is a maintainer policy call.

TonyBagnall:
all done, maintainer call on 3) is to let it in

alexbanwell1 (approved):
Re-reviewed after the latest changes — both actionable points from my previous review are addressed, CI is green, approving.

### Changes since last review
1. **ACF error-message space fixed.** Now renders `"...calculated with just 0 observations."` correctly, using `X.shape[1] - n_lags`. The new test asserts on `match="0 observations"`.

2. **`impute_missing=False` + NaN now raises a clear error.** The new guard in `_compute_components_` is placed *before* the Box–Cox block, so it also closes the silent-NaN-through-Box–Cox hole I flagged. `test_mstl_rejects_missing_values_without_imputation` is parametrized over `boxcox_lambda=[None, 0.5]`, covering both branches. Thorough fix.

3. **Misleading test name corrected** — `test_mstl_with_jumps_reconstructs_input` → `..._by_construction`, which honestly reflects that the reconstruction invariant holds structurally rather than validating decomposition quality.

### Remaining (non-blocking)
- Removing `use_numba`/`stl_use_numba` (released in v1.4/v1.5) without a deprecation cycle is a breaking API change. It's intentional and flagged in the PR description; since the default never actually enabled Numba, practical impact is minimal — worth noting in the release notes as breaking.
- Trivial: `np.isnan(z).any()` is evaluated twice (guard + impute branch); negligible.

### CI
All checks pass (pytest across macOS/Ubuntu/Windows × Python 3.12–3.14, doctests, pre-commit, no-soft-deps, notebooks).

LGTM.

PR #3653, opened 2026-07-16
[MNT] Deprecation of StatsModelsACF and StatsModelsPACF

### LINKED ISSUE #3529: [ENH] Deprecate StatsModelsACF and PACF

we have these legacy series transformers kicking around, not used anywhere in the toolkit. We have our own AutoCorrelationSeriesTransformer, true we dont have a PACF one

I think deprecate before release 1.5, I think we should remove the statsmodels soft dep in time, not many uses of it

TonyBagnall:
Reference Issues/PRs

part of #3562 
closes #3529 

deprecates two series transformers, StatsModelsACF and StatsModelsPACF. Neither are used in aeon, so are provided purely as very thin wrappers for statsmodels. We have very few thin wrappers like this left. We have an ACF series transformer, PACF easy to implement if anyone wants it.

### DIFF
--- a/aeon/transformations/series/_acf.py
+++ b/aeon/transformations/series/_acf.py
@@ -4,6 +4,7 @@ __maintainer__ = ["TonyBagnall"]
 __all__ = ["AutoCorrelationSeriesTransformer", "StatsModelsACF", "StatsModelsPACF"]
 
 import numpy as np
+from deprecated.sphinx import deprecated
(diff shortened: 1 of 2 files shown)


MatthewMiddlehurst (approved):
supportive of this. if someone ends up needing these theres time to revert.

I would give it a bit before merging just in case.

PR #3655, opened 2026-07-16
[ENH] Speed up Arsenal classifier

### LINKED ISSUE #3371: [ENH] Dont use Parallel when n_jobs is 1

TonyBagnall:
This PR improves Arsenal fit and prediction performance, best to hope for is 20% speed up, but every little helps. 

part of #3616 
closes #3371

For context, Arsenal is now slower than TDE, about twice as fast as STC and still 10x faster than DrCIF, the new bottleneck. 

  ### Changes

  - Normalize ROCKET input once per Arsenal operation.
  - Add efficient OOB-index selection using boolean masks.
  - Add generalized parallel job handling, including a direct path for n_jobs=1. Note this copies the function from #3631 
  - Reduce vote-aggregation overhead by working with class indices.
  - Fuse training/OOB estimate generation into the fit pass.
  - Exception-safe ROCKET thread restoration.
  - Vectorised one-hot probability fallbacks for the ROCKET-family and Hydra classifiers.
  -  fixed a minor bug: arsenal was weighting by error instead of by accuracy on one unused path, but it made little difference

  ### Validation

  - Arsenal and parallel utility tests: 9 passed
  - Ruff checks and formatting passed
  - git diff --check passed


aeon/classification/convolution_based/_hydra.py
```diff
@@ -137,6 +137,14 @@ def _fit(self, X, y):
     def _predict(self, X) -> np.ndarray:
         return self.clf_.predict(X)
 
+    def _predict_proba(self, X) -> np.ndarray:
+        """Return one-hot probabilities from Hydra's class predictions."""
+        preds = self._predict(X)
+        class_indices = np.searchsorted(self.classes_, preds)
+        dists = np.zeros((len(preds), self.n_classes_))
+        dists[np.arange(len(preds)), class_indices] = 1
+        return dists
```
MatthewMiddlehurst: Is this different from our default? If it is functionally the same should just speed up the default if it makes a difference.

aeon/classification/convolution_based/tests/test_convolution_classifier_probabilities.py
MatthewMiddlehurst, line 1: Is this adding much? feel like I may be missing something with the proba changes. They mostly fell back to the default right?

aeon/classification/convolution_based/tests/test_rocket_classifier.py
MatthewMiddlehurst, line 1: Seems to be the same as other tests added?

MatthewMiddlehurst (commented):
Are all of the rocket/minirocket/multirocket changes related to threading? bit confused on what is meant by normalisation why the structure has changed rather than just calling _transform if we want to skip validation.

If we normalise the series in every one and I just missed that then ok.

Have not gone through the full Arsenal code yet but do not see anything massively off. Do you have some benchmark results similar to TDE and RotF? Especially nice to see since expected results changed here.

aeon/classification/convolution_based/_hydra.py
TonyBagnall: it can go in base, its for a tiny speed up and the current Arsenal over rides predict_proba.

aeon/classification/convolution_based/tests/test_convolution_classifier_probabilities.py
TonyBagnall: no it can go.

aeon/classification/convolution_based/tests/test_rocket_classifier.py
TonyBagnall: not dsure why I drifted into other convolution based, I think because of the generalised pass through

TonyBagnall:
went a little too far here, was meant to stick to arsenal

TonyBagnall:
move the fallback predict proba up, not sure why arsenal had an over ride method there, removed changes to non arsenal classifiers and removed the normalisation for multi-rocket, which has an argument rather than always normalise

TonyBagnall:
> Are all of the rocket/minirocket/multirocket changes related to threading? bit confused on what is meant by normalisation why the structure has changed rather than just calling _transform if we want to skip validation.

the normalisation fix is to avoid renormalising every ensemble member. 
so they all handle normalisation differently: Rocket normalises by default, (normalise=True). MultiRocket defaults to False and MiniRocket has no normalise option. 

We could just normalise once then set Rocket(normalise=False), and its simpler, but each estimator is a pipeline in estimators_. If we set it to false but there is a risk then of calling individual estimators without normalisation. 
Arsenal afc = Arsenal.fit(rawX) <- will normalise
afc.estimators_.predict(rawX) <- wont normalise

Granted, its a edge case, and there is no reason for it to touch the others, was overreaching

TonyBagnall:
image
speed up 10%

TonyBagnall:
actually makes arsenal a tiny bit better to published results 
image

aeon/transformations/collection/convolution_based/_rocket.py
```diff
@@ -136,12 +143,37 @@ def _transform(self, X, y=None):
...
+        The kernels-only part of ``_transform``: callers that normalise input
+        themselves (or use ``normalise=False``) can call this directly, so
+        ensembles that normalise once can treat every rocket transformer
+        uniformly.
+        """
+        # convolve in single precision: the kernel weights are float32 and
+        # the output features are float32, so float64 input only adds
+        # per-element promotion in the hot loop. asarray avoids a copy if X
+        # is already float32.
+        X = np.asarray(X, dtype=np.float32)
```
MatthewMiddlehurst, line 160: minor (optional) but could also do this in Arsenal to avoid doing it for each ensemble member

aeon/classification/convolution_based/_arsenal.py
```diff
@@ -300,102 +343,138 @@ def _fit_arsenal(self, X, y, keep_transformed_data=False):
...
         scaler = StandardScaler(with_mean=False)
-        scaler.fit(transformed_x, y)
+        # scoring="accuracy" makes best_score_ the LOO CV accuracy used to
+        # weight this member; with the default scorer it is the negative LOO
+        # mean squared error, which inverts the weighting
         ridge = RidgeClassifierCV(
-            alphas=np.logspace(-3, 3, 10), class_weight=self.class_weight
+            alphas=np.logspace(-3, 3, 10),
+            class_weight=self.class_weight,
+            scoring="accuracy",
```
MatthewMiddlehurst: For the scoring change,  I think this introduces a bug that causes binary problems to go unweighted. This was picked up by AI so not 100%, but it is an sklean issue with the default cv method. 

I am inclined to believe as the expected results for our dummy problem do not look weighed by the multiclass one does. If verified I think this version is better than the old one, but should probably create an issue.

aeon/transformations/collection/convolution_based/_rocket.py
```diff
@@ -275,11 +301,26 @@ def _apply_kernels(X, kernels):
...
-    return _X.astype(np.float32)
+# one kernel body, two compilations: a parallel dispatcher for standalone
+# n_jobs > 1 use, and a serial nogil dispatcher (prange degrades to range
+# without parallel=True) that never enters numba's threading layer, so
+# ensemble members can transform concurrently from joblib threads. The
+# transform kernels are compiled without fastmath so both compilations
+# produce identical results (measured as free here); fastmath would let
+# them reassociate differently and break n_jobs invariance.
+_apply_kernels = njit(parallel=True, cache=True)(_apply_kernels_impl)
+_apply_kernels_serial = njit(nogil=True, cache=True)(_apply_kernels_impl)
```
MatthewMiddlehurst: Tested this and the cache name is _apply_kernels_impl, not the names here, which means whatever you run first will be cached first and then loaded in the future regardless of changing n_threads input.

aeon/classification/convolution_based/_arsenal.py
TonyBagnall: https://github.com/scikit-learn/scikit-learn/issues/34942

got a fix I'm testing, if it works could go in here

TonyBagnall: mitigated as follows:
                                                                                                                                                                              
 For binary problems, this change works out the leave-one-out accuracy from the predictions RidgeClassifierCV already stores, taking the predicted class from their sign. It then refits at the best alpha, which costs one extra ridge fit per member. Multiclass is untouched and bit-identical.                                                                                    

I ran the lot and it makes Arsenal better but not HC2.                             
                                                                                                                                                                                            
The expected results for univariate Arsenal and HIVECOTEV2 have been regenerated

aeon/transformations/collection/convolution_based/_rocket.py
TonyBagnall: I've dropped the split: there's a single parallel=True kernel again, as on main, and all calls go through the shared lock. That still stops ensemble threads entering numba's threading layer at the same time.

aeon/transformations/collection/convolution_based/_rocket.py
TonyBagnall: good point, done: Arsenal casts to float32 once after normalising, so each member's np.asarray no longer copies. Only for Rocket, and outputs are unchanged.

MatthewMiddlehurst (approved):
LGTM. May be a couple of small things but in a good enough state for a release.

PR #3656, opened 2026-07-16
[ENH] Add parameters to method timer and stop `fit` timer from overwriting the attribute

MatthewMiddlehurst:
Allows estimators to set their own fit time if they want. Mostly useful for estimators which load from results files.


TonyBagnall (approved):
lgtm, although BaseSimilaritySearch still documents fit_time_millis_ as the wall-clock time “set automatically by the @method_timer decorator.” That is no longer necessarily true, but no blocker

PR #3664, opened 2026-07-21
[MNT] dimensions -> channels

TonyBagnall:
deprecates two parameters in ordinal from dimensions -> channels and corrects package wide references of dimensions to channels


baraline (approved):
Lgtm, deprecation notice where needed and got them all.

PR #3665, opened 2026-07-22
[ENH] Skip redundant per-slice input validation in interval feature extraction

### LINKED ISSUE #3599: [ENH] Move collection variance check out of _check_X so it only validates top-level estimator inputs

## Describe the issue

#3315 added `check_collection_variance()` to reject collections with near-zero but non-constant variance (`std <= 1e-7`, `range != 0`), because some numba utilities treat such series as effectively constant. It was wired into `BaseCollectionEstimator._check_X`, which runs on every call to a collection estimator's public API, including internal calls made by composite estimators on derived data.

This caused #3570: HIVE-COTE v2 crashed intermittently in the `hybrid.ipynb` notebook. The trigger is DrCIF's first-order-difference representation of `ItalyPowerDemand`: case 15 has a locally linear stretch, so `np.diff(X[15, 0])[13:16]` is constant up to float rounding (`std ≈ 5.3e-9`, `range ≈ 1.3e-8`). Whenever `RandomIntervals` drew that interval and passed the slice to `Catch22.fit_transform`, the variance check raised and aborted the whole HC2 fit.

Near-constant subseries are normal for interval, shapelet, and dictionary pipelines, and the error's advice ("rescale or normalise your data") is not appropriate: `ItalyPowerDemand` is already z-normalised.

As an interim fix, PR #3598 demoted the check in `_check_X` to a `UserWarning`. That stops the crashes but is not the final solution: users can now get spurious low-variance warnings caused by internal slices of valid data

For a test, `np.diff` of `ItalyPowerDemand` case 15, columns 13:16, is a real-world near-constant slice. A seeded `DrCIFClassifier` fit on `ItalyPowerDemand` covering that interval would catch any reintroduction.

Adding two data points from the HC2 speedup work. First, the cost side: DrCIF makes over a thousand validated Catch22 calls per fit/predict cycle (per tree, per representation), each paying the full _check_X  
  ▎ including the per-case std scan of check_collection_variance — validation inside the hot loop of the dominant HC2 component. Second, option 2 now has a precedent: the rocket transformers expose a private      
  ▎ kernels-only entry point (_transform_kernels) that Arsenal calls after validating once at ensemble level, keeping public calls fully checked and stored pipelines self-contained. The same pattern would fit the 
  ▎ interval forests' internal Catch22 usage, alongside restoring the top-level check to an error per option 1.                                                                                                      
                                                                                                                  

## Related

* #3315
* #3570
* PR #3598 , interim warning fix


The check should distinguish top-level user input from internal component calls. Since internal components currently use the same public `fit`/`fit_transform` API as users, this needs one of the following:

1. Move the check up into the entry-point validation of `BaseClassifier`, `BaseRegressor`, `BaseClusterer`, and `BaseCollectionTransformer`, for example `_fit_setup`, out of `_check_X`, so it runs once on the data the user actually passed.

2. Have composite estimators call component transformers through a private path that skips input revalidation. This would also avoid redundant conversion and metadata work on every slice, which would be a performance win for interval forests.

3. Gate the check behind a tag or constructor flag that composite estimators disable on their internal components.

Option 1 seems the smallest coherent change. Option 2 is more invasive but fixes the underlying layering problem. Whichever is chosen, the warning added in PR #3598  should then be restored to an error at the top level only.


_No response_

### Additional context

_No response_

TonyBagnall:
closes #3599

This came about because of a test for flat series triggering multiple warnings in DrCIF. 

This lead us to look at RandomIntervals and SupervisedIntervals. These extract features on interval slices of data already validated at the top-level fit. They called the features  public fit_transform/transform, paying a full _check_X (missing-value and variance scans) on every slice - thousands of times per DrCIF fit/predict.                                                                                                                                         

Had to adjust  SupervisedIntervals to stick with Numpy 3D. Then its simply a matter of using the private _fit_transform/_transform  functions which assume pre-validated inner-type input, eliminating the per-slice checks. SupervisedIntervals expands its 2D per-channel slices to the numpy3D inner type first. Non-collection transformer features keep the public path. Neither RandomIntervals and SupervisedIntervals can work with unequal length, so will always deal correctly with numpy3D.                                                                                                                                    
                                      
Output is bit-identical (DrCIF, RSTSF, SupervisedIntervals) and n_jobs invariant is preserved. 

As well as removing annoying warnings, the real pay off is the speed up. DrCIF is 12% faster fit and 18% predict.  to 1.2, supervised is about 3x faster when a transformer feature is used.  The DrCIF gain scales with the validation share: more channels or longer series means more/larger slices thus larger gain; this univariate case is a conservative point. Multivariate DrCIF would benefit more. 

Now off to look to see if this happens elsewhere


TonyBagnall:
as an extension, I looked if this slow pattern was repeated elswhere. tl;dr: there are small further gains for interval based, possible larger gains for dictionary based. 

Might be worth a follow up PR, especially for WEASEL and MUSE, but requires further investigation, I'll make two issues later.                                                                                                                                                                              
                                                                                                                                                                                                                     
  1. Interval forests — same #3599 pattern, outer call site. DrCIF, STSF, CIF, TSF, RISE, RSTSF all re-validate the same stored representation via intervals[r].fit_transform(Xt[r]) once per tree × representation  
  (base_interval_forest.py:1012/1086). At default n_estimators (200–500) × ~3 representations that's 600–1500 redundant validations of already-validated derived data. This is the cheapest win: the fast-path       
  primitives now exist, and it's one shared base class. (One subtlety: the representations are derived — e.g. first-order-difference — so unlike the original X they're first seen here; the variance/missing scan is
  still redundant against the top-level guarantee, but that's worth stating precisely.)                                                                                                                              
                                                                                                                                                                                                                     
  2. Dictionary-based — heaviest by call count, different mechanism. MUSE, BOSS/cBOSS, WEASEL, TDE re-validate full (or subsampled) data through an SFA/SFAFast per ensemble member / window / channel. MUSE is the  
  standout: 204 SFAFast validations even at a tiny config, because it fits one SFA per channel×window; on real multivariate data this is thousands. WEASEL re-validates on every window at both fit and predict.

TonyBagnall:
further exploring, number of calls

- DrCIF (and interval forests), after fix: ≈ 3·E + 2 — linear in E (n_estimators), independent of both n_cases and n_timepoints.                                                                                   
  - DrCIF, before fix: ≈ (3 + 3·I(M))·E + 2 where I(M) = intervals/representation ≈ O(√M) — the 3·I(M)·E inner term is what was removed.                                                                             
  - Dictionary (BOSS/WEASEL/MUSE): constant in n_cases, grows with n_timepoints (one SFA per window/channel, n_windows ≈ O(M)); no n_estimators knob (capped by max_ensemble_size/window grid).                      
                                                                                                                                                                                                                     
  So: constant in n_cases everywhere ,linear in n_estimators for ensembles; growing in n_timepoints for anything windowed or interval-based — and the DrCIF fix on this PR specifically collapsed the n_timepoints growth to  flat and dropped the n_estimators slope ~7×. The dictionary family still has the full n_timepoints growth, which is why MUSE/WEASEL are the strongest remaining candidates.

MatthewMiddlehurst (commented):
Looks fine, but is there a need to do it differently got both transforms? I would other go with the inline or function for both if possible.

TonyBagnall:
> Looks fine, but is there a need to do it differently got both transforms? I would other go with the inline or function for both if possible.

yes fair enough, Ive gone with functions rather than inlining and generalised to a shared helpers, _fit_feature,                      
_fit_transform_feature and _transform_feature. Each takes the 2D single-channel slice, expands it to numpy3D and routes aeon collection transformers to the private method.

One thing worth flagging, since a shared helper would otherwise have quietly decided it: the two transforms already differed in how they call a feature that isn't a collection transformer. RandomIntervals passes the expanded 3D slice and y; SupervisedIntervals passes the 2D slice and no y. That predates this PR — both did it before I touched them. I've preserved both exactly via an expand_fallback argument rather than silently unifying them, since that would be a behaviour change on a path this PR isn't otherwise about. 

Output is unchanged: my bot diffed 36 output arrays across both transforms plus DrCIF and RSTSF — univariate and multivariate, n_jobs 1 and 2, callable and transformer features, dilation, normalise_for_search=False — against the previous commit, and they're bit-identical.

MatthewMiddlehurst (approved):

PR #3667, opened 2026-07-22
[BUG] Fix iterative forecast when horizon >1

### LINKED ISSUE #3666: [BUG]  iterative_forecast is incorrect for horizon > 1
### Describe the bug

Found  #2823  trawling old issues so looked at this and found a bug:                                                                                                                                                                                                 
                                                                                                                                                                                                                                                          
iterative_forecast (the IterativeForecastingMixin default) fits once, then repeatedly calls one-step predict, appending each prediction back onto y. It never uses self.horizon, it implicitly assumes a one-step model as that is part of the contract. However, any forecaster that sets horizon > 1 and applies iterative forecast through the default method produces unintuitive results. 
       

clearest is seasonal naive, where the correct answer is unambiguous and direct is ground truth 

(y = 1..12, seasonal_period=3, last season  [10,11,12]):
  
```python
  import numpy as np
  from aeon.forecasting import NaiveForecaster
  
  y = np.arange(1, 13, dtype=float)  
  f = NaiveForecaster(strategy="seasonal_last", seasonal_period=3, horizon=2)
  print(f.direct_forecast(y, 5))     # [10. 11. 12. 10. 11.]  correct
  print(f.iterative_forecast(y, 5))  # [11. 12. 11. 12. 11.]  wrong  
```  

 

### Expected results

I would expect  [10. 11. 12]

### Actual results

[11. 12. 11. 12. 11.]

### Suggested fix

I think it doesnt make sense to call iterative on a forecaster trained on horizon > 1, so we should raise

### Versions

_No response_

TonyBagnall:
fixes #3666 

raise an exception when you call iterative with a forecaster constructed with a horizon > 1, since it doesnt really make sense to expect iterative in this case, or at least it would be offset, which is potentially very confusing behaviour

### DIFF
--- a/aeon/forecasting/base.py
+++ b/aeon/forecasting/base.py
@@ -303,6 +303,8 @@ class IterativeForecastingMixin:
         Raises
         ------
         ValueError
+            If the forecaster was configured with ``horizon`` other than 1.
+        ValueError
             If ``prediction_horizon`` is less than 1.
         ValueError
             If only one of ``exog`` and ``future_exog`` is provided.
@@ -320,6 +322,15 @@ class IterativeForecastingMixin:
         >>> f.iterative_forecast(y, 2)
         array([3., 2.])
         """
+        if self.horizon != 1:
+            raise ValueError(
+                "iterative_forecast requires the forecaster to be configured with "
+                f"horizon=1, but {self.__class__.__name__} has horizon="
+                f"{self.horizon}. Iterative forecasting recursively applies a "
+                "one-step-ahead model, feeding each prediction back as the next "
+                "observation, so a horizon greater than 1 has no meaning here. Use "
+                "direct_forecast for multi-step forecasting, or set horizon=1."
+            )
         y, exog, future_exog = self._check_iterative_forecast_inputs(
             y, prediction_horizon, exog, future_exog
         )
--- a/aeon/forecasting/tests/test_base.py
+++ b/aeon/forecasting/tests/test_base.py
@@ -245,6 +245,44 @@ def test_iterative_forecast_rejects_scalar_y():
         f.iterative_forecast(np.array(1.0), prediction_horizon=1)
 
 
+@pytest.mark.parametrize(
+    "forecaster",
+    [
+        RegressionForecaster(window=4, horizon=3),
+        NaiveForecaster(strategy="seasonal_last", seasonal_period=3, horizon=2),
+    ],
+)
+def test_iterative_forecast_requires_unit_horizon(forecaster):
+    """iterative_forecast rejects a forecaster configured with horizon > 1.
+
+    Iterative forecasting recursively feeds each prediction back as the next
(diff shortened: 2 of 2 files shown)


alexbanwell1 (approved):
## Review: [BUG] Fix iterative forecast when horizon > 1

Small, well-targeted fix. `iterative_forecast` recursively feeds each one-step prediction back as the next observation, so a forecaster built with `horizon > 1` previously produced an offset/ill-defined sequence. The new guard raising `ValueError` when `self.horizon != 1` is the right call.

### Correctness — verified
- **Guard is sound.** `self.horizon` is always set by `BaseForecaster.__init__`, and every forecaster mixing in `IterativeForecastingMixin` also extends `BaseForecaster`, so the attribute access is safe.
- **ETS/ARIMA are unaffected.** Both override `iterative_forecast` (via `_iterative_forecast_from_fitted`) and never reach this guard, so their own multi-step handling is preserved.
- **`test_iterative_matches_direct_for_seasonal_naive` is correct.** With `seasonal_period=3`, `y=[1..12]`, both paths yield `[10,11,12,10,11,12]`. Direct clones per step with `horizon=i+1` (`idx=(h-1)%3`); iterative keeps `horizon=1` (`idx=0`) but the appended predictions walk the window forward, so the phases stay in lockstep. A genuine cross-check of the `horizon=1` path, not a tautology.
- **`test_iterative_forecast_requires_unit_horizon`** correctly parametrizes two `horizon>1` forecasters.

### Regression check — clean
Grepped every `iterative_forecast` caller and `horizon=<n>` construction:
- The `horizon = 3/4/5` occurrences in the arima/dotm/ets tests are local `prediction_horizon` values, not forecaster config.
- The `horizon>1` `NaiveForecaster` instances in `test_naive.py` only call `predict`/`forecast`, never `iterative_forecast`.
- Ensemble/SCUM internal callers forward to user-supplied components; surfacing this error there is the intended new behavior.

No existing test or internal path breaks.

### Design observation (non-blocking)
There's an intentional asymmetry with the sibling `direct_forecast`, which *ignores* the configured horizon entirely (it clones and sets `f.horizon = i+1` internally). So `RegressionForecaster(window=4, horizon=3)` works with `direct_forecast` but now raises with `iterative_forecast`. Raising is defensible — silently overriding a user-set horizon would be just as confusing, and the explicit message pointing to `direct_forecast` is good UX. Worth a sentence in the changelog so users of the direct/iterative pair understand why one is lenient and the other strict.

### Minor nits (optional)
- The guard runs before `_check_iterative_forecast_inputs`, so the horizon error precedes `prediction_horizon`/`exog` validation. Reasonable — it's a construction-level misconfiguration.
- The error message is thorough and helpful but fairly long (6 lines); could trim, though I'd keep the actionable guidance.
- Docstring `Raises` section is updated correctly and the new entry's order matches the actual raise order.

Correct fix for #3666 with good, meaningful test coverage and no regressions. LGTM.

PR #3669, opened 2026-07-22
[MNT] replace REDCOMETS imblearn oversampling with aeon SMOTE/ROS

### LINKED ISSUE #3654: [MNT] Remove dependency on class-imbalance

scikit class imbalance soft dependency is only really used in RedComets for SMOTE and over sampling. We can simply replace this with our own SMOTE

aeon’s native SMOTE and ADASYN currently use imbalanced-learn only as a parity oracle in:

  - aeon/transformations/collection/imbalance/tests/test_smote.py
  - aeon/transformations/collection/imbalance/tests/test_adasyn.py
  - 
REDCOMETS currently uses:

  `from imblearn.over_sampling import RandomOverSampler, SMOTE`

  For imbalanced data it:

  1. Normalizes each series.
  2. Caps the minimum class count at six.
  3. Constructs NearestNeighbors(n_neighbors=min_class_count - 1).
  4. Passes that estimator to imbalanced-learn SMOTE.
  5. Falls back to RandomOverSampler when SMOTE raises ValueError.
  6. Uses sampling_strategy="all" to balance every class to the majority count.

  Because imbalanced-learn removes the sample itself from the neighbour matrix, the effective number of candidate neighbours is currently:

  min(minimum_class_count, 6) - 2

  This detail should be preserved unless a separate bug fix explicitly changes it.

  ## Implementation checklist

- [ ] Replace SMOTE with aeons version
- [ ] Implement simple RandomOverSampler
- [ ] deal with shape - 2D - 3D etc
- [ ] verify no change in behaviour
- [ ] replace test imports with hard coded results

_No response_

BB0813:
## Fix

Fixes #3654

### Problem

`imbalanced-learn` was only needed at runtime by `REDCOMETS` (SMOTE + RandomOverSampler). aeon already has a ported collection `SMOTE`, so the soft dependency is unnecessary for this path.

### Solution

1. Add `RandomOverSampler` to `aeon.transformations.collection.imbalance` (matches imblearn `sampling_strategy="all"`, with multivariate support).
2. Replace REDCOMETS oversampling with aeon `SMOTE` / `RandomOverSampler`.
3. Neighbour mapping preserves the previous imblearn behaviour:
   - `m = min(min_class_count, 6)`
   - `NearestNeighbors(n_neighbors=m-1)` + self-exclusion ⇒ synthesis neighbours `m-2`
   - `SMOTE(n_neighbors=m-2, distance="euclidean", random_state=...)`
4. Remove REDCOMETS `python_dependencies: imblearn` tag.
5. Catch only `ValueError` around aeon SMOTE (boundary / neighbour failures); do not swallow broader exceptions.

`imbalanced-learn` remains an optional soft dependency for SMOTE/ADASYN **parity tests** only; removing it from `pyproject.toml` can be a follow-up once those tests use hard-coded fixtures.

### Equivalence evidence (correctness only)

Resampling + REDCOMETS `variant=1`, `n_trees=25`, `random_state=0` on UCR train/test:

| dataset | resample mode | bit-exact | acc old | acc new | Δacc |
|---------|---------------|-----------|---------|---------|------|
| ItalyPowerDemand | smote | yes | 0.9417 | 0.9417 | 0 |
| GunPoint | smote | yes | 0.9667 | 0.9667 | 0 |
| Wine | smote | yes | 0.5556 | 0.5556 | 0 |
| ECG200 | smote | yes | 0.8600 | 0.8600 | 0 |
| ArrowHead | none (balanced) | yes | 0.6800 | 0.6800 | 0 |
| Beef | none | yes | 0.7333 | 0.7333 | 0 |
| Coffee | none | yes | 1.0000 | 1.0000 | 0 |
| OliveOil | smote | no | 0.9000 | 0.9000 | 0 |
| Lightning7 | smote | no | 0.7260 | 0.7260 | 0 |
| Trace | smote | no | 0.9500 | 0.9500 | 0 |

Notes:
- Binary / larger-minority cases are often bit-identical after neighbour mapping.
- Multiclass / small-minority cases (OliveOil, Lightning7, Trace) are not always bit-identical (as expected), but **test accuracy was identical** on this suite.
- An expanded 56-dataset follow-up and a multi-seed re-check of the few accuracy deltas are in the PR comments (correctness only; no speed claims).

### Changes

- `imbalance/_random_over_sampler.py` (new, multivariate-capable)
- `imbalance/__init__.py` export
- `_redcomets.py` oversampling path
- tests for ROS (incl. multivariate) + REDCOMETS oversampling / no-imblearn tag

### Testing

- unit: ROS balances classes, matches imblearn ROS when available, multivariate whole-case copy
- unit: REDCOMETS SMOTE path matches imblearn on a synthetic binary case
- unit: ROS fallback on tiny minority
- unit: REDCOMETS end-to-end fit/predict without imblearn
- UCR accuracy equivalence table above

Happy to extend the UCR suite further if useful.

### DIFF
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -18,7 +18,7 @@ from sklearn.utils import check_random_state
 from aeon.classification.base import BaseClassifier
 from aeon.transformations.collection import Normalizer
 from aeon.transformations.collection.dictionary_based import SAX, SFAFast
-from aeon.transformations.collection.imbalance import RandomOverSampler, SMOTE
+from aeon.transformations.collection.imbalance import SMOTE, RandomOverSampler
 from aeon.utils.validation import check_n_jobs
 
 
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -228,7 +228,10 @@ class REDCOMETS(BaseClassifier):
                     distance="euclidean",
                     n_jobs=self._n_jobs,
                 ).fit_transform(X_3d, y)
-            except (ValueError, IndexError):
+            except ValueError:
+                # aeon SMOTE raises ValueError when n_neighbors exceeds the
+                # minority class size; do not catch broader exceptions that
+                # could hide real bugs by silently falling back to ROS.
                 X_smote, y_smote = RandomOverSampler(
                     random_state=self.random_state
                 ).fit_transform(X_3d, y)
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -13,14 +13,13 @@ import numpy as np
 from joblib import Parallel, delayed
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.model_selection import cross_val_score
-from sklearn.neighbors import NearestNeighbors
 from sklearn.utils import check_random_state
 
 from aeon.classification.base import BaseClassifier
 from aeon.transformations.collection import Normalizer
 from aeon.transformations.collection.dictionary_based import SAX, SFAFast
+from aeon.transformations.collection.imbalance import RandomOverSampler, SMOTE
 from aeon.utils.validation import check_n_jobs
-from aeon.utils.validation._dependencies import _check_soft_dependencies
 
 
 class REDCOMETS(BaseClassifier):
@@ -92,7 +91,6 @@ class REDCOMETS(BaseClassifier):
     """
 
     _tags = {
-        "python_dependencies": "imblearn",
         "capability:multivariate": True,
         "capability:multithreading": True,
         "algorithm_type": "dictionary",
@@ -195,15 +193,6 @@ class REDCOMETS(BaseClassifier):
             List of ``(RandomForestClassifier(), weight)`` tuples fitted on `SAX`
             transformed training data
         """
-        _check_soft_dependencies(
-            "imbalanced-learn",
-            package_import_alias={"imbalanced-learn": "imblearn"},
-            severity="error",
-            obj=self,
-        )
-
-        from imblearn.over_sampling import SMOTE, RandomOverSampler
-
         X = Normalizer().fit_transform(X).squeeze(1)
 
         if self.variant in [1, 2, 3]:
@@ -221,21 +210,29 @@ class REDCOMETS(BaseClassifier):
             y_smote = y
 
         else:
+            # Cap the minority count used for neighbour selection, matching the
+            # previous imblearn path: NearestNeighbors(n_neighbors=min_c-1)
+            # with self-exclusion leaves (min_c-2) synthesis neighbours.
             if min_neighbours > 5:
                 min_neighbours = 6
+            n_neighbors = min_neighbours - 2
+            X_3d = X[:, np.newaxis, :]
             try:
+                if n_neighbors < 1:
+                    raise ValueError(
+                        "Not enough minority samples for SMOTE neighbour search"
+                    )
                 X_smote, y_smote = SMOTE(
-                    sampling_strategy="all",
-                    k_neighbors=NearestNeighbors(
-                        n_neighbors=min_neighbours - 1, n_jobs=self._n_jobs
-                    ),
+                    n_neighbors=n_neighbors,
                     random_state=self.random_state,
-                ).fit_resample(X, y)
-
-            except ValueError:
+                    distance="euclidean",
+                    n_jobs=self._n_jobs,
+                ).fit_transform(X_3d, y)
+            except (ValueError, IndexError):
                 X_smote, y_smote = RandomOverSampler(
-                    sampling_strategy="all", random_state=self.random_state
-                ).fit_resample(X, y)
+                    random_state=self.random_state
+                ).fit_transform(X_3d, y)
+            X_smote = np.squeeze(X_smote, 1)
 
         lenses = self._get_random_lenses(X_smote, n_lenses)
         sfa_lenses = lenses[: n_lenses // 2]
--- a/aeon/classification/dictionary_based/tests/test_redcomets_oversampling.py
+++ b/aeon/classification/dictionary_based/tests/test_redcomets_oversampling.py
@@ -37,7 +37,7 @@ def _redcomets_resample_aeon(X2d, y, random_state=0, n_jobs=1):
             distance="euclidean",
             n_jobs=n_jobs,
         ).fit_transform(X_3d, y)
-    except (ValueError, IndexError):
+    except ValueError:
         X_smote, y_smote = RandomOverSampler(random_state=random_state).fit_transform(
             X_3d, y
         )
--- a/aeon/classification/dictionary_based/tests/test_redcomets_oversampling.py
+++ b/aeon/classification/dictionary_based/tests/test_redcomets_oversampling.py
@@ -1,0 +1,116 @@
+"""Tests for REDCOMETS oversampling without imbalanced-learn."""
+
+from collections import Counter
+
+import numpy as np
+from sklearn.neighbors import NearestNeighbors
+from sklearn.utils import check_random_state
+
+from aeon.classification.dictionary_based import REDCOMETS
+from aeon.transformations.collection import Normalizer
+from aeon.transformations.collection.imbalance import RandomOverSampler, SMOTE
+from aeon.utils.validation._dependencies import _check_soft_dependencies
+
+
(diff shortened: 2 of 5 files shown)


BB0813:
### Expanded UCR equivalence follow-up

I re-ran a broader old-vs-new comparison after the initial 10-dataset check.

#### Protocol
- Univariate UCR train/test
- `REDCOMETS(variant=1, n_trees=25, random_state=0, n_jobs=1)` for the full suite
- Old path = previous imblearn SMOTE/ROS (monkeypatched into the same REDCOMETS code)
- New path = aeon `SMOTE` / `RandomOverSampler` (this PR)
- Neighbour mapping unchanged: `m = min(min_class, 6)`, synthesis neighbours `m-2`

#### Full suite (`n_trees=25`, seed `0`)
- **56** univariate datasets completed (train size ≤ 800)
- Class counts after resampling: **56/56 match**
- Bit-exact resampled matrices: **39/56**
- Identical accuracy: **46/56**
- Accuracy changed: **10/56**
- Mean Δacc (`new - old`): **≈ -0.0006**
- Max |Δacc| on this pass: **0.0519** (`MiddlePhalanxTW`)
- Mean end-to-end fit speedup (old/new): **≈ 1.12×**

Datasets with accuracy change on this single-seed pass (all non-bit-exact multiclass SMOTE cases):

| dataset | Δacc | old | new |
|---------|------|-----|-----|
| MiddlePhalanxTW | -0.0519 | 0.5779 | 0.5260 |
| OSULeaf | +0.0248 | 0.5785 | 0.6033 |
| CBF | +0.0200 | 0.9200 | 0.9400 |
| Adiac | +0.0179 | 0.7289 | 0.7468 |
| ProximalPhalanxTW | -0.0146 | 0.7805 | 0.7659 |
| MedicalImages | -0.0132 | 0.7395 | 0.7263 |
| Fish | -0.0114 | 0.8171 | 0.8057 |
| DistalPhalanxOutlineAgeGroup | -0.0072 | 0.7122 | 0.7050 |
| FacesUCR | +0.0049 | 0.8366 | 0.8415 |
| SwedishLeaf | -0.0032 | 0.9088 | 0.9056 |

Binary / bit-exact cases remained identical.

#### Re-check of the 10 changed datasets (`n_trees=100`, seeds `0/1/2`)
Averaging over 3 seeds reduces the apparent gaps substantially:

| dataset | mean Δacc (3 seeds) |
|---------|---------------------|
| MiddlePhalanxTW | **+0.0000** |
| OSULeaf | +0.0110 |
| CBF | -0.0007 |
| Adiac | +0.0068 |
| ProximalPhalanxTW | -0.0033 |
| MedicalImages | -0.0053 |
| Fish | -0.0095 |
| DistalPhalanxOutlineAgeGroup | **+0.0000** |
| FacesUCR | -0.0011 |
| SwedishLeaf | -0.0005 |

Across these 30 runs:
- overall mean Δacc ≈ **-0.0003**
- max single-run |Δacc| ≈ **0.045** (MiddlePhalanxTW, seed 0 only; other seeds positive)
- ≈ **77%** of runs have |Δacc| ≤ 0.01
- ≈ **93%** of runs have |Δacc| ≤ 0.02

#### Interpretation
- No stable systematic degradation from the aeon oversampling path.
- Remaining differences are concentrated in non-bit-exact multiclass SMOTE cases and look seed/noise sensitive once more trees are used.
- Happy to add a small Wilcoxon / average-rank summary, or expand further, if that would help review.

Also pushed an isort-only fix for import order (`SMOTE, RandomOverSampler`) after the pre-commit failure.

TonyBagnall (changes requested):
Thanks for this. I have reviewed the change and reproduced the key claims locally. The mechanism is sound and the neighbour mapping is correct: aeon `SMOTE` adds one internally (`_SingleClassKNN(n_neighbors=self.n_neighbors + 1)`) then drops the self column, so `n_neighbors = min_c - 2` reconstructs the same `min_c - 2` synthesis neighbours as the old imblearn estimator path. Resampled output is bit-exact (1e-10) on the binary and ROS-fallback cases, labels and shapes match, multiclass is expectedly non-bit-exact with unchanged accuracy, and the new tests pass.

Three changes needed before I approve:

**1. `RandomOverSampler` should support multivariate input.** It samples complete cases so it is already channel-agnostic, but without `"capability:multivariate": True` the base class rejects 3D multi-channel input:

```
ValueError: Data seen by instance of RandomOverSampler has multivariate series, but RandomOverSampler cannot handle these characteristics.
```

Adding the tag, it balances correctly and preserves all channels (duplicated cases are exact whole-case copies). Please add the tag and a multivariate test. Note this does not apply to `SMOTE`, which squeezes the channel axis and interpolates on the collapsed series, so it is correctly univariate-only.

**2. Do not catch `IndexError` around the SMOTE call.** 
This is actually useful, looking at this PR has highlighted an area of improvement in our imbalance module, so hurrah. 
My bot did this: I swept every minority size this path can reach (min_c 3 to 10; the `n_neighbors < 1` guard already routes min_c <= 2 to the fallback) and aeon SMOTE never raises. Provoking the boundary failure raises `ValueError` (`Expected n_neighbors <= n_samples_fit`), not `IndexError`. So there is no expected path that raises `IndexError`, and catching it can only swallow a genuine internal bug and silently degrade to random oversampling. Please narrow it to `except ValueError`. If a specific case does raise `IndexError`, name it and catch it narrowly at source with a comment.

**3. The speed table does not reproduce, and I think it is measuring the wrong thing.** Timing the genuine path REDCOMETS runs, `SMOTE(...).fit_transform` on 3D input, warm (numba precompiled, median of 25), I get close to the opposite of the table:

| case | imblearn | aeon | old/new |
| --- | --- | --- | --- |
| binary min 20, len 60 | 1.10 ms | 2.19 ms | 0.50x |
| binary min 8, len 24 | 0.98 ms | 2.00 ms | 0.49x |
| multiclass, len 200 | 1.81 ms | 3.86 ms | 0.47x |
| tiny min 2 (ROS fallback) | 1.55 ms | 0.85 ms | 1.83x |

The aeon SMOTE path is roughly 1.5-2x slower across sizes, not 2-4x faster. The only path that wins is the ROS fallback. I suspect the table was timed cold (the first aeon call pays a few seconds of numba JIT) or against a stripped kernel rather than the transformer. Even with the wrapper removed, aeon's core only overtakes sklearn on large series, and there the transformer overhead eats the win.

This is not a blocker for the mechanism, since resampling is a trivial fraction of a 25-tree fit, but an unsubstantiated 2-4x claim should not sit in the PR record. Please either:

1. commit a benchmark script that times the real `fit_transform` path warm and reproduces the numbers, or
2. drop the speed column and keep the equivalence table as correctness-only.

I have a self-contained script that times the genuine path warm and checks equivalence in the same run, happy to add it or you can adapt it. Run with no arguments for the synthetic suite; imblearn is optional so it does not reintroduce the dependency.

One scope question: `imbalanced-learn` is still in `pyproject.toml` because the SMOTE/ADASYN parity tests import it, so the dependency is not actually removed yet, only removed from the REDCOMETS runtime path. Do you want the pyproject removal and the parity-test fixtures done here, or tracked as the follow-up you mention? The issue title is "remove the dependency", so worth being explicit about what actually closes #3654.

BB0813:
@TonyBagnall Thanks for the careful review — all three points make sense.

**1. Multivariate `RandomOverSampler`**
Added `"capability:multivariate": True` and a multi-channel test that checks whole-case duplication preserves all channels.

**2. `except ValueError` only**
Dropped `IndexError` from the SMOTE fallback. The expected boundary failure is `ValueError` (`Expected n_neighbors <= n_samples_fit`); broader catching would risk hiding real bugs behind silent ROS fallback.

**3. Speed claims**
Agreed — the earlier 2–4× figure was not timing the real warm `fit_transform` path. I've dropped the speed claim from the narrative. Correctness/equivalence evidence stands; resampling is a small fraction of a 25-tree fit either way.

**Scope of #3654**
This PR removes imblearn from the **REDCOMETS runtime path** only. `imbalanced-learn` remains in `pyproject.toml` for the SMOTE/ADASYN parity tests. I'd prefer to track hard-coded fixtures + full soft-dep removal as a follow-up so this PR stays focused on the REDCOMETS swap + equivalence evidence. Happy to do that follow-up next if you'd like.

TonyBagnall (changes requested):
fixes look good, but could you remove the speed up claims on the text of the PR in the table? I dont think they stand up, but have been helpful in alerting us to a possible issue with SMOTE

BB0813:
@TonyBagnall Done — removed the speedup column and any remaining speed claims from the PR description. The equivalence table is correctness-only now. Thanks again for the catch.

TonyBagnall (approved):
lgtm, thanks for this

PR #3670, opened 2026-07-23
[ENH] Add PrevalidatedRidgeClassifier

### LINKED ISSUE #3419: [ENH] Implement Prevalidated Ridge Regression classifier for high-dimensional feature transforms

note this issue has been assigned, will formally assign when possible.

Implement a PrevalidatedRidgeClassifier in aeon, based on:

Dempster, Webb and Schmidt, "Prevalidated Ridge Regression is a Highly-Efficient Drop-In Replacement for Logistic Regression for High-Dimensional Data", arXiv:2401.15610.

Upstream reference implementation:

https://github.com/angus924/preval

aeon TSC transformers often produce dense high-dimensional feature matrices which are used to train linear classifiers. Fitting and tuning logistic regression can be expensive in high-dimensional settings and it is designed to produce predictions rathe than probability estimates.

Prevalidated Ridge Regression, PreVal, aims to provide logistic-regression-like classification accuracy and probabilistic predictions while being substantially faster to train. This makes it a good candidate as a classifier head for high-dimensional time series feature transforms.

The upstream repository is GPL-3.0, while aeon is BSD-3-Clause, but we have the permission from the authors to adapt the code for the aeon interface. 

Classifier, name to be agreed, possibly

PrevalidatedRidgeClassifier

Location:

aeon.classification.sklearn

further discussion below on details and interface. 

_No response_

### Additional context

_No response_

adam-goose:
#### Reference Issues/PRs

Closes #3419 

This PR adds `PrevalidatedRidgeClassifier` to
`aeon.classification.sklearn`. The classifier is based on Dempster, Webb and Schmidt,
*Prevalidated Ridge Regression is a Highly-Efficient Drop-In Replacement for Logistic Regression for High-Dimensional Data*
([arXiv:2401.15610](https://arxiv.org/abs/2401.15610)), and adapts the authors' [reference implementation](https://github.com/angus924/preval) to aeon's estimator interface.

- supports binary and multiclass targets, including non-numeric labels;
- selects a ridge penalty from a configurable positive lambda grid using
  efficient leave-one-out predictions;
- calibrates the ridge predictions with a fitted scale factor and returns class
  probabilities through `predict_proba`;
- uses the smaller of the feature-space and case-space eigendecompositions for
  `n >= p` and `n < p` respectively; and
- removes constant and very low-variance feature columns before fitting.

The classifier exposes the usual aeon fitted state and prediction interface,
including learned classes, selected lambda, scale, coefficients, intercept,
and probability estimates. It has also been added to the generated
classification API reference.

##### Verification

The implementation was checked directly against the standalone reference using
identical copied inputs and lambda grids. Exact equality was obtained for the
selected lambda, scale, coefficients, intercept, predicted probabilities, and
predicted labels in all cases:

- **12 deterministic synthetic cases**, covering multiple seeds, binary and
  multiclass targets, integer and string labels, `n >= p` and `n < p`, varied
  dataset dimensions and lambda grids, constant/duplicated/near-constant
  columns, and correlated or mildly ill-conditioned features;
- **6 bundled real datasets** using official train/test splits: GunPoint,
  ItalyPowerDemand, ArrowHead, BasicMotions, OSULeaf, and ACSF1.
  These cover binary through 10-class problems, 24 to 1,460 transformed
  features, and both the n >= p and n < p computation paths.

The aeon unit tests cover binary and multiclass estimator lifecycles, string and
integer labels, both shape branches, low-variance feature removal, fitted
attribute dimensions, probability validity, prediction/probability agreement,
and invalid lambda grids. All **7 focused unit tests** pass, as do aeon's
centralized estimator checks.

A repeated fit-time comparison used warm-up runs, alternating execution order,
and seven measured repetitions on four synthetic and the six real datasets.
The median aeon/reference runtime ratio across the ten cases was **1.064**
(**+6.4%**). The largest median slowdown was **1.084x (+8.4%)**. This indicates
a small, consistent interface overhead with no substantial computational
regression relative to the reference implementation in the tested cases.

No. The implementation uses `scipy.optimize.minimize` and `sklearn.preprocessing.LabelBinarizer`;
SciPy and scikit-learn are  lready core aeon dependencies.

#### Any other comments?

The detailed correctness verification and runtime scripts are maintained separately from the aeon PR
so that the upstream contribution remains focused on the estimator, its unit tests, and API documentation.

##### For all contributions


TonyBagnall:
thanks for the PR @adam-goose, some comments. We have approval from the author for the port, but we should acknowledge it in the code, especially as they are BSD-3. Add a citation, link to the original code and give a permission note
Mirror what MiniRocket already does:

adam-goose:
@TonyBagnall yes, of course. sorry, missed that bit. I've added it now.
Let me know if there's any further changes required.

TonyBagnall:
@adam-goose sorry for the delay, it seems a faithful implementation, but the public params coef_ and intercept_ are a bit confusing, since coef_[0] is the coefficient for the artificial column of ones, so the effective intercept is intercept_ + coef_[0], and low-variance features are also omitted from coef_.

you could Could we either retain the reference representation under more explicit fitted attribute names, e.g. B_/B0_, or expose conventional coef_ and intercept_ derived from the reference representation? I would prefer the latter if we want this to behave like the other sklearn-style linear classifiers.

```self.intercept_ = B0 + B[0]``
so rathr than
```coef_ =
[[ 0.2,  0.4],
 [ 1.1, -0.7],
 [ 0.6,  0.3]]

intercept_ = [-0.5, 0.1]
```
you have 
```coef_ =
[[ 1.1, 0.0,  0.6],
 [-0.7, 0.0,  0.3]]

intercept_ =
[-0.3, 0.5]
```

TonyBagnall:
I would like a hard coded test against the reference implementation in the tests. Give commented out code on how to generate example output, hard code the output and test aeon implementation reproduces it

TonyBagnall:
also please expand the public docstring. The paper/reference implementation expects appropriately centred/standardised features, and the upstream example uses StandardScaler. I dont think we should hard code it but we should document, preferably with a short example. Please also document the fitted attributes in an Attributes section.

TonyBagnall (changes requested):
see three comments for requested changes

TonyBagnall (changes requested):
Looks good now, a faithful reproduction. 

One small aeon/style point: self.c is fitted state without a trailing underscore, and c is also not a very informative name. Since this is the fitted calibration scale, could you use a local variable such as calibration_scale during fit and then assign it to self.scale\_? Or just use self.calibration_scale_, dont mind. 
There doesn't seem to be any need to retain self.c.

adam-goose:
Yes, good point. I just used c because that's what the paper called it.
Updated the calibration scale to use a local calibration_scale variable during fitting before assigning to self.scale_. Also cleaned up the fitted LabelBinarizer to use self.label_binarizer_ consistently rather than retaining the duplicate private _lb attribute.

TonyBagnall (approved):
great, thanks for this

PR #3680, opened 2026-07-24
[MNT]: change param

### LINKED ISSUE #3678: [MNT] Change the parameter `znormalized` to `znormalize` in SAX transformation
### Describe the issue

### Additional context

- update tests

unfettered-one:
#### Reference Issues/PRs

Fixes #3678 

 - Just the param change name from znormalized to znormalize
 - Tests and documentation(doc-strings) changed accordingly

No

##### For all contributions


patrickzib (commented):
I guess this falls under the deprecation policy, as at least one classifier is affected?

https://www.aeon-toolkit.org/en/latest/developer_guide/deprecation.html

TonyBagnall:
thanks for the pr, but as per the issue, this should be done after #3649 and requires deprecation, please read our deprecation policy. Also could you make the prefix of the title [MNT] not MNT? thanks

unfettered-one:
> thanks for the pr, but as per the issue, this should be done after #3649 and requires deprecation, please read our deprecation policy. Also could you make the prefix of the title [MNT] not MNT? thanks

Sure , will do the deprecation changes and ping here (title already changed)

unfettered-one:
@TonyBagnall , @patrickzib , I made the deprecation changes, but the ETS tests are failing on assert, Is this supposed ti happen, that if I add any other imports the tests will fail?

TonyBagnall:
@hadifawaz1999 any thoughts? I think you created the issue?

hadifawaz1999 (changes requested):
Thanks for this. However, the idea is not simply to replace the name of the parameter but also its functionality, the old parameter is called `znormalized` which assumes the input is already znormalzied when its value is True, and not when its value is False. Meaning with the old parameter, when the value is False we normalize and when True nothing happens. The idea is that the new parameter name `normalize` is an order, meaning if True, normalize, and if False do nothing. The PR should take into consideration these changes

unfettered-one:
> Thanks for this. However, the idea is not simply to replace the name of the parameter but also its functionality, the old parameter is called `znormalized` which assumes the input is already znormalzied when its value is True, and not when its value is False. Meaning with the old parameter, when the value is False we normalize and when True nothing happens. The idea is that the new parameter name `normalize` is an order, meaning if True, normalize, and if False do nothing. The PR should take into consideration these changes

Okay, Thanks for the explanation, Will do the respective changes.

hadifawaz1999:
Can you change the PR title to something more detailed ? thanks

PR #3684, opened 2026-07-25
Rename TDE/IndividualTDE parameter 'max_dims' to 'max_channels' (#3418)

SpiliosDimakopoulos:


TonyBagnall:
thanks, but I assigned this and its also part of #3625

PR #3685, opened 2026-07-25
Fix checkpoint fallback 3595

SpiliosDimakopoulos:


TonyBagnall:
thanks for this, please start the PR title with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV] as per template instructions and actually describe your PR

PR #3686, opened 2026-07-25
[DOC] finish remaining docstring clean-up in unequal_length (#3496)

### LINKED ISSUE #3496: [DOC] Typos in unequal length package

TonyBagnall:
closes #3496, which the earlier doc PRs (#3528, #3536, #3573) left open and the bots now ignore. If you want bot solutions, make smaller issues than this, god knows why but PRs went up that did one or two of these. Its all part of the learning curve, I wont accept partial solutions any more. 

- add docstrings to the _get_min_length/_get_max_length helpers
- add a _fit docstring to Resizer, matching Padder and Truncator, and align the three _fit summaries
- reword the Resizer class summary to the same form as the other two
- document that error_on_long=False (Padder) and error_on_short=False (Truncator) leave out-of-range series unchanged, so min/max respectively can return an unequal length list
- drop the stray double space in the padded_length/resized_length parameter names


baraline (approved):
LGTM, and yes, I think likewise. Multiplying PRs to close one issue makes it really hard to track progress. Either make the issues smaller or don't accept PRs that don't fix the issue.

PR #3690, opened 2026-07-25
[DEP] deprecate unused dataset loaders (#2835)

### LINKED ISSUE #2835: [ENH] Revise datasets in datasets/data

the size of the package is growing, and a lot of this is in the shipped datasets, a lot of which are there for legacy reasons.

https://github.com/aeon-toolkit/aeon/issues/2224

We should go through and consider if we can ditch or replace any.

Collections:
ACSF1, ArrowHead, BasicMotions., CardanoSentiment, Covid3Months, GunPoint, ItalyPowerDemand, JapaneseVowels, OSULeaf, PLAID, UnitTest

Series for forecasting:
Airline, Longley, Lunx, m1_yearly_dataset, PBS_dataset, solar, Uschange

Others
Daphnet_SO6RO2EO
KDD_TSAD_135

_No response_

### Additional context

_No response_

TonyBagnall:
Closes #2835

1. Deprecate forecasting datset loaders  not used elsewhere. load_lynx, load_shampoo_sales and load_PBS_dataset have no uses anywhere in aeon outside their own loader, their entry in the API reference and a test of the loader itself. They are legacy series inherited from the bad place and are carried only so the loaders have something to load. 
2. Remove arff files not touched anywhere. We removed the arff loader, may as well remove the data
3. Deprecate OSULeaf thats equal len

Deprecated in v1.6.0 for removal in v1.7.0, following the PLAID precedent. The data files stay until removal so the loaders keep working during the deprecation window; a parameterised test asserts the FutureWarning fires and the series still loads.

### Forecasting datasets

This is a dead-API question, and the clearer win. All eight datasets together are only **59 KB**, so this is not about disk usage.

The table below shows usage within the package, excluding `aeon/datasets` itself. We could also remove `Longley`, `solar`, and `Uschange`, but this will do for now.

| Dataset | In-package uses |
|---|---|
| `Airline` | Approximately 35 uses: tests for Box-Cox, log, Yeo-Johnson, N-BEATS, TCN, DeepAR and plotting; doctests in ARIMA, ETS, ACF, STRAY and pipelines |
| `m1_yearly_dataset` | 3 uses: fixture for the `.tsf` reader |
| `Longley` | 1 doctest in `_pca.py`; 2 further uses are commented out |
| `Uschange` | 1 doctest in `_dobin.py` |
| `Lynx` | 0 |
| `ShampooSales` | 0 |
| `PBS_dataset` | 0 |
| `solar` | 0 in-package; used by `smoothing_filters.ipynb` |


on other datasets, deprecating Daphnet for these claude generated reasons, hard to argue:

that ships in aeon is one file, Daphnet_S06R02E0.csv (449 KB) — subject 6, run 2. It reaches aeon the same way ecg-diff-count-3 does: via the TimeEval benchmark suite (#1588), in TimeEval's CSV format.                                                 
                                                                                                                                                                                                                                                                                                                                                                                          
● What it is: 7,040 timepoints × 9 channels — tri-axial accelerometers at ankle, leg and trunk (ankle_horiz_fwd, ankle_vert, … trunk_horiz_lateral), sampled at ~256 Hz, so roughly 27 seconds of walking. Source 
  is the UCI Daphnet Freezing of Gait repository (Roggen, Plotnik & Hausdorff, 2013), routed through TimeEval like ecg-diff-count-3. The intended task is unsupervised anomaly detection, with freeze-of-gait     
  episodes as the anomalies.                                                                                                                                                                                      
                                                                                                                                                                                                                  
  Two problems:                                                                                                                                                 
                                                                                                                                                                                                                  
  1. The shipped segment contains no anomalies. Every one of the 7,040 labels is 0:                                                                                                                               
                                                                                                                                                                                                                  
So aeon ships an anomaly-detection example dataset with zero anomalies in it. Any detector evaluated on it scores against an all-negative ground truth — precision, recall and AUC are undefined or degenerate. As a demonstration or benchmark fixture it can't do the job its docstring describes.                                                                                                                                                   
  2. The docstring's shape is wrong. Returns claims (28800, 9) and (28800,); the actual return is (7040, 9) and (7040,). The Notes section correctly says "Series length: 7040", so Returns contradicts it. That's the identical bug I found in load_ecg_diff_count_3, which claims (10000,2) multivariate but returns (10000,) univariate — both introduced together in #1588, so it looks like copy-paste from the full upstream datasets rather than the excerpts actually shipped.                                                                                                                                                             
                                                                                                                                                                                                                  
Where that leaves it: Daphnet is 449 KB, has zero references outside its own test and the API reference, and doesn't work for its stated purpose. That's a stronger case for removal than anything else in the  audit — it's not just unused, it's misleading. The alternative, if the Daphnet data is worth keeping at all, is swapping S06R02E0 for a run that actually contains freeze episodes, which changes the shipped file rather than deleting it.  

 Acts on the audit in #2835. 
                                                                                                                                                                                                                  
  - 12ebfd954 — deprecate load_lynx, load_shampoo_sales, load_PBS_dataset (v1.6.0, removal v1.7.0)                                                                                                                
  - 819aca52f — delete four orphaned .arff duplicates (993 KB; no public API, so no deprecation cycle)                                                                                                            
  - 296c2e599 — deprecate Daphnet, fix its Returns shapes, record that the excerpt is unlabelled                                                                                                                  
  - 0e53cc36d — deprecate ecg-diff-count-3, fix its Returns shape


then on classification, deprecating OSULeaf and AFCS1 because they represent 50% of the size of the shipped datasets and are not used


MatthewMiddlehurst (approved):
Fine with this. Agree think we can clean further. Should have a think about what each of the remaining ones adds.

PR #3701, opened 2026-08-01
[ENH] Add SSH Index to similarity search.

baraline:

- This adds a second approximate index to `similarity_search.whole_series`, alongside `SimHashIndexANN`, and factors out what the two indexes now share.

Sketch, Shingle & Hash (Luo & Shrivastava, arXiv:1610.07328), is an index whose bucket collisions correlate with DTW similarity rather than with cosine similarity. What it actually hashes is the weighted Jaccard similarity of shingle multisets, which is invariant to *where* a pattern occurs in the series, so unlike SimHash it is not defeated by a time shift. 

- **`rerank_distance` on `SimHashIndexANN`**

`SimHashIndexANN` gets an optional `rerank_distance`. `None` keeps the existing collision-count ranking and its `1 / collision_count` proxy distances; setting it scores the candidates (and only the candidates) with a real distance and returns true distances, which also resolves the collision-count ties.

This was needed to benchmark the underlying buckets created by the hash with/out reranking, as SSH always rerank.

- **Shared bucket tables (`_commons.py`)**

Both indexes now build their tables through `_build_hash_tables` and probe them through `_tally_bucket_collisions`. The tables are held in a `HashTables` NamedTuple using a compressed sparse row layout (`table_offsets`, `bucket_keys`, `bucket_starts`, `case_indices`) instead of a list of dicts. With selective buckets there is nearly one bucket per case per table, so the old layout created `n_cases * n_tables` Python objects at fit time, which dominated fit time and cost far more memory than the indices themselves. `_bucket_dicts` materializes the old `{bucket key: case indices}` view for inspection.

- **Fix: `normalize` is now frozen at fit time**

`NaiveSeriesSearch` and `SimHashIndexANN` store `X_` on the scale `normalize` selects, so honouring a later change of the flag compared a raw query against a normalized collection: the self-match of a fitted series landed at a large distance instead of zero. Both estimators now freeze the flag as `self._normalize` at fit time and read it from there, with a regression test.

Also `BaseSimilaritySearch.fit` now runs `_validate_fit_params` *before* storing `X_`, so a failed fit does not leave the whole collection referenced by an estimator that never fitted. The base class docstring now documents `X_`, `n_cases_`, `n_channels_` and `n_timepoints_`, including the rule that `X_` is always on the scale its distances are computed on.

- **Documentation**

- New `examples/similarity_search/comparing_indexes.ipynb`: what the two indexes hash, why the hash only chooses the candidates once both re-rank, an aligned-query regime where SimHash finds the better candidates, a misaligned-query regime wher only the shingled index survives, and a short guide on which one to use.
- `code_speed.ipynb` gains SSH sections on trading recall for query speed with `n_tables` / `n_hashes_per_table`, and on tuning the sketch with `window_length`, `shift` and `shingle_size`.
- `simhash_index.ipynb` updated for the new `tables_` layout.

None.

##### For all contributions

- [X] I've added the estimator/function to the online [API documentation](https://www.aeon-toolkit.org/en/latest/api_reference.html).


baraline:
Todo for after merging and the recent bug issue are fixed :

Raise an issue to add the UCR-Suite branch-and-bound DTW pruning inside the candidate set reranking. We are missing it compared to the paper.

TonyBagnall:
not sure where this conflict came from @baraline ?

baraline:
@TonyBagnall It was from #3812 causing conflicts on the exemple notebook

TonyBagnall:
hi @baraline this is looking good, Ive had a look through myself but its a lot to get your head around. The only thing claude could find was this, one thing to fix and the rest up to you, Im happy to approve this if you fix the one thing, others optional

What I ran: fetched pull/3701/head and diffed three-dot against fresh main (19 files, +4281/−479). Ran aeon/similarity_search (347 passed, 4 failed, see below) and check_estimator on SSH, SimHash and Naive (25/25/17 passed).

**Verified**

1. The SSH pipeline matches a pure-Python oracle bit for bit. I reimplemented sketch → shingle → occurrence-ranked MinHash → key folding from the docstring's description, with arbitrary-precision integers and a Counter instead of the open-addressing table. Keys were identical across four configurations, including shingle_size=64 (the 64-bit packing edge), shingle_size=1, window_length equal to the series length, and shift larger than window_length.

2. The weighted-Jaccard claim holds. With 2000 single-hash tables, the per-pair collision rate matched the weighted Jaccard of the shingle multisets (computed independently with Counter) within 0.017, against a binomial sd of about 0.011 across 15 pairs.

3. The parallel kernel is split-invariant. Keys were identical for every n_tasks from 1 to n_cases, for 1, 2, 7, 17 and 40 cases. This tests the stamp-and-reuse trick in the occurrence table directly.

4. Candidate gathering and re-ranking are exact. Over 20 queries, the candidate set was exactly the set of cases sharing a key in some table, collision counts were correct, the returned distances equal dtw_distance with the same distance_params, and results were sorted.

5. The SimHash refactor is behaviour-preserving. Against main, 75 of 75 queries across three configurations (default, normalize=False, uniform with 40×12) gave identical indexes and distances.
6. The normalize freeze fix works. After set_params(normalize=False) on a fitted estimator, the self-match is at distance 0 for all three estimators.
7. Other edge cases are handled. The sketch gives identical keys for float32 and float64 input despite fastmath=True; there are no NaN guards for fastmath to elide, and NaN input is rejected at fit. Parameter validation is in _validate_fit_params, not __init__.

**Should fix before merge**

test_hash_collection_is_thread_count_invariant fails whenever numba has only one thread. It calls set_num_threads(max(2, ...)), which raises ValueError on a single-core machine or with NUMBA_NUM_THREADS=1. It passes on CI but will fail for contributors running locally in those setups. Since n_tasks is already a kernel argument, the test can call _hash_collection_kernel with each n_tasks from 1 to n_cases directly. That tests the invariant on any machine, without touching thread settings.

**Non-blocking**

1. The notebook's conclusion is stronger than its own evidence. It says SSH "is the only one that still retrieves anything once the queries are displaced". But the notebook's own re-rank plot shows SSH offers the source while DTW then doesn't return it in the top k. My runs agree with the plot, not the sentence:
2. Rolled random walks at matched budgets: SimHash with DTW re-ranking had higher recall (0.74 vs 0.41 at about 50 candidates).
3. Planted motifs at random offsets, at about 40 candidates: recall against exact DTW tied at 0.34, and SSH was ahead only on same-motif precision (0.61 vs 0.50).
Something like "is the only one whose candidate set still contains the source" would be accurate.
4. Empty results. With selective settings, a query can gather zero candidates and predict returns empty arrays. This is documented ("n_found may be smaller than k"), but a UserWarning when nothing collides would help users tuning n_hashes_per_table.
5. Stray spec files. Deleting docs/superpowers/specs/*.md is right, since they look like agent planning files that got merged by mistake. Adding docs/superpowers/ to .gitignore would stop it recurring.

baraline:
Thanks, will take a look at these

PR #3704, opened 2026-08-02
[BUG] Fix PeriodogramTransformer crash on unequal-length input for non-constant pad_with

Hrafz:
## Title
[BUG] Fix `PeriodogramTransformer` crash on unequal-length input for non-`"constant"` `pad_with`

`PeriodogramTransformer` declares `"capability:unequal_length": True` and its
docstring documents `pad_with` as any `numpy.pad` mode, but the unequal-length
code path crashes for every mode except the default `"constant"`.

```python
import numpy as np
from aeon.transformations.collection import PeriodogramTransformer

X = [np.random.rand(2, 20), np.random.rand(2, 33)]
PeriodogramTransformer(pad_series=True, pad_with="mean").fit_transform(X)
```

Expected: a list of two arrays, shapes `(2, 16)` and `(2, 32)`.

Actual:
```
ValueError: unsupported keyword arguments for mode 'mean': {'constant_values'}
```

The unequal-length branch of `_transform` always passes `constant_values` to
`np.pad`, regardless of `mode`; the equal-length (`numpy3D`) branch already
guards this correctly with the same series, so the fix mirrors that branch's
kwargs construction in the unequal-length one.

`pad_with="mean"` is the only non-default value used anywhere in aeon itself
(`RandomIntervalSpectralEnsembleClassifier`/`Regressor` hard-code it), so it's
a realistic value for a user to copy — though RISE itself doesn't declare
`capability:unequal_length` and can't reach this branch today.

## Tests

`aeon/transformations/collection/tests/test_periodogram.py`, parametrized
over every `numpy.pad` mode: checks output shape and that the unequal-length
path agrees with the equal-length path on the same data under the same
`pad_with`.

Before fix:
```
9 failed, 1 passed in 1.34s
```
(every mode but `"constant"` raised the `ValueError` above)

After fix:
```
10 passed in 1.08s
```

Full `aeon/transformations/collection/tests/` suite after the fix: `80
passed, 1 skipped` (skip is a pre-existing optional-dependency skip,
unrelated to this change) — no regressions versus the 71-passing baseline
plus these 10 new tests.

No open issue or PR covers this; happy to open an issue first if preferred.
No behavior change to the already-working `"constant"` path.

### DIFF
--- a/aeon/transformations/collection/_periodogram.py
+++ b/aeon/transformations/collection/_periodogram.py
@@ -86,6 +86,9 @@ class PeriodogramTransformer(BaseCollectionTransformer):
             Xt = np.abs(np.fft.fft(X)[:, :, : int(X.shape[2] / 2)])
         else:
             Xt = []
+            kwargs = {"mode": self.pad_with}
+            if self.pad_with == "constant":
+                kwargs["constant_values"] = self.constant_value
             for x in X:
                 if self.pad_series:
                     len = int(
@@ -100,8 +103,7 @@ class PeriodogramTransformer(BaseCollectionTransformer):
                                 len,
                             ),
                         ),
-                        mode=self.pad_with,
-                        constant_values=self.constant_value,
+                        **kwargs,
                     )
                 Xt.append(np.abs(np.fft.fft(x)[:, : int(x.shape[1] / 2)]))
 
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
nice, thanks for this, all looks good to me

TonyBagnall:
@all-contributors add @Hrafz for bug

PR #3707, opened 2026-08-02
[BUG] Znorm utils do not preserve float precision

### LINKED ISSUE #3706: [BUG] z-normalisation numba utilities silently upcast float32 input to float64
### Describe the bug

The z-normalisation helpers in `aeon/utils/numba/general.py` allocate their output with `np.zeros(X.shape)`, which always returns a float64 array regardless of theinput dtype. A float32 collection is therefore silently upcast to float64, doubling its memory.

Four functions are affected:

- `z_normalise_series_2d`
- `z_normalise_series_2d_with_mean_std`
- `z_normalise_series_3d`
- `normalise_subsequences`

`NaiveSeriesSearch(normalize=True)` stores the result, so the upcast is permanent for the lifetime of the fitted estimator: `X_` holds twice the memory of the collection the caller passed in. Other callers allocate a float64 copy transiently rather than storing it, so they pay the memory during `fit`/`predict` instead: 

`SimHashIndexANN(normalize=True)`, `KShape(z_normalise=True)`, `RandomDilatedShapeletTransform` (via `normalise_subsequences`), `SupervisedIntervals(normalise_for_search=True)` and `NaiveSubsequenceSearch`.

The values themselves are correct, so this is a memory and dtype-contract issue rather than a numerical one.

```python
import numpy as np

from aeon.similarity_search.whole_series import NaiveSeriesSearch
from aeon.utils.numba.general import z_normalise_series_3d

X = np.random.default_rng(0).standard_normal((100, 3, 500)).astype(np.float32)
print("input                :", X.dtype, X.nbytes, "bytes")

# The shared numba utility drops the caller's precision.
print("z_normalise_series_3d:", z_normalise_series_3d(X).dtype)

# It is stored, so the fitted estimator holds twice the memory it needs.
est = NaiveSeriesSearch(normalize=True).fit(X)
print("NaiveSeriesSearch.X_ :", est.X_.dtype, est.X_.nbytes, "bytes")

assert est.X_.dtype == X.dtype, f"expected {X.dtype}, got {est.X_.dtype}"
```

### Expected results

A float32 collection stays float32 through normalisation, so the fitted estimato holds no more memory than the array the caller passed in, and the assert passes:

input                : float32 600000 bytes
z_normalise_series_3d: float32
NaiveSeriesSearch.X_ : float32 600000 bytes

Integer input should keep promoting to float64, as it does today.

### Actual results

The output is float64 and the stored collection is twice the size of the input:

```
input                : float32 600000 bytes
z_normalise_series_3d: float64
NaiveSeriesSearch.X_ : float64 1200000 bytes
  File "repro.py", line 16, in <module>
    assert est.X_.dtype == X.dtype, f"expected {X.dtype}, got {est.X_.dtype}"
           ^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: expected float32, got float64
```

### Suggested fix

Give the four functions an allocation that follows the input's floating precision.

Numba supports neither `np.promote_types` nor a comparison against `X.dtype` (both raise TypingError), so the output dtype can be taken from a true division of two scalars of the input dtype. 

### Versions

```
System:
    python: 3.14.6 (main, Jun 11 2026, 04:05:02) [MSC v.1944 64 bit (AMD64)]
executable: C:\Users\antoine\Documents\aeon\.venv\Scripts\python.exe
   machine: Windows-11-10.0.26200-SP0
Python dependencies:
         aeon: 1.5.0
          pip: None
   setuptools: None
 scikit-learn: 1.8.0
        numpy: 2.3.5
        numba: 0.63.1
        scipy: 1.17.1
       pandas: 2.3.3
```

baraline:
#### Reference Issues/PRs

Fixes #3706 

Add a helper to preserve dtype precision of the input at the output of the normalization functions


TonyBagnall:
python, weak typing except primitives where its militant. dtype=type(one / one)  ... this is like next gen c programming,

so my assistants tell me this

1. z_normalise_series itself still returns float64 for float32 input because the current mean and standard-deviation calculations use float64. This means the 1D and collection normalisation utilities will have different dtype behaviour after this change Is that intentional?

2 the 1d entry points still upcast (consistency gap). z_normalise_series, z_normalise_series_with_mean_std, and z_normalise_series_with_mean are untouched and still return float64 for float32 input (they compute (X - mean)/s where stats.mean/std return float64). So after this PR, z_normalise_series_2d(float32) → float32 but z_normalise_series(float32) → float64.

3. fastmath=True on _zeros_normalised is pointless. The helper only does a trivial one / one and allocates zeros; no FP computation benefits, and fastmath's nnan/ninf flags are exactly the family that silently break isnan/isinf guards elsewhere in the codebase. Harmless here (I confirmed it doesn't affect the type derivation), but I'd drop it for clarity and keep cache=True.

TonyBagnall (dismissed):
lgtm, see comment for minor quibbles

baraline:
Thanks I'll probably change some stuff based on these comments before merging to round some corners

baraline:
On the overall package, there would be other things to do to conserve precision of the input and not upcast it, but I'll raise another issue for that. We can go with this.

aeon/utils/numba/general.py
```diff
@@ -154,6 +154,33 @@ def first_order_differences_3d(X: np.ndarray) -> np.ndarray:
...
+@njit(fastmath=True, cache=True, nogil=True)
+def _as_normalised_float(X: np.ndarray) -> np.ndarray:
+    """Return a float copy of ``X``, in the dtype its z-normalisation should output.
+
+    ``np.zeros(X.shape)`` always returns float64, which silently doubles the memory
+    of a float32 input. Dividing by one is the operation the normalisation performs,
+    so numpy's promotion rules give the intended output dtype: float32 stays float32,
+    and every integer or boolean dtype normalises in float64. Numba supports neither
+    ``np.promote_types`` nor ``np.result_type``, so it is also the simplest way to
+    ask numpy for that rule from a jitted function.
```
baraline, line 166: As a self note, we cannot keep a int32 input to a float32, as for exemple we have

int32 value            : 16777217
round-tripped via f32  : 16777216   <- lost
round-tripped via f64  : 16777217   <- exact
int32 max 2147483647 vs largest exact float32 integer 16777216

So for this case we have to upcast. Following the same reasoning, int16 could go to float32, but we never deal with less than 32bit in numba

TonyBagnall (approved):

aeon/utils/numba/general.py
TonyBagnall: I'll resolve your own conversation so we can merge :)

PR #3711, opened 2026-08-05
[BUG] `DiscreteFourierApproximation`: keep conjugate-symmetric terms

### LINKED ISSUE #3710: [BUG] `DiscreteFourierApproximation` halves the amplitude of every term it retains
### Describe the bug

`DiscreteFourierApproximation._transform` masks the two-sided `np.fft.fft`
spectrum and keeps the lowest `keep = int(r * n_timepoints)` bins. For a real
series the spectrum is conjugate symmetric, so bins `k` and `n - k` together carry
one frequency. Keeping `k` without `n - k` and then taking `.real` of the inverse
returns half of that term's amplitude, so a sinusoid lying entirely inside the
retained band comes back at half its amplitude instead of unchanged.

The halving is exact on the fluctuation about the mean whenever
`max(int(r * n_timepoints), 1) <= n_timepoints / 2`, i.e. for `r <= 0.5`, which
includes the default:

```
z[n] - mean = 0.5 * (correct_reconstruction[n] - mean)
```

DC is the only retained term with no conjugate partner, which is why the mean
survives while the variance comes out quartered.

Above `r = 0.5` the retained set starts to include both partners of the
frequencies in `(n - keep, keep)`, so those return at full amplitude while the
rest stay halved, and the distortion becomes a frequency-dependent gain rather
than a uniform factor. Feeding unit cosines in bin by bin at `n_timepoints=40`,
`r=0.75`, the measured gain is 0.5 on bins 1-10 and 1.0 on bins 11-20: a two-tone
input with unit cosines at bins 2 and 15 comes back with amplitudes 0.5 and 1.0.
Low frequencies are attenuated relative to high, which is the opposite of
smoothing.

There is a second consequence at the default `r=0.5`. `keep` counts bins of the
two-sided spectrum, so for even `n_timepoints` the retained bins are
`0 .. n/2 - 1`, every positive frequency below Nyquist. The default therefore
discards only the Nyquist term and does almost no smoothing; it essentially just
scales the series halfway to its mean. That is the series the `DFA Default` panel
of `examples/transformations/smoothing_filters.ipynb` plots as "Smoothed Series".

Finally, `sort=True` usually escapes the halving, because the largest-`|X[k]|`
bins come in conjugate pairs and so that path normally retains both partners; it
is halved only where `keep` splits a pair. The two settings therefore disagree by
a factor of two on the same frequency in the common case.

```python
import numpy as np

from aeon.transformations.series.smoothing import DiscreteFourierApproximation

# 1. a cosine entirely inside the retained band
n = np.arange(64)
x = np.cos(2 * np.pi * 3 * n / 64)  # frequency bin 3

out = DiscreteFourierApproximation(r=0.25).fit_transform(x)[0]
print(np.abs(out).max())            # 0.5
print(np.abs(out - 0.5 * x).max())  # 5.3e-16 -- exactly half the input

# 2. sort=True and sort=False disagree by a factor of two on the same input
for sort in (False, True):
    out = DiscreteFourierApproximation(r=0.5, sort=sort).fit_transform(x)[0]
    print(sort, np.abs(out).max())  # False 0.5 / True 1.0

# 3. r=1.0 is exact, so the error is confined to r < 1
out = DiscreteFourierApproximation(r=1.0).fit_transform(x)[0]
print(np.abs(out - x).max())        # 3.3e-16

# 4. the default configuration, on the x_signal series from
#    examples/transformations/smoothing_filters.ipynb
np.random.seed(42)
np.random.random(128)
t1, t2 = np.linspace(0, 64, 256), np.linspace(0, 32, 256)
x_signal = np.sin(t1) + np.cos(t2) + (np.random.random(256) - 0.5)

out = DiscreteFourierApproximation().fit_transform(x_signal)[0]
print(out.std() / x_signal.std())            # 0.499954
print(np.corrcoef(out, x_signal)[0, 1])      # 0.99990776
```

### Expected results

A band-limited approximation reproduces the retained frequencies at their own
amplitude, so case 1 prints `1.0`, case 2 prints `1.0` for both `sort` settings,
and case 4 gives a standard-deviation ratio near 1 with the output visibly
decorrelated from the noise.

### Actual results

No exception is raised; the values are wrong.

```
0.5                       # 1. peak amplitude, expected 1.0
5.342948306008566e-16     #    the output is exactly 0.5 * x
False 0.49999999999999994 # 2. same input, same r, the two paths disagree
True 1.0000000000000002   #    by a factor of two
3.3306690738754696e-16    # 3. r=1.0 is correct
0.49995387834371807       # 4. std ratio at the default, expected ~1.0
0.9999077566874363        #    a smoother should decorrelate from the noise
```

The existing `test_dft` only asserts that a channel transformed on its own matches
the same channel transformed inside a 2-channel array, a consistency check that
passes with or without this, since both channels are scaled identically. No
assertion pinned the output amplitude.

### Suggested fix

Use `np.fft.rfft`/`np.fft.irfft`, which carry the `n_timepoints // 2 + 1`
non-redundant terms of a real series. Amplitudes are then preserved with no
mirroring logic and `r=1` reconstructs the series exactly. This also makes `r` a
proportion of those `n_timepoints // 2 + 1` independent terms, which is the only
count that is meaningful for a real series, but it moves the cut-off for a given
`r`, so `r=0.5` would cut at `fs/4` rather than just below Nyquist. That is a
change in what `r` selects, and worth a decision before I open anything.

The narrower alternative fixes the amplitude alone and leaves `r` untouched: keep
`fft`/`ifft` and mirror the mask onto the conjugate partners with one added line,
`mask[:, 1:] |= mask[:, :0:-1]`. The cost is that the default `r=0.5` then really
does retain everything except Nyquist, so it stays an almost-identity transform.

I have both versions working locally, along with three value assertions for
`test_dft.py` (in-band amplitude preserved for both `sort` settings, `r=1.0`
exact for odd and even lengths, and a term above the cut-off removed rather than
attenuated). Three of those cases fail on `main` and pass with either fix. Happy
to open a PR with whichever of the two you prefer.

### Versions

<details>

```
System:
    python: 3.12.3 [GCC 13.3.0]
   machine: Linux-x86_64-with-glibc2.39

Python dependencies:
         aeon: 1.5.0
 scikit-learn: 1.8.0
        numpy: 2.3.5
        numba: 0.63.1
        scipy: 1.17.1
       pandas: 2.3.3
```

</details>

Reproduced on `main` at 3d067d5.

Hrafz:
#### Reference Issues/PRs

Fixes #3710.

`_transform` masked the two-sided `np.fft.fft` spectrum and kept the lowest
`keep = int(r * n_timepoints)` bins. For a real series the spectrum is conjugate
symmetric, so bins `k` and `n - k` together carry one frequency; keeping `k`
without `n - k` and then taking `.real` of the inverse returns half of that term.
A cosine lying entirely inside the retained band comes back halved rather than
unchanged:

```python
import numpy as np
from aeon.transformations.series.smoothing import DiscreteFourierApproximation

n = np.arange(64)
x = np.cos(2 * np.pi * 3 * n / 64)

out = DiscreteFourierApproximation(r=0.25).fit_transform(x)[0]
print(np.abs(out).max())            # 0.5 on main, expected 1.0
print(np.abs(out - 0.5 * x).max())  # 5.3e-16 on main: exactly half the input
```

DC is the only retained term with no partner, so the mean survives while the
variance comes out quartered. `sort=True` usually escapes it, because the
largest-`|X[k]|` bins come in conjugate pairs, so the two settings disagree by a
factor of two on the same input.

The change is to use `rfft`/`irfft`, which carry the `n_timepoints // 2 + 1`
non-redundant terms: amplitudes are preserved with no mirroring logic and `r=1`
reconstructs the series exactly. `r` becomes a proportion of those independent
terms, which does move the cut-off for a given `r`, `r=0.5` cuts at `fs/4` rather
than just below Nyquist, so the docstring now says which set `r` is a proportion
of. The measurements and the narrower alternative, mirroring the mask onto the
conjugate partners so that `r` keeps its current meaning, are in #3710; happy
to switch to that one instead.

No.

#### Any other comments?

`test_dft.py` previously only asserted that a channel transformed alone matches
the same channel transformed inside a 2-channel array, a consistency check that
passes with or without the bug. No assertion pinned the output amplitude. Three
added: in-band amplitude preserved, for `r` in {0.25, 0.5, 1.0} and both `sort`
settings; `r=1.0` exact for odd and even lengths; and a term above the cut-off
removed rather than attenuated. On `main` that file is 3 failed, 11 passed, the
two `sort=False` amplitude cases and the high-frequency test, while the
`sort=True` cases and `r=1.0` pass, which localises the problem to the two-sided
mask.

`aeon/transformations/series/` goes from 209 passed, 1 skipped to 217 passed, 1
skipped: the eight new cases, nothing else changing colour. The 19
`check_estimator` cases for `DiscreteFourierApproximation` pass before and after.
`pre-commit` clean.

##### For all contributions

### DIFF
--- a/aeon/transformations/series/smoothing/_dfa.py
+++ b/aeon/transformations/series/smoothing/_dfa.py
@@ -18,7 +18,8 @@ class DiscreteFourierApproximation(BaseSeriesTransformer):
     Parameters
     ----------
     r : float, default=0.5
-        Proportion of Fourier terms to retain [0, 1]
+        Proportion of the ``n_timepoints // 2 + 1`` Fourier terms to retain [0, 1].
+        ``r=1`` reconstructs the series exactly.
     sort : bool, default=False
         Sort the Fourier terms by amplitude to keep most important terms
 
@@ -62,8 +63,11 @@ class DiscreteFourierApproximation(BaseSeriesTransformer):
         -------
         transformed version of X
         """
-        # Compute DFT
-        dft = np.fft.fft(X)
+        # Compute the DFT. X is real, so its spectrum is conjugate symmetric and
+        # rfft returns the n_timepoints // 2 + 1 non-redundant terms. Masking the
+        # two-sided fft instead drops the conjugate partner of every retained
+        # term, which halves its amplitude in the reconstruction.
+        dft = np.fft.rfft(X)
 
         # Mask array of terms to keep and number of terms to keep
         mask = np.zeros_like(dft, dtype=bool)
@@ -79,6 +83,6 @@ class DiscreteFourierApproximation(BaseSeriesTransformer):
             mask[:, 0:keep] = True
 
         # Invert DFT with masked terms
-        X_ = np.fft.ifft(dft * mask).real
+        X_ = np.fft.irfft(dft * mask, n=X.shape[1])
 
         return X_
--- a/aeon/transformations/series/smoothing/tests/test_dft.py
+++ b/aeon/transformations/series/smoothing/tests/test_dft.py
@@ -31,3 +31,39 @@ def test_dft(r, sort):
 
     np.testing.assert_almost_equal(x_1[0], x_12[0], decimal=4)
     np.testing.assert_almost_equal(x_2[0], x_12[1], decimal=4)
+
+
+@pytest.mark.parametrize("r", [0.25, 0.50, 1.00])
+@pytest.mark.parametrize("sort", [True, False])
+def test_dft_preserves_retained_amplitude(r, sort):
+    """A sinusoid inside the retained band must come back with its amplitude."""
+    n_timepoints = 64
+    k = 3  # frequency bin, well inside the retained band for every r above
+    n = np.arange(n_timepoints)
+    x = 2.0 + np.cos(2 * np.pi * k * n / n_timepoints)
+
+    dft = DiscreteFourierApproximation(r=r, sort=sort)
+    x_ = dft.fit_transform(x)[0]
+
+    np.testing.assert_allclose(x_, x, atol=1e-8)
+
+
+def test_dft_full_r_is_identity():
+    """r=1.0 retains every term, so the series must be reconstructed exactly."""
+    rng = np.random.RandomState(0)
+    for n_timepoints in (63, 64):
+        x = rng.normal(size=n_timepoints)
+        x_ = DiscreteFourierApproximation(r=1.0).fit_transform(x)[0]
+        np.testing.assert_allclose(x_, x, atol=1e-8)
+
+
+def test_dft_discards_high_frequencies():
+    """A term above the cut-off must be removed, not merely attenuated."""
+    n_timepoints = 64
+    n = np.arange(n_timepoints)
+    low = np.cos(2 * np.pi * 2 * n / n_timepoints)
+    high = np.cos(2 * np.pi * 30 * n / n_timepoints)
+
+    x_ = DiscreteFourierApproximation(r=0.25, sort=False).fit_transform(low + high)[0]
+
+    np.testing.assert_allclose(x_, low, atol=1e-8)


TonyBagnall (approved):
thanks for this, well found, #3710 is indeed a bug and solution looks sound. Was slightly concerned on impact elsewhere, but it seems this is a standalone function, not embedded in classifiers ets

PR #3726, opened 2026-08-09
[BUG] Preserve float precision in Numba helpers

### LINKED ISSUE #3720: [BUG] utils/numba/general.py returns float64 for float32 input in six functions
### Describe the bug

Same class as the z_normalise_series* fix (#3706), but these outputs have a different shape from X, so the X / 1 fix<html>

Function | File:line | Cause | Output size
-- | -- | -- | --
slope_derivative | aeon/utils/numba/general.py:710 | np.zeros(X.shape[0] - 2) | 0.97x input
slope_derivative_2d | aeon/utils/numba/general.py:757 | np.zeros((X.shape[0], X.shape[1] - 2)) | 0.97x
slope_derivative_3d | aeon/utils/numba/general.py:807 | np.zeros((X.shape[0], X.shape[1], X.shape[2] - 2)) | 0.97x
sliding_mean_std_one_series | aeon/utils/numba/general.py:550-551 | mean and std = np.zeros((n_channels, n_subs)) | 0.78x
get_subsequence | aeon/utils/numba/general.py:420 | np.zeros((n_channels, length)) | per call, in a loop
get_subsequence_with_mean_std | aeon/utils/numba/general.py:456-458 | explicit dtype=np.float64 | per call, in a loop

### Suggested fix

Could use a similar fix like np.zeros(shape, (X[:0] / 1).dtype)

### Versions

_No response_

cakeni:
#### Reference Issues/PRs

Fixes #3720. See also #3707, which addresses the same dtype class in separate z-normalisation helpers.

Allocates outputs using the dtype produced by true division of the input. This keeps `float32` inputs as `float32` across all six affected Numba helpers while preserving the existing `float64` output for integer and `float64` inputs. A focused regression test covers every returned array from the six helpers.

The calculation loops are unchanged; only output allocation dtype is corrected. If #3707 merges first, its changes touch the same module but this fix does not depend on them.

No.

#### Any other comments?

Validation performed locally:

- `.venv\Scripts\python.exe -m pytest -o addopts='' aeon/utils/numba/tests/test_general.py -q` — 42 passed
- `.venv\Scripts\python.exe -m pytest -o addopts='' aeon/utils/numba/tests/test_general.py -q --nonumba true` — 42 passed
- `.venv\Scripts\python.exe -m pytest -o addopts='' aeon/distances/elastic/tests/test_distance_correctness.py -q` — 2 passed
- `.venv\Scripts\python.exe -m pytest -o addopts='' aeon/similarity_search/subsequence/tests/test_mass.py -q` — 9 passed
- `.venv\Scripts\pre-commit.exe run --files aeon/utils/numba/general.py aeon/utils/numba/tests/test_general.py` — all hooks passed

##### For all contributions
- [x] I will use `@all-contributors` after this PR has been merged.


baraline:
Hi, thanks for working on this.

I'll wait for #3707 to get in first, as we would want to re-use (or adapt) some functions declared there to initialize arrays with the correct dtype, instead of re-declaring the X / 1 trick everywhere.

Other than that it looks correct, will review better when the other PR is in.

baraline:
hi @cakeni the first PR is in. Could you try to uniformize what you done to use the same reasoning ?

cakeni:
Done — I’ve updated the dtype initialization to reuse `_as_normalised_float` from #3707 and synced the branch with the latest `main`. The focused tests are passing.

aeon/utils/numba/tests/test_general.py
```diff
@@ -185,6 +188,29 @@ def test_sliding_mean_std_one_series(dtype):
         mean, std = sliding_mean_std_one_series(X, 100, 3)
 
 
+@pytest.mark.parametrize("dtype", DATATYPES)
+def test_float_output_dtype_follows_input_precision(dtype):
+    """Test helpers preserve float precision and promote integer input."""
+    X = np.arange(24, dtype=dtype).reshape(2, 12)
```
baraline: Could you just replace this with the testing.utils function to generate data ? (and cast it to the dtype)

aeon/utils/numba/general.py
```diff
@@ -417,7 +417,7 @@ def get_subsequence(
         The resulting subsequence.
     """
     n_channels, _ = X.shape
-    values = np.zeros((n_channels, length))
+    values = np.zeros((n_channels, length), dtype=_as_normalised_float(X[:0]).dtype)
```
baraline, line 420: It's not "pretty",  but doing things like :

```
def _as_normalised_float_dtype(input_dtype):
   if input_dtype == np.float32 return np.float64 else return np.float32
``` 
Is not possible inside numba functions (or at least not without overcomplicating things), so we'll go with this for now.

baraline (changes requested):
Just one small comment on the test data and we can merge this, good job !

cakeni:
@baraline I updated the dtype tests to use `make_example_2d_numpy_series(...).astype(dtype)` as requested. The current CI is green; could you take another look when you have a chance?

baraline (approved):
LGTM, thanks !

PR #3727, opened 2026-08-10
[DEP] Deprecate znormalized in favor of znormalize in SAX

### LINKED ISSUE #3678: [MNT] Change the parameter `znormalized` to `znormalize` in SAX transformation

Dilraj07:
Fixes #3678.

Replaces the parameter `znormalized` in `SAX` with `znormalize` (default `True`), adds deprecation handling for `znormalized`, and updates unit tests.

No.

#### PR checklist
- [x] The PR title starts with [MNT]

> The code has been reviewed by a human.

### DIFF
--- a/aeon/transformations/collection/dictionary_based/_sax.py
+++ b/aeon/transformations/collection/dictionary_based/_sax.py
@@ -3,6 +3,8 @@
 __maintainer__ = []
 __all__ = ["SAX", "_invert_sax_symbols"]
 
+import warnings
+
 import numpy as np
 import scipy.stats
 from numba import get_num_threads, njit, prange, set_num_threads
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
thanks for the PR, but this will require deprecation I think, see our policy here
https://www.aeon-toolkit.org/en/latest/developer_guide/deprecation.html
its quite simple

Dilraj07:
> thanks for the PR, but this will require deprecation I think, see our policy here https://www.aeon-toolkit.org/en/latest/developer_guide/deprecation.html its quite simple

Thanks for the guidance! I have updated the PR to adhere to aeon's deprecation policy:
- Set `znormalized="deprecated"` default in `SAX.__init__` with the `# TODO remove 'znormalized' in v1.7.0` removal comment.
- Preserved backward compatibility by raising a `FutureWarning` (`stacklevel=2`) and mapping `znormalized` to `znormalize`.
- Updated docstrings and added unit tests for the deprecation warning, default sentinel, and value mapping.
- Updated the PR title prefix to `[DEP]`.

hadifawaz1999:
We have duplicate PRs for this #3680

MatthewMiddlehurst:
See the review and functionality changes requested in https://github.com/aeon-toolkit/aeon/issues/3680 if you wish to proceed please. This is more then a rename.

TonyBagnall:
Thanks for working on this. I’m going to close this in favour of #3756 so we have a single PR for #3678.

The main reason for choosing #3756 is backward compatibility: it keeps the deprecated znormalized parameter in its existing position in the constructor. In this PR, moving znormalized to the end and putting znormalize in its old position means existing positional calls could silently acquire the opposite normalization behaviour.

#3756 also updates the ReCOMETS usage while preserving its existing behaviour. We’ll finish the remaining documentation/example and deprecation details there.

Thanks for the contribution.

Issue #2374, opened 2024-11-20
[ENH] Change fitted estimators and ensemble collections from protected to public


MatthewMiddlehurst:
### Describe the feature or idea you want to propose

Currently in a lot of estimators we fit an sklearn estimator i.e. `_estimator` or form an ensemble of estimators i.e. `_ensemble` or `_classifiers` but these are protected (starting with an `_`).

Theres not much reason for them to be protected IMO.

### Describe your proposed solution

Change fitted estimator attributes from protected to public using our fitted attribute convention i.e. from `_estimator` top `estimator_`.

These should then be properly documented in the estimator docstring under "Attributes"

TonyBagnall:
see https://github.com/aeon-toolkit/aeon/pull/2350 for an example

vedpawar2254:
can you assign this issue to me

aryan0931:
@aeon-actions-bot assign @aryan0931

MatthewMiddlehurst:
@vedpawar2254 the comment has to be made in a new post, just saw this now. See the contributor guide on the webpage.

aryan0931:
yes I am working on that.

Sharathchenna:
@aeon-actions-bot assign @Sharathchenna

TonyBagnall:
had a look at this

21 files already use the public estimator_/estimators_/ensemble_ convention, while 38 are private. The _estimator → estimator_ rename dominates; the odd ones out are _pipeline, _clf, _transformers, _estimators, _model. Made a list, if I make an old issue a GFI, will the bots go for it? Lets see. Do one group at a time not all at once, e.g. classification/shapelet_based, and avoid any involved in a large existing PR

 ### classification/ #3714 #3713 
  - [x] `shapelet_based/_sast.py` — `_pipeline`
  - [x] `shapelet_based/_rsast.py` — `_pipeline` 
  - [x] `shapelet_based/_rdst.py` — `_estimator`
  - [x] `early_classification/_teaser.py` — `_estimators`, `_estimator`
  - [x] `early_classification/_probability_threshold.py` — `_estimators`, `_estimator`
  - [x] `hybrid/_hivecote_v1.py` — `_estimators`
  - [x] `interval_based/_rstsf.py` — `_transformers`
  - [x] `interval_based/_quant.py` — `_estimator`
  - [x] `interval_based/_interval_pipelines.py` — `_estimator`
  - [x] `convolution_based/_rocket.py` — `_estimator`
[...]

PR #3729, opened 2026-08-11
[MNT] Make model_ attribute public in ARIMA and ETS

aashoday:
#### Reference Issues/PRs

Related to #2374 - Renames `_model` to `model_` across `forecasting/stats/` (`ARIMA` and `ETS`), removes early attribute initialization from `__init__`, and documents each in its class docstring.

- Renames the internal `_model` to `model_` across `forecasting/stats/` (ARIMA, ETS, CES, etc.) and documents `model_` under Attributes in each class docstring.

- Removes the early initialization of `self.model_ = []` from `__init__` in `ARIMA` and `ETS` so that fitted attributes are created strictly during `fit()`.

- Adds unit tests (`test_arima_model_attribute_lifecycle` and `test_ets_model_attribute_lifecycle`) to verify that `model_` is only set after fitting and that the `_model` attribute no longer exists.

None

#### Any other comments?

Existing test suites across `forecasting/stats/` pass cleanly with all attribute assertions updated to use `model_`:

- `forecasting/stats/tests/test_arima.py`

- `forecasting/stats/tests/test_ets.py`

- `forecasting/stats/tests/test_ces.py`

##### For all contributions

### DIFF
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -68,6 +68,8 @@ class ARIMA(BaseForecaster, IterativeForecastingMixin):
         Coefficients for autoregressive terms (length p).
     theta_ : np.ndarray
         Coefficients for moving average terms (length q).
+    model_ : np.ndarray
+        Array containing the structure of the fitted model (constant, p, q).
 
     References
     ----------
--- a/aeon/forecasting/stats/_arima.py
+++ b/aeon/forecasting/stats/_arima.py
@@ -102,7 +102,7 @@ class ARIMA(BaseForecaster, IterativeForecastingMixin):
         self.residuals_ = []
         self.fitted_values_ = []
         self.aic_ = 0
-        self._model = []
+        self.model_ = []
         self._parameters = []
         self.exog_ = None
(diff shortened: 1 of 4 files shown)


aashoday:
Hi @TonyBagnall,

While reviewing `ARIMA` , `AutoARIMA`, `ETS`, and `AutoETS`, I noticed several other trailing-underscore attributes are currently being initialized there as well.

To keep this PR focused on the scope of `#2374`, I left those untouched.  

Let me know if you'd like me to clean up those additional initializations in this PR as well.

TonyBagnall:
so this is my mistake, but on looking I don't think _model in ARIMA/ETS is actually in scope for #2374. That issue is intended to expose fitted estimators/collections currently stored in protected attributes. Here _model is just an internal integer representation passed to the numerical routines ([constant, p, q] for ARIMA and [error, trend, seasonality, seasonal_period] for ETS), rather than a fitted estimator. Making this model_ exposes an implementation detail as part of the public API. AutoARIMA.final_model_ and AutoETS.wrapped_model_ are the actual fitted estimator objects and are already public. I would leave _model protected.

PR #3731, opened 2026-08-12
[MNT] Make early_classification estimator_ and estimators_ attributes public

aashoday:
#### Reference Issues/PRs

Related to #2374 - renames internal `_estimator` and `_estimators` attributes to public conventions (`estimator_`, `estimators_`) across the early classification module (`TEASER` and `ProbabilityThresholdEarlyClassifier`), per the module's convention, and documents each in its class docstring.

* Renames `_estimator` to `estimator_` and `_estimators` to `estimators_` across both files.
* Removes `self._estimators = []` from the `__init__` methods to ensure fitted attributes are strictly created during `fit()`.
* Documents `estimator_` and `estimators_` under the `Attributes` section of each class docstring.
* Adds lifecycle regression tests (`test_probability_threshold_estimator_attribute_lifecycle` and `test_teaser_estimator_attribute_lifecycle`) to their respective test files to ensure the attributes are absent before fitting and populated after fitting.

No

#### Any other comments?

Verified using the dedicated test suite for early classification.

##### For all contributions

### DIFF
--- a/aeon/classification/early_classification/_probability_threshold.py
+++ b/aeon/classification/early_classification/_probability_threshold.py
@@ -76,8 +76,6 @@ class ProbabilityThresholdEarlyClassifier(BaseEarlyClassifier):
         the results of previous method calls.
         Records in order: the time stamp index, the number of consecutive decisions
         made, the predicted class and the series length.
-    estimator_ : BaseEstimator
-        The cloned base estimator used for early classification.
     estimators_ : list of BaseEstimator
         The fitted estimators for each time stamp.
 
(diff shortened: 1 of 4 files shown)


TonyBagnall (changes requested):
so this requires a little more thought, for issues not described in the original issue. I think there are two issues here.

estimator_ is not actually a fitted estimator. It is just the resolved base estimator/template, which is subsequently cloned and fitted into estimators_. If the user passes an estimator, estimator_ will still refer to that unfitted object. I would therefore leave this as an internal attribute and only expose estimators_, which contains the actual fitted estimators.

For TEASER, _one_class_classifiers also contains fitted estimators and seems to fall under #2374. I think this should become something like one_class_classifiers_ and be documented as a fitted attribute.

The _estimators → estimators_ changes themselves look fine.

aashoday:
Hey @TonyBagnall,

Thanks for the review. I initially made estimator_ public because it was listed on the #2374 checklist, so I assumed it should be exposed.

I will revert it back to the protected estimator state and only keep estimators public. I will also make one_class_classifiers public in TEASER and update the docstrings.

I will push these changes shortly.

aashoday:
Hi @TonyBagnall

I have made the requested changes.
Let me know if anything else needs changes.

TonyBagnall (approved):
LGTM, thanks

PR #3732, opened 2026-08-13
[MNT] Make estimators_ attribute public in HIVE-COTE classifiers

aashoday:
#### Reference Issues/PRs

Related to #2374 - renames internal `_estimators` attribute to `estimators_` across the hybrid classification module (`_BaseHIVECOTE`, `HIVECOTEV1`, and `HIVECOTEV2`), per the module's convention, and documents it in the public class docstrings.

* Renames `_estimators` to `estimators_` in the `_fit` methods of `HIVECOTEV1` and `HIVECOTEV2`.
* Updates the private base class `_BaseHIVECOTE` to pull from `self.estimators_` rather than `self._estimators` to prevent the underlying `fit` loop from crashing.
* Documents `estimators_` under the `Attributes` section of the `HIVECOTEV1` and `HIVECOTEV2` class docstrings.
* Adds a lifecycle regression test (`test_hivecote_estimator_attribute_lifecycle`) to `test_hc.py` checking both V1 and V2.

No

#### Any other comments?

Verified using the dedicated test suite for hybrid classification. 

> Note: Although #2374 only listed `_hivecote_v1.py`, I've updated `_BaseHIVECOTE` and `_hivecote_v2.py` as well. Since both inherit their fitting loop from the same base class, updating only V1 would have caused the base class `getattr` to fail.

##### For all contributions


TonyBagnall (changes requested):
as with #3731, this is a little more complex that straight swap. I don't think estimators_ is the right attribute to expose here. The objects stored in _estimators are the unfitted component templates. _BaseHIVECOTE._fit clones and fits these, with the actual fitted estimators already stored publicly in fitted_estimators_.

I would therefore leave _estimators protected and make sure fitted_estimators_ is documented in the Attributes sections of HIVECOTEV1 and HIVECOTEV2.

Also, there is a typo in the new test name:

test_hivecote_estimator_attribute_lifecycly

which should presumably be lifecycle.

aashoday:
Hey, Thanks for the review.

I will revert estimators_ back to the protected _estimators state, fix the lifecycly typo in the test, and ensure fitted_estimators is fully documented in the Attributes sections for V1 and V2.

i'll push these updates soon

TonyBagnall:
last one @aashoday thanks for doing this

aashoday:
Hey @TonyBagnall,
Thanks for being patient with me! This was my first issue and I really loved working on it. I'm definitely looking forward to contributing more to the project.

TonyBagnall (approved):

PR #3739, opened 2026-08-15
[MNT] Expose YeoJohnsonTransformer through public API

mkzung:
#### Reference Issues/PRs

None.

`YeoJohnsonTransformer` lives in `aeon/transformations/series/_yeojohnson.py` and is missing from
`aeon/transformations/series/__init__.py`, so it cannot be imported from
`aeon.transformations.series` the way its neighbours can, and it does not appear in the API
reference.

It is also invisible to `all_estimators`, which walks the public modules, so none of the estimator
contract checks have ever run against it.

Adds it to `__all__` and to `docs/api_reference/transformations.md`, and extends the existing test to
import it from the package rather than from the private module.

`all_estimators(include_sklearn=False)` returns 247 before and 248 after.

No.

#### Any other comments?

Reverting `__init__.py` and keeping the test gives a collection error rather than a failure, since
the import is at module level.

### DIFF
--- a/aeon/transformations/series/__init__.py
+++ b/aeon/transformations/series/__init__.py
@@ -17,6 +17,7 @@ __all__ = [
     "StatsModelsPACF",
     "BKFilter",
     "BoxCoxTransformer",
+    "YeoJohnsonTransformer",
     "ScaledLogitSeriesTransformer",
     "PCASeriesTransformer",
     "STLSeriesTransformer",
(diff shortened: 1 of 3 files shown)


TonyBagnall (changes requested):
hi, thanks for this, nice find, I dont really think we need the regression test test_is_discoverable, could you take it out?

mkzung:
Taken out, thanks. The import in the existing test moved to the package, so it still fails if the export goes away.

TonyBagnall (approved):
LGTM, thanks

TonyBagnall:
@all-contributors add @mkzung for maintenance

PR #3740, opened 2026-08-15
[BUG] ScaledLogitSeriesTransformer: apply the scaled logit when a bound is zero

mkzung:
#### Reference Issues/PRs

None.

`ScaledLogitSeriesTransformer` picks its formula with a truthiness test:

```python
if self.upper_bound and self.lower_bound:
```

A bound of zero is falsy, so a bound the user set is treated as one they did not. With
`lower_bound=0` the transformer falls through to the upper-bound-only branch and returns
`-log(upper - X)` instead of `log((X - 0) / (upper - X))`. The same happens to `upper_bound=0`,
which lands on the lower-bound-only branch.

Zero is an ordinary bound here. A series bounded below by zero is the common case for anything
non-negative, and both `_transform` and `_inverse_transform` carry the same test.

Changed to `is not None`, which is what the two `elif` branches underneath already use.

No.

#### Any other comments?

Two test cases, one per bound, because the two are separate branches and a fix proved on one says
nothing about the other. Reverting `_scaled_logit.py` and keeping them turns both red.


TonyBagnall:
thanks, test fail seems unrelated, I'll take a look

TonyBagnall:
I suspect fail was a numba caching issue from #3707

TonyBagnall (approved):
Nice find. 

One optional thought: a single-bound zero case, e.g. upper_bound=0, lower_bound=None, would round out the edge-case coverage. That path wasn't affected by this bug though, so I wouldn't consider it necessary for this PR, just thinking about coverage

mkzung:
Checked that rather than take it on trust: with `upper_bound=0, lower_bound=None` the output is identical before and after the change, since `0 and None` and `0 is not None and None is not None` both send it down the upper-bound-only branch. So the path really is untouched here.

Happy to add the case if you want the coverage, either on this branch or a follow-up, whichever keeps this one simple.

TonyBagnall:
> Checked that rather than take it on trust: with `upper_bound=0, lower_bound=None` the output is identical before and after the change, since `0 and None` and `0 is not None and None is not None` both send it down the upper-bound-only branch. So the path really is untouched here.
> 
> Happy to add the case if you want the coverage, either on this branch or a follow-up, whichever keeps this one simple.

add as a follow up, maybe look at coverage for these generally as you are here? coverage is ok on series transformers, but some gaps exist

TonyBagnall:
@all-contributors add @mkzung for bug

PR #3741, opened 2026-08-15
[BUG] reverse_windowing: make the psutil import optional

mkzung:
#### Reference Issues/PRs

None.

`reverse_windowing` chooses between a vectorised and an iterative implementation by asking how much
memory is free, and `_has_enough_memory_for_vectorized_entire` imports `psutil` to find out.

`psutil` is not a dependency of aeon. It arrives through the developer extras, so it is present in
CI and absent for anyone who installed the package normally, and `reverse_windowing` raises
`ModuleNotFoundError` for them on the default path.

The import is wrapped and returns `False` when it fails, which selects the memory-bounded iterative
implementation. That is the same branch the function already takes when memory is short, so the
answer is unchanged, only slower.

The docstring for `force_iterative` says so, since it is the parameter a reader would reach for.

No. It removes a hard requirement on an undeclared one.

#### Any other comments?

The test hides `psutil` through `sys.modules` and checks the function still returns the right
answer. Reverting `windowing.py` and keeping it gives one failure.


MatthewMiddlehurst:
Good catch. Think this is primarily used for anomaly detection, but not super familiar with this code. 

We typically use soft dependency utilities in these cases and add it to the pyproject.toml. Not 100% sure how best to deal with this case but can discuss.

MatthewMiddlehurst:
Think this was implemented by @SebastianSchmidl if they have time? Ignore if not.

mkzung:
psutil is only in pyproject as `pytest-xdist[psutil]`, so it is not declared for runtime use, and `windowing.py` is the only place that imports it. The same function already imports `dask.distributed` under a try/except twenty lines further down, and dask is not declared either, so this patch makes psutil match its neighbour.

Happy to switch to `_check_soft_dependencies` and add psutil to `all_extras` if you prefer that. One thing worth knowing if we do: `_has_enough_memory_for_vectorized_entire` runs on every `reverse_windowing` call, and `_check_soft_dependencies` measures 57.5 us against the 1.8 us of the `psutil.virtual_memory()` call it would guard. So it wants to be a module-level flag like `_hydra.py:137` rather than a call inside the function. I have that version ready if you want it.

TonyBagnall (approved):
this lgtm, thanks for the prs, I think this is fine for this function given it already does it with dask

PR #3742, opened 2026-08-15
[BUG] BaseCollectionEstimator: enforce the capability:univariate tag

mkzung:
#### Reference Issues/PRs

None.

`BaseCollectionEstimator._check_X` reads four capability tags and enforces three of them.
`capability:multivariate`, `capability:missing_values` and `capability:unequal_length` all raise when
the data needs something the estimator says it cannot do. `capability:univariate` is read nowhere,
so an estimator declaring `False` is handed univariate data anyway.

Adds the missing branch, in the same shape as the multivariate one directly above it.

Of 178 collection estimators, `MUSE` is the only one setting `capability:univariate: False`, and
enforcing the tag showed that the tag is wrong. `_muse.py:210` warns on univariate input and carries
on, saying WEASEL is recommended instead, and the class docstring's own example runs MUSE on
`load_unit_test`, which is univariate, with `use_first_order_differences=False` so it stays that way.
A class that warns and proceeds is capable; the tag said it was not. Corrected to `True`, so the
doctest passes and the recommendation stays a recommendation.

If MUSE is meant to refuse univariate input rather than warn, that is the opposite fix and a
deliberate behaviour change, and I would rather you made that call than have me guess it. Either way
the tag and the code should agree, and until now nothing checked that they did.

No.

#### Any other comments?

The new test sets the tag on a mock classifier rather than on `MUSE`, so it tests the base class
rather than one estimator's configuration. Reverting `_base_collection.py` and keeping it gives one
failure; putting MUSE's tag back to `False` turns its doctest red.


TonyBagnall:
thanks, this has also highlighted a hole in the testing, as MUSE is tested on univariate as well, will need to adjust tests. @MatthewMiddlehurst any thoughts, your domain really :)

mkzung:
Thanks. I went looking for the tests that would need adjusting and I think the answer is
none of them, so here is what I checked in case it saves you the same pass.

The checking framework already respects the tag. `_get_datatypes_for_estimator` gates the
univariate datatypes behind `if univariate:`, so with the tag back to `False` and the
enforcement in place, `check_estimator(MUSE)` is 21 checks and 21 pass. Nothing in the
battery hands MUSE univariate data.

MUSE's stored probabilities are in `multivariate_expected_results` only, so
`check_classifier_against_expected_results` returns early on the UnitTest run with "no
stored results" rather than comparing anything.

TEASER builds `MUSE(support_probabilities=True)` only when `n_channels_ > 1` and `WEASEL`
otherwise, so it never passes univariate data down.

`test_muse.py` is multivariate, `n_channels=3`.

The one place univariate data reaches MUSE is its own class docstring, which runs
`load_unit_test` with `use_first_order_differences=False`. With the tag at `False` and the
enforcement on, `fit` raises "Data seen by instance of MUSE has univariate series, but MUSE
cannot handle these characteristics" and the doctest fails. With the tag at `True` it
passes. That doctest is the whole of the exposure.

On whether the branch is the right shape at all: `BaseSeriesEstimator._check_X` already
does this, at `_base_series.py:204`, in the same four-tag block and raising the same way.
The collection base was the one that skipped it, so this is the two bases agreeing rather
than a new rule.

I also checked the three other estimators that declare `capability:univariate: False`,
`GreedyGaussianSegmenter`, `HidalgoSegmenter` and `InformationGainSegmenter`. All three are
series estimators, so the series base already refuses their univariate input. I bypassed
the gate to see whether the tag was wrong the way MUSE's was: on a series with a clean
change at t=60, GGS returns a boundary at 0, IGTS returns 2, and Hidalgo raises on a
missing attribute. Their tags are right.

So MUSE is the only disagreement between a tag and the code behind it, and the docstring is
the only test that would have caught it.

MatthewMiddlehurst:
Happy with the check. Think it is both new and rarely used currently, so must have slipped through! Thanks.

@patrickzib will let you call on the MUSE tag. IMO wouldnt allow univariate but probably some other changes required.

mkzung:
Thanks. Leaving the MUSE tag alone here then, so this stays just the enforcement, and @patrickzib can decide whether the tag or the estimator is the thing to change.

patrickzib:
> Thanks. Leaving the MUSE tag alone here then, so this stays just the enforcement, and @patrickzib can decide whether the tag or the estimator is the thing to change.

Hi thank you for the changes. 

I think, it is fine to allow MUSE to run on univariate data. There is no technical reason to prevent it.

mkzung:
Thanks, that settles it. The tag stays `True`, which is what this branch already has, so nothing needs changing and the pull request is just the enforcement plus MUSE keeping the capability it has.

For the record on why the flip and the enforcement are one commit: with the tag at `False` and the base class enforcing it, MUSE's own docstring example raises

    ValueError: Data seen by instance of MUSE has univariate series, but MUSE cannot
    handle these characteristics.

`test_muse.py` passes either way, so the unit tests never see it. The `doctests` job does: `pr_pytest.yml:161` runs `python -m pytest -n logical --doctest-only`, and narrowed to that one file it goes 1 passed to 1 failed on the tag alone.

52 pass in `test_base_collection.py`.

patrickzib:
> Thanks, that settles it. The tag stays `True`, which is what this branch already has, so nothing needs changing and the pull request is just the enforcement plus MUSE keeping the capability it has.

Yes, all good with me. 

If @MatthewMiddlehurst is fine with the other changes, this PR is good to go in.

MatthewMiddlehurst (approved):

PR #3746, opened 2026-08-15
[BUG] CollectionToSeriesWrapper: return a series, not a collection of one

mkzung:
#### Reference Issues/PRs

Follows #3740, where you asked for a look at coverage on the series transformers.

`CollectionToSeriesWrapper` exists to let a collection transformer take a single series. It
reshapes the 2D series into a collection of one case going in, and nothing reshaped the result
back, so a series transformer handed the caller a 3D array:

```python
X = np.random.normal(size=(1, 20))
CollectionToSeriesWrapper(MockCollectionTransformer()).fit_transform(X).shape
# (1, 1, 20)
```

Each of the three methods that reshape on the way in now unwraps the single case on the way out.

That unwrap is guarded rather than a plain `[0]`, because a collection transformer need not
return a collection. `SFAWhole.transform` returns a tuple of two arrays, and indexing it keeps
the first and drops the second in silence. Only a 3D array holding one case, or a list of one,
is unwrapped; anything else is returned untouched, so `SFAWhole` through the wrapper raises
exactly as it does on `main`.

I came here from the coverage question rather than the other way round. The only test on this
file checks tag inheritance and never calls `fit`, `transform`, `fit_transform` or
`inverse_transform`, so all four sat at 0 and nobody had looked at the shape. The docstring
example prints the shape now too, which is the thing this class exists to get right.

No.

#### Any other comments?

`inverse_transform` gets the same treatment and no test. `CollectionId` is the only collection
transformer with an inverse that is not this wrapper, and through the wrapper it raises
`NotFittedError`: with `fit_is_empty` true the wrapper calls `self.transformer`, the instance
the caller passed, which it never fits. That looks like a separate defect.

Two of the low coverage numbers nearby are not gaps. `_yeojohnson.py` reads 52% with numba on
and 96% under `NUMBA_DISABLE_JIT=1`, since coverage cannot see inside an `njit` kernel, and
`_matrix_profile.py` reads 50% only where `stumpy` is missing and its test skips. The package
sits at 92% rather than the 73% a plain run reports. `_collection_wrapper.py` at 54% was the
figure that meant what it said.

### DIFF
--- a/aeon/transformations/series/_collection_wrapper.py
+++ b/aeon/transformations/series/_collection_wrapper.py
@@ -16,10 +16,8 @@ from aeon.transformations.series.base import (
 def _single_case(Xt):
     """Unwrap a collection of one case, and leave anything else alone.
 
-    The wrapper feeds the collection transformer one case, so a collection
-    comes back holding one. Not every collection transformer returns a
-    collection though: SFAWhole returns a tuple of two arrays, and indexing
-    that would keep the first and drop the second.
+    Not every collection transformer returns a collection. SFAWhole returns a
+    tuple of two arrays, and indexing that drops the second in silence.
     """
     if isinstance(Xt, np.ndarray) and Xt.ndim == 3 and Xt.shape[0] == 1:
         return Xt[0]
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
LGTM, thanks for fixing this

PR #3747, opened 2026-08-15
[BUG] Fix Catch22 TypingError on float32 input with outlier features

### LINKED ISSUE #3721: [BUG] Catch22 and 6 dependent estimators raise TypingError on float32 input
### Describe the bug

Will appear after #3706 is fixed.

_transform_case_features pre-declares typed placeholders before its feature loop. The outlier_series placeholder is fixed to float64, while its real value comes from z_normalise_series_with_mean(series, smean), whose dtype follows the input. When the two differ, numba cannot unify the branches:

```
Cannot unify array(float64, 1d, C) and array(float32, 1d, C) for 'outlier_series.2'
```

Affected: Catch22 (transformations.collection.feature_based), Catch22Classifier (classification.feature_based), Catch22Clusterer (clustering.feature_based), CanonicalIntervalForestClassifier and DrCIFClassifier (classification.interval_based), RISTClassifier and HIVECOTEV2 (classification.hybrid).

```python
import numpy as np
from aeon.transformations.collection.feature_based import Catch22
X = np.random.default_rng(0).standard_normal((6, 1, 40)).astype(np.float32)
Catch22().fit_transform(X)      # TypingError
```

### Expected results

No typing error

### Actual results

typing error

### Suggested fix

_No response_

### Versions

_No response_

Nandinisingh07:
#### Reference Issues/PRs
Fixes #3721

`Catch22` raised a `numba.core.errors.TypingError` for any float32 input
when `outlier_norm=True` (the default), via the compiled 3D-input path
`_transform_case_numba`.

The `outlier_series` placeholder used for features 13/14
(`outlier_include` / `outlier_include_n`) was hardcoded to `np.float64`.
When the real z-normalised series was computed for a float32 input series,
numba's nopython mode could not unify the float64 placeholder type with
the float32 array being assigned to it, and compilation failed.

The fix derives the placeholder's dtype from the input series itself via
`series[:0] / 1`, matching the same dtype-promotion rule already used by
`_as_normalised_float` elsewhere in the codebase (float32 stays float32,
every other dtype normalises to float64).

No.

- Correctness of the dtype-promotion fix (`series[:0] / 1`) versus the
  original `np.empty(0, np.float64)` placeholder, and whether this
  aligns with `_as_normalised_float`'s documented promotion rule.
- Whether the regression test adequately covers the failure mode
  (float32 input, `outlier_norm=True` and `outlier_norm=False`).

#### Any other comments?

Verified manually against all 6 estimators listed in #3721 as affected:
`Catch22`, `Catch22Classifier`, `Catch22Clusterer`,
`CanonicalIntervalForestClassifier`, `DrCIFClassifier`, `RISTClassifier`,
`HIVECOTEV2` — all pass cleanly with no dtype errors on float32 input.

#### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF],
  [DEP] or [GOV]
- [x] Have you updated the tests to account for these changes?
- [x] Have you written new tests for your core changes, as applicable?
- [ ] Have you updated the documentation to account for these changes?


Nandinisingh07:
Hii!
@MatthewMiddlehurst @TonyBagnall @baraline, just following up on this PR. All checks are now passing. Whenever you get a chance, I’d really appreciate a review. Thank you! 😊

aeon/transformations/collection/feature_based/tests/test_catch22.py
```diff
@@ -1521,3 +1521,24 @@ def track_n(s):
...
+
+def test_catch22_float32_output_dtype():
+    """Test Catch22 does not raise on float32 input with outlier features.
+
+    Regression test: the outlier_series placeholder in _transform_case_numba was
+    hardcoded to float64, which numba's nopython mode could not unify with a
+    float32 z-normalised series produced for features 13/14 (outlier_include /
+    outlier_include_n), raising a TypingError for float32 input.
+    """
+    X = np.random.default_rng(0).standard_normal((6, 1, 40)).astype(np.float32)
```
baraline: This needs to be replaced with the existing testing utils (make_example_3d_numpy) to generate 3D data. You need to keep the float32 cast

aeon/transformations/collection/feature_based/_catch22.py
baraline: This will be float64 by default, extracting/storing features with float64 precision on a float32 input is against our goal I think.
Any thoughts on this ? @TonyBagnall @MatthewMiddlehurst

aeon/transformations/collection/feature_based/_catch22.py
```diff
@@ -1934,7 +1934,10 @@ def _transform_case_numba(
         # typed placeholders; real values are assigned before first use. These
         # must be dtype-fixed (not aliases of series) so the kernel also
         # compiles for integer input arrays.
-        outlier_series = np.empty(0, np.float64)
+        # / 1 applies the same dtype promotion z_normalise_series_with_mean
+        # uses internally (float32 stays float32, everything else becomes
+        # float64), so this placeholder always matches the real value's dtype.
+        outlier_series = series[:0] / 1
```
baraline: Even if we override the value afterward, I'd rather want a helper function to init an empty array to the correct type/shape.

baraline (changes requested):
Hi, thanks for working on this, some comments

aeon/transformations/collection/feature_based/tests/test_catch22.py
Nandinisingh07: Updated to use make_example_3d_numpy with the float32 cast kept — see db44123.

aeon/transformations/collection/feature_based/_catch22.py
Nandinisingh07: Fixed in 7a5a783 — c22 is now built via _zeros_promoted(X[0], n_feats * len(X)), so it keeps the input's dtype instead of defaulting to float64.

aeon/transformations/collection/feature_based/_catch22.py
Nandinisingh07: Added the _zeros_promoted helper as requested in 7a5a783 — it's used both here and for the c22 buffer above.

Nandinisingh07:
@baraline @TonyBagnall @MatthewMiddlehurst 
I’ve pushed both requested changes. The test now uses `make_example_3d_numpy`, and `_catch22.py` now has a `_zeros_promoted` helper that’s used for both the `outlier_series` placeholder and the `c22` output buffer, so the input dtype is preserved.

All checks are passing. Ready for another look whenever you get a chance!

TonyBagnall:
@baraline I think your three review points have now been addressed, could you take another quick look when you get a chance and confirm whether or not you are happy with the changes? I dont really have a strong take or opinion on this

baraline (approved):
LGTM, thanks for the changes !

aeon/transformations/collection/feature_based/tests/test_catch22.py
```diff
@@ -1521,3 +1522,26 @@ def track_n(s):
...
+
+    # outlier_norm=True exercises the fixed code path (features 13/14)
+    c22 = Catch22(outlier_norm=True, replace_nans=True)
+    result = c22.fit_transform(X)
+    assert result.shape == (6, 22)
+
+    # outlier_norm=False takes the other branch -- should still work
+    c22_no_norm = Catch22(outlier_norm=False, replace_nans=True)
+    result_no_norm = c22_no_norm.fit_transform(X)
+    assert result_no_norm.shape == (6, 22)
```
baraline, line 1547: Forgot a point sorry, could you also add a test to check that the OUTPUT is float 32 when passed float 32 and float 64 when passed float 64 ?

baraline (changes requested):
@Nandinisingh07 I Forgot one small thing on the testing sorry, we will be good to go after that.

TonyBagnall:
@Nandinisingh07 one small fix and we are good to go

PR #3748, opened 2026-08-16
[DOC] Clear stale ImportError output from data_unequal notebook

### LINKED ISSUE #3385: [DOC] Example cell fails with ImportError and renders traceback on page (all_estimators import)

**Description**

When viewing the documentation page on handling unequal length time series, the example code cell fails to execute and the rendered documentation shows an ImportError instead of expected output.

**Location**
https://www.aeon-toolkit.org/en/stable/examples/datasets/data_unequal.html#Loading-data-with-unequal-length-series-or-missing-values

**Expected behavior**
The documentation example should execute successfully and display a list of estimators supporting capability:unequal_length.

**Environment note**
Occurs on documentation website (not local environment)
Likely during docs build execution 

This appears to be a documentation execution failure rather than a user-side installation issue, as the traceback is rendered directly in the docs page.

_No response_

AnkitKumarIISERB:
Closes #3385

The rendered documentation page was showing a stale ImportError traceback 
from a previous notebook execution with an incompatible sklearn version. 
The underlying code in `_continuous_interval_tree.py` has since been updated 
to use `validate_data` from a compatible sklearn version.

This PR clears the cached cell output so the docs page no longer renders 
the error traceback.


TonyBagnall:
Thanks for looking at this. I don't think simply clearing the cached output fixes the issue though. Our docs use nbsphinx_execute = "never", so notebooks are not rerun during the documentation build

Could you rerun this notebook against the current aeon environment and commit the refreshed outputs instead? In particular, the all_estimators(...) cell should display the expected estimator list rather than have outputs: [].

It also looks like this change has cleared the outputs from all the other cells in the notebook, which we should retain.

TonyBagnall (changes requested):
see comment

AnkitKumarIISERB:
Done! set up a Python 3.12 conda environment (numba crashes the kernel 
on Python 3.13), reran the notebook, and committed the fresh outputs. 
The all_estimators cell now shows the expected estimator list and all 
5 code cells have fresh outputs. No ImportError or traceback remains.

Note: the failing CI check is examples/visualisation/plotting_estimators.ipynb 
which is unrelated to this PR — data_unequal.ipynb passes in 7s.

TonyBagnall (approved):
Thanks, this addresses my previous comment. The notebook now has the refreshed outputs and looks good to me.

PR #3749, opened 2026-08-16
BUG: row_* functions in numba stats upcast float32 input to float64

### LINKED ISSUE #3724: [MNT] utils/numba/stats.py row_* functions return float64 arrays for float32 input
The 13 row_* functions allocate with an untyped np.zeros(X.shape[0]). Low priority: the output is one value per channel, so this is dtype consistency rather than a memory problem.

row_mean, row_std, row_quantile, row_median, row_quantile25, row_quantile75, row_numba_min, row_numba_max, row_slope, row_iqr, row_ppv, row_count_above_mean, row_count_mean_crossing.

The scalar counterparts (mean, std, quantile, …) would also need to be changed, but need to test the impact.

Nandinisingh07:
#### Reference Issues/PRs
Closes #3724

The `row_*` functions in `aeon/utils/numba/stats.py` (`row_mean`,
`row_count_mean_crossing`, `row_count_above_mean`, `row_median`,
`row_quantile25`, `row_quantile75`, `row_std`, `row_numba_min`,
`row_numba_max`, `row_slope`, `row_iqr`, `row_ppv`) allocated their
output array with `np.zeros(X.shape[0])`, which always defaults to
`float64` regardless of the input array's dtype. For `float32` input
this silently upcast the result to `float64`, doubling memory usage
for the returned array.

This PR replaces the untyped allocation with one derived from the
input dtype, so `float32` input now returns `float32` output and
`float64` input continues to return `float64`.

Also includes a cherry-picked fix for a related `numba.TypingError`
in `Catch22._transform` (`_outlier_include`), which surfaced while
verifying this change against downstream estimators that consume
these row functions (it failed to unify `float64`/`float32` arrays
for `outlier_series`).

No.

#### Any other comments?

Added a regression test, `test_row_functions_preserve_dtype`, in
`aeon/utils/numba/tests/test_stats.py` covering all affected
functions for both `float32` and `float64` input.

Verified no downstream regressions by running the fix against
`CanonicalIntervalForestClassifier`/`Regressor`, `DrCIFClassifier`/
`Regressor`, `SupervisedTimeSeriesForest`, `RISTClassifier`,
`SevenNumberSummary`, `RandomIntervals`, and `SupervisedIntervals` —
all pass. Also ran the full `test_stats.py`,
`interval_based`, and `feature_based`/`catch22` test suites locally
(42 passed, 6 skipped as expected).


TonyBagnall (changes requested):
Thanks. The Catch22 fix itself looks sensible to me: deriving the placeholder from series[:0] / 1 gives the same promotion behaviour as the z-normalisation path, so float32 and integer inputs should compile consistently.

However, the PR description and title no longer match the PR . Could you rewrite the description to reflect the changes

Issue #3723, opened 2026-08-09
[BUG] Six transformers return float64 for float32 input


baraline:
### Describe the bug

Each returns a data-shaped float64 array for float32 input. One or two lines per method, so this can be one PR or split per file.

Estimator | File:line | Function | Cause
-- | -- | -- | -- 
DWTTransformer | aeon/transformations/collection/_dwt.py:54 | _transform | np.array(X, dtype=np.float64), unconditional 
Resizer | aeon/transformations/collection/unequal_length/_resize.py:107 | _transform | np.zeros((x.shape[0], length)) 
SupervisedIntervals | aeon/transformations/collection/interval_based/_supervised_intervals.py:254 | _transform | np.zeros((X.shape[0], len(transform))) 
ClaSPTransformer | aeon/transformations/series/_clasp.py:66 | _sliding_dot_product | untyped np.zeros inside np.concatenate 
PLASeriesTransformer | aeon/transformations/series/_pla.py | _transform / _sliding_window | segments accumulated as Python-float lists, then np.asarray 
SlopeTransformer | aeon/transformations/collection/_slope.py:73 | _transform | Python-float list, then np.asarray 

References for review — same packages, already correct: DownsampleTransformer, Truncator, PAA, Normalizer, Centerer, MinMaxScaler, SimpleImputer, Tabularizer, PeriodogramTransformer, DifferenceTransformer, MovingAverage, SavitzkyGolayFilter, GaussianFilter.

4nmus:
Hello, I’d like to work on DWTTransformer and ensure float32 input preserves dtype. Is anyone already working on that part? I read the contribution guide.

baraline:
Hey, you are clear to go ! Noone is working on this right now. Looking at how we dealt with some of the issues in #3707  might be useful. You can post message on discord or on the issue here if you have any questions.

4nmus:
Thank you!  #3707 indeed helped. It seems that  `dtype = np.float32 if X.dtype == np.float32 else np.float64` is enough with additional tests added to check that float32 => float32 and other dtypes => float64 for _dwt.py in tests. If it's correct and expected behavior, i can follow up with other files as well and make a PR.

4nmus:
@aeon-actions-bot assign @4nmus

PR #3752, opened 2026-08-17
[BUG] Preserve float32 dtype in affected transformers.

4nmus:
#### Reference Issues/PRs

Fixes part of #3723.

Preserves `float32` output precision in the following transformers:
DWTTransformer
Resizer
SlopeTransformer
SupervisedIntervals
PLASeriesTransformer

Previously, these transformers could unnecessarily promote float32 input to float64 through explicit casts or default NumPy allocations.
The changes preserve the following dtype behavior:
-float32 input produces float32 output
-float64 input produces float64 output
-integer input is promoted to float64

Regression tests were added or updated to verify dtype behavior for float32, float64, and integer inputs.

ClaSPTransformer, which is mentioned in #3723, is not included in this PR because its existing float64 conversion is related to a separate Numba typing constraint.

No.

#### Any other comments?

ClaSPTransformer still returns float64 since it forces the input to be float64 due to Numba. 

##### For all contributions


aeon/transformations/collection/interval_based/tests/test_intervals.py
```diff
@@ -56,3 +59,39 @@ def test_supervised_transformers():
...
 
     assert X_t.shape == (X.shape[0], 8)
+
+
+@pytest.mark.parametrize(
+    "dtype, expected_dtype",
+    [
+        (np.float32, np.float32),
+        (np.float64, np.float64),
+        (np.int64, np.float64),
```
baraline: need to also add int32, float64

aeon/transformations/collection/interval_based/_supervised_intervals.py
```diff
@@ -208,7 +208,8 @@ def _fit_transform(self, X, y=None):
         for i in range(1, self.n_intervals):
             Xt = np.hstack((Xt, transformed_intervals[i]))
 
-        return Xt
+        out_dtype = np.float32 if X.dtype == np.float32 else np.float64
+        return Xt.astype(dtype=out_dtype, copy=False)
```
baraline, line 212: We shouldn't neet to cast the input here, it must natively used float32 (or 64) in the generate intervals function.

What we want is that if we have a float 32 input, computation are made on float 32 to benefit from the reduced memory footprint (and computation time to some extent)

aeon/transformations/collection/interval_based/_supervised_intervals.py
```diff
@@ -251,7 +252,8 @@ def _transform(self, X, y=None):
             for i in range(len(self.intervals_))
         )
 
-        Xt = np.zeros((X.shape[0], len(transform)))
+        out_dtype = np.float32 if X.dtype == np.float32 else np.float64
+        Xt = np.zeros((X.shape[0], len(transform)), dtype=out_dtype)
```
baraline: Same comment as above, we want the interval computation to use 32 bit precision, not simply cast the output.

So `_transform_intervals` itself should return the appropriate dtype already.

aeon/transformations/collection/tests/test_dwt.py
```diff
@@ -111,3 +111,20 @@ def test_dwt_performs_correcly_along_each_dim():
...
     )
     np.testing.assert_array_almost_equal(res, orig)
+
+
+@pytest.mark.parametrize(
+    "dtype, expected_dtype",
+    [
+        (np.float32, np.float32),
+        (np.float64, np.float64),
+        (np.int64, np.float64),
```
baraline: similar to before, add int32, float64

aeon/transformations/collection/tests/test_dwt.py
```diff
@@ -111,3 +111,20 @@ def test_dwt_performs_correcly_along_each_dim():
...
+    ],
+)
+def test_dwt_preserves_float_precision(dtype, expected_dtype):
+    """Check dwt preserves float32 and promotes proper dtype output."""
+    X = np.arange(16, dtype=dtype).reshape(2, 1, 8)
```
baraline: use the testing utils functions to generate testing data, as you did in the inerval tests (make_example_3d_numpy)

aeon/transformations/collection/tests/test_slope_transformer.py
```diff
@@ -75,3 +75,23 @@ def test_slope_performs_correcly_along_each_dim():
...
     )
     np.testing.assert_array_almost_equal(res, orig, decimal=5)
+
+
+@pytest.mark.parametrize(
+    "dtype, expected_dtype",
+    [
+        (np.float32, np.float32),
+        (np.float64, np.float64),
+        (np.int64, np.float64),
```
baraline: also int32, float64

aeon/transformations/collection/unequal_length/tests/test_resize.py
```diff
@@ -87,3 +87,23 @@ def test_incorrect_arguments():
...
     with pytest.raises(ValueError, match="resized_length must be"):
         resizer.fit_transform(X)
+
+
+@pytest.mark.parametrize(
+    "dtype, expected_dtype",
+    [
+        (np.float32, np.float32),
+        (np.float64, np.float64),
+        (np.int64, np.float64),
```
baraline: also int32,float64

aeon/transformations/collection/tests/test_slope_transformer.py
```diff
@@ -75,3 +75,23 @@ def test_slope_performs_correcly_along_each_dim():
...
+    ],
+)
+def test_slope_preserves_float_precision(dtype, expected_dtype):
+    """Check that Slope preserved float32 and promotes proper dtype output."""
+    X = np.array(
+        [[[4, 6, 10, 12, 8, 6, 5, 5]]],
+        dtype=dtype,
+    )
```
baraline: again, use the testing utils to generate data

aeon/transformations/collection/unequal_length/tests/test_resize.py
```diff
@@ -87,3 +87,23 @@ def test_incorrect_arguments():
...
+    ],
+)
+def test_resizer_preserves_float_precision(dtype, expected_dtype):
+    """Test Resizer preserved float32 and promotes proper dtype output."""
+    X = [
+        np.array([[0, 1, 2, 3]], dtype=dtype),
+        np.array([[0, 2, 4, 6, 8]], dtype=dtype),
```
baraline, line 100: I don't think the current testing util to generate variable length list allows dtype args, so this one is fine for now.

baraline (changes requested):
Thanks for working on this, the issue is a bit more complex than simply casting the output, we want the internal computation to be performed in the appropriate precision, to benefit from the reduced memory footprint.

I didn't comment on _dwt, _slope and _pla as the issues are the same that those pointed out in intervals.

If you need any pointers don't hesitate to ask questions.

4nmus:
Thank you for the review! I updated implementation so that computations itself preserve float32 within in `_pla`, `_supervised_intervals` and `_slope`. However, there is an issue within _supervised_intervals:` row_*` feature funcitons force float64 regardless of changes, it's already covered in #3724 / #3749  so I kept the SupervisedIntervals working arrays and interval buffers in X.dtype and just normalized feature outputs at that boundary. I also added fixedd and added int32 in tests. I hope that is what was missing.

aeon/transformations/collection/interval_based/tests/test_intervals.py
4nmus: Added

aeon/transformations/collection/interval_based/_supervised_intervals.py
4nmus: _transform_intervals now returns values in the working X.dtype, and _transform allocates its result directly with that dtype instead of correcting the dtype

aeon/transformations/collection/interval_based/_supervised_intervals.py
4nmus: I removed final cast. The input is promoted once and interval  buffers now use X.dtype, so float32 is preserved throughout the SupervisedIntervals.

aeon/transformations/collection/tests/test_dwt.py
4nmus: added

aeon/transformations/collection/tests/test_slope_transformer.py
4nmus: Updated to use make_example_3d_numpy

aeon/transformations/collection/tests/test_slope_transformer.py
4nmus: added

aeon/transformations/collection/tests/test_dwt.py
4nmus: now uses  make_example_3d_numpy

baraline:
Great ! Overall looks fine but I'll try to have a closer look at it this week. 

Thanks for continuing to work on this.

baraline:
@MatthewMiddlehurst, do you think it would be worth having a internal estimator attribute (would tag system be appropriate to declare the capability ?) with the input precision (32/64) set during fit and reused across all the estimator declarations ?
It would avoid using the same dtype everytime and would make a clean standard across all estimators to declare and use the precision.

We can set it during input preprocessing similarly to how @4nmus wrote it in the test :
```
internal_dtype = np.float32 if input_dtype == np.float32 else dtype = np.float64
```

MatthewMiddlehurst:
Yes could be useful. Lots of possible ways to handle dtypes which we haven't really done anything with really.

aeon/transformations/collection/interval_based/_supervised_intervals.py
```diff
@@ -241,6 +241,8 @@ def _fit(self, X, y=None):
         return self
 
     def _transform(self, X, y=None):
+        X = X / 1
+
```
MatthewMiddlehurst, line 245: why?

aeon/transformations/collection/_slope.py
```diff
@@ -62,17 +60,20 @@ def _transform(self, X, y=None):
         # Get information about the dataframe
         n_cases, n_channels, n_timepoints = X.shape
         self._check_parameters(n_timepoints)
-        full_data = []
+
+        X = X / 1
```
MatthewMiddlehurst, line 64: same here (and everywhere else)

aeon/transformations/collection/interval_based/_supervised_intervals.py
4nmus: This would establish proper dtype before computation. Float32 and float64 is preserved while int input promoted to float64 for further computation using numpy's promotion rules.It was used in #3707 like return X / 1 to preserve the dtype so i followed same approach as in `_as_normalized_float` in #3707. It does not cast the final result back to requested dtype. I can also change it to a helper function but it's seems to be a cleaner version.

MatthewMiddlehurst: not really sure I like the X/1 thing for both really, seems a bit hacky.

4nmus: Okay, i can do it with `dtype = np.result_type(X.dtype, np.float32)`. So it's going to be something like that:
```
dtype = np.result_type(X.dtype, np.float32)
Xt = np.empty((n_cases, n_channels, n_intervals), dtype=dtype)
Y = np.arange(interval_size, dtype=dtype)
```
Is this okay?

PR #3753, opened 2026-08-20
[BUG] Fix `MUSE` with `scikit-learn` 1.8 and drop deprecated parameters

MatthewMiddlehurst:
Fixes muse which was missed in #3250 which pushed a fix for WEASEL.

Removes deprecated parameters. This is functionally the same as live, as the current options do not change defaults or any internal operations.


patrickzib (approved):
Thank you, @MatthewMiddlehurst

PR #3754, opened 2026-08-20
[ENH] Replace the esig soft dependency with RoughPy

MatthewMiddlehurst:
This PR swaps `esig` for `roughpy`. `esig` is no longer maintained, is mostly a wrapper for `roughpy`, and has caused issues in the past for installs. This allows us to keep the algorithms which use it, and hopefully is a bit less troublesome. 

Includes some `esig` functions which are required as a bridge, and fixes the previous monkey patch for an unfixed bug.

AI found some issues with ARM Linux installs, so excluded that for now.


TonyBagnall (approved):
lgtm

PR #3755, opened 2026-08-22
[BUG] Guard default RidgeClassifierCV against LAPACK SVD overflow

### LINKED ISSUE #3737: [BUG] Large memory jobs precision issue
### Describe the issue

I'm running classifiers on the multiverse, including large ones. Many classifiers OOM, but some fail in code due to precision issues. 
                                                                                                                            
  Several aeon classifiers fail on valid large classification datasets because their                                          
  default `RidgeClassifierCV` backend attempts an SVD on a transformed feature                                                
  matrix containing more elements than 32-bit LAPACK can index.                                                               
                                                                                                                              
  This is not an out-of-memory error. SciPy raises the following exception before                                             
  the classifier completes:                                                                                                   
                                                                                                                              
  ```text                                                                                                                     
  ValueError: Indexing a matrix of 3365692186 elements would incur an integer                                                 
  overflow in LAPACK. Try using numpy.linalg.svd instead.                                                                     
```
  For RDSTClassifier on AustraliaRainfall_disc, the matrix passed to ridge                                                    
  fitting corresponds to approximately 112,186 cases and 30,000 transformed                                                   
  features. The error reports 3,365,692,186 elements, including the augmented                                                 
  intercept column.  

 ### Confirmed affected combinations                                                                                         
                                                                                                                              
   aeon classifier               Dataset                   Outcome                                                            
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━                                           
   RDSTClassifier                AustraliaRainfall_disc    LAPACK integer overflow                                            
  ────────────────────────────  ────────────────────────  ─────────────────────────                                           
   RDSTClassifier                Tiselac                   LAPACK integer overflow                                            
  ────────────────────────────  ────────────────────────  ─────────────────────────                                           
   MultiRocketHydraClassifier    Tiselac                   LAPACK integer overflow                                            
  ────────────────────────────  ────────────────────────  ─────────────────────────                                           
   RocketClassifier              AustraliaRainfall_disc    LAPACK integer overflow                                            
                                                                                                                              
  MultiRocketHydraClassifier on AustraliaRainfall_disc also exhausted 128 GB                                                  
  while constructing or holding its transformed data. If it reaches the default 

  ### Relevant traceback                                                                                                      
                                                                                                                              
  File ".../aeon/classification/shapelet_based/_rdst.py", line 223, in _fit                                                   
      self._estimator.fit(X_t, y)                                                                                             
  File ".../sklearn/linear_model/_ridge.py", line 2068, in _svd_decompose_design_matrix                                       
      U, singvals, _ = linalg.svd(X, full_matrices=0)                                                                         
  File ".../scipy/linalg/_decomp_svd.py", line 152, in svd                                                                    
      raise ValueError(...)                                                                                                   
  ValueError: Indexing a matrix of 3365692186 elements would incur an integer                                                 
  overflow in LAPACK. Try using numpy.linalg.svd instead.                                                                     
                                                                                                                              
  The corresponding failures in RocketClassifier and                                                                          
  MultiRocketHydraClassifier also occur when their transformed data are passed                                                
  to RidgeClassifierCV.  

### Expected behaviour                                                                                                      
                                                                                                                              
  The default classifiers should either:                                                                                      
                                                                                                                              
  1. support a scalable fitting route for feature matrices above the LAPACK                                                   
     indexing limit;                                                                                                          
                                                                                                                              
  2. detect the unsupported matrix size before fitting and raise an informative                                               
     aeon-level error; or                                                                                                     
                                                                                                                              
  3. document and expose a supported alternative estimator.                                                                   
                                                                                                                              
  A silent estimator change may be undesirable because it could change published                                              
  algorithm behaviour and reproducibility.                                                                                    
                                                                                                                              
  RDSTClassifier and RocketClassifier already expose an estimator parameter.                                                  
  MultiRocketHydraClassifier currently constructs RidgeClassifierCV                                                           
  internally and does not expose an equivalent estimator parameter.   

                                                                              
  The experiments used the classifiers' default aeon configurations, one CPU      
  thread, and up to 128 GB of RAM. Increasing memory does not resolve the         
  confirmed                                                                       
  LAPACK failures because they arise from integer indexing limits rather than     
  memory allocation.                                                              
                                       




_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by aashoday**
#### Reference Issues/PRs
Fixes #3737.

Several classifiers (`RDSTClassifier`, `RocketClassifier`, `MultiRocketHydraClassifier`)
fit a default `RidgeClassifierCV` on transformed feature matrices that can exceed the
32-bit LAPACK indexing limit for SVD, causing an unhelpful scipy `ValueError` on large
datasets.

This PR:
- Adds `check_lapack_svd_safe()` to `aeon.utils.validation`, which raises an
  informative `ValueError` before fitting when the default estimator is used and
  the transformed matrix would overflow 32-bit LAPACK's SVD indexing.
- Wires this check into `RDSTClassifier` and `RocketClassifier`, ahead of their
  default `RidgeClassifierCV` fit.
- Exposes an `estimator` parameter on `MultiRocketHydraClassifier` (previously
  hardcoded to `RidgeClassifierCV`), matching the pattern already used by
  `RDSTClassifier` `RocketClassifier`, and applies the same check.
- The check only runs when the default estimator is in use; behaviour for
  existing custom-estimator users is unchanged.
- `RocketClassifier`'s internal pipeline was split into 
  transform/scale/fit steps (it previously used `make_pipeline`), since the
  check needs the transformed matrix shape before fitting.

No.

#### Any other comments?
This PR targets only the LAPACK indexing bug not the separate 128GB OOM reported for `MultiRocketHydraClassifier` on `AustraliaRainfall_disc`. 

AI was used to help understand parts of the codebase and LAPACK, scipy internals
while working through this fix; all changes were written and reviewed by me.

##### For all contributions

### CODE REVIEW

File: aeon/utils/validation/__init__.py [hunk 1]
```diff
@@ -34,3 +37,38 @@ def check_n_jobs(n_jobs: int) -> int:
         return max(1, os.cpu_count() + 1 + n_jobs)
     else:
         return n_jobs
+
+
+def check_lapack_svd_safe(n_samples: int, n_features: int, estimator_name: str) -> None:
+    """Raise an informative error if a matrix is too large for LAPACK SVD.
+
+    Matrices with more than ``2**31 - 1`` elements may overflow 32-bit integer
```
aadya940, line 45: Not sure how relevant this is in this context but it is not necessary that LAPACK will overflow for more than 2**31 - 1 elements, it is only the case in the default LP64 scipy build. However, now SciPy supports ILP64 builds which can accomodate far large matrices (upto 64 bit integer indices), see [these](https://discuss.scientific-python.org/t/ann-scipy-1-18-0-release/2394) release notes, if we add these checks it will mostly work okay in the default SciPy but will restrict user who specifically use ILP64 SciPy builds.

Therefore, the more robust way would be to query scipy:

```python
import scipy

try:
    config = scipy.show_config(mode='dicts')
    is_ilp64 = config['Build Dependencies']['blas']['cython_blas_ilp64']
    print(f"Is SciPy ILP64? {is_ilp64}")
except KeyError:
    print("Is SciPy ILP64? False (Default LP64 profile active)")
```

PR #3756, opened 2026-08-22
Rename SAX znormalized to znormalize, default to normalizing

### LINKED ISSUE #3678: [MNT] Change the parameter `znormalized` to `znormalize` in SAX transformation

AmjadAAYD:
Fixes #3678. znormalized=True meant input is already normalized (skip normalization), which is confusing since True reads as 'yes, normalize'. Added znormalize (default True, means 'do normalize') as the new parameter. znormalized is kept for backward compatibility with default sentinel 'deprecated', emits a FutureWarning when explicitly passed, and is internally mapped to the equivalent znormalize value (note the meaning is inverted: znormalized=True == znormalize=False).

Updated the one internal caller in _redcomets.py to use znormalize=False (behavior-preserving equivalent of its old znormalized=True call). Updated docstrings and the SAX.inverse_sax error message wording. Updated the existing test asserting the old default (test_sax_default_znormalized_is_true) to reflect the new default, and added a new test for the deprecation warning + value mapping.

Verified: full test_sax.py suite (48 tests) and both redcomets test files pass. Also ran a standalone check that get_params()/clone() still round-trip correctly for both the new and legacy parameter, per sklearn's estimator contract.

### DIFF
--- a/aeon/classification/dictionary_based/_redcomets.py
+++ b/aeon/classification/dictionary_based/_redcomets.py
@@ -280,7 +280,7 @@ class REDCOMETS(BaseClassifier):
             sfa_clfs.append((rf, weight))
 
         sax_transforms = [
-            SAX(n_segments=w, alphabet_size=a, znormalized=True) for w, a in sax_lenses
+            SAX(n_segments=w, alphabet_size=a, znormalize=False) for w, a in sax_lenses
         ]
 
         sax_clfs = []
--- a/aeon/transformations/collection/dictionary_based/_sax.py
+++ b/aeon/transformations/collection/dictionary_based/_sax.py
@@ -3,6 +3,8 @@
 __maintainer__ = []
 __all__ = ["SAX", "_invert_sax_symbols"]
 
+import warnings
+
 import numpy as np
 import scipy.stats
 from numba import get_num_threads, njit, prange, set_num_threads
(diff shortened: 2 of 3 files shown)


AmjadAAYD:
Checked everything, all checks are passing.

AmjadAAYD:
Checked everything, all checks are passing now.

patrickzib:
Hmm, if I am not mistaken, we have two (three including this one) duplicate PRs for this now:

#3680 
#3727

MatthewMiddlehurst:
Yes this is a duplicate, and runs into the same issue as previous. See the review and functionality changes requested in https://github.com/aeon-toolkit/aeon/issues/3680 if you wish to proceed please. This is more then a rename.

TonyBagnall:
perhaps we should change the issue to make it clearer its not just a name change?

TonyBagnall (changes requested):
Thanks, this now implements the intended behaviour from #3678 correctly. znormalize=True means perform z-normalization, znormalize=False means do not normalize, and the deprecated znormalized values are mapped with the inverse semantics. The ReCOMETS change also preserves its existing behaviour.

A few changes before merge please:

1. This will now be deprecated in v1.7.0, so znormalized should be marked for removal in v1.8.0. Please update the warning/docstring accordingly and add the usual TODO, e.g. # TODO: remove 'znormalized' in v1.8.0.

2. #3678 also explicitly asks for the example notebook to be updated. Please update examples/transformations/SAX.ipynb to use znormalize and the new semantics.
Since this introduces a deprecation, please change the PR title to use the [DEP] prefix.

Otherwise the implementation looks good to me.

PR #3760, opened 2026-08-26
Fix merlin standard deviation check being inverted

### LINKED ISSUE #3759: [BUG] Merlin STD check the wrong way around
### Describe the bug

MERLIN checks this on line 87 

```python
       for i in range(X.shape[0] - self.min_length + 1):
            if std(X[i : i + self.min_length]) > AEON_NUMBA_STD_THRESHOLD:
                warnings.warn(
                    "There is region close to constant that will cause the results "
                    "to be unstable. It is suggested to delete the constant region "
                    "or try again with a longer min_length.",
                    stacklevel=2,
                )
```

this check is the wrong way around and will always warn, but warnings are probably supressed in testing. 

run with warnings not supressed

### Expected results

warn on flat

### Actual results

all warnings

### Suggested fix

simple fix, invert test, and also put a break, no point warning repeatedly
### Versions

_No response_

DresdenGman:
In `aeon/anomaly_detection/series/distance_based/_merlin.py`, the condition `std(X[i : i + self.min_length]) > AEON_NUMBA_STD_THRESHOLD` was incorrectly warning when the sequence *wasn't* flat. This commit inverses the check so it correctly warns when a sequence *is* flat. Additionally, added a test case `test_merlin_constant_region_warning` to explicitly verify this fix. Fixes #3759

### DIFF
--- a/aeon/anomaly_detection/series/distance_based/_merlin.py
+++ b/aeon/anomaly_detection/series/distance_based/_merlin.py
@@ -85,13 +85,14 @@ class MERLIN(BaseSeriesAnomalyDetector):
             )
 
         for i in range(X.shape[0] - self.min_length + 1):
-            if std(X[i : i + self.min_length]) > AEON_NUMBA_STD_THRESHOLD:
+            if std(X[i : i + self.min_length]) <= AEON_NUMBA_STD_THRESHOLD:
                 warnings.warn(
                     "There is region close to constant that will cause the results "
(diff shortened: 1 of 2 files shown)


TonyBagnall (changes requested):
thanks, but this test is too much. Could you simplify it? We should only need a short synthetic series containing a constant region and pytest.warns(...) to verify the warning is raised, and possibly another varying series to test it is not raised

PR #3763, opened 2026-08-27
[BUG] Fix inverted MERLIN near-constant std check

### LINKED ISSUE #3759: [BUG] Merlin STD check the wrong way around

VedantMadane:
#### Reference Issues/PRs

Fixes #3759.

MERLIN warned when subsequence `std` was *above* `AEON_NUMBA_STD_THRESHOLD`, i.e. the opposite of a near-constant region. Invert the comparison to `<=` so the warning fires on flat/near-constant windows, and `break` after the first warning so it is not repeated for every window.

Add `test_merlin_constant_region_warning` covering the constant-region path.

No.

#### Any other comments?

Suggested fix from the issue: invert the test and stop after one warning.

##### For all contributions

### DIFF
--- a/aeon/anomaly_detection/series/distance_based/_merlin.py
+++ b/aeon/anomaly_detection/series/distance_based/_merlin.py
@@ -85,13 +85,14 @@ class MERLIN(BaseSeriesAnomalyDetector):
             )
 
         for i in range(X.shape[0] - self.min_length + 1):
-            if std(X[i : i + self.min_length]) > AEON_NUMBA_STD_THRESHOLD:
+            if std(X[i : i + self.min_length]) <= AEON_NUMBA_STD_THRESHOLD:
                 warnings.warn(
                     "There is region close to constant that will cause the results "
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
LGTM, thanks

TonyBagnall:
@all-contributors add @VedantMadane for bug

PR #3768, opened 2026-08-29
[BUG] Accept numeric binary regression targets, reject non-numeric targets

alexbanwell1:
### Reference Issues/PRs

None (spotted while running window-based forecasters on short M4 series).

### What does this implement/fix? Explain your changes.

`check_regression_y` accepted `"continuous"` and `"multiclass"` targets but
**rejected `"binary"`** ones with a `ValueError`. However, a *numeric* target with
only one or two unique values is reported by
`sklearn.utils.multiclass.type_of_target` as `"binary"` even though it is a valid
regression target — e.g. a short or first-differenced integer series whose
windowed targets happen to take only a couple of distinct values. This made
`RegressionForecaster` (and forecasters built on it) crash on such series.

Following review, this PR keys the check on **whether the target is numeric**
rather than on the `type_of_target` label alone:

- numeric targets (continuous, multiclass **and** binary/constant) are accepted,
  with no warning;
- non-numeric targets raise a `ValueError`. `type_of_target` reports strings and
  datetimes as `"binary"`/`"multiclass"` just like integers, so this is enforced
  with an explicit numeric-dtype check. This also closes a gap where non-numeric
  *multiclass* targets (e.g. `datetime64` with 3+ unique values) were previously
  accepted.

### Does your contribution introduce a new dependency? If yes, which one?

No.

### Any other comments?

Tests in `aeon/utils/validation/tests/test_labels.py`:
- numeric targets (multiclass, binary, constant) accepted with **no** warning;
- non-numeric targets (string, `datetime64`, boolean) raise `ValueError`.

Local `test_labels.py` passes (85 passed).

### PR checklist

- [x] The PR title starts with either [ENH], [MNT], [DOC], [BUG], [REF], [DEP] or [GOV].
- [x] I've added unit tests and made sure they pass locally.


TonyBagnall (changes requested):
I think in principle yes, but if it is numeric and binary I dont think we need a warning, and if its categorical and binary, I think its an error?

y = np.array(["2020-01-01", "2020-01-02"], dtype="datetime64[D]")
check_regression_y(y)

should be an error (I think) and its now a warning?

alexbanwell1:
Thanks @TonyBagnall - good points, I've reworked it accordingly:

- **Numeric binary: no warning.** Numeric binary/constant targets are now just accepted as valid regression targets, no `UserWarning`.
- **Categorical binary: error.** The check now requires the target to be numeric (`np.issubdtype(y.dtype, np.number)`) rather than trusting the `type_of_target` label. Your example now raises:
  ```python
  y = np.array(["2020-01-01", "2020-01-02"], dtype="datetime64[D]")
  check_regression_y(y)
  # ValueError: y is not numeric, cannot fit a regressor. ...
  ```
- As a bonus this also closes a gap where non-numeric **multiclass** targets (e.g. datetimes with 3+ unique values) were previously accepted.

Tests updated to cover numeric (accepted, no warning) and non-numeric string/datetime/bool (rejected). I'll update the PR title to match.

alexbanwell1:
Should now have been updated and good to go!

TonyBagnall (approved):
Thanks, lgtm

Issue #3326, opened 2026-02-28
[MNT] tsfresh multithreaded testing failures


MatthewMiddlehurst:
### Describe the issue

Occasionally tsfresh related estimators fail the multithreading test as the produce results different for n_jobs==1 and n_jobs>1. I am not sure whether we can fix this on our side if it is the transform, but needs investigation whether it is the transform or estimators (i.e. regression is the most likely IMO)

### Suggest a potential alternative/fix

Test how often this happens. If it is the transform, open an issue or comment on thier repo.

Amiirhosseini:
Hi — would you like me to investigate this?

**Idea / plan:**
1. Reproduce the flaky `n_jobs==1` vs `n_jobs>1` mismatch for tsfresh-related estimators with a small repeatable test script.
2. Determine whether the divergence comes from the transform layer (`tsfresh` feature extraction) or from downstream estimators (regression/classification path).
3. If it's transform-side, open/comment on an upstream `tsfresh` issue with a minimal repro; if it's aeon-side, propose a fix (e.g. deterministic ordering / seed handling) with tests.
4. Document how often the failure occurs across repeated CI-style runs.

I'm an ECE student (UofC) with Python and testing experience. Please let me know if you'd like me to take this on.

@aeon-actions-bot assign @Amiirhosseini

MatthewMiddlehurst:
Sure, feel free to have a go.

Amiirhosseini:
Hi @MatthewMiddlehurst — investigation update + PR opened.

**Findings:**
1. The flaky failures come from the **tsfresh transform layer** (`tsfresh.extract_features`), not from downstream sklearn estimators. Classifiers/regressors create a fresh `TSFresh` at `fit` time with the correct `n_jobs`.
2. When `n_jobs > 1`, tsfresh parallel extraction can return slightly different floating-point values vs `n_jobs=1` — this is intermittent (hence PR #3325's temporary skip).
3. Locally reproduced the CI multithreading check path on test data; no mismatches in limited runs, but the CI flakes are real.

**Proposed fix (PR #3770):**
- Set `non_deterministic: True` on all tsfresh-based estimators (`TSFresh`, `TSFreshRelevant`, `TSFreshClassifier/Regressor/Clusterer`, `FreshPRINCEClassifier/Regressor`).
[...]

PR #3770, opened 2026-08-29
[MNT] Mark tsfresh estimators non-deterministic for multithreading tests

Amiirhosseini:
## Summary
- Investigated intermittent `check_estimator_multithreading` failures for tsfresh-related estimators (#3326).
- Root cause: `tsfresh` parallel feature extraction (`n_jobs > 1`) can produce slightly different floating-point results vs `n_jobs=1` — this is transform-side, not downstream sklearn estimators.
- Mark all tsfresh-based estimators with `non_deterministic: True` and remove the blanket `EXCLUDED_TESTS` skips so the multithreading check still validates `n_jobs` / `_n_jobs` plumbing without asserting identical outputs.

## Investigation notes
- Reproduced the CI test path locally (`_clone_estimator` + `set_params(n_jobs=2)` + fit/predict) on `EqualLengthUnivariate-Classification-numpy3D` test data — no mismatches in 3 runs with `efficient`/`minimal` feature sets, but CI flakes are intermittent (PR #3325 previously skipped these tests for that reason).
- Downstream classifiers/regressors (FreshPRINCE, TSFreshClassifier) create a fresh `TSFresh` transformer at `fit` time with the correct `n_jobs`, so the divergence originates in `tsfresh.extract_features`, not aeon's sklearn wrappers.

## Test plan
- [x] `pytest aeon/transformations/collection/feature_based/tests/test_tsfresh.py` — pass
- [x] `pytest aeon/classification/feature_based/tests/test_tsfresh.py` — pass
- [ ] CI `test_all_estimators` with `--enablethreading True` (tsfresh soft dependency)

### DIFF
--- a/aeon/classification/feature_based/_fresh_prince.py
+++ b/aeon/classification/feature_based/_fresh_prince.py
@@ -85,6 +85,7 @@ class FreshPRINCEClassifier(BaseClassifier):
         "capability:multithreading": True,
         "capability:train_estimate": True,
         "algorithm_type": "feature",
+        "non_deterministic": True,
         "python_dependencies": "tsfresh",
     }
 
--- a/aeon/classification/feature_based/_tsfresh.py
+++ b/aeon/classification/feature_based/_tsfresh.py
@@ -88,6 +88,7 @@ class TSFreshClassifier(BaseClassifier):
         "capability:multithreading": True,
         "capability:unequal_length": True,
         "algorithm_type": "feature",
+        "non_deterministic": True,
         "python_dependencies": "tsfresh",
     }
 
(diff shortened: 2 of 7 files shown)


Amiirhosseini:
Still waiting on maintainer review — pytest and pre-commit are green on my side. Happy to adjust if anything is needed for #3326.

TonyBagnall:
thanks for this, but I would like to investigate why it produces different results when threaded rather than just tag as non deterministic, since afaik it should be deterministic. Is it a problem with TSFresh package, or just a tiny tolerance issue? What features are effected?

MatthewMiddlehurst:
Agreed that just adding the tag it not really what we want to do here.

TonyBagnall:
going to close this as we dont think adding a tag is the right way to do this, happy to discuss on discord if you want to continue

PR #3782, opened 2026-09-01
[BUG] Install graphviz for the notebook example runs

TonyBagnall:
The run-notebook-examples job has failed on every PR since 27 Aug, on examples/visualisation/plotting_estimators.ipynb:

    plot_network(input_shape=(2, 12), network=mlp, show_layer_names=True)
    ImportError: You must install graphviz ... for `plot_model` to work.

plot_network calls keras plot_model, which needs the graphviz "dot" binary as well as the pydot package. macos-14 runners do not provide it and no workflow installed it, so add it next to the existing bash install in both workflows that run the notebooks.

keras only raises when "IPython.core.magics.namespace" is absent from sys.modules, printing and returning otherwise, which is why this surfaces in CI but not always locally.

Also fix the handler that masked the cause. It replaced keras's precise message with "Either graphviz or pydot should be installed", which is wrong twice over: both are required, not either, and graphviz is a system install rather than a package. It now says so and chains the original.

### DIFF
--- a/.github/workflows/periodic_tests.yml
+++ b/.github/workflows/periodic_tests.yml
@@ -57,6 +57,13 @@ jobs:
           brew install bash
           /opt/homebrew/bin/bash --version
 
+      - name: Install graphviz
+        # keras plot_model needs the graphviz "dot" binary, not just pydot;
+        # plotting_estimators.ipynb calls it through plot_network
+        run: |
+          brew install graphviz
+          dot -V
+
       - name: Setup Python 3.12
         uses: actions/setup-python@v7
         with:
--- a/.github/workflows/pr_examples.yml
+++ b/.github/workflows/pr_examples.yml
@@ -28,6 +28,13 @@ jobs:
           brew install bash
           /opt/homebrew/bin/bash --version
 
+      - name: Install graphviz
+        # keras plot_model needs the graphviz "dot" binary, not just pydot;
+        # plotting_estimators.ipynb calls it through plot_network
+        run: |
+          brew install graphviz
+          dot -V
+
       - name: Setup Python 3.12
         uses: actions/setup-python@v7
         with:
--- a/aeon/visualisation/estimator/_network_plot.py
+++ b/aeon/visualisation/estimator/_network_plot.py
@@ -64,8 +64,11 @@ def plot_network(
                 show_layer_activations=show_layer_activations,
                 dpi=dpi,
             )
-        except ImportError:
-            raise ImportError("Either graphviz or pydot should be installed.")
+        except ImportError as e:
+            raise ImportError(
+                "plot_network requires both the pydot package and a graphviz "
+                "installation providing the 'dot' binary."
+            ) from e
 
     else:
         file_name_ = "model" if file_name is None else file_name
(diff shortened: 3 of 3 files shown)


baraline (commented):
I'm a bit confused, is this some MacOS specific install issue where the dot binary is not shipped alongside pydot ?

TonyBagnall:
> I'm a bit confused, is this some MacOS specific install issue where the dot binary is not shipped alongside pydot ?

it was not installing graphviz on the runners, failing every PR 
https://github.com/aeon-toolkit/aeon/actions/runs/33252924465/job/99264688383?pr=3768

not afaik mac specific, due to some other upgrade

baraline:
If it is Independent of the OS, then shouldn't we add this to the pyproject toml in all extras/docs then ?

TonyBagnall:
dont mind how its done, but its failing all PRs atm.

claude tells me this. 
 plot_network calls keras plot_model, which needs the graphviz dot binary, not just the pydot package. The macos-14 runners don't ship it and no workflow installed it. This adds brew install graphviz alongside the existing bash install in both workflows that run the notebooks — pr_examples.yml and periodic_tests.yml.  

It looks intermittent because keras only raises when "IPython.core.magics.namespace" is absent from sys.modules, printing a message and returning otherwise — so the same missing dependency can pass locally and fail in CI.

Also corrects the handler in _network_plot.py that masked the cause. It replaced keras's precise message with "Either graphviz or pydot should be installed", wrong twice over: both are required rather than either, and graphviz is a system install rather than a package. It now says so and chains the original exception.

I guess thats the change that broke it elsewhere.

TonyBagnall:
just checking about OS specific

TonyBagnall:
to quote the oracle, so yes maybe a wider fix would be better, this is a solution to main failing though, and  seems a minimal solution for now

"The bug isn't OS-specific; the CI failure is — because macOS is the only place the notebooks run.

The failure. run-notebook-examples is the only job that executes notebooks, and it's macos-14 in both workflows. Everything else — pytest, doctests, codecov, pre-commit — is ubuntu-24.04 or the pytest OS matrix, and none of them touch plot_network. So the red X only ever appears on macOS. The docs build doesn't rescue or reproduce it either: docs/conf.py sets nbsphinx_execute = "never", so Read the Docs renders the stored outputs without running anything.   
  
  The underlying problem is cross-platform. Nothing about a missing dot binary is macOS-specific — pydot is a Python package that shells out to graphviz, which is a separate system install on every OS. My own Windows box has pydot 4.0.1 and no dot, and hit exactly the same keras condition; it printed instead of raising only because Python.core.magics.namespace happened to be in sys.modules. Same missing  dependency, different symptom.

Also worth knowing: plot_network guards with _check_soft_dependencies(["tensorflow", "pydot"]), which only checks Python packages and never the dot binary. That's the real gap — the guard passes on any  OS without graphviz and the failure surfaces later inside keras. Installing graphviz in CI fixes the symptom on the one platform that runs notebooks; making that check verify the binary would fix it for users everywhere. I'd keep that out of #3782 to hold the scope, but it's a reasonable follow-up
"

baraline (approved):
Fine with it going through as a minimal fix, but we need an issue to keep track of the wider fix then.

PR #3785, opened 2026-09-03
[MNT] Declare joblib and threadpoolctl as direct dependencies

Obila34:
@-


TonyBagnall (approved):
wow that was quick, 6 mins from posting, thanks :)

PR #3787, opened 2026-09-03
[MNT] Allow scikit 1.9

### LINKED ISSUE #3788: [MNT] Support scikit-learn 1.9 and raise the upper bound
The scikit-learn upper bound has been raised experimentally from <1.9.0 to <1.10.0 #3787 

test suite exposes an incompatibility with the callback API introduced in scikit-learn 1.9.

  ## Minimal reproduction

  import numpy as np
  from sklearn.pipeline import make_pipeline
  from sklearn.preprocessing import StandardScaler

  from aeon.transformations.collection import Tabularizer

  X = np.ones((3, 1, 4))
  make_pipeline(Tabularizer(), StandardScaler()).fit(X)

  This raises:

  AttributeError: 'Tabularizer' object has no attribute '_parent_callback_ctx'

  The error originates from:

  sklearn/callback/_callback_context.py:524
  del sub_estimator._parent_callback_ctx

  ## Cause

  Scikit-learn 1.9 attaches _parent_callback_ctx to sub-estimators while fitting pipelines, searches, and other meta-estimators.

  Aeon estimators call BaseAeonEstimator.reset() when fitting. reset() removes non-hyperparameter instance attributes, including _parent_callback_ctx. Scikit-learn subsequently
  tries to remove the temporary attribute itself and raises AttributeError.

  ## Observed scope

  The PR-scoped suite produced:

  47 failed, 7744 passed, 286 skipped, 10 xfailed, 2 errors

  Representative affected areas include:

  - Aeon transformers inside sklearn.pipeline.Pipeline
  - GridSearchCV, RandomizedSearchCV, and halving searches
  - Rocket and MultiRocket classifiers and regressors
  - SAST and RSAST
  - Signature transforms and SignatureClassifier
  - ElasticEnsemble
  - Generic aeon estimator checks
  - sklearn compatibility tests

  Representative reruns across these areas reproduced the same callback-context error.

  ## Required changes

  - Make BaseAeonEstimator.reset() preserve state temporarily owned by scikit-learn’s callback context.
  - Review the new warning that SVC(probability=True) is deprecated in 1.9 and will be removed in 1.11. This is not currently a test blocker.

  ## Reproduction environment

  Python: 3.12.10
  aeon: 1.5.0 development checkout
  scikit-learn: 1.9.0
  NumPy: 2.3.5
  SciPy: 1.15.3
  Platform: Windows

TonyBagnall:
fixes #3788  
test to generate fails for upgrade to 1.9

1. testing against scikit-learn 1.9 identified a compatibility issue with its new callback context. Scikit-learn temporarily adds _parent_callback_ctx to estimators, but `BaseAeonEstimator.reset()` removed it during  fitting, causing cleanup to raise an `AttributeError`. The fix preserves this scikit-learn-owned attribute during reset and adds regression coverage. 

2. Scikit now parses strings that look like ints, so '1'/'2' is parsed into 1/2, whereas 'A'/'B' remains a string. This happens with `ExtraTrees` when `class_weight="balanced"` is set to true, 
ExtraTreesClassifier(class_weight="balanced").fit(X, np.array(["1","2",...]))→ ValueError: The classes, [1, 2], are not in class_weight
only RSTSF sets this by default, so its the only one to fail, but all are vulnerable. It actually effects random forest too, RandomForestClassifier so effects

`ExtraTreesClassifier: RSTSF,  QUANTClassifier, RISTClassifier, QUANTRegressor, ExtraTreesRegressor`

`RandomForestClassifier: Catch22Classifier, SummaryClassifier, TSFreshClassifier,  SignatureClassifier`

the fix is to switch to swap to `compute_sample_weight` but thats only valid when balanced is True. 

This is what the problem is, basically a bug in scikitlearn imo 

It's a scikit-learn 1.9 bug, and it only bites forest ensembles. Here's the actual mechanism.

  compute_class_weight has three branches. When you hand it the string "balanced", it takes a dedicated branch that uses a LabelEncoder — perfectly safe. When you hand it a dict, it takes this branch (sklearn/utils/class_weight.py:104):
                                                             
 ```python
 for i, c in enumerate(classes):                                                       
      try:
          c = int(c)          # "1" -> 1
      except ValueError:      # `classes` contains strings
          c = str(c)
      if c in class_weight:   # 1 in {"1": 1.0, "2": 1.0}  ->  False
          weight[i] = class_weight[c]
      else:
          unweighted_classes.append(c)
```
  That int(c) coercion is the bug, not the nicest use of exceptions either. Hits in forests because they do this
the forests do with "balanced"
  (sklearn/ensemble/_forest.py:861):
```python
  # Computing class_weight (dict or list) for the "balanced" option.
  class_weight_k_vect = compute_class_weight("balanced", classes=self.classes_[k], ...)
  class_weight_k = {key: val for (key, val) in zip(self.classes_[k], class_weight_k_vect)}
  ...
  expanded_class_weight = compute_sample_weight(class_weight, y_original)
```
  The forest doesn't pass "balanced" through. It round-trips: converts "balanced" into an explicit dict keyed by the original labels — {"1": 1.0, "2": 1.0}, string keys — then feeds that dict back in. Second time round it lands in the dict branch, int("1") gives 1, the lookup misses every key, and every class is declared unweighted:

It needs three things at once: a forest (nobody else round-trips through a dict), "balanced",
and labels that are strings parsing as integers. "a"/"b" survive because int("a") raises and falls back to str. Real integers survive because np.int64(1) hashes equal to 1. It's precisely the "1"/"2" labels that most UCR/UEA datasets ship with.

  ValueError: The classes, [1, 2], are not in class_weight

The solution

  Never let the forest do the round-trip. Compute the balanced weights ourselves and pass them as  sample_weight — which is what the forest was going to derive anyway, just without the lossy detour through a string-keyed dict:

3. Warning but may as well fix now
sklearn/svm/_base.py:239: FutureWarning: The `probability` parameter was deprecated in 1.  and will be removed in version 1.11. Use CalibratedClassifierCV(SVC(), ensemble=False  instead of SVC(probability=True)                                                      
its only in a test, so small change
stc = ShapeletTransformClassifier(estimator=SVC(probability=False))

4. Deprecate "friedman_mse" 
scikit-learn 1.9 deprecates criterion="friedman_mse" for DecisionTreeRegressor,ExtraTreeRegressor,RandomForestRegressor and ExtraTreesRegressor (#32708), with removal in 1.11. The criterion was intended for gradient boosting but was never correctly implemented in scikit-learn's trees — it always behaved identically to "squared_error". scikit-learn's own warning states this outright: "It maps to squared_error as both were always
 equivalent."

aeon used it in three places, all now switched to "squared_error":
  - aeon/transformations/collection/dictionary_based/_sfa.py — _igb_mae
  - aeon/transformations/collection/dictionary_based/_sfa_fast.py — _igb_mae
  - aeon/base/_estimators/sklearn/tests/test_rotation_forest.py


MatthewMiddlehurst (approved):
Kinda wish we did not have to do the class_weight change but not anyone heres fault and not sure what an alternative would be.

I have kinda wanted to deprecate a lot of the `class_weight` parameters anyway as they mostly just get passed to the estimator, which in most cases can just be replaced using an `estimator` parameter to do the same thing. Feel we should only really include it is `estimator` is not viable for whatever reason or it does something more than just pass through. Not entirely relevant to updating the sklearn version though 🙂.

Issue #2801, opened 2025-05-14
[ENH] Test `anomaly_detection` estimator `_fit_predict` in general testing


MatthewMiddlehurst:
### Describe the feature or idea you want to propose

This is currently not tested. This should only test estimators when the `_fit_predict` function is overridden from the base class, no need to test default functionality. Should check that the overridden functions runs as expected, and produced the same output as `.fit().predict()`.

### Describe your proposed solution

Write a test and add it the AD checks in the `testing` module

notaryanramani:
Hey, can I take on this issue?

SebastianSchmidl:
Please wait until #2326 is merged because it introduces a new base class hierarchy for AD, which involves changes to the test architecture as well.

Nandinisingh07:
Hello!
I’m interested in working on this once #2326 is merged. I’ve been contributing to aeon recently (type hints and documentation), and I’d like to help with anomaly detection testing as well.

Please let me know if I should prepare anything in advance or look at specific parts of the code meanwhile.

PR #3789, opened 2026-09-03
[ENH] Check series anomaly detectors discriminate on the labelled fix…

### LINKED ISSUE #3771: [ENH] Estimator checks let a constant-score or random anomaly detector pass
The series anomaly detector checks in `_yield_series_anomaly_detection_checks.py` verify output shape and dtype but never compare scores to the labels that `UNIVARIATE_SERIES_ANOMALY` already provides. A detector that returns the same float at every timepoint passes every check today, and so does one that returns random noise. The metric layer already treats constant scores as a failure (`roc_auc_score` returns 0.0 with a warning), so the estimator checks are weaker than the benchmark metrics that sit beside them. I would like to add one check that fits on the labeled fixture and asserts that scores are not constant, that ranked scores place the labeled anomalies above the prevalence baseline, and for binary detectors that the positive rate is neither zero nor one. Detectors that cannot rank the synthetic fixture would opt out through an `EXCLUDED_TESTS` entry in `aeon/testing/testing_config.py`. This is the degenerate-detector guard from the FDES benchmark specification I maintain, which came out of a shipped detector that returned identical scores for 220 days until a user report, not a test, caught it (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46988, reported by pathcl). This is separate from #2801, which asks for a `_fit_predict` consistency check in the same yield function and has been waiting since #2326 merged. If this is welcome I will fold that check into the same PR so both land together. Happy to adjust the fixture or the baseline form to whatever the anomaly module maintainers prefer.

mateenali66:
#### Reference Issues/PRs

Fixes https://github.com/aeon-toolkit/aeon/issues/3771

Related, not addressed here: https://github.com/aeon-toolkit/aeon/issues/2801

`check_series_anomaly_detector_output` only checks the shape and dtype of the returned scores. A detector that returns a constant score passes it. So does one that returns scores unrelated to the input. Nothing in the general checks asks whether the detector discriminates.

This adds `check_series_anomaly_detector_discriminates`. It fits and predicts on the labelled fixture the suite already uses, then measures the area under the ROC curve of the scores against that fixture's labels. It requires a gap of at least 0.05 from a constant-score baseline computed on the same labels, which is chance level by construction.

On @TonyBagnall's question in the issue about whether a constant score is always wrong. It is not. On a constant input series a constant score is the right answer. This check never sees such a series. It only runs on the labelled fixture, which does contain anomalies, so a constant score on that fixture is a real failure. The docstring says this so the next reader does not have to find the issue thread.

Design notes:

- Rank based, so no threshold and no assumption about the scale of the scores.
- The gap is two-sided. `OneClassSVM` and `IsolationForest` both score anomalies low rather than high on this fixture, area under the curve 0.000 with `random_state=0`. They discriminate perfectly, the sign is inverted. That may be worth a separate look, but a one-sided check would fail them here for the wrong reason.
- The baseline is read off the fixture rather than hard coded at 0.5, so it survives changes to the fixture's anomaly rate.
- Only yielded for `anomaly_output_type == "anomaly_scores"`. Binary output is a label, not a ranking, and an all-zero label vector is a legitimate prediction.
- Runs on one datatype per instance, so it adds one fit and predict per detector.
- `random_state=0` is pinned, following the existing convention in the clustering, classification and multithreading check files. Without it `IsolationForest` and `ExtendedIsolationForest` fall under the floor in roughly 2 runs in 100.

No. `roc_auc_score` comes from scikit-learn, already a core dependency.

#### Any other comments?

Three things need review.

**`MockAnomalyDetector` had to change.** It returned `np.zeros(...)` with `anomaly_output_type: "anomaly_scores"`, so it was exactly the degenerate detector this check rejects, and it is run through `parametrize_with_checks` in `test_check_estimator.py`. It now returns the distance of each time point from its channel mean. Still trivial, but it depends on the input. No test asserted its values, only shape and dtype. If you would rather the mock stay a no-op, the alternative is an `EXCLUDED_TESTS` entry, but the developer guide calls those temporary so I did not go that way.

**The floor is low on purpose, and it will not catch random scores.** The labelled fixture is 20 time points with exactly one anomaly, so the area under the curve only takes values `k/19`. A floor high enough to reliably catch random scores would fail real detectors. Measured over 1000 seeds a random-score detector is caught 11% of the time. A constant score is caught every time. The worst margin among current detectors is STOMP at 0.132 against the 0.05 floor. Raising the floor would need a longer fixture with more labelled anomalies, which touches shared data and belongs in its own PR.

**Not extended to collection anomaly detectors.** They use per-case classification labels, which is different semantics. Happy to follow up if you want it.

On #2801, I looked at whether it fits here and it does not. I compared `fit_predict(X)` against `fit(X).predict(X)` for the ten series detectors that override `_fit_predict`. Five differ: CBLOF, ExtendedIsolationForest, IsolationForest, LOF and PyODAdapter. The pyod adapter path returns `decision_scores_`, the in-sample training scores, while `predict` calls `decision_function`, and LOF also flips `novelty` between them. So the test as described in that issue would fail half the detectors today, and whether that is a bug or intended is a design decision I did not want to make inside this PR. I will write it up on #2801.

### DIFF
--- a/aeon/testing/estimator_checking/_yield_series_anomaly_detection_checks.py
+++ b/aeon/testing/estimator_checking/_yield_series_anomaly_detection_checks.py
@@ -3,10 +3,17 @@
 from functools import partial
 
 import numpy as np
+from sklearn.metrics import roc_auc_score
 
 from aeon.base._base import _clone_estimator
 from aeon.base._base_series import VALID_SERIES_INNER_TYPES
 from aeon.testing.testing_data import FULL_TEST_DATA_DICT
+
+# Smallest gap from the constant score baseline that a detector must reach on the
+# labelled testing fixture. The fixture is short and holds a single labelled
+# anomaly, so this floor is deliberately low. It is here to catch a detector which
+# does not discriminate at all, not to grade how good a detector is.
+MIN_DISCRIMINATION_MARGIN = 0.05
 
 
 def _yield_series_anomaly_detection_checks(
@@ -21,6 +28,19 @@ def _yield_series_anomaly_detection_checks(
 
     # test class instances
     for i, estimator in enumerate(estimator_instances):
+        # binary output is a label rather than a ranking, so there is no score to
+        # rank the time points by. estimator is None when a soft dependency is
+        # missing and the instance could not be created
+        if (
+            estimator is not None
+            and estimator.get_tag("anomaly_output_type") == "anomaly_scores"
+        ):
+            yield partial(
+                check_series_anomaly_detector_discriminates,
+                estimator=estimator,
+                datatype=datatypes[i][0],
+            )
+
         # test all data types
         for datatype in datatypes[i]:
             yield partial(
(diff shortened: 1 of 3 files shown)


MatthewMiddlehurst:
Thanks. Interesting idea but will have to have a think about it. 

I do not think we check other types of algorithms like this in the general testing. I think the closest would be comparisons to expected results, which usually pick up issues like this when you put in the the expected result, though it is not mandatory as we have simple benchmarks (e.g. majority class for classification which will always be bad).

The main purpose of the general testing is that the algorithms match the framework rather than check performance IMO. The fact we have had to alter the class for testing framework is not a good sign I would say.

mateenali66:
Thank you! Fair point. The per-detector tests already assert the anomaly lands in the right window, which covers this better than a general check and needs no change to the mock. I'll close this and #3771.

PR #3797, opened 2026-09-05
[BUG] samplers: ask for the labels transform needs instead of failing on them

mkzung:
#### Reference Issues/PRs

None.

`SMOTE().fit(X, y).transform(X)` dies inside the sampler on `AttributeError: 'NoneType' object has no attribute 'shape'`, rather than saying the labels are required. ADASYN, ESMOTE, OHIT and RandomOverSampler do the same.

The five resample `X` and `y` together, so `transform` cannot work without `y`. The `requires_y` tag marks what `fit` needs and nothing on the class marks what `transform` needs, so there was no other way for a caller to find out.

Each `_transform` now raises `ValueError` naming the estimator and both ways out:

```
SMOTE resamples X and y together, so transform needs the labels:
pass y, or call fit_transform.
```

The guard sits in the five samplers rather than in `BaseCollectionTransformer`, where a first attempt broke five legitimate estimators that transform without labels.

`aeon/transformations/collection/imbalance` runs 19 passed, 3 skipped. The new test file adds 10 cases: the five refusals, the five estimators still resampling when `y` is passed.

### DIFF
--- a/aeon/transformations/collection/imbalance/_adasyn.py
+++ b/aeon/transformations/collection/imbalance/_adasyn.py
@@ -83,6 +83,11 @@ class ADASYN(SMOTE):
         )
 
     def _transform(self, X, y=None):
+        if y is None:
+            raise ValueError(
+                f"{self.__class__.__name__} resamples X and y together, so transform "
+                "needs the labels: pass y, or call fit_transform."
+            )
         X = np.squeeze(X, axis=1)
         X_resampled = [X.copy()]
         y_resampled = [y.copy()]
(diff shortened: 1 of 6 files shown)


mkzung:
One correction to the description: the attribute is `copy`, not `shape`.

```
>>> SMOTE().fit(X, y).transform(X)
AttributeError: 'NoneType' object has no attribute 'copy'
```

Same for ADASYN, ESMOTE, OHIT and RandomOverSampler on `origin/main`. Nothing else moves; the guard and the tests are unchanged.

TonyBagnall (approved):
thanks for this, not sure how this got through, but resampling is still experimental, thats my excuse :) I think it works with
`sampler.fit_transform(X, y)` and thats how we used it. The whole requires y thing is a little unsettled

PR #3783, opened 2026-09-03
[BUG] Preserve final DisjointCNN filter features

kiwoongyoon:
Title: [BUG] Preserve final DisjointCNN filter features

## Summary

- Apply the DisjointCNN axis permutation only between convolution blocks, matching the reference architecture.
- Preserve all final convolution filters through global average pooling instead of collapsing the projection input to one feature.
- Add a focused network-shape regression assertion.

Fixes #3775

## Test evidence

- Python compilation passed for the changed implementation and test.
- A focused source assertion verified that `Permute` is applied only when another block follows.
- `git diff --check` passed.
- The TensorFlow-backed pytest test was not run locally because this environment lacks pytest and TensorFlow; CI should run the added assertion.

## Risks or notes for maintainers

- The network is shared by the classifier and regressor, so both now retain the final filter dimension as intended.
- No dependency is introduced.
- A full ERing accuracy benchmark was not run locally; the regression test covers the tensor-shape root cause isolated in the issue.

This pull request includes code written with the assistance of AI.
The code has **not yet been reviewed** by a human.

Issue #3775, opened 2026-08-31
[BUG] DisjointCNN performs poorly on problems > 2 classes


TonyBagnall:
### Describe the bug

I'm running classifiers on the multiverse core 

https://github.com/aeon-toolkit/multiverse/

and something wrong with DisjopintCNN I think. 

Handwriting (26 classes) scores 0.055 against a 1/26 = 0.038 chance rate, and
PhonemeSpectra (39 classes) scores 0.075 against 1/39 = 0.026.

### Published vs obtained

Published values are Table 3 of the Disjoint-CNN paper; ours are single runs on the
default train/test split with `random_state=0` and otherwise default parameters.

| dataset | published | aeon | difference |
|---|---|---|---|
| ArticularyWordRecognition | 0.984 | 0.277 | -0.708 |
| UWaveGestureLibrary | 0.924 | 0.366 | -0.559 |
| Libras | 0.916 | 0.533 | -0.383 |
| ERing | 0.964 | 0.589 | -0.375 |
| LSST | 0.547 | 0.270 | -0.277 |
| Handwriting | 0.321 | 0.055 | -0.266 |
| Cricket | 0.999 | 0.750 | -0.249 |
| PhonemeSpectra | 0.282 | 0.075 | -0.207 |
| PEMS-SF | 0.817 | 0.642 | -0.176 |
| AtrialFibrillation | 0.375 | 0.200 | -0.175 |
| EigenWorms | 0.593 | 0.420 | -0.174 |
| StandWalkJump | 0.436 | 0.267 | -0.169 |
| SelfRegulationSCP1 | 0.903 | 0.737 | -0.166 |
| Heartbeat | 0.759 | 0.615 | -0.145 |
| NATOPS | 0.967 | 0.828 | -0.139 |
| DuckDuckGeese | 0.644 | 0.520 | -0.124 |
| RacketSports | 0.846 | 0.763 | -0.083 |
| EthanolConcentration | 0.309 | 0.232 | -0.077 |
| MotorImagery | 0.569 | 0.500 | -0.069 |
| HandMovementDirection | 0.324 | 0.257 | -0.068 |
| PenDigits | 0.995 | 0.962 | -0.033 |
| FaceDetection | 0.567 | 0.538 | -0.028 |
| Epilepsy | 0.989 | 0.971 | -0.018 |
| **mean** | **0.697** | **0.494** | **-0.203** |

For context, on the 55 datasets where we have results for several deep classifiers,
[...]

PR #3806, opened 2026-09-10
[BUG] DisjointCNNNetwork: stop permuting the final convolution block

breken-ai:
#### Reference Issues/PRs

None. `.github/PULL_REQUEST_TEMPLATE.md` asks contributors to use linking keywords where an issue exists rather than requiring one; this was found while reading the code, not filed as an issue first.

**Symptom.** Every network built by `DisjointCNNNetwork` throws away its learned representation immediately before the projection head. Global average pooling emits a **single** feature per sample regardless of how many filters the last convolution block produced. Built on `main` with `n_layers=2, n_filters=[4, 8], input_shape=(12, 3)`:

```
Activation              in (None, 12, 1, 8) -> out (None, 12, 1, 8)
Permute                 in (None, 12, 1, 8) -> out (None, 12, 8, 1)
MaxPooling2D            in (None, 12, 8, 1) -> out (None, 2, 8, 1)
GlobalAveragePooling2D  in (None, 2, 8, 1)  -> out (None, 1)
Dense                   in (None, 1)        -> out (None, 128)
```

The 128-unit projection head is fed one scalar. With the fix:

```
Activation              in (None, 12, 1, 8) -> out (None, 12, 1, 8)
MaxPooling2D            in (None, 12, 1, 8) -> out (None, 2, 1, 8)
GlobalAveragePooling2D  in (None, 2, 1, 8)  -> out (None, 8)
Dense                   in (None, 8)        -> out (None, 128)
```

**Root cause.** The axis permutation is applied inside the block helper rather than between blocks. On `main`, `_one_plus_one_d_convolution_layer` ends at `aeon/networks/encoder/_disjoint_cnn.py:308` with:

```python
spatial_conv = tf.keras.layers.Permute((1, 3, 2))(spatial_conv)
```

The permutation exists so that the *next* block's spatial convolution operates across the previous block's filters. Because it lives at the end of the helper, it also fires after the final block, where there is no next block — it just swaps the filter axis out of the feature position right before `MaxPooling2D`/`GlobalAveragePooling2D`. Pooling then reduces over the filters and leaves a last-axis size of 1.

**Change.** Move the `Permute((1, 3, 2))` out of the helper and into the `build_network` loop, guarded by `if i < self.n_layers - 1` (`_disjoint_cnn.py:254-255` on this branch). Inter-block behaviour is unchanged — the same permutation happens in the same place between blocks. Only the trailing one is gone. Two lines added to the loop, two removed from the helper; the class signature and public API are untouched.

**Trade-off a reviewer should weigh — this is a real behaviour change, not a refactor.** Networks built after this change have a different graph, so **existing DisjointCNN checkpoints will not load**. Anyone with saved weights has to retrain. That is the price of the fix; there is no way to restore the final-filter representation while keeping the old output shape. I have not re-benchmarked accuracy against the published DisjointCNN numbers (see *What was not verified*), so if aeon's stored results for this architecture were produced with the permuted graph, they were produced against the defect.

`_one_plus_one_d_convolution_layer` is private and is referenced nowhere outside `_disjoint_cnn.py` (checked with grep across `aeon/`), so moving the permute out of it cannot affect another network.

No.

#### Any other comments?

The new test asserts a graph property (`model.layers[-1].input.shape[-1] == 8`) rather than a numeric output, because the defect is structural. The expected `8` is hard-coded from the test's own `n_filters=[4, 8]` argument rather than read back off the model, so it is not tautological. It reuses the TensorFlow soft-dependency skip already used by the other test in the file.

### Testing

Environment: Python 3.12.12, local `.venv`, TensorFlow 2.21.0 installed — so the new test executed rather than skipping.

**Revert proof — new test fails against unpatched source, passes with the fix.**

Fix in place:

```
$ .venv/bin/python -m pytest aeon/networks/encoder/tests/test_disjoint_cnn.py -p no:randomly -q
2 passed in 0.47s
```

Source restored to the base commit, new test left in place (`git checkout <base> -- aeon/networks/encoder/_disjoint_cnn.py`):

```
>       assert model.layers[-1].input.shape[-1] == 8
E       assert 1 == 8
FAILED aeon/networks/encoder/tests/test_disjoint_cnn.py::test_disjoint_cnn_final_projection_receives_final_filter_count
1 failed, 1 passed in 2.39s
```

`assert 1 == 8` is the defect stated numerically: 1 feature reaching the projection head where `n_filters[-1] == 8` belongs.

**Surrounding suites:**

```
$ .venv/bin/python -m pytest aeon/networks/encoder/tests/ aeon/classification/deep_learning/tests/test_deep_classifier_base.py -p no:randomly -q
168 passed, 2 xfailed, 2 warnings in 16.90s
```

**General estimator checks on both estimators that use this network:**

```
DisjointCNNClassifier Counter({'PASSED': 33}) 33 checks
DisjointCNNRegressor  Counter({'PASSED': 33}) 33 checks
```

**No stored expected results are invalidated:** `grep -rn "DisjointCNN" aeon/testing/expected_results/` returns nothing, so there is no results table to update.

**Lint** (on the two touched files):

```
All checks passed!
2 files already formatted
$ git diff --check <base>
(no output)
```

### What was not verified

- **No accuracy benchmark, no training run on real data.** The evidence here is a graph-shape assertion plus `check_estimator`, which fits tiny synthetic data and asserts API conformance — it does not measure whether the corrected network learns better. I did not reproduce the published DisjointCNN results before or after, and I cannot tell you the accuracy delta.
- **Weight compatibility was not tested, only reasoned about.** I did not attempt to load an old checkpoint into the new graph; I am asserting incompatibility from the shape change, not from an observed load failure.
- **One TensorFlow version only.** TensorFlow 2.21.0, macOS, CPU. No GPU, no older TF, and nothing was run on aeon's CI matrix.
- **Full `pre-commit` was not run.** `pre-commit`, `black`, `isort`, `pyupgrade` and `codespell` are not installed in this environment. Only `ruff check` and `ruff format --check` ran, on the touched files.

##### For all contributions

Not applicable — no new estimator or public function is added.

### DIFF
--- a/aeon/networks/encoder/_disjoint_cnn.py
+++ b/aeon/networks/encoder/_disjoint_cnn.py
@@ -251,6 +251,8 @@ class DisjointCNNNetwork(BaseDeepLearningNetwork):
                 activation=self._activation[i],
                 kernel_initializer=self._kernel_initializer[i],
             )
+            if i < self.n_layers - 1:
+                x = tf.keras.layers.Permute((1, 3, 2))(x)
 
         max_pool_layer = tf.keras.layers.MaxPooling2D(
             pool_size=(self.pool_size, 1),
@@ -304,7 +306,5 @@ class DisjointCNNNetwork(BaseDeepLearningNetwork):
 
         spatial_conv = tf.keras.layers.BatchNormalization()(spatial_conv)
         spatial_conv = tf.keras.layers.Activation(activation=activation)(spatial_conv)
-
-        spatial_conv = tf.keras.layers.Permute((1, 3, 2))(spatial_conv)
 
         return spatial_conv
--- a/aeon/networks/encoder/tests/test_disjoint_cnn.py
+++ b/aeon/networks/encoder/tests/test_disjoint_cnn.py
@@ -20,3 +20,17 @@ def test_disjoint_cnn_netowkr_kernel_initializer():
 
     assert len(output_layer.shape) == 2
     assert len(input_layer.shape) == 3
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["tensorflow"], severity="none"),
+    reason="Tensorflow soft dependency unavailable.",
+)
+def test_disjoint_cnn_final_projection_receives_final_filter_count():
+    """Final block is not permuted before pooling."""
+    import tensorflow as tf
+
+    network = DisjointCNNNetwork(n_layers=2, n_filters=[4, 8], kernel_size=[3, 3])
+    inputs, outputs = network.build_network((12, 3))
+    model = tf.keras.Model(inputs, outputs)
+    assert model.layers[-1].input.shape[-1] == 8


breken-ai:
Closing in favour of #3783, which makes the same change for #3775 and was opened first.

PR #3812, opened 2026-09-10
[DOC] Use a shipped dataset in the SimHash index notebook

### LINKED ISSUE #3811: [BUG] notebook simhash_index relies on download from zenodo
### Describe the bug

examples/similarity_search/simhash_index.ipynb can fail in notebook CI because it downloads ECG5000 from Zenodo at runtime.

The failing cell contains:

X, _ = load_classification("ECG5000")

now we should probably have retries for downloads, but that is a separate issue, I dont think notebooks should download.

Failing CI run:

https://github.com/aeon-toolkit/aeon/actions/runs/34451589066

### Expected results

passes

### Actual results

In the current CI failure, the request for the Zenodo metadata timed out, causing the whole notebook job to fail with:

ValueError: Could not download files for ECG5000 from Zenodo record 11186692:
ValueError: Failed to retrieve Zenodo metadata for record 11186692:
The read operation timed out

### Suggested fix

notebooks shouldnt download from zenodo. Either a) change example to use current baked in data or b) bake in ECG5000 and ship with aeon

### Versions

_No response_

L4XB:
#### Reference Issues/PRs

Fixes #3811. The notebook was added in #3717.

`examples/similarity_search/simhash_index.ipynb` loaded ECG5000 through `load_classification`, i.e. a Zenodo download at run time, which is what made the notebook CI job fail when the Zenodo metadata request timed out.

The collection is now `load_italy_power_demand()` (1096 series, shipped with aeon), so the notebook needs no network. I checked the shipped datasets against what the walkthrough wants to show (a collection large and diverse enough that a query examines clearly fewer than all series), with the notebook's index parameters (`n_tables=20, n_bits_per_table=26, random_state=0`):

| dataset | series | buckets in table 0 | series 0's bucket | candidates examined | approx vs exact top-5 |
|---|---|---|---|---|---|
| ECG5000 (before) | 5000 | – | – | 1046 / 5000 | 3 of 5 shared |
| **ItalyPowerDemand** | 1096 | 209 | 5 | 180 / 1096 | 4 of 5 shared |
| OSULeaf | 442 | 431 | 1 | 1 / 442 | only the query itself |
| ACSF1 | 200 | 78 | 64 | 114 / 200 | 3 of 5 shared |
| GunPoint | 200 | 59 | 1 | 34 / 200 | 3 of 5 shared |
| ArrowHead | 211 | 43 | 75 | 152 / 211 | 3 of 5 shared |

ItalyPowerDemand keeps the narrative of every cell intact (selective buckets, a small shared bucket for series 0, a query that examines about a sixth of the collection). Only the code of that one cell changes; the outputs of the three data-dependent cells (the load/fit cell, the bucket-keys cell and the query cell) were re-executed, and the bucket-keys cell now has outputs where it had none. The other cells, including the plots, are untouched.

`nbqa black` / `nbqa isort` (the pre-commit hooks) pass on the notebook.

No.

#### Any other comments?

`examples/similarity_search/code_speed.ipynb` also downloads ECG5000/Wafer/FordA, but it is already on the excluded list for the PR run, so I left it alone.

### DIFF
--- a/examples/similarity_search/simhash_index.ipynb
+++ b/examples/similarity_search/simhash_index.ipynb
@@ -420,21 +420,15 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "collection shape: (5000, 1, 140)\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "random projections (hash_funcs_flat_): (520, 140)\n",
-      "boolean signatures: (5000, 520) bool\n",
-      "the 26 bits forming series 0's key in table 0: [0 1 1 0 0 1 0 1 0 1 0 0 1 0 1 1 0 0 1 1 0 1 1 0 0 0]\n"
+      "collection shape: (1096, 1, 24)\n",
+      "random projections (hash_funcs_flat_): (520, 24)\n",
+      "boolean signatures: (1096, 520) bool\n",
+      "the 26 bits forming series 0's key in table 0: [0 1 0 0 1 1 1 0 1 0 0 0 1 0 1 1 1 1 1 1 1 0 0 1 0 1]\n"
      ]
     }
    ],
    "source": [
-    "from aeon.datasets import load_classification\n",
+    "from aeon.datasets import load_italy_power_demand\n",
     "from aeon.similarity_search.whole_series import SimHashIndexANN\n",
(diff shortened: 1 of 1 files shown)


baraline (approved):
Nice catch, should avoid further issue with this problem. Notebook looks good.

PR #3831, opened 2026-09-16
[BUG] Return float32 from WEASELTransformerV2

### LINKED ISSUE #3830: [BUG] WEASELTransformerV2 output needs to be turned into float32
### Describe the bug

PR #3250 fixed issues about scikit-learn regression for RidgeClassifierCV. However, for WEASEL_V2, this fix has been applied to the classifier (WEASEL_V2) instead of the transformer (WEASELTransformerV2). I was playing with `WEASELTransformerV2` and obtained very bad classification results because of this.

**Note: I did not check the other algorithms mentioned in the PR.**

```python
import sklearn; assert sklearn.__version__ == "1.8.0"
from aeon.classification.dictionary_based._weasel_v2 import WEASELTransformerV2
from aeon.datasets import load_unit_test
from sklearn.linear_model import RidgeClassifierCV

X_train, y_train = load_unit_test(split="train")
X_test, y_test = load_unit_test(split="test")

transformer = WEASELTransformerV2(feature_selection="chi2_top_k", random_state=0, n_jobs=1)
Xt_train = transformer.fit_transform(X_train, y_train)
Xt_test = transformer.transform(X_test)

for label, Xt_tr, Xt_te in [
    ("as-is (uint32, what WEASELTransformerV2 actually returns)", Xt_train, Xt_test),
    ("cast to float32", Xt_train.astype(np.float32), Xt_test.astype(np.float32)),
    
]:
    clf = RidgeClassifierCV(alphas=np.logspace(-1, 5, 10))
    clf.fit(Xt_tr, y_train)
    y_pred = clf.predict(Xt_te)
    accuracy = (y_pred == y_test).mean()
    print(f"[{label}]")
    print(f"    coef_ all zero : {np.allclose(clf.coef_, 0)}")
    print(f"    unique preds   : {np.unique(y_pred, return_counts=True)}")
    print(f"    accuracy       : {accuracy:.3f}")
    print()
```

### Expected results

Same performance.

### Actual results

```python
[as-is (uint32, what WEASELTransformerV2 actually returns)]
    coef_ all zero : True
    unique preds   : (array(['1'], dtype='<U1'), array([22]))
    accuracy       : 0.545

[cast to float32]
    coef_ all zero : False
    unique preds   : (array(['1', '2'], dtype='<U1'), array([14,  8]))
    accuracy       : 0.909
```

### Suggested fix

Apply the fix to WEASELTransformerV2 instead of WEASEL_V2:
```python
words = words.astype(np.float32, copy=False)
```

### Versions

<details>
System:
    python: 3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:11:29) [Clang 20.1.8 ]
executable: /opt/anaconda3/bin/python
   machine: macOS-26.5.1-arm64-arm-64bit-Mach-O
Python dependencies:
         aeon: 1.5.0
          pip: 25.3
   setuptools: 80.9.0
 scikit-learn: 1.8.0
        numpy: 2.3.5
        numba: 0.63.1
        scipy: 1.17.1
       pandas: 2.3.3
</details>

MatthewMiddlehurst:
#### Reference Issues/PRs

Fixes #3830. Follow-up to #3250.

scikit-learn 1.8.0 `_RidgeGCV.fit` (used by `RidgeCV`/`RidgeClassifierCV` with the default `cv=None`) casts `coef_`, `intercept_` and `dual_coef_` back to the input dtype. For integer input this truncates the coefficients to zero, so every prediction is the same class. This was fixed upstream in scikit-learn/scikit-learn#33020 (released in 1.9.0), but our bound is `<1.9.0` so 1.8.0 users are affected.

#3250 added the `float32` cast in `WEASEL_V2._fit`, so the classifier worked but `WEASELTransformerV2` still returned `uint32`. This moves the cast into `WEASELTransformerV2.fit_transform` and `transform` (same itemsize as `uint32`, so no extra memory in the output) and removes the now redundant cast from `WEASEL_V2`.

Added `test_weasel_v2_transform_float_output`, which checks the output dtype and that `RidgeClassifierCV` fitted on the transform has non-zero coefficients and the expected accuracy. It fails on `main` and passes with this change on scikit-learn 1.8.0 and 1.7.2.


MatthewMiddlehurst:
@patrickzib Doesnt really change the classifier side of things. It seems to have been fixed upstream, but this version will still be dodgy. Can create an issue to eventually turn them back when we drop these versions of sklearn. Not sure you would need to have that many words though!

MatthewMiddlehurst:
Possibly some other transforms where this still applied, but again unless someone wants to create an issue not sure we should change all our transform output dtypes for this one bug.

johannfaouzi:
The bug is only present in a single version of scikit-learn (`1.8.0`), so we could technically check scikit-learn version and only perform the fix if it's `1.8.0`. But it does not change the fact that this fix should probably be removed when scikit-learn 1.8.0 is no longer supported.

patrickzib:
Fine with me. Thank you for fixing @MatthewMiddlehurst

patrickzib (approved):
lgtm

PR #3833, opened 2026-09-17
[MNT] Update `numba` upper bound and fix numpy 2.4 incompatibilities

MatthewMiddlehurst:
#### Reference Issues/PRs

Split out of #3832. See also #3835 (tslearn and sphinx) and #3834 (pandas).

Raises the upper bound to `numba<0.68.0`.

numba 0.63 required `numpy<2.4`, so this bump also lets numpy 2.4 install. Our own numpy bound already allowed it.

numpy 2.4 now raises a `TypeError` when converting an array with `ndim > 0` to a Python scalar:

- **`plot_significance`:** the column indices were a list of 1-element arrays, so `scores[:, indices]` was 3D, with shape `(n_datasets, n_estimators, 1)`. That meant `wilcoxon_test` got a 2D column per estimator and returned array p-values. It now takes the first matching index, so `ordered_scores` is 2D and the `.flatten()` is no longer needed. This fixes the 5 failures in `test_significance.py`.
- **`examples/networks/deep_learning.ipynb`:** `int(np.random.choice(..., size=1))` is now `np.random.choice(..., size=1).item()`, the same fix as #3795.

Both changes also work with numpy < 2.4.

numba 0.67 causes small floating point changes on Linux CI:

- **ROCKAD doctest:** scores changed by about 1e-7, just beyond the doctest float tolerance. The example now rounds to 5 decimals instead of 7. That gives the same output under numba 0.63.1/numpy 2.3.5 and numba 0.67.0/numpy 2.4.6.
- **`MultiRocketRegressor` expected results:** on CardanoSentiment, the float32 predictions come from terms of about 2.5e5 that cancel to about 0.1 (ridge intercept ≈ -71,061). They round to multiples of 1/128, so tiny feature changes move 2 of the 10 predictions by 1/64. Computing the same model in float64 differs by up to 0.03, so the stored values were float32 rounding rather than a meaningful reference. `check_regressor_against_expected_results` is added to `EXCLUDED_TESTS` for `MultiRocketRegressor` in `testing_config.py`, as was done for HIVE-COTE 2 and TDE after the numba 0.63 update. The stored expected results are unchanged. Exclusions match on the check name, so this also skips the (stable) Covid3Month comparison. The float32 precision issue itself should be fixed in a separate PR.


baraline (approved):
LGTM, not surprised with small numba float changes.

PR #3834, opened 2026-09-17
[MNT] Update `pandas` upper bound to allow pandas 3

MatthewMiddlehurst:
#### Reference Issues/PRs

Split out of #3832. See also #3835 (tslearn and sphinx) and #3833 (numba).

Raises the upper bound to `pandas<3.1.0`.

- **Validation:** pandas 3 stores string data in the new `str` dtype by default. That's a pandas extension dtype, not a numpy dtype, so `np.issubdtype` raises `TypeError: Cannot interpret ... as a data type` before our "must contain numeric values only" error is reached. That breaks `test_get_type_errors` in `test_series.py` and `test_collection.py`. The dtype checks on pandas objects in `get_type`, `_is_pd_wide` and `_is_pd_multiindex` now check `isinstance(dtype, np.dtype)` before calling `np.issubdtype`. The shared check is `_is_float_or_int_dtype` in `aeon/utils/validation/__init__.py`. Results for numpy dtypes are unchanged. Extension dtypes (`str`, `category`, nullable `Float64`/`Int64`) get our error instead of numpy's `TypeError`.
- **Integer `pd.Series` (bug fix, can be split out):** `series.get_type` only accepted float dtypes for a `pd.Series`. Its check was `floating and not integer`, while the DataFrame and ndarray branches reject anything that is `not floating and not integer`. So `is_series(pd.Series([1, 2, 3]))` was `False`, and `plot_series` rejected integer Series, even though integer ndarrays and DataFrames pass and `BaseSeriesEstimator` accepts integer Series. The Series branch now uses the same shared check. Tests are added for integer inputs and for a string `pd.Series`.
- **Doctests:** `load_airline`, `load_lynx` and `load_shampoo_sales` printed `type(y)`. pandas 3 sets `Series.__module__` to `"pandas"`, so this now prints `<class 'pandas.Series'>`. The class itself still lives in `pandas.core.series`. The doctests now check `type(y).__name__`.
- **Read-only loader output:** with copy-on-write in pandas 3, `.values` and `.to_numpy()` can return read-only views, so in-place edits such as `y -= y.mean()` would raise. On pandas 3 this affected `load_shampoo_sales`, `load_lynx`, `load_airline`, `load_solar`, `load_PBS_dataset`, `y` from `load_from_tsv_file`, and univariate `X` and `y` from `load_from_timeeval_csv_file`. Every loader that returns an array from pandas now uses `to_numpy(copy=True)`. This includes `load_uschange`, `load_longley` and the results loaders, which are only writeable today because of how pandas lays out their data internally. The existing loader tests now assert `flags.writeable`, and they fail on pandas 3 without this change.

These changes also work with pandas 2.x.


TonyBagnall (approved):
## Review by @TonyBagnall
well, review by Claude really, I dont think anyone will believe otherwise :)

Verified empirically on pandas 3.0.2 and 2.3.3 (each bug reproduced on `main`
first, then confirmed fixed on the branch). Approve, non-blocking notes only.

### Confirmed fixes

**1. Extension-dtype guard (`_is_float_or_int_dtype`).** On `main` + pandas 3,
`get_type(pd.Series(["a","b"]))` raises the raw
`TypeError: Cannot interpret '<StringDtype(...)>' as a data type` before our
message is reached. On the branch, `str`, object-string, `category`, nullable
`Int64` and nullable `Float64` all return the clean
`ERROR pd.Series must contain numeric values only`. Numeric numpy dtypes are
unchanged. Correct.

**2. Integer `pd.Series` bug.** Reproduced: on `main`,
`is_series(pd.Series([1,2,3]))` is `False` and `plot_series` rejects an integer
Series. On the branch both work, matching the ndarray/DataFrame branches and
`BaseSeriesEstimator`. Good catch splitting this out as a separable fix.

**3. Read-only loader output.** On `main` + pandas 3, `load_airline`,
`load_lynx` and `load_shampoo_sales` return read-only arrays and `y -= y.mean()`
raises `ValueError`. `load_uschange`/`load_longley` are writeable only by
accident of layout today. On the branch all seven loaders (plus `load_solar`,
`load_PBS_dataset`) return writeable arrays. Pre-emptively copying uschange/
longley/results-loaders is the right call rather than relying on internal layout.

**4. Doctests.** Confirmed `type(y).__module__ == "pandas"` on pandas 3, so the
old `<class 'pandas.core.series.Series'>` expectation fails. The
`type(y).__name__ == 'Series'` rewrite passes on both pandas 2 and 3.

### Test run

138 passed across the touched validation/loader/benchmarking test files
(including the new `flags.writeable` and integer/string-Series assertions). The
only non-passes were environmental: missing `mocker` fixture, then the archive-
download TSAD tests that need network. `check_estimator` on
`KNeighborsTimeSeriesClassifier` is clean on pandas 3 (26/26). "Also works with
pandas 2.x" confirmed by re-running on 2.3.3.

### Scope check

No remaining `np.issubdtype`-on-a-pandas-dtype landmines in the validation path:
the other `np.issubdtype` calls in `series.py`/`collection.py` are the
ndarray/np-list branches (guaranteed numpy dtypes), and `collection.get_type`
routes every pandas check through the two patched helpers (`_is_pd_multiindex`,
`_is_pd_wide`, `df-list` via `_is_pd_wide`).

### Non-blocking

- Nullable numeric dtypes (`Int64`/`Float64`) are now rejected with the
  numeric-values error rather than the raw numpy `TypeError`. This is a strict
  improvement and is documented, but it does leave pandas masked/nullable
  numeric Series unsupported. Fine to leave as-is; worth a follow-up only if a
  user actually wants masked-array support.
- The bare `np.issubdtype` pair in the ndarray branches could reuse
  `_is_float_or_int_dtype` for symmetry, but those dtypes are always numpy so
  it is safe and not worth the churn.

Clean, tightly-scoped unit. Full pandas-3 green CI depends on the companion
splits (#3833 numba, #3835 tslearn/sphinx) landing too, which is the right way
to cut this.

PR #3835, opened 2026-09-17
[MNT] Update `tslearn` and `sphinx` upper bounds

MatthewMiddlehurst:
#### Reference Issues/PRs

Split out of #3832, which bumps four dependencies at once. numba and pandas need code changes and are in #3833 and #3834.

Raises the upper bounds to `tslearn<0.10.0` and `sphinx<9.1.0`. Neither needs code changes.

- **tslearn 0.9:** `KernelKMeans` with `kernel_params={"sigma": "auto"}` now passes `random_state` through when estimating sigma. Results for that setting change once, then stay reproducible.
- **Sphinx 9** (and docutils 0.22): the docs build produces the same warnings as with Sphinx 8.2.3. `docs/_sphinxext/aeon_mini_gallery.py` uses `env.app`, which Sphinx 11 removes. That needs a fix before a future bump, but it doesn't affect this one.


MatthewMiddlehurst:
Would prefer if #1719 goes in first, but not essential

TonyBagnall (approved):
claude again: 

Verified empirically against tslearn 0.9.0 and Sphinx 9.0.4 / docutils 0.22.4.
Approve, one non-blocking follow-up.

### tslearn 0.5.2 -> 0.9.0

The only runtime use of tslearn in aeon is the `TimeSeriesKernelKMeans` wrapper
(`aeon/clustering/_kernel_k_means.py`); everything else is citations or
adapted-code comments (soft-DTW, SAX, bounding matrix, dataset loader).

- All 4 bundled tests in `test_kernel_k_means.py` pass on tslearn 0.9.0 with
  their existing hardcoded expected labels/inertia unchanged. The GAK test uses
  the default kernel, which resolves to `sigma=1.0` (deterministic), not the
  `sigma="auto"` estimation path, so the random_state-threading change in 0.9
  doesn't touch it. That's why no test update is needed, which matches the
  "results change only for that setting" note.
- `check_estimator(TimeSeriesKernelKMeans)` is clean (21/21) on 0.9.0.
- Confirmed the opt-in path the PR describes: `kernel_params={"sigma": "auto"}`
  is now reproducible across fits under 0.9.0 (identical labels on repeat fit
  with the same `random_state`). aeon passes user `kernel_params` straight
  through and doesn't test that setting with hardcoded values, so the one-time
  result change is invisible to the suite.

### sphinx <8.3.0 -> <9.1.0 (docutils 0.22)

aeon ships two local extensions, `aeon_mini_gallery` and
`sphinx_remove_toctrees`. Under Sphinx 9.0.4 + docutils 0.22.4:

- Both import and their `setup()` register without error; a minimal
  `sphinx-build` loading both produces HTML cleanly. No removed-API usage in
  either (no `add_stylesheet`/`app.info`/`app.warn`/etc.).
- The flagged `env.app` at `aeon_mini_gallery.py:174` still exists on
  `BuildEnvironment` in Sphinx 9, confirming it's a Sphinx 11 removal and this
  bound is clear of it. Correct to defer that fix.

### Non-blocking

- The `env.app` future-break is worth a tracked issue rather than only a line in
  this PR body, so it's actionable before anyone attempts Sphinx 10/11. Needs
  the app/env API, so not a good-first-issue.
- I could not reproduce the full docs build here (nbsphinx notebook execution,
  pandoc, and the third-party extension stack: numpydoc, myst_parser,
  sphinx_design, sphinxext.opengraph, versionwarning). The "same warnings as
  Sphinx 8.2.3" claim covers those third-party extensions and rests on the local
  build plus the docs CI job, not independent reproduction here. Worth a glance
  that the docs workflow ran green on this branch.

Clean split. Together with #3833/#3834 this closes out the pandas/numba/tslearn/
sphinx bumps from #3832.

PR #3402, opened 2026-05-12
[ENH] Implement Tselect channel selection algorithm

TonyBagnall:
#### Reference Issues/PRs

resolves #3389

Implements the TSelect channel selection algorithm in the aeon channel_selection package. Algorithm  is described in
    .. [1] Nuyts, L., Perini, L. and Davis, J. "TSelect: selecting relevant and
        non-redundant channels for multivariate time series classification."
        Data Mining and Knowledge Discovery, 39, 76, 2025.
        https://doi.org/10.1007/s10618-025-01132-4

based on the implementation here
https://github.com/ML-KULeuven/TSelect

PR #3836, opened 2026-09-17
[ENH] Add KGMTP algorithm

### LINKED ISSUE #3800: [ENH] KG-MTP algorithm for time series classification

I was wondering if you would be interested in adding KG-MTP to the list of convolution-based classification algorithms? The paper was published last year (December 2025) in Expert Systems with Applications (https://www.sciencedirect.com/science/article/abs/pii/S0957417425023115). The source code is publicly available (https://github.com/WangPanJie2024/KG-MTP).

The original source code is a bit messy, so I rewrote it (both for readability and runtime speed). In the original publication, the authors used 30 resamples, but different from the resamples of the Bake off redux paper. Thus, I ran the algorithm with the original 30 resamples so that the comparisons are fair, both on the 112 datasets and the 142 datasets (the additional 30 from "redux"). The results are quite good:

| Datasets | Accuracy | Balanced accuracy | F1 |
|:--------:|:---------:|:---------------------:|:---:|
| 112 | 0.888882 | 0.871193 | 0.875145 |
| 142 | 0.876170 | 0.854799 | 0.856987 |

In terms of accuracy and balanced accuracy, the scores are slightly higher than the ones of MR-Hydra:

Image

If you are interested, I can work on a PR.

_No response_

### Additional context

_No response_

johannfaouzi:
#### Reference Issues/PRs

Fixes #3800 

This PR adds the KGMTP algorithm:
* Paper: [Kernel grouping for time series classification with multiple transformations and pooling operators](https://www.sciencedirect.com/science/article/pii/S0957417425023115)
* Original code: https://github.com/WangPanJie2024/KG-MTP

No 

#### Any other comments?

Two comments:
* I did not update the example about [Convolution based time series classification in aeon](https://www.aeon-toolkit.org/en/stable/examples/classification/convolution_based.html).
* The code does not allow a legacy `numpy.random.RandomState` instance for `random_state` because [Numba supports `numpy.random.Generator()` objects](https://numba.readthedocs.io/en/stable/reference/numpysupported.html#generator-objects) but [but does not allow you to create individual `RandomState ` instances](https://numba.readthedocs.io/en/stable/reference/numpysupported.html#randomstate-and-legacy-random-number-generation).

##### For all contributions


TonyBagnall:
great, thanks for this. Could you post some evaluation of equivalence to the reference implementation, ideally with timing comparison too? Something like this? https://github.com/aeon-toolkit/aeon/pull/3402

johannfaouzi:
Good news! The KGMTP transformation in this PR is bit-to-bit (at float32 precision) identical to the original code. The predictions are thus also strictly identical (I didn't change the code for this part, it's StandardScaler -> RidgeClassifierCV).

### Results

| dataset | train_fit_transform_original (s) | train_fit_transform_proposed (s) | MAE_X_train | test_transform_original (s) | test_transform_proposed (s) | MAE_X_test | accuracy_original | accuracy_proposed |
|---|---|---|---|---|---|---|---|---|
| Chinatown | 0.0022 | 0.0035 | 0.0 | 0.0156 | 0.0133 | 0.0 | 0.9796 | 0.9796 |
| Rock | 0.0563 | 0.0453 | 0.0 | 0.0944 | 0.0691 | 0.0 | 0.7200 | 0.7200 |
| Crop | 0.5484 | 0.3820 | 0.0 | 1.1952 | 0.7500 | 0.0 | 0.6829 | 0.6829 |
| StarLightCurves | 0.5408 | 0.4571 | 0.0 | 4.2453 | 3.5169 | 0.0 | 0.9773 | 0.9773 |

### `scipy.signal.hilbert` (vectorized) versus `scipy.fftpack.hilbert` (non-vectorized)

Regarding the runtimes improvements, they are relatively minor and depend on the size of the dataset:
* `scipy.fftpack.hilbert` is faster than `scipy.signal.hilbert` for large series lengths and a given number of series.
* `scipy.signal.hilbert` is faster than `scipy.signal.hilbert` for a large number of series a given series length.

On my laptop (MacBook Air M1, 2020), the vectorized version (`scipy.signal.hilbert`) won on 98/112 datasets, and summed over the full archive it took 0.98s total vs. 1.75s for the loop (1.78x aggregate). The 14 losses were concentrated in long series (all at series_length >= 1250) and never worse than about 0.63x (ACSF1, length 1460); wins on short series reached up to 32x (SmoothSubspace, length 15).

### Reproducibility

There are two small, self-contained scripts to run from the `aeon` repo root, on this branch, after `pip install -e .` (you need to copy them, they are provided below):
* **`compare_kgmtp.py`** — runtime, numerical parity, and accuracy, original vs. proposed, on real UCR datasets.
* **`hilbert_transform_benchmark.py`** — a narrower, standalone timing comparison of `KGMTP._hilbert_transform`'s two candidate implementations (the row-by-row `scipy.fftpack.hilbert` loop the original paper uses, vs. the batched `scipy.signal.hilbert(X, axis=-1)` call this PR actually uses).

**`compare_kgmtp.py`**
```python
"""Compare aeon's KGMTP against the original paper implementation.

Run from the aeon repo root, on the `add-kgmtp` branch, after `pip install -e .`:

    python compare_kgmtp.py

Clones johannfaouzi/KG-MTP@aeon-benchmark-support to a temp directory to get the
original implementation (see that branch for the two small changes it carries over
the paper's own code: a threadable RNG parameter, and a timed/importable wrapper
around what expri_code/run_KGMTP.py already does) and compares it against aeon's own
`KGMTP` transformer. Needs network access to clone that branch and to download any
dataset not already cached locally by aeon.

For each dataset, reports:
  - train_fit_transform_{original,proposed}_runtime: seconds for fit(X_train) +
    transform(X_train).
  - MAE_X_{train,test}: mean absolute error between the two implementations' raw,
    pre-scaling features (each implementation's own internal-scaling choice is
    deliberately excluded, to isolate "did the port change the math").
  - test_transform_{original,proposed}_runtime: seconds for transform(X_test) alone.
  - accuracy_{original,proposed}_test: downstream classification accuracy, using each
    implementation's own real pipeline.
"""

import argparse
import os
import subprocess
import sys
import tempfile
import time

import numpy as np
from sklearn.linear_model import RidgeClassifierCV
from sklearn.metrics import accuracy_score
from sklearn.pipeline import make_pipeline
from sklearn.preprocessing import StandardScaler

from aeon.datasets import load_classification
from aeon.transformations.collection.convolution_based import KGMTP as ProposedKGMTP

KGMTP_FORK_URL = "https://github.com/johannfaouzi/KG-MTP.git"
KGMTP_FORK_BRANCH = "aeon-benchmark-support"
# One UCR dataset per corner of the (n_cases, series_length) space, rather than an
# arbitrary handful of small/fast ones -- so a single default run gives some evidence
# about all four regimes instead of only the "small and fast" corner:
#   Chinatown        n_cases=  363  length=   24  (low N,  low L)
#   Rock              n_cases=   70  length= 2844  (low N,  high L)
#   Crop              n_cases=24000  length=   46  (high N, low L)
#   StarLightCurves   n_cases= 9236  length= 1024  (high N, high L)
DEFAULT_DATASETS = ["Chinatown", "Rock", "Crop", "StarLightCurves"]
ALPHAS = np.logspace(-3, 3, 10)


def clone_kgmtp_fork(dest):
    subprocess.run(
        [
            "git", "clone", "--branch", KGMTP_FORK_BRANCH, "--single-branch",
            "--depth", "1", KGMTP_FORK_URL, dest,
        ],
        check=True,
    )


def mae(a, b):
    return float(np.mean(np.abs(np.asarray(a) - np.asarray(b))))


def proposed_raw_features(model, X2d):
    X_hilbert = model._hilbert_transform(X2d)
    X_diff = np.diff(X2d, 1)
    return model._transform_branches(X2d, X_hilbert, X_diff)


def run_proposed_pipeline(X_train_3d, y_train, X_test_3d, y_test, x_train_2d, x_test_2d, n_kernels, seed):
    model = ProposedKGMTP(
        n_kernels=n_kernels, max_dilations_per_kernel=32, n_features_per_kernel=5,
        n_jobs=-1, random_state=seed,
    )

    t0 = time.perf_counter()
    Xt_train = model.fit_transform(X_train_3d)
    train_rt = time.perf_counter() - t0

    t0 = time.perf_counter()
    Xt_test = model.transform(X_test_3d)
    test_rt = time.perf_counter() - t0

    clf = make_pipeline(StandardScaler(), RidgeClassifierCV(alphas=ALPHAS))
    clf.fit(Xt_train, y_train)
    preds = clf.predict(Xt_test)
    acc = accuracy_score(y_test, preds)

    raw_train = np.concatenate(proposed_raw_features(model, x_train_2d), axis=1)
    raw_test = np.concatenate(proposed_raw_features(model, x_test_2d), axis=1)
    return {
        "train_fit_transform_runtime": train_rt,
        "test_transform_runtime": test_rt,
        "accuracy_test": acc,
        "raw_train": raw_train,
        "raw_test": raw_test,
    }


def warm_up(run_kgmtp_pipeline):
    """Run both implementations once on a tiny toy dataset, results discarded."""
    print("Warming up Numba JIT compilation on a tiny toy dataset (discarded) ...")
    rng = np.random.default_rng(0)
    n, length = 20, 24
    x = rng.standard_normal((n, length))
    y = rng.integers(0, 2, n)
    x_3d = x[:, np.newaxis, :]
    run_kgmtp_pipeline(x, y, x, y, n_kernels=1200, seed=0)
    run_proposed_pipeline(x_3d, y, x_3d, y, x, x, 1200, 0)
    print()


def print_table(rows):
    columns = list(rows[0].keys())
    widths = {c: max(len(c), *(len(f"{r[c]:.4g}" if isinstance(r[c], float) else str(r[c])) for r in rows)) for c in columns}
    header = " | ".join(c.ljust(widths[c]) for c in columns)
    print(header)
    print("-+-".join("-" * widths[c] for c in columns))
    for r in rows:
        cells = [f"{r[c]:.4g}" if isinstance(r[c], float) else str(r[c]) for c in columns]
        print(" | ".join(cell.ljust(widths[c]) for cell, c in zip(cells, columns)))


def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--n-kernels", type=int, default=1200)
    parser.add_argument("--datasets", type=str, default=",".join(DEFAULT_DATASETS))
    parser.add_argument("--seed", type=int, default=0)
    parser.add_argument("--out", type=str, default="kgmtp_original_vs_proposed.csv")
    args = parser.parse_args()

    with tempfile.TemporaryDirectory() as tmp:
        kgmtp_dir = os.path.join(tmp, "KG-MTP")
        print(f"Cloning {KGMTP_FORK_URL}@{KGMTP_FORK_BRANCH} ...")
        clone_kgmtp_fork(kgmtp_dir)
        sys.path.insert(0, kgmtp_dir)
        from kgmtp_pipeline import run_kgmtp_pipeline  # noqa: E402

        warm_up(run_kgmtp_pipeline)

        rows = []
        for name in [d.strip() for d in args.datasets.split(",") if d.strip()]:
            print(f"[{name}] running original + proposed ...")
            X_train, y_train = load_classification(name, split="TRAIN")
            X_test, y_test = load_classification(name, split="TEST")
            x_train_2d = X_train[:, 0, :].astype(np.float64)
            x_test_2d = X_test[:, 0, :].astype(np.float64)

            original = run_kgmtp_pipeline(
                x_train_2d, y_train, x_test_2d, y_test,
                n_kernels=args.n_kernels, seed=args.seed,
            )
            proposed = run_proposed_pipeline(
                X_train, y_train, X_test, y_test, x_train_2d, x_test_2d,
                args.n_kernels, args.seed,
            )

            row = {
                "dataset": name,
                "train_fit_transform_original_runtime": original["train_fit_transform_runtime"],
                "train_fit_transform_proposed_runtime": proposed["train_fit_transform_runtime"],
                "MAE_X_train": mae(original["raw_train"], proposed["raw_train"]),
                "test_transform_original_runtime": original["test_transform_runtime"],
                "test_transform_proposed_runtime": proposed["test_transform_runtime"],
                "MAE_X_test": mae(original["raw_test"], proposed["raw_test"]),
                "accuracy_original_test": original["accuracy_test"],
                "accuracy_proposed_test": proposed["accuracy_test"],
            }
            rows.append(row)
            print(
                f"  MAE_X_train={row['MAE_X_train']:.3e}  MAE_X_test={row['MAE_X_test']:.3e}"
                f"  acc_original={row['accuracy_original_test']:.4f}"
                f"  acc_proposed={row['accuracy_proposed_test']:.4f}"
            )

        with open(args.out, "w") as f:
            f.write(",".join(rows[0].keys()) + "\n")
            for r in rows:
                f.write(",".join(str(v) for v in r.values()) + "\n")
        print(f"\nSaved results to {args.out}\n")
        print_table(rows)


if __name__ == "__main__":
    main()

```

**`hilbert_transform_benchmark.py `**
```python
"""Loop (`scipy.fftpack.hilbert`) vs. vectorized (`scipy.signal.hilbert`) timing.

`KGMTP._hilbert_transform` (in `aeon/transformations/collection/convolution_based/
_kgmtp.py`) batches the row-by-row `scipy.fftpack.hilbert` loop into a single
`scipy.signal.hilbert(X, axis=-1)` call. The two functions use opposite sign
conventions (`fftpack.hilbert(x) == -signal.hilbert(x, axis=-1).imag`), and once that
sign is corrected and the same float32 round-trip is applied, their outputs are
numerically identical -- that part is a fact, not a benchmark result, and doesn't
depend on hardware.

Whether the vectorized version is actually *faster* does depend on hardware (CPU,
core count, BLAS/FFT backend) and on the data shape (series length matters far more
than instance count). Rather than bake a specific speedup number into the code or the
PR description -- which would only be true on whatever machine produced it -- this
script lets anyone reproduce the comparison on their own hardware and their own choice
of datasets.

Point `--datasets-dir` at a local UCR-archive-style directory (one subfolder per
dataset, each containing `<name>_TRAIN.ts` and `<name>_TEST.ts` -- e.g. a checkout of
https://github.com/johannfaouzi/KG-MTP, which bundles all 112 UCR archive datasets
under `datasets/`) -- only each dataset's (N, L) shape is read from the `.ts` files
and timed on synthetic random data of that shape; the actual values are never used,
since only the Hilbert transform's runtime (not its output) is being compared here.

Usage
-----
    python hilbert_transform_benchmark.py --datasets-dir DIR [--reps N] [--out CSV]
"""

import argparse
import csv
import os
import platform
import time

import numpy as np
import scipy
import scipy.fftpack
import scipy.signal


def get_shape(ts_path):
    """Read (n_instances, series_length) from a `.ts` file's header/data, cheaply."""
    n = 0
    length = None
    with open(ts_path, "r") as f:
        in_data = False
        for line in f:
            line = line.rstrip("\n")
            if not in_data:
                if line.lower().startswith("@serieslength"):
                    length = int(line.split()[1])
                elif line.lower().startswith("@data"):
                    in_data = True
                continue
            if line.strip():
                n += 1
                if length is None:
                    length = len(line.split(":")[0].split(","))
    return n, length


def loop_hilbert(X):
    """Apply the original row-by-row `scipy.fftpack.hilbert` loop."""
    out = np.empty_like(X)
    for i in range(X.shape[0]):
        out[i] = scipy.fftpack.hilbert(X[i])
    return out


def vec_hilbert(X):
    """Apply the vectorized `scipy.signal.hilbert` call, sign-corrected."""
    return -scipy.signal.hilbert(X, axis=-1).imag


def time_call(fn, X, reps):
    """Best-of-`reps` wall-clock time, to reduce scheduling noise."""
    best = None
    for _ in range(reps):
        t0 = time.perf_counter()
        fn(X)
        dt = time.perf_counter() - t0
        if best is None or dt < best:
            best = dt
    return best


def main():
    """Run the loop-vs-vectorized Hilbert transform timing comparison."""
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument(
        "--datasets-dir",
        required=True,
        help="Directory with one subfolder per dataset, each holding "
        "<name>_TRAIN.ts/<name>_TEST.ts -- e.g. a checkout of "
        "https://github.com/johannfaouzi/KG-MTP, which bundles all 112 UCR "
        "archive datasets under datasets/.",
    )
    parser.add_argument(
        "--reps",
        type=int,
        default=5,
        help="Repeats per timing (best-of-N is kept).",
    )
    parser.add_argument("--seed", type=int, default=0)
    parser.add_argument(
        "--out",
        default=None,
        help="Optional path to save a per-dataset CSV of the results.",
    )
    args = parser.parse_args()

    print(f"platform: {platform.platform()}")
    print(f"processor: {platform.processor() or platform.machine()}")
    print(f"cpu_count: {os.cpu_count()}")
    print(f"numpy: {np.__version__}  scipy: {scipy.__version__}")
    print()

    names = sorted(
        d
        for d in os.listdir(args.datasets_dir)
        if os.path.isdir(os.path.join(args.datasets_dir, d))
    )
    if not names:
        raise SystemExit(f"No dataset subfolders found under {args.datasets_dir}")

    rng = np.random.default_rng(args.seed)
    rows = []
    for name in names:
        ddir = os.path.join(args.datasets_dir, name)
        train_ts = os.path.join(ddir, f"{name}_TRAIN.ts")
        test_ts = os.path.join(ddir, f"{name}_TEST.ts")
        if not (os.path.exists(train_ts) and os.path.exists(test_ts)):
            continue
        n_train, l_train = get_shape(train_ts)
        n_test, l_test = get_shape(test_ts)
        if l_train != l_test or l_train is None:
            # Only equal-length, univariate shapes are meaningful for this
            # shape-only timing comparison -- skip anything else.
            continue
        n, length = n_train + n_test, l_train
        X = rng.standard_normal((n, length)).astype(np.float64)
        t_loop = time_call(loop_hilbert, X, args.reps)
        t_vec = time_call(vec_hilbert, X, args.reps)
        speedup = t_loop / t_vec
        rows.append((name, n, length, t_loop, t_vec, speedup))
        print(
            f"{name:35s} N={n:6d} L={length:5d}  loop={t_loop:8.4f}s  "
            f"vec={t_vec:8.4f}s  speedup={speedup:6.2f}x"
        )

    total_loop = sum(r[3] for r in rows)
    total_vec = sum(r[4] for r in rows)
    wins_vec = sum(1 for r in rows if r[5] > 1.0)
    wins_loop = sum(1 for r in rows if r[5] < 1.0)
    print()
    print(f"datasets benchmarked: {len(rows)}")
    print(f"vectorized wins: {wins_vec}   loop wins: {wins_loop}")
    print(f"total loop time:       {total_loop:.4f}s")
    print(f"total vectorized time: {total_vec:.4f}s")
    print(
        "aggregate speedup (sum of loop times / sum of vectorized times): "
        f"{total_loop / total_vec:.2f}x"
    )

    if args.out:
        with open(args.out, "w", newline="") as f:
            w = csv.writer(f)
            w.writerow(
                ["dataset", "N", "L", "loop_seconds", "vectorized_seconds", "speedup"]
            )
            w.writerows(rows)
        print(f"\nSaved per-dataset results to {args.out}")


if __name__ == "__main__":
    main()

```

rdancer:
@johannfaouzi 

Hi Johann, @TonyBagnall has just suggested I review this - we use KG-MTP as the base of our classifier; it's been recently used by https://github.com/azizalasfour1/ALEC which our classifier is based on.

Off of the top of my head, the only changes that ALEC makes, compared to the original KG-MTP code, are small, but important; they make the transform reproducible.

I'll look at it over the next few days.

rdancer:
https://github.com/aeon-toolkit/aeon/pull/3836

@johannfaouzi @TonyBagnall @MatthewMiddlehurst 

The PR looks good, I like the speedup of the Hilbert transform.

Will look more deeply over the week-end; I've had a look at it briefly now, here are my comments:

## Should be split into the KG-MTP transform proper, and Ridge

- The classifier should be split into a separate transform (which is what it really is), and a separate Ridge head, same as the `HydraClassifier` already is split into `HydraTransform` and Ridge (The ROCKET family of which KG-MTP is a member is a transform -> Ridge architecture; they are not classifiers in their own right, and fitting different heads to them is an area of active research, notably recent [Dempster et al. 2025](https://doi.org/10.1007/s10994-026-07059-1). And it's just less confusing that way.

## There is some code that should be deduplicated

- The [`_SparseScaler` from `aeon.classification.convolution_based.HydraClassifier`](https://github.com/aeon-toolkit/aeon/blob/83e8a9a4ea3cf03d73256c51196c5175156e5c53/aeon/classification/convolution_based/_hydra.py#L142) should be promoted to a public class, and reused in the KG-MTP class, instead of having duplicate code with identical functionality.

-  `_check_random_state` custom implementation should be replaced with `from sklearn.utils import check_random_state`, to bring it in line with [the rest of the library](https://github.com/search?q=repo%3Aaeon-toolkit%2Faeon+check_random_state&type=code)

## There are some naming issues, perhaps because of language barrier

-  `s/Hydra scaler/sparse scaler (used by Hydra)/ ` (sparse scaler is not necessarily Hydra-specific)
- `s/n_ppv_features_/n_pooling_features_/` -- there are other features apart from Proportion of Positive Values (PPV); the paper calls all the non-Hydra features (somewhat unfortunately in my humble opinion) "pooling features"

-----

It's nicely commented, which is always good, and as far as I can tell it follows the aeon general structure, and overall is an improvement over the original implementation with respect to readability. We are looking forward to using this version in our upcoming implementation of [ALEC](https://github.com/signalsphinx/ALEC).

johannfaouzi:
Thank you for your review @rdancer!

## Should be split into the KG-MTP transform proper, and Ridge

> - The classifier should be split into a separate transform (which is what it really is), and a separate Ridge head, same as the `HydraClassifier` already is split into `HydraTransform` and Ridge (The ROCKET family of which KG-MTP is a member is a transform -> Ridge architecture; they are not classifiers in their own right, and fitting different heads to them is an area of active research, notably recent [Dempster et al. 2025](https://doi.org/10.1007/s10994-026-07059-1). And it's just less confusing that way.

I genuinely don't understand this point. There is:

* a transformer named `KGMTP` in [`aeon/transformations/collection/convolution_based/_kgmtp.py`](https://github.com/johannfaouzi/aeon/blob/03471375cecfcf562f287ecca53c22e73f67b052/aeon/transformations/collection/convolution_based/_kgmtp.py#L462-L811) in the public API, and
* a classifier named `KGMTPClassifier` in [`aeon/classification/convolution_based/_kgmtp.py`](https://github.com/johannfaouzi/aeon/blob/03471375cecfcf562f287ecca53c22e73f67b052/aeon/classification/convolution_based/_kgmtp.py#L21-L194) in the public API.

I also follow the (inconsistent) naming conventions in `aeon`:

* `Rocket` / `RocketClassifier`
* `MiniRocket` / `MiniRocketClassifier`
* `MultiRocket` / `MultiRocketClassifier`
* `HydraTransformer` / `HydraClassifier`

I can add the `Transformer` suffix to `KGMTP` if needed, but the class is already in the `transformations` subpackage. I have no strong opinion on this.

## There is some code that should be deduplicated

> - The [`_SparseScaler` from `aeon.classification.convolution_based.HydraClassifier`](https://github.com/aeon-toolkit/aeon/blob/83e8a9a4ea3cf03d73256c51196c5175156e5c53/aeon/classification/convolution_based/_hydra.py#L142) should be promoted to a public class, and reused in the KG-MTP class, instead of having duplicate code with identical functionality.

It's not usable as it is because it's in PyTorch and not in NumPy. The original class needs to be updated to handle both inputs, or the whole transform should be done in PyTorch and not in NumPy / Numba. But the original code is in NumPy / Numba.

> -  `_check_random_state` custom implementation should be replaced with `from sklearn.utils import check_random_state`, to bring it in line with [the rest of the library](https://github.com/search?q=repo%3Aaeon-toolkit%2Faeon+check_random_state&type=code).

I agree, but `numpy.random.RandomState` cannot be used directly in Numba as aforementioned. It's a very old [issue in scikit-learn](https://github.com/scikit-learn/scikit-learn/issues/16988) (2020) that has never been addressed, but `numpy.random.RandomState` is legacy. We can "bypass" this issue by using `sklearn.utils.check_random_state` (or the equivalent in `aeon` if there is one), but it needs to return a Generator, therefore the `numpy.random.RandomState` would not be used directly for the random number generation: it would be used to generate the seed of the Generator that would be used for the random number generation.

## There are some naming issues, perhaps because of language barrier

> -  `s/Hydra scaler/sparse scaler (used by Hydra)/ ` (sparse scaler is not necessarily Hydra-specific)

That's the same convention used in Hydra. The argument in the class is called [`scale_hydra`](https://github.com/johannfaouzi/aeon/blob/03471375cecfcf562f287ecca53c22e73f67b052/aeon/transformations/collection/convolution_based/_kgmtp.py#L499-L503), so there is no public mention of `sparse`.

> - `s/n_ppv_features_/n_pooling_features_/` -- there are other features apart from Proportion of Positive Values (PPV); the paper calls all the non-Hydra features (somewhat unfortunately in my humble opinion) "pooling features"

I agree with this. I have no strong opinion about the solution:

* a single `int` attribute: `n_pooling_features_`,
* five `int` attributes: `n_ppv_features`, `n_mpv_features_`, `n_mipv_features_`, `n_lscv_features_`, `n_zcc_features_`,
* a single attribute but as a tuple or a dictionary to provide more information.

However, there are no such public attributes in the other relevant classes (`Rocket`, `MiniRocket`, `MultiRocket`).

PR #3840, opened 2026-09-17
[MNT] Fix PR estimator subsampling for Python 3.14

MatthewMiddlehurst:
#### Reference Issues/PRs

Follow-up to #3245, which added Python 3.14 without updating the PR subsampling remap.

The PR pytest matrix gives each OS/Python version pair a third of the estimators by mapping the Python minor version to 0/1/2 and adding an OS offset. The matrix now runs 3.12, 3.13 and 3.14, but the remap still listed 3.10/3.12/3.13. This resulted in each OS missing some estimators tests in PR runs. Every test was run in PRs and on the nightly tests, this just caused PRs to have a very poor spread.

To stop this from going stale again, a new `test_ci_workflows.py` reads the `pytest` job matrix from `pr_pytest.yml`, drops entries excluded for PR testing, and asserts that every runner OS and every Python version covers all three subsamples. It runs in about 10ms. Against the old remap it fails with:

```
AssertionError: PR runs for ubuntu-24.04 in pr_pytest.yml only test estimator subsamples [0, 2], update _get_pr_subsample_index so that every subsample is tested on each OS and Python version.
```

The test skips when `.github` is not present, i.e. when the tests are not run from a repository checkout. Within a checkout it fails rather than skips if the workflow file or the matrix cannot be read, so a renamed or restructured workflow is not silently ignored.

`pyyaml` is added to the `dev` extra to parse the workflow. It was already installed in every environment running the tests through `pre-commit`.


TonyBagnall (approved):
lgtm

PR #3841, opened 2026-09-17
[MNT] Fix PR welcome comment label updates

MatthewMiddlehurst:
#### Reference Issues/PRs

Noticed on #3834, where the labels set at PR creation were never ticked in the bot comment and adding `codecov actions` later did nothing.

- **Label edits never updated the comment:** `pr_welcome_updater.py` looked for a comment containing ``## Thank you for contributing to `aeon-eval` ``. Our comment says `aeon`, so the loop never matched, the script exited 0 and the "PR label edited" workflow passed while doing nothing. This has been the case since the script was added in #3024. It now looks for the right heading, and only edits the comment when a box actually changes, so it does not trigger the comment edit workflow for nothing.
- **Labels set when the PR is opened:** these are applied before the welcome comment exists, so the label edit workflow has nothing to find and the comment is created with every box unticked. `pr_open_commenter.py` now ticks the boxes for labels already on the PR, using the same `label_options` list the other scripts use. It re-reads the labels once more after creating the comment, which covers a label added while the workflow is running. Any label added after the comment exists is handled by the label edit workflow as before.
- **Label colours:** label names are rendered as shields.io badges again, e.g. `maintenance`. The latex colouring removed in #2878 does render on desktop and mobile web again, but it uses the label colour as *text* colour, and every label in the repo fails WCAG AA on at least one theme that way. Nine of the labels that can appear in the comment are effectively invisible on one of them: `enhancement`, `forecasting`, `examples`, `governance`, `refactor` and `visualisation` on the light theme, `datasets`, `deprecation` and `testing` on the dark theme. A badge uses the colour as the background and picks a readable text colour for it, and works on the mobile app where latex may not. The image alt text is the label name, so the sentence still reads correctly if the badge does not load.
- `str(list).strip("[]").replace("'", "")` is replaced with a join. That idiom put each entry through `repr`, which is why the old latex output had doubled backslashes.

The comment now embeds images from `img.shields.io`, proxied by GitHub. No Python dependency is added.

#### Any other comments?

Tested by running both scripts against a stubbed `PyGithub` client: labels present at open, a label added mid-run, no labels, a label change with nothing to do, and the no-title-labels and large-content-set message paths. Every badge URL generated from the repo's current labels returns 200, including the escaped `full pre-commit` and the emoji label.

These workflows run from `main`, so #3834 will not update until this is merged.


MatthewMiddlehurst:
This ones a full AI PR FYI. Mostly because its boring and I messed it up last time doing it manually apparently 🙂. Checked over it and looks fine.

TonyBagnall:
well my bot says your bot is correct

Approve, nothing blocking. Ran both scripts against a stubbed `PyGithub` rather than reading the diff.

Verified:
- **Heading fix is real and complete.** The opening comment is headed ``## Thank you for contributing to `aeon` ``, so the `aeon-eval` search in `pr_welcome_updater.py` never matched, the loop found no comment, and the `pull_request_target: [labeled, unlabeled]` workflow exited 0 doing nothing. `git blame` puts the wrong string at 2025-08-13. `pr_welcome_edited.py` already used the correct heading, so this was the only affected script.
- **Open-time ticking + mid-run re-read.** With a `full pre-commit` label present at open, its box is ticked in the created comment. Adding a `codecov actions` label between the two `pr.get_labels()` calls produces exactly one follow-up `comment.edit`. No-labels and a non-CI label (tested an emoji label) neither crash nor edit. The `!= comment.body` guards make an unchanged label set a genuine no-op, so no empty comment-edit is triggered.
- **Badge escaping.** Matches shields.io rules (`-`→`--`, `_`→`__`, `quote` for spaces): `full pre-commit` → `full%20pre--commit`, emoji and multi-word labels encode fine, and the doubled dashes keep the single name/colour separator unambiguous. Alt text is the label name, so the line degrades gracefully.
- **The `join` replacing `str(list).strip("[]")...`** removes the `repr` pass that was doubling the backslashes in the old latex output.

Two non-blocking notes, neither introduced by this PR:
- `_commons.label_options` still carries `("stop pre-commit fixes", "Stop automatic `pre-commit` fixes ...")`, but there is no such checkbox in the comment template, so it is a dead no-op in all three scripts including the new `tick_label_boxes`. Since you are in here anyway, might be worth dropping the orphaned entry (or restoring its box). The other two template boxes without a `label_options` entry, "Regenerate expected results" and "Push an empty commit", are correctly handled in `pr_welcome_edited.py`, so those are fine.
- The second-check `new_comment_body != comment.body` assumes the API echoes the created body verbatim, which it does; worst case if it ever normalised whitespace is one redundant edit, never a wrong result. Fine as-is.
-

TonyBagnall (approved):

PR #3842, opened 2026-09-17
[ENH] RDST samplix fix & perf improvements

baraline:

This PR speeds up `RandomDilatedShapeletTransform`, (without changing the features it produces), and fixes a bug in the filtering of the shapelets that could not be sampled.

- Vectorised distance computation. The subsequences of a series are copied to a `(n_channels, length, n_subsequences)` array by the new `get_dilated_subsequences` and `normalise_dilated_subsequences` utilities, so the inner loop of the distance kernel runs over contiguous subsequences and is vectorised by numba. The parallel loop of transform runs over (series, length, dilation) combinations instead of series only, which balances the work between threads.
- Sampling from admissible point counts. During fit, the number of admissible start points is kept per series and shapelet length and updated together with the alpha similarity mask. Sampling a shapelet now only scans the mask of the selected series instead of running `np.where` on every series, so the fit time no longer grows with the number of series. The same random draws are made, so the shapelets are unchanged for a given seed.
- Bug fix. `random_dilated_shapelet_extraction` is compiled with `fastmath=True`, which folds its `values == np.inf` check to False. The shapelets that could not be sampled once the alpha similarity mask had exhausted the search space were returned with inf values, and gave all zero features after the "invalid values" warning of transform. They are now tracked with a flag and dropped, so `n_shapelets_` can be lower than `max_shapelets` as documented. The alpha similarity mask update also no longer wraps negative indices to the end of the series, which changes the sampled shapelets only when several lengths share a dilation.

#### Any other comments?

Comparison with main on 30 UCR datasets, using the RDSTClassifier pipeline with default parameter. The single thread numbers are means over 3 seeds. Times are the totals over the 30 datasets.

| | Fit (sampling and train transform) | Test transform | Accuracy |
|---|---|---|---|
| 1 thread | 41.3 s to 15.6 s, 3.0x | 80.9 s to 28.8 s, 3.4x | identical on the 30 datasets |
| 6 threads | 23.9 s to 8.2 s, 3.0x | 41.6 s to 9.1 s, 5.1x | identical on the 30 datasets |

The accuracies are identical because the sampled shapelets are the same and the shapelets returned by the bug were all zero features, which the classifier ignores. They still cost time on main: on small datasets most of the 10000 shapelets cannot be sampled (Chinatown keeps 90 of them), which explains the largest speedups. The fit time is now flat in the number of series: with 2000 series of 150 points and 10000 shapelets, fit goes from 5.7 s to 0.2 s single threaded.


TonyBagnall:
Nice! I'll review next week, and also will rerun on the UCR for speed comparison and sanity checks, RDST does well on the multiverse https://github.com/aeon-toolkit/multiverse

aeon/transformations/collection/shapelet_based/_dilated_shapelet_transform.py
TonyBagnall: Shapelets : tuple
    The returned tuple contains 7 arrays describing the shapelets parameters:
        - values : array, shape (max_shapelets, n_channels, max(shapelet_lengths))
            Values of the shapelets.
        - startpoints : array, shape (max_shapelets)
            Start points parameter of the shapelets
        - lengths : array, shape (max_shapelets)
            Length parameter of the shapelets
        - dilations : array, shape (max_shapelets)
            Dilation parameter of the shapelets
        - threshold : array, shape (max_shapelets)

I think its now 9 and should now be n_shapelets not max_shapelets?

TonyBagnall (changes requested):
comment correction, and #3726 changes sliding_mean_std_one_series(float32) to return float32, while #3842 still allocates inv_stds as float64, so I its failing

couple of other points from bots 
- The test should use `assert_allclose` (e.g. `rtol=1e-6` for float32) instead of `assert_array_equal`. Even with the dtype fix it is still 1 ulp off here. Exact equality depends on whether LLVM turns the fastmath division into a reciprocal multiply, and that is not stable across compilers or platforms.

- ### Non-blocking

- `compute_shapelet_dist_vector` and `compute_shapelet_features` are public names. Their `X_subs` layout has changed from `(n_sub, n_channels, length)` to `(n_channels, length, n_sub)`, so any external caller will get wrong results silently rather than an error. Nothing in aeon depends on the old layout: the visualisation module has its own copy. Even so, I'd either rename them with a leading underscore or add a line to the changelog.
- The comment "each row of X_new is written by one thread" is not guaranteed under `prange`. There is no race, because each `(i_x, pair)` iteration writes disjoint columns, so the comment should say that instead.
- The non-normalised path now copies subsequences where it used to take a strided view. The copy is `n_channels * length * n_sub` floats per iteration, which is fine for UCR/UEA but could get large for high-channel data. This is just a note.
- Follow-up rather than this PR: `_features_from_dist_vector` and the `values` initialisation still rely on `np.inf` sentinels inside `fastmath=True`. It works today (the oracle agrees), but `ninf` semantics don't guarantee it. The same bug class was just fixed in the mask filter. Using `fastmath={"nsz", "arcp", "contract", "afn", "reassoc"}` on these kernels would make this safe.

TonyBagnall (commented):
random_dilated_shapelet_extraction still says “contains 7 arrays”, although it returns 9. also now that unsuccessful candidates are filtered, its return shapes should say n_shapelets, not max_shapelets. There are similar max_shapelets references for startpoints/classes in the transform docstring.

PR #3843, opened 2026-09-18
[BUG] Make rocket transform fitting thread-safe and consistent

MatthewMiddlehurst:
#### Reference Issues/PRs

Follows #3655. Fixes the flaky `test_arsenal_n_jobs_does_not_change_output` and `test_parallel_verbose_fit_preserves_predictions` in the `NUMBA_DISABLE_JIT=1` codecov job.

**Threaded fits now match sequential fits.** Arsenal fits its members in joblib threads, and each rocket transform seeds `np.random` per member.

- MiniRocket and MultiRocket chose channels with `np.random.seed` and `np.random.*` in plain Python, i.e. numpy's global `RandomState`, which all threads share. On multivariate data, concurrent fits (e.g. `Arsenal(rocket_transform="minirocket", n_jobs=4)`) got different channels and biases from sequential fits, with JIT enabled. They now draw from their own `check_random_state(seed)`, which gives the same numbers when seeded and no longer reseeds the user's global RNG.
- `np.random` inside `@njit` functions is thread-local when compiled, but with `NUMBA_DISABLE_JIT=1` it is numpy's global RNG, so Rocket's kernel generation raced as well. That is the cause of the flaky Arsenal tests. The seeded numba generators now run under `_NUMBA_RANDOM_LOCK`; compiled, they hold the GIL anyway, so it costs nothing.
- MultiRocket's transform functions called `np.random.seed` without drawing anything; with JIT disabled this could reseed another thread's fit. Removed.

**MultiRocket made consistent with MiniRocket and Rocket.**

- `n_kernels < 84` raised `ZeroDivisionError`. Both reference implementations ([MiniRocket](https://github.com/angus924/minirocket), [MultiRocket](https://github.com/ChangWeiTan/MultiRocket)) need at least 84, one feature per base kernel (`num_features // 84`), and neither guards it. aeon's MiniRocket already documents raising smaller values to 84, so MultiRocket now does the same, and both store the value used as `n_kernels_`. The MultiRocket docstring's feature count is corrected (49,728 by default, a multiple of 672 rather than 336).
- Fitting on a single case failed, as `X.squeeze()` also dropped the case axis.
- With `normalise=True`, fit normalised after the float32 cast and transform before it, so the biases were fitted to slightly different series from those transformed (about 10% of BasicMotions train features differed, by up to 1.03). Both now normalise then cast, as Rocket does.

**Threading.**

- New `_numba_threads(n_jobs)` sets numba's thread count (which is per Python thread), caps it at `NUMBA_NUM_THREADS` and restores it on exit. Rocket had no cap, so `Rocket(n_jobs=...)` above the thread pool size raised in `transform`.
- Rocket keeps `_NUMBA_PARALLEL_LOCK`. MiniRocket and MultiRocket deliberately do not take it: it makes Arsenal members transform one at a time, which I measured at 2-3x slower fit and predict with `n_jobs=4`, as #3655 found. They remain exposed to the workqueue concurrent-launch abort, as on main.
- Arsenal casts to float32 once for every transform rather than only for Rocket, and MiniRocket and MultiRocket use `np.asarray`, so there is no per-member copy.

No.

#### Any other comments?

Output changes, checked against main:

- JIT enabled: fitted parameters and transforms of Rocket, MiniRocket and MultiRocket (`normalise=False`) are bit-identical, as are Arsenal outputs at `n_jobs=1` and all 25 stored expected-results pipelines for rocket-based estimators (run with the Linux-only guard bypassed). Arsenal at `n_jobs=2` now matches `n_jobs=1`; on main it did not for multivariate MiniRocket and MultiRocket.
- Intended changes: MultiRocket with `normalise=True`; and with JIT disabled, multivariate MultiRocket's differenced representation, whose channel draws no longer depend on the bias fitting having reseeded the global RNG. The latter moves the two stored CardanoSentiment results checked by the JIT-disabled codecov job (MultiRocketRegressor, MultiRocketHydraRegressor) closer to the JIT-enabled values (max difference 0.058 → 0.015 and 0.043 → 0.006). That job only runs these checks on Linux, so it is the one to watch.

Timing, median of 5 paired runs in fresh processes, JIT enabled, 200 cases x 300 timepoints, Arsenal with `n_kernels=2000, n_estimators=10`. Every configuration measured (Arsenal and standalone transforms, 1 and 3 channels, `n_jobs` 1 and 4) is within -6% to +3% of main.

| | main fit / predict (s) | PR fit / predict (s) |
|---|---|---|
| Arsenal minirocket, 3 channels, n_jobs=4 | 1.06 / 0.82 | 1.07 / 0.77 |
| Arsenal multirocket, 1 channel, n_jobs=4 | 2.71 / 1.96 | 2.69 / 1.99 |
| Arsenal multirocket, 3 channels, n_jobs=4 | 4.61 / 3.61 | 4.52 / 3.60 |
| MultiRocket transform, n_jobs=1 | 0.04 / 1.78 | 0.04 / 1.84 |

Tests: no dedicated threaded-fit test; the Arsenal `n_jobs` tests and `check_estimator_multithreading` cover `n_jobs` invariance. Small tests cover the fixes the general checks miss (`n_kernels < 84`, single-case fit, normalise consistency, the thread cap), and each fails on main.

##### For all contributions


TonyBagnall:
still draft, but I ran it through claude who likes it

Verified against numba 0.67 / numpy 2.4.4 / sklearn 1.8. Single-core sandbox
(numba pool = 1, OMP layer), so the RNG, consistency, preprocessing and capping
logic is run directly; the concurrent-launch/workqueue behaviour is inspection
only. Approve, one design question below.

### Reproducibility preserved (JIT on, i.e. production)

Bit-identical main vs this branch, same random_state:
- Rocket, MiniRocket, MultiRocket (univariate and multivariate), and
  `Arsenal.predict_proba` for all three `rocket_transform` values.

Only change: `MultiRocket(normalise=True)` shifts by 3.8e-6, which is the
fit/transform fix below, and `normalise` defaults to False so no default user is
affected. The channel-selection refactor (isolated `check_random_state` per
transform, shared across MultiRocket's two representations) reproduces the old
draws exactly under JIT.

### The consistency fix is real

MultiRocket multivariate, JIT on vs JIT off (NUMBA_DISABLE_JIT=1):
- main: max elementwise diff **31.3** (not allclose). The channel draws are
  polluted by the bias-fitting draws off the global RNG when JIT is disabled.
- branch: **7.6e-6** (float32 order only). Consistent, and equal to main's
  JIT-on output, so the corrected coverage-mode result matches production.

### Each new test catches a genuine pre-existing bug (ran the bodies on main)

1. `test_multirocket_normalise_matches_normalised_input`: on main, `_fit` casts
   to float32 then normalises, while `_transform` normalises then casts, so
   biases are fitted to slightly different series than are transformed (3.8e-6).
   `_prepare_input` (normalise then cast, both paths, matching Rocket) fixes it.
2. single-series `MultiRocket.fit_transform`: crashes on main with
   `ValueError: not enough values to unpack` because `X.squeeze()` collapses the
   case axis on a (1, 1, T) input. `squeeze(1)` fixes it. MiniRocket was already
   fine.
3. `MultiRocket(n_kernels < 84)`: crashes on main with `ZeroDivisionError` in
   `_fit_dilations`. `max(n_kernels, 84)` plus the new `n_kernels_` fixes it and
   aligns with MiniRocket.

### Other checks

- `_numba_threads` caps at `NUMBA_NUM_THREADS` and restores. Concrete receipt:
  with the pool at 1, `Rocket(n_jobs=2)` raises
  `ValueError: number of threads must be between 1 and 1` on main and is fine on
  the branch. Mini/MultiRocket didn't crash on main only because they capped at
  `multiprocessing.cpu_count()`, which is the wrong ceiling; `NUMBA_NUM_THREADS`
  is correct.
- The `seed` argument removed from `_transform_uni/_transform_multi` was unused
  for the output (the features are deterministic given the fitted parameters).
  Transform output is unchanged; removal only drops a reseed of the interpreter's
  global RNG that happened under JIT-disabled.
- `check_estimator` clean: Rocket/MiniRocket/MultiRocket 20 each, Arsenal 24.
  Full `test_all_rockets.py` + `test_parallel.py`: 58 passed.
- The "thread count is per Python thread" rationale for `_numba_threads` matches
  numba's documented contract for `set_num_threads`. I couldn't do a value-based
  concurrent test (pool = 1 can't hold two distinct counts).

### One design question (non-blocking, pre-existing)

Rocket holds `_NUMBA_PARALLEL_LOCK` around its transform launch; MiniRocket and
MultiRocket deliberately don't, for a 2-3x Arsenal speedup. Both hit the same
hazard: `Arsenal` runs member transforms concurrently in joblib threads
(`_run_jobs(..., prefer="threads")`), each entering a numba `prange` region. Now
that `_numba_threads` has removed the set-threads race for all three, the open
question is whether a concurrent parallel *launch alone* still aborts the
workqueue layer. The PR answers both ways: Rocket's comment says concurrent
entry aborts workqueue (launch alone is dangerous), yet Mini/Multi drop the lock
(launch alone is safe). If the former, Mini/Multi are exposed whenever numba
falls back to workqueue (no tbb or omp present); if the latter, Rocket's lock is
now vestigial and could go too, recovering Rocket's Arsenal transform speedup.
Worth settling one way, or documenting the threading-layer assumption. This
asymmetry already exists on main (Rocket locked, Mini/Multi not); the branch just
makes the reasoning explicit, which is why it's not blocking. I couldn't
reproduce the workqueue abort here (single core, OMP layer).

### Minor

- MultiRocket now rounds `n_kernels` up to 84; below that it crashed before, so
  no real user is affected, and `n_kernels_` documents it.
- `X.astype` to `np.asarray(..., dtype=np.float32)` copy-avoidance and the
  transform docstring return-type corrections (were "pandas DataFrame", actually
  ndarray) are right.

PR #3846, opened 2026-09-20
[DOC] Update documentation style and small fixes

baraline:

- Migrate sphinx theme to pydata, used by other large packages
- Fix outdated infos and dead links.
- Update multiple styles for a more modern look
- Restyle API reference and add search option
 
To discuss between core developers to see if the style match what we would want to go for.


baraline:
As we said in discord, this is a baseline open for discussion.

Still need to make some review myself to make sure we do not break any compatibilities. And maybe some work on the font part.

It would require the go of multiple core developers before we move this forward.

TonyBagnall:
@baraline as its for discussion, can we make this draft?

Issue #3818, opened 2026-09-15
[MNT] Remove estimators excluded from `check_non_state_changing_method`


MatthewMiddlehurst:
#3412 changed `check_non_state_changing_method` so it catches in-place changes inside nested attributes (lists, dicts, fitted sub-estimators). The old version compared a shallow copy of `__dict__` and missed these. The stricter check found four more estimators that change their own state in `predict`-type methods, on top of three that were already excluded. All seven are currently skipped via `EXCLUDED_TESTS` in `aeon/testing/testing_config.py`.

Methods like `predict`, `predict_proba`, `predict_scores` and `transform` should not change an estimator's attributes. If they do, results can depend on how many times and in what order these methods were called. The fitted object also stops being the one actually used.

This issue tracks removing those exclusions, with one sub-issue per estimator. To close a sub-issue, fix the estimator, remove its entry from `EXCLUDED_TESTS`, and check that `check_non_state_changing_method` passes.

| Estimator | What changes during `predict` |
|---|---|
| `FLUSSSegmenter` | results stored on `self` (`found_cps`, `profiles`, `scores`, `profile`) |
| `ClaSPSegmenter` | results stored on `self` (`found_cps`, `profiles`, `scores`) |
| `HMMSegmenter` | `num_states`, `states`, `num_obs`, `trans_prob`, `trans_id` stored on `self` |
| `InformationGainSegmenter` | inner `_IGTS` object writes `intermediate_results_` |
| `GreedyGaussianSegmenter` | inner `_GGS` object writes `_intermediate_change_points` / `_intermediate_ll` |
| `ROCKAD` | query statistics inside the fitted `NearestNeighbors` KD-tree |
| `COPOD` | the wrapped PyOD `COPOD` writes `U_l`, `U_r`, `U_skew`, `O` in `decision_function` |

**Out of scope:** `TEASER` and `ProbabilityThresholdEarlyClassifier` are also excluded from this check, but early classifiers keep information between predict calls on purpose.

MatthewMiddlehurst:
Please note this is not as simple as ripping up the attributes and making the test pass. Some of these will be informative outputs that we want to keep, but do not meet the current framework. Whether we can move them about to meet the framework or the framework needs to change somehow to meet this is a discussion to be had.

Simple AI solutions which are likely to break these and remove informative outputs without discussion may just be closed.

PR #3850, opened 2026-09-21
[BUG] Stop COPOD predict from modifying the fitted PyOD model

### LINKED ISSUE #3825: [BUG] `COPOD` predict modifies the fitted PyOD model
Part of #3818.

`COPOD.predict` calls `fitted_pyod_model_.decision_function` through `PyODAdapter`. PyOD's `COPOD.decision_function` writes `self.U_l`, `self.U_r`, `self.U_skew` and `self.O` every time it runs, joining the training data with the new data. This happens upstream in PyOD, but it means `predict` changes the fitted state of the aeon estimator.

Separately, `COPOD.__init__` imports PyOD, builds the model and sets `self._n_jobs`, so `set_params(n_jobs=...)` has no effect on the model that is actually used.

Vinayak19112003:
Closes #3825 (sub-issue of #3818).

**Why:** `COPOD.predict` calls PyODs `COPOD.decision_function` on the fitted model, and PyOD overwrites `U_l`, `U_r`, `U_skew` and `O` on every call (computed over training data joined with the new input). So `predict` changed the fitted state of the aeon estimator — `check_non_state_changing_method` failed for `COPOD`.

**What changed:**
- `PyODAdapter._predict` now scores windows through a new `_get_model_for_predict()` hook (default: the fitted model, unchanged behavior for other detectors).
- aeons `COPOD` overrides the hook to return `deepcopy(self.fitted_pyod_model_)`, since the mutation happens upstream in PyOD and cant be avoided there. A copy (rather than saving/restoring the four attributes) doesnt depend on PyODs internal attribute names and covers the `n_jobs > 1` parallel path too.
- Removed `COPOD` from `EXCLUDED_TESTS` in `aeon/testing/testing_config.py`.
- Added regression test `test_copod_predict_does_not_change_fitted_state`.

**How tested:** new test fails on unfixed code, passes with the fix; `check_estimator(COPOD, checks_to_run="check_non_state_changing_method")` passes; full `aeon/anomaly_detection/` suite: 107 passed, 4 skipped (missing optional deps); pre-commit clean on touched files. Predicted scores are unchanged vs pre-fix and vs raw PyOD.

**Follow-up (not in this PR):** the issue also notes `set_params(n_jobs=...)` has no effect because the PyOD model is built eagerly in `__init__`; thats separate param plumbing and left out to keep this fix minimal.


aeon/anomaly_detection/series/_pyodadapter.py
```diff
@@ -107,12 +107,22 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
...
+    def _get_model_for_predict(self) -> BaseDetector:
+        """Return the fitted PyOD model used to score windows in ``_predict``.
+
+        Some PyOD models change their fitted state when ``decision_function``
+        is called. Subclasses wrapping such models should override this to
+        return a copy, so that ``predict`` does not modify the fitted
+        estimator.
+        """
+        return self.fitted_pyod_model_
+
```
MatthewMiddlehurst: Is a method required for this? If its cheap enough could just copy so we dont run into this issue on any adapter method. keep in mind this class can be used for any pyod class really

aeon/anomaly_detection/series/distribution_based/_copod.py
```diff
@@ -61,6 +67,12 @@ def _predict(self, X: np.ndarray) -> np.ndarray:
     def _fit_predict(self, X: np.ndarray, y: np.ndarray | None = None) -> np.ndarray:
         return super()._fit_predict(X, y)
 
+    def _get_model_for_predict(self) -> "BaseDetector":
+        # PyOD's COPOD.decision_function overwrites the fitted attributes U_l,
+        # U_r, U_skew and O on every call, so score a copy to leave the fitted
+        # model unchanged, see #3825.
+        return deepcopy(self.fitted_pyod_model_)
```
MatthewMiddlehurst: Would a regular copy work instead of a deep copy? Would not be ideal to do that every predict.

aeon/anomaly_detection/series/distribution_based/tests/test_copod.py
```diff
@@ -25,6 +28,24 @@ def test_copod_default():
...
+    series = make_example_1d_numpy(n_timepoints=80, random_state=0)
+
+    copod = COPOD(window_size=10, stride=1)
+    copod.fit(series, axis=0)
+    fitted_state_before = deepcopy(copod.fitted_pyod_model_.__dict__)
+
+    preds = copod.predict(series, axis=0)
+
+    assert preds.shape == (80,)
+    assert deep_equals(fitted_state_before, copod.fitted_pyod_model_.__dict__)
```
MatthewMiddlehurst: This is the same as the test you just removed from the exclude list it seems?

MatthewMiddlehurst (changes requested):

MatthewMiddlehurst:
Also noticed `self.n_jobs = n_features` which is not being caught to do being multithreading related..

aeon/anomaly_detection/series/distribution_based/_copod.py
Vinayak19112003: Yep, switched to a shallow copy. I checked PyOD’s COPOD source — the mutations in `decision_function` (`U_l`/`U_r`/`U_skew`/`O` and `n_jobs`) are all attribute re-assignments, so a shallow copy is enough to protect the original and it keeps predict cheap. Covered by the tests.

aeon/anomaly_detection/series/_pyodadapter.py
Vinayak19112003: Good point — done. I dropped the hook method entirely; `_predict` (and `_inner_predict`) now just score a copy of the fitted model, so this holds for any PyOD class wrapped by the adapter, not just COPOD. COPOD doesn’t need an override anymore.

aeon/anomaly_detection/series/distribution_based/tests/test_copod.py
Vinayak19112003: You’re right — I verified `check_non_state_changing_method` fails on the unfixed code, so the re-enabled check covers it. Removed the explicit test.

Vinayak19112003:
Good catch — that reassignment happens in PyOD’s parallel branch of `decision_function`, and the estimator check never hits it since it runs with `n_jobs=1`. With the copy now happening in the adapter’s `_predict`, the reassignment lands on the copy, so the fitted model keeps its `n_jobs`. I added a small regression test for that path since the check doesn’t cover it.

PR #3852, opened 2026-09-22
BUG Fix ClaSPSegmenter prediction state mutation

### LINKED ISSUE #3820: [BUG] `ClaSPSegmenter` stores results on self in `predict`
Part of #3818.

`ClaSPSegmenter` has `fit_is_empty=True`, but it stores its results on the estimator in several places:

- `_predict` sets `self.found_cps`, `self.profiles` and `self.scores`, including the branch that sets them to empty lists.
- `predict_scores` and `_run_clasp` set the same three attributes.

These attributes have no trailing underscore, and `get_fitted_params` is overridden to return them even though nothing is fitted.

`FLUSSSegmenter` has the same problem (#3819), so the two fixes should look the same.

Solaris-star:
## Summary
- make `ClaSPSegmenter` prediction helpers return computed results without storing them on the estimator
- return no fitted parameters for the `fit_is_empty=True` estimator
- re-enable the generic non-state-changing estimator check

Fixes #3820

## Validation
- `pre-commit run black --files aeon/segmentation/_clasp.py aeon/segmentation/tests/test_clasp.py aeon/testing/testing_config.py`
- `pre-commit run ruff --files aeon/segmentation/_clasp.py aeon/segmentation/tests/test_clasp.py aeon/testing/testing_config.py`
- `pytest -q aeon/segmentation/tests/test_clasp.py`
- generic `check_non_state_changing_method` for `ClaSPSegmenter`
- `git diff --check`


MatthewMiddlehurst:
Not sure I like this way of going about it. Not sure what those attributes contain, but having to replace the attribute calls with a private equivalent of predict is not an upgrade IMO. Needs more discussion.

PR #3853, opened 2026-09-22
[BUG] FLUSSSegmenter no longer stores results on self in predict

### LINKED ISSUE #3819: [BUG] `FLUSSSegmenter` stores results on self in `predict`
Part of #3818.

`FLUSSSegmenter` has `fit_is_empty=True`, but it stores its results on the estimator in several places:

- `_predict` and `predict_scores` set `self.found_cps`, `self.profiles` and `self.scores`.
- `_run_fluss` sets `self.profile` and `self.scores`.

These attributes have no trailing underscore, and `get_fitted_params` is overridden to return them even though nothing is fitted.

`ClaSPSegmenter` has the same problem (#3820), so the two fixes should look the same.

Vinayak19112003:
Closes #3819 (sub-issue of #3818).

`_predict` and `predict_scores` used to stash their results on the estimator (`found_cps`, `profiles`, `scores`), and `_run_fluss` also wrote `profile`/`found_cps`/`scores` onto `self`. Since `fit_is_empty=True`, none of that is fitted state, and repeated calls could leave stale attributes behind. The helpers now just return the computed values and nothing is written to `self`; `get_fitted_params` returns `{}` since there is nothing fitted (same shape as the ClaSP fix in #3852). Also removed the `FLUSSSegmenter` entry from `EXCLUDED_TESTS` so the generic check runs again.

The docstring example referenced the removed `fluss.profiles`/`fluss.scores` attributes, so it now uses `predict_scores` instead. Public behaviour is otherwise unchanged: `predict` still returns the found change points and `predict_scores` the scores.

Done criteria from #3818:
- `check_non_state_changing_method` passes for `FLUSSSegmenter`
- new regression test `test_fluss_predict_does_not_change_state` (fails on main, passes here) asserting `predict`/`predict_scores` leave `__dict__` untouched
- `pytest aeon/segmentation/tests/test_fluss.py` passes (2 passed), pre-commit (black/ruff/isort/codespell) clean


MatthewMiddlehurst:
Similar to https://github.com/aeon-toolkit/aeon/pull/3852 I think this needs more discussion first. I don't think just dropping the data is good, or at least needs some reasoning to go with it on why we aren't losing much.

Vinayak19112003:
Fair question — here’s the reasoning. The short version: what `predict` was storing was never working API, and the useful outputs are still available.

The attributes `predict` set (`found_cps`, `profiles`/`profile`, `scores`) were never documented — there’s no Attributes section in the docstring, and the naming wasn’t even consistent (`profiles` vs `profile`). `get_fitted_params` returned `{"profile": self.profile}`, but `profile` only existed after `predict` had run, so after a plain `fit` it just raised `AttributeError`. Nothing that worked reliably is being taken away.

What users actually need is still there: the change points are `predict`’s return value, and the scores remain available through `predict_scores(X)`. The one genuine loss is the full matrix profile for anyone inspecting it post-hoc after `predict` — but keeping that means `predict` keeps mutating state, which is exactly what #3818 and the estimator check rule out. Computing it in `fit` isn’t an option either, since FLUSS’s fit is empty by design.

Same situation as #3852, I know — happy to align with whatever you decide there. And if you’d rather keep the profile reachable, I can have a method return it without storing anything, so `predict` stays pure. Just say the word.

PR #3854, opened 2026-09-22
[BUG] GreedyGaussianSegmenter no longer mutates its inner _GGS object in predict

### LINKED ISSUE #3823: [BUG] `GreedyGaussianSegmenter` modifies its inner `_GGS` object in `predict`
Part of #3818.

`GreedyGaussianSegmenter.__init__` builds a `_GGS` dataclass and stores it as `self.ggs`. `_fit` only calls `self.ggs.initialize_intermediates()`, and `_predict` calls `self.ggs.find_change_points(X)`, which overwrites `_intermediate_change_points` and `_intermediate_ll`.

As with IGTS, `set_params` does not reach the `_GGS` object built in `__init__`.

Vinayak19112003:
Closes #3823 (sub-issue of #3818).

`GreedyGaussianSegmenter.__init__` used to build a `_GGS` work object once and keep it on `self.ggs`. `_predict` then called `find_change_points` on it, which overwrites `_intermediate_change_points`/`_intermediate_ll` on every call, so results depended on call history. It also meant `set_params(k_max=...)` never reached the object actually used for segmentation.

The `_GGS` work object is now built fresh from the current parameters for each `predict` call (new `_get_ggs` helper, also used by `__repr__`), so `predict` cannot mutate estimator state and `set_params` always takes effect. `_fit` is a no-op since the estimator is stateless. Also removed the `GreedyGaussianSegmenter` entry from `EXCLUDED_TESTS` so the generic check runs again, and dropped the stale `Attributes` docstring section that documented `_GGS` internals as if they were estimator attributes (they never were).

Done criteria from #3818:
- `check_non_state_changing_method` passes for `GreedyGaussianSegmenter`
- new regression tests `test_GreedyGaussianSegmenter_predict_does_not_change_state` and `test_GreedyGaussianSegmenter_set_params_reaches_algorithm` (both fail on main, pass here)
- `pytest aeon/segmentation/tests/test_ggs.py` passes (4 passed), pre-commit (black/ruff/isort/codespell) clean


Vinayak19112003:
Same design question as #3855 — I’ve laid out the trade-off and options there (https://github.com/aeon-toolkit/aeon/pull/3855#issuecomment-5795874128), covering both PRs. Happy to go whichever way you prefer.

PR #3855, opened 2026-09-22
[BUG] InformationGainSegmenter no longer mutates its inner _IGTS object in predict

### LINKED ISSUE #3822: [BUG] `InformationGainSegmenter` modifies its inner `_IGTS` object in `predict`
Part of #3818.

`InformationGainSegmenter.__init__` builds an `_IGTS` object and stores it as `self._igts`. `_predict` then calls `self._igts.find_change_points(X)`, which resets and fills `intermediate_results_` on that object. The old shallow check missed this; #3412 catches it.

Building the object in `__init__` causes another problem: calling `set_params(k_max=...)` does not update the `_IGTS` object that is actually used.

Vinayak19112003:
Closes #3822 (sub-issue of #3818).

`InformationGainSegmenter.__init__` used to build an `_IGTS` object once and keep it on `self._igts`. `_predict` then called `find_change_points` on it, which resets and refills `intermediate_results_` on every call, so the fitted object was no longer the one actually used and results depended on call history. It also meant `set_params(k_max=...)` never reached the object used for segmentation.

The `_IGTS` work object is now built fresh from the current parameters for each `predict` call (new `_get_igts` helper, also used by `__repr__`), so `predict` cannot mutate estimator state and `set_params` always takes effect. Also removed the `InformationGainSegmenter` entry from `EXCLUDED_TESTS` so the generic check runs again, and dropped the stale `Attributes` docstring section that documented `_IGTS` internals as if they were estimator attributes (they never were).

Done criteria from #3818:
- `check_non_state_changing_method` passes for `InformationGainSegmenter`
- new regression tests `test_InformationGainSegmenter_predict_does_not_change_state` and `test_InformationGainSegmenter_set_params_reaches_algorithm` (both fail on main, pass here)
- `pytest aeon/segmentation/tests/test_igts.py` passes (7 passed), pre-commit (black/ruff/isort/codespell) clean


MatthewMiddlehurst:
Both this and https://github.com/aeon-toolkit/aeon/pull/3854/ need further discussion on whether we want these attributes to be available rather than just removing them.

Vinayak19112003:
Good question — let me lay out what’s actually there, because the docstrings promise more than the code ever delivered. This applies to #3854 as well.

On both estimators the documented `change_points_` attribute was never set on the estimator — `_predict` used it as a plain local variable. And on IGTS there’s even a commented-out line (`# self.intermediate_results_ = self._igts.intermediate_results_`) where someone deliberately chose not to expose the intermediates. So those Attributes sections were stale; removing them is a docs fix either way.

What really disappears: the per-k intermediate results that used to be reachable on the inner work object after `predict` (`_intermediate_change_points`/`_intermediate_ll` for GGS, `intermediate_results_` for IGTS) — but only as a side effect of `predict` mutating shared state, on private attributes. That part is genuinely useful (e.g. picking k from the likelihood curve), so it’s a real loss, just never a supported one.

Options as I see them:

1. Accept the removal (these PRs as-is) — the intermediates were private and only available by accident of the bug.
2. Expose them properly through a stateless method that returns them without storing anything on `self`, along the lines of FLUSS’s `predict_scores` — `predict` stays pure, the data stays reachable. That’s new API though, so bigger than a bug fix.

One thing I’d keep regardless: building the work object fresh per `predict` also fixes `set_params` never reaching the algorithm (it used to be built once in `__init__`), which was the second half of #3822/#3823. Happy to implement option 2 if you’d like it, or leave these as pure bug fixes.

Issue #827, opened 2023-10-13
[ENH] checkpointing interface


TonyBagnall:
### Describe the feature or idea you want to propose

it would be good to have a checkpointing (aka "warm start" interface to allow the restarting of algorithms after pausing mid way. This will become more important with the 60 hr time limit on HPC.

### Describe your proposed solution

we could just define an interface (akak mixin) that models the java version perhaps, with pickling for default save and load. Might get complex with threading

```java
public interface Checkpointable extends Serializable {

    boolean setCheckpointPath(String path);
    boolean setCheckpointTimeHours(int t);
    void saveToFile(String filename);
    void loadFromFile(String filename);
```
```python
class Checkpointing:
   @abstract
    def set_path(path):
   ...
   def set_type(hours):
   ...
   def save(file_name):
   ...
  def load(file_name):
   ...
```

TonyBagnall:
bumping this as it will soon become a priority, given that my university has a three day job limit on the cluster

PR #3856, opened 2026-09-22
[ENH] Adds Checkpointable mixin with Arsenal

TonyBagnall:
got round to first go at Checkpointing #827 because I want to have a go at some big problems. Open to redesign. The biggest change is a resume_fit, this is required because base class fit resets all parameters, so rather than alter that we have a resume method. 

basic model is

1. `fit()` initializes training, without checkpoints behaviour the same. Fit with checkpoints

```python
  clf = Arsenal(
      n_estimators=100,
      checkpoint_path="arsenal.pkl", # current directory or absolute path ok, directory must exist
      checkpoint_interval=30, # Minutes
      random_state=0,
  )
  clf.fit(X, y)
```

2. `resume_fit()` continues existing training. 

```python
clf = Arsenal.load_checkpoint("arsenal.pkl")
clf.resume_fit(X, y)
```

Ive kept it external to fit for cleanliness, rather than look in fit if a pkl exists etc. not least because fit completely resets the classifier state, and its messy to embed it. User controls the checkpoint structure. Also need to maintain fit/_fit structure. 

3. Checkpointing in _fit and  settings control when progress is saved through is _checkpoint_is_due

4. Arsenal is the example for checkpointable. Classifier is made checkpointable in _fit and _resume_fit

Arsenal does this
_fit() → initialize ensemble and RNGs → _continue_arsenal()
_resume_fit() → _continue_arsenal()

we need to store all state is self now for pickling, and the checkpoint boundaries are algorithm specific. 
Arsenal calls the base class self._checkpoint_if_due() after each batch. Obviously checkpoint time is approximate and algorithm specific, trade off between saving too often and losing work on a deadline. 

5. Works with contracting. So you can contract a model and come back to it whenever you want.

```python
  clf = Arsenal(
      time_limit_in_minutes=3500,
      contract_max_n_estimators=1000,
      checkpoint_path="arsenal.pkl",
      checkpoint_interval=30,
      random_state=0,
  )
  clf.fit(X, y)
```
  A later job can continue with:

```python
  clf = Arsenal.load_checkpoint("arsenal.pkl")
  clf.resume_fit(X, y)
```

6. predict and predict proba are not checkpointed, I would split that externally for the slow predictors, i.e. split test_X into separate calls

7. For cluster runs, we can just adjust tsml-eval to check if  a path exists to a partly trained model and continue. But we might need some work on that to get internals

8. As a sort of side effect you can use resume_fit in code without checkpointing, say by increasing the number of trees then resuming. Not something to encourage, but its harmless at the moment imo.


aeon/base/_checkpoint.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        if self.get_tag("cant_pickle"):
+            raise ValueError(
+                "Default checkpoint persistence requires cant_pickle=False."
+            )
+        if not getattr(self, "_checkpoint_ready", False):
+            raise ValueError("No safe continuation state is available to checkpoint.")
+        filepath = Path(filepath)
+        temporary_path = None
+        try:
```
TonyBagnall, line 139: the reason for this is to avoid overwriting an existing file before writing the new is complete

The temporary file protects the previous checkpoint. Writing directly to arsenal.pkl would truncate it immediately; a failed write or scheduler termination could then leave no usable checkpoint. With this approach, the previous file remains untouched until the replacement is ready.

aeon/base/_checkpoint.py
```diff
@@ -0,0 +1,189 @@
...
+        estimator : CheckpointableMixin
+            The restored estimator.
+
+        Raises
+        ------
+        TypeError
+            If the checkpoint does not contain an instance of ``cls``.
+        """
+        with Path(filepath).open("rb") as file:
+            metadata = pickle.load(file)
```
TonyBagnall: save_checkpoint() writes two separate pickle objects sequentially into the same file:

  pickle.dump(metadata, file)
  pickle.dump(self, file)

  Each pickle.load(file) reads the next object and advances the file position:

  1. The first loads the metadata.
  2. The second loads the estimator.

  This lets us check the format and aeon version before attempting to reconstruct the estimator, which might fail under an incompatible version.

aeon/base/_checkpoint.py
```diff
@@ -0,0 +1,189 @@
...
+                name: value
+                for name, value in self.get_params(deep=False).items()
+                if name not in self._checkpoint_mutable_params
+            }
+        )
+
+    def _validate_checkpoint(self, X, y):
+        if not getattr(self, "_checkpoint_ready", False):
+            raise ValueError("No safe continuation state is available; call fit first.")
+        if joblib_hash((X, y), coerce_mmap=True) != self._checkpoint_data_signature:
```
TonyBagnall, line 281: we use the hash to make sure the data is the same on subsequent calls to resume_fit

aeon/base/_checkpoint.py
```diff
@@ -0,0 +1,189 @@
...
+            raise ValueError(
+                "checkpoint_interval must be None or positive finite minutes."
+            )
+        if self.checkpoint_path is not None:
+            Path(self.checkpoint_path)
+            if self.get_tag("cant_pickle"):
+                raise ValueError("Default checkpointing requires cant_pickle=False.")
+        self._checkpoint_last_time = time.monotonic()
+
+    def _checkpoint_if_due(self, force=False):
```
TonyBagnall, line 316: (force=True) is used for the final checkpoint when fit is complete, so it overrides the timer requirement

aeon/classification/convolution_based/tests/test_arsenal_checkpoint.py
```diff
@@ -0,0 +1,191 @@
...
+    )
+    assert joblib_hash(expected._train_estimates) == joblib_hash(
+        actual._train_estimates
+    )
+
+
+@pytest.mark.parametrize("n_jobs", [1, 2])
+@pytest.mark.parametrize("contracted", [False, True])
+@pytest.mark.parametrize("train_estimates", [False, True])
+def test_arsenal_resume(
```
TonyBagnall, line 42: we can generalise this to any checkpointable later

aeon/base/_checkpoint.py
TonyBagnall: this has changed, it now writes a json for header, as its safer

TonyBagnall:
Fixed final-only checkpointing to retain normal batching; small batches are used for time contracts or periodic checkpoints. Checkpoint paths now serialize as strings, avoiding platform-specific
  Path unpickling failures without changing the live parameter.

TonyBagnall:
added a shared fixture for  checkpointing in conftest.py, imagining wider roll out., This also could tidy up the following subsequently but later, its cosmetic
image

